PR #1310, opened 2017-02-01
Fixed argument type for $siteHandle

carlcs:
This was causing an issue with switching sites on edit entry pages.


takobell:
Thanks, Carl.  Cleaned up docblock and added changelog item in https://github.com/craftcms/cms/commit/fc31c8706e83b20d6beed8b9aebd62ff8fa2702f

PR #1317, opened 2017-02-01
Add `url` option to DB config

wesrice:
PaaS solutions (Heroku, Dokku, CloudFoundry, etc) often provide database credentials in the form of a single string via an environment variable. This string contains the driver, username, password, server, port and database name.

This patch defines the `url` option in the config, parses the `url` option parts if defined, and then assigns the parts to existing DB config options when the DB config is loaded, having no impact on existing calls to the DB config.

### DIFF
--- a/src/config/defaults/db.php
+++ b/src/config/defaults/db.php
@@ -41,6 +41,11 @@ return [
      */
     'schema' => 'public',
     /**
+     * The database credentials in a string format following the standard uri syntax.
+     * Format: <driver>://<user>:<password>@<server>:<port>/<database>
+     */
+    'url' => null,
+    /**
      * If you're sharing Craft installs in a single database (MySQL) or a single
      * database and using a shared schema (PostgreSQL), then you can set a table
      * prefix here to avoid table naming conflicts per install. This can be no more than 5
--- a/src/services/Config.php
+++ b/src/services/Config.php
@@ -915,6 +915,34 @@ class Config extends Component
             }
         }
 
+        if ($category === self::CATEGORY_DB && $configSettings['url']) {
+            $url = parse_url($configSettings['url']);
+
+            if (isset($url['scheme'])) {
+                $configSettings['driver'] = $url['scheme'];
+            }
+
+            if (isset($url['user'])) {
+                $configSettings['user'] = $url['user'];
+            }
+
+            if (isset($url['password'])) {
+                $configSettings['database'] = $url['pass'];
+            }
+
+            if (isset($url['host'])) {
+                $configSettings['server'] = $url['host'];
+            }
+
+            if (isset($url['port'])) {
+                $configSettings['port'] = $url['port'];
+            }
+
+            if (isset($url['path'])) {
+                $configSettings['database'] = trim($url['path'], '/');
+            }
+        }
+
         $this->_configSettings[$category] = $configSettings;
     }
 


takobell:
I know Craft runs on PaaS providers already (have seen it on Heroku at least).  How is it currently connecting to the database in those cases?  Do they also support fallback to DSN like strings?

wesrice:
The issue really isn't that Craft won't run on PaaS providers, but rather the way that most PaaS services attaches database services exposes the database credentials through a uri string, which out of the control of the user.

This patch is really just a nicety to limit parsing by CMS user, making Craft a bit more friendly to PaaS users.

Below are some use cases of DB credentials being bound to an environment variable in this format through a variety of DB services. Note that not all services attach the the `DATABASE_URL` environment variable. However, the Craft user can just map `DB_URL` to `getenv('WHATEVER')` and still benefit from the functionality.

https://devcenter.heroku.com/articles/heroku-postgresql#connecting-in-php
https://devcenter.heroku.com/articles/cleardb#using-cleardb-with-php
https://devcenter.heroku.com/articles/jawsdb#using-jawsdb-with-php
https://devcenter.heroku.com/articles/jawsdb-maria#using-jawsdb-maria-with-php
https://github.com/dokku/dokku-postgres/blob/master/README.md
https://github.com/dokku/dokku-mariadb/blob/master/README.md
https://cloud.gov/docs/services/relational-database/#bind-to-an-application

takobell:
> Format: <driver>://<user>:<password>@<server>:<port>/<database>

Is that a standardized format across providers?  Looks through some of the links and I'm not seeing a reference to it.

themccallister:
@takobell this is a common format that I think is a standard (could not tell you where that is specified though). I'm pretty sure that OpenShift has the same implementation of the URL string with credentials...

wesrice:
@takobell I researched this a bit more last night. The format is a an implementation of [RFC 1738](https://www.ietf.org/rfc/rfc1738.txt). However, I could not find any documentation by a PaaS explicitly outlining the implementation of the RFC as it pertains to a database resource. However, Googling '[rfc 1738 "database_url"](https://www.google.com/webhp?sourceid=chrome-instant&rlz=1C5CHFA_enUS702US702&ion=1&espv=2&ie=UTF-8#q=rfc+1738+%22database_url%22)' nets several results of people referencing that the database string format is using the RFC format.

On thing that I did notice is that the <driver> string provided by PaaS DB providers might not align with the necessary driver string for Yii. `mysql` aligns, however, it appears PaaS DB providers use `postgres` and Yii uses `pgsql`. I will need to add a map to this before any further consideration is given.

takobell:
https://github.com/cakephp/cakephp/issues/2569 led me to https://adam.herokuapp.com/past/2010/3/30/urls_are_the_uniform_way_to_locate_resources/, which is pretty good to.  Going to add some thoughts, request some changes using the GH review tool thingy.

src/config/defaults/db.php
```diff
@@ -41,6 +41,11 @@
      */
     'schema' => 'public',
     /**
+     * The database credentials in a string format following the standard uri syntax.
+     * Format: <driver>://<user>:<password>@<server>:<port>/<database>
```
takobell: Much like the 'dsn' setting where you can just explicitly define the whole DSN string, we should probably note that if you set this to anything other than `null`, Craft will ignore all of the other DSN related config setting.

src/services/Config.php
```diff
@@ -915,6 +915,34 @@ private function _loadConfigSettings(string $category)
             }
         }
```
takobell: I think the place this logic should happen is dynamically in `config/components/common.php`, when we're initializing the `db` component (much like we do for `dsn`).

wesrice: @takobell I originally had the logic there, but realized that `Config::CATEGORY_DB` is reference throughout the application, and not just in `config/components/common.php`. These files include:

- `base/ApplicationTrait.php`
- `db/Connection.php`
- `db/mysql/Schema.php`
- `db/pgsql/Schema.php`
- `et/EtTransport.php`
- `helpers/ChartHelper.php`
- `migrations/Install.php`

Putting the logic in `services/Config.php` pretty much centralizes the processing so that by the time the other classes need to retrieve specific variables, such as the driver, the database url is already parsed and there is not need to re-parse.

I'm still open to putting this in a better place; this just seemed like the most central place.

brandonkelly:
We talked this over last week and agreed it's a good idea.

We've made some significant changes to the Config service for Beta 8, so I couldn't merge the PR in directly, though the gist is now there.

Thanks @wesrice!

Issue #1314, opened 2017-02-01
'redis' not actually available as a cache method


ghost:
### Description
In the comments in the config files, it mentions redis as a cache option, but it's coming back as an unsupported cache type.

Missing in https://github.com/craftcms/cms/blob/develop/src/config/components/common.php#L129

### Steps to reproduce

1. Set cache type in configs as "redis"
2. Get `\InvalidArgumentException`

### Additional info

- Craft version: Craft3 Beta
- PHP version: 7.0+
- Database driver & version: MySQL 5.6+
- Plugins & versions: n/a

brandonkelly:
Turns out Redis support was removed from Yii 2, and moved into a separate [extension](https://github.com/yiisoft/yii2-redis). It should be possible to Composer-install that extension yourself and make the application config adjustments it requires via [`config/app.php`](https://github.com/craftcms/docs/blob/master/en/configuration.md#application-config), so I just updated the `cacheMethod` comment to not include `redis` anymore.

rsanchez:
@brandonkelly FYI, you can't use that Redis Yii extension because Craft validates the cacheMethod [here](https://github.com/craftcms/cms/blob/develop/src/config/GeneralConfig.php#L764-L767):

```php
// Validate cacheMethod
if (!in_array($this->cacheMethod, [self::CACHE_METHOD_APC, self::CACHE_METHOD_DB, self::CACHE_METHOD_FILE, self::CACHE_METHOD_MEMCACHE, self::CACHE_METHOD_WINCACHE, self::CACHE_METHOD_XCACHE, self::CACHE_METHOD_ZENDDATA], true)) {
    throw new InvalidConfigException('Unsupported cacheMethod value: '.$this->cacheMethod);
}
```

Any chance of making redis an first-class option again?

brandonkelly:
There’s no need to change the `cacheMethod` value. Your `config/app.php` file will be completely overriding Craft’s [`cache` component definition](https://github.com/craftcms/cms/blob/006c9f65037f36f397d57df377d4cb7599a1c153/src/config/app/main.php#L196-L254), which is the only place that checks the `cacheMethod` setting.

Here’s what your `config/app.php` might look like:

```php
return [
    'components' => [
        'cache' => [
            'class' => yii\redis\Cache::class,
            'redis' => [
                'hostname' => 'localhost',
                'port' => 6379,
                'database' => 0,
            ],
        ],
    ],
];
````

PR #1320, opened 2017-02-01
Adding redis support

fmolina24:
Regarding #1314 
- Added yiisoft/yii2-redis package.
- Added redis as a cache option in common.php
- Added redis default config.
- Added redis as a category to Config service.


takobell:
Closing this per: https://github.com/craftcms/cms/issues/1314#issuecomment-276812639

wesrice:
Hey @takobell and @brandonkelly,

For a bit of meta discussion, I guess there are a few philosophies that P&T could take on a pull request like this, and others (like [my pull request](https://github.com/craftcms/cms/pull/1317)).

On one hand, I understand the complete hesitance of creating new features in the core of the CMS. That could possibly open pandora's box for P&T and should be guarded. In my opinion, and from what I can tell of yours, new features should strive to exist in plugins, rather than pull requests.

Is adding cache drivers something that can be accomplished via a plugin now? Should there be a first party Redis plugin to bridge v2 installs into the v3 era? This could be useful in transitioning an app from v2 to v3, given that v2 has first party support for Redis.

brandonkelly:
Yes, a plugin could change the caching driver by doing this from `init()`:

```php
Craft::$app->set('cache', [
    'class' => yii\redis\Cache::class,
    'redis' => [
        'hostname' => Craft::$app->config->get('hostname', 'redis'),
        'port' => Craft::$app->config->get('port', 'redis'),
        'database' => Craft::$app->config->get('database', 'redis')
    ]
];
```

The only caveat is, there's some possibility that another plugin will get initialized before this one, and it may need to read from/write to the data cache from its `init()` method. Which would go to whatever the configured cache driver is (i.e. File).

Alternatively, it would be easy enough to implement Redis without a plugin.

    composer require yiisoft/yii2-redis

Then add a `config/app.php` file if there isn't one already, and add this:

```php
<?php

return [
    'components' => [
        'cache' => [
            'class' => yii\redis\Cache::class,
            'redis' => [
                'hostname' => getenv('REDIS_HOSTNAME'),
                'port' => getenv('REDIS_PORT'),
                'database' => getenv('REDIS_DB'),
            ]
        ],
    ],
    // ...
];
```

takobell:
@wesrice Generally speaking, we'll prefer plugins rather than bloating core.

PR #1408, opened 2017-02-21
Allow Plugin Class to override composer hasSettings config by using hasSettings method.

dalewpdevph:
Would it be possible to have feature where you can override composer extra: hasSettings config by adding a hasSettings method that returns a boolean?
At the moment when i am trying to pass a hasSettings property on the plugin class it won't override the composer config. It will still use the composer config hasSettings value. On your documentation creating a `public $hasSettings = true;` will have you an option to display setting link but this does not work. It will always get the config from the composer.


brandonkelly:
You can either set `hasSettings` with a public property on your plugin class, **or** via `extra.hasSettings` in `composer.json`, but you shouldn’t do both – the `composer.json` value will take precedence (as you’ve found).

If you really want to set it in `composer.json` and in your plugin class (for some reason), just set it from `init()`:

```php
public function init()
{
    parent::init();

    $this->hasSettings = true;

    // ...
}
```

dalewpdevph:
@brandonkelly Thanks

PR #1458, opened 2017-03-04
some Chinese translation

xssworm:


narration-sd:
That's a lot of work, xssworm -- and a lot better than could be done with Google Translate, though a few I checked look nice there. 

I do have a question, though: on this one, 如果类别不存在URLs，那么此处可以保持空白, there is seemingly a change from the imperative language of the original: Leave blank if categories don’t have URLs.

This could be an important difference, if it is not necessary by cultural patterns, as this instruction is  something the person needs to do -- or maybe it's just that Google didn't do an accurate job of translating the sense.

In this one, though, you made extra clarity, which is probably particularly important across languages: 以站点为单位翻译（不依据语言来翻译, Google say it is: Translate by site (not translated by language).

Appreciated from here, and I'm sure will be by the team, especially coming this early..

xssworm:
"Leave blank if categories don’t have URLs. " in Chinese is can be translated to  "此处可以保持空白如果类别不存在URLs", but usually we don't translate like this. In Chinese, we usually put conditional statements in front. So it's "如果类别不存在URLs，那么此处可以保持空白" here.

Google translation can‘t accurately understand the meaning of "for" in Chinese. Paraphrase "Translate for each site" in Chinese is "为每一个站点翻译" or "翻译每个网站", but this is ambiguous in Chinese. So I have to made extra clarity.

narration-sd:
> "Leave blank if categories don’t have URLs. " in Chinese is can be translated to "此处可以保持空白如果类别不存在URLs", but usually we don't translate like this. The English's declarative sentence usually put "when", "if", "which" on the back. In Chinese we usually put conditional statements in front. So it's "如果类别不存在URLs，那么此处可以保持空白" here.

Yes, that kind of thing I understand, and is why I appreciate how you are doing this work.

But my point is about the 'imperative' nature of the original: it is a (lightweight) command. 

You **must** leave the field blank, if there are not categories. That is the sense in English, which I think needs to come across -- with respect for northeast Asian courtesy (I lived once in Korea).

xssworm:
Thank you for your suggestion. I will be improving this translation. Chinese is my first language and I live in Shanghai, China.

narration-sd:
Thank you for the careful listening also  ;)

My father grew up in Nanking (as used to say it), spoke Mandarin/Putonghua, was not Chinese. His father a university professor there, who stayed through the Japanese invasion and much later -- a long story...

takobell:
Thanks @xssworm and @narration-sd!

PR #1500, opened 2017-03-11
Fix casing in Dashboard.js asset

luwes:
The asset failed to load on my Nginx server because of the casing.


takobell:
Thanks, @luwes!

PR #1534, opened 2017-03-15
adding onAfterAction event to element index

nateiler:
Adding an onAfterAction event which is triggered after a successful action has executed.  This would allow for additional logic to be applied against the list of elements that were selected.


brandonkelly:
Thanks!

Issue #1480, opened 2017-03-10
Allow BaseElementIndex.js to easily extend the POSTed endpoints


nateiler:
Can the POSTed action paths be configurable via the settings so one could point to their own controller without having to override the entire method.

This would apply to the following method: 
- updateElements
- submitAction
- refreshSources

brandonkelly:
I don't see why not!

nateiler:
I can make a pull request if you would like ... let me know.

brandonkelly:
Resolved w/ ceffc96655bea651f4239af9b481d53ae61468b1

PR #1536, opened 2017-03-15
Moving post action paths into settings

nateiler:
enhancement: #1480


nateiler:
Should be good to go with the settings property change from *Path to *Action

PR #1578, opened 2017-03-29
Fixes Safari font rendering in Live Preview

carlcs:


brandonkelly:
@carlcs Is there an issue related to this PR? Not sure what the CP CSS has to do with how fonts are rendered within the Live Preview iframe.

carlcs:
The issue is not with fonts in the iframe, it’s the CP part (left part) of Live Preview that has much lighter font weight than normal.

carlcs:
Edit Entry page:
bildschirmfoto 2017-04-05 um 00 52 28

Live Preview:
bildschirmfoto 2017-04-05 um 00 52 42

carlcs:
Sorry that I missed to add a better description to the commit.

brandonkelly:
Ahh, thanks. I do see that the Live Preview font is thinner in Safari, but do you have any idea why? We are not setting `-webkit-font-smoothing` in either context, and the text properties are identical. Here’s the two labels’ computed `font` styles side-by-side, for example:

screen shot 2017-04-11 at 12 19 13 pm

carlcs:
It might have to do with the animations (font rendering using GPU, see https://www.youtube.com/watch?v=9Woaz-cKPCE&hd=1), but I have also seen this issue with elements that are simply `position: fixed;`.

brandonkelly:
Ah, makes sense. Thanks!

PR #1590, opened 2017-04-02
Revoke the last commit for Chinese translation

xssworm:
it covered the previous translation work.
file changlog: https://github.com/craftcms/cms/commits/develop/src/translations/zh/app.php


takobell:
Thanks!

PR #1616, opened 2017-04-11
* Add $view->registerJsonLd()

khalwat:
* Add $view->registerJsonLd()
* Add ability to render `<title>` tag if `$view->title` is set

Resolves this: https://github.com/yiisoft/yii2/issues/11018

It's resolved in such a way that even if they do decide to implement some method of adding non `application/javascript` script tags, it won't conflict.


khalwat:
Closed by popular demand

PR #1617, opened 2017-04-11
Add $view->registerScript()

khalwat:
* Add $view->registerScript()
* Add render <title> tag if $view->title is set

Resolves this: yiisoft/yii2#11018

It's resolved in such a way that even if they do decide to implement some method of adding non application/javascript script tags, it won't conflict.

### DIFF
--- a/src/web/View.php
+++ b/src/web/View.php
@@ -675,9 +675,6 @@ class View extends \yii\web\View
     protected function renderHeadHtml()
     {
         $lines = [];
-        if (!empty($this->title)) {
-            $lines[] = '<title>' . Html::encode($this->title) . '</title>';
-        }
         if (!empty($this->_scripts[self::POS_HEAD])) {
             $lines[] = implode("\n", $this->_scripts[self::POS_HEAD]);
         }
--- a/src/web/View.php
+++ b/src/web/View.php
@@ -651,15 +651,22 @@ class View extends \yii\web\View
     /**
      * Registers a generic <script> code block.
      * @param string $script the generic <script> code block to be registered
+     * @param int $position the position at which the generic <script> code block should be inserted
+     * in a page. The possible values are:
      *
+     * - [[POS_HEAD]]: in the head section
+     * - [[POS_BEGIN]]: at the beginning of the body section
+     * - [[POS_END]]: at the end of the body section
+     *
+     * @param array $options the HTML attributes for the <script> tag.
      * @param string $key the key that identifies the generic <script> code block. If null, it will use
      * $script as the key. If two generic <script> code blocks are registered with the same key, the latter
      * will overwrite the former.
      */
-    public function registerScript($script, $options = [], $key = null)
+    public function registerScript($script, $position = self::POS_END, $options = [], $key = null)
     {
         $key = $key ?: md5($script);
-        $this->_scripts[$key] = Html::script($script, $options);
+        $this->_scripts[$position][$key] = Html::script($script, $options);
     }
 
     /**
@@ -669,7 +676,10 @@ class View extends \yii\web\View
     {
         $lines = [];
         if (!empty($this->title)) {
-            $lines[] = '<title>' . Html::encode($this->title) . '</title>' . "\n";
+            $lines[] = '<title>' . Html::encode($this->title) . '</title>';
+        }
+        if (!empty($this->_scripts[self::POS_HEAD])) {
+            $lines[] = implode("\n", $this->_scripts[self::POS_HEAD]);
         }
 
         $html = parent::renderHeadHtml();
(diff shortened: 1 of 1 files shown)


src/web/View.php
```diff
@@ -643,6 +649,48 @@ public function clearJsBuffer(bool $scriptTag = true)
...
+    }
+
+    /**
+     * @inheritdoc
+     */
+    protected function renderHeadHtml()
+    {
+        $lines = [];
+        if (!empty($this->title)) {
+            $lines[] = '<title>' . Html::encode($this->title) . '</title>' . "\n";
```
brandonkelly: This doesn't seem related to the rest of the PR, unless I’m missing something?

src/web/View.php
```diff
@@ -643,6 +649,48 @@ public function clearJsBuffer(bool $scriptTag = true)
...
 
     /**
+     * Registers a generic <script> code block.
+     * @param string $script the generic <script> code block to be registered
+     *
+     * @param string $key the key that identifies the generic <script> code block. If null, it will use
+     * $script as the key. If two generic <script> code blocks are registered with the same key, the latter
+     * will overwrite the former.
+     */
+    public function registerScript($script, $options = [], $key = null)
```
brandonkelly: There should be a `$position` argument here just like `registerJs()`, but only which only supports `POS_HEAD`, `POS_BEGIN`, and `POS_END`.

src/web/View.php
```diff
@@ -643,6 +649,48 @@ public function clearJsBuffer(bool $scriptTag = true)
...
+    }
+
+    /**
+     * @inheritdoc
+     */
+    protected function renderBodyEndHtml($ajaxMode)
+    {
+        $lines = [];
+        if (!empty($this->_scripts)) {
+            $lines[] = implode("\n", $this->_scripts);
```
brandonkelly: This would result in

```php
$lines = [
    [
        '<script ... />',
        // ...
    ]
];
```

rather than

```php
$lines = [
    '<script ... />',
    // ...
];
```

src/web/View.php
khalwat: It's something else that I'm hoping can be added. I can do a separate PR if you like; it emulates what Yii2 does in their default template `yiisoft/yii2-debug/views/layouts/main.php`

If `$title` is not set (as is the case by default), then nothing happens. If it is set, then it renders the `<title>` tag based on `$view->title`

src/web/View.php
khalwat: Well, the array ends up looking like this:

`Craft::dd($lines);`

```
[
    0 => '<script type=\"application/ld+json\">{
    \"@context\": \"http://schema.org\",
    \"@type\": \"Article\",
    \"name\": \"Some Blog\",
    \"url\": \"https://nystudio107.com/blog\"
}</script>'
]```

File: src/web/View.php (hunk 2 above)
Comment (khalwat) [inline src/web/View.php]: yep, okay. Requires extending a couple more methods, but I'd agree, it's better this way.

File: src/web/View.php (hunk 3 above)
Comment (khalwat) [inline src/web/View.php]: ah, I see. Nevermind.

Comment (khalwat) [inline src/web/View.php]: Okay so for whatever reason, this is the code pattern that Yii2 uses, so I just followed it. They end up doing an `implode("\n", $lines)` at the end, which results in it being formatted properly.

Example from `yii\web\view`:

```
    protected function renderHeadHtml()
    {
        $lines = [];
        if (!empty($this->metaTags)) {
            $lines[] = implode("\n", $this->metaTags);
        }

        if (!empty($this->linkTags)) {
            $lines[] = implode("\n", $this->linkTags);
        }
        if (!empty($this->cssFiles)) {
            $lines[] = implode("\n", $this->cssFiles);
        }
        if (!empty($this->css)) {
            $lines[] = implode("\n", $this->css);
        }
        if (!empty($this->jsFiles[self::POS_HEAD])) {
            $lines[] = implode("\n", $this->jsFiles[self::POS_HEAD]);
        }
        if (!empty($this->js[self::POS_HEAD])) {
            $lines[] = Html::script(implode("\n", $this->js[self::POS_HEAD]), ['type' => 'text/javascript']);
        }

        return empty($lines) ? '' : implode("\n", $lines);
    }
```

khalwat:
Requested changes made

src/web/View.php
brandonkelly: I'd rather see this as a separate PR; it's confusing in this context.

src/web/View.php
brandonkelly: OK gotcha. It's a bit confusing at first glance, but I guess consistency with Yii is a good thing.

brandonkelly:
All looks great, and I'm not opposed to the `<title>` thing, but that belongs in a separate PR.

khalwat:
Removed the `<title>` rendering (it's just 2 lines of code). Will submit via a separate PR.

brandonkelly:
Thank you sir!

PR #1671, opened 2017-04-26
allow install migrations to use their own plugin

nateiler:
During a plugin install migration, you cannot get the instance via `Plugin::getInstance()` on the plugin you're trying to install.

### DIFF
--- a/src/services/Plugins.php
+++ b/src/services/Plugins.php
@@ -471,6 +471,8 @@ class Plugins extends Component
             'plugin' => $plugin
         ]));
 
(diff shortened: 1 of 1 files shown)


brandonkelly:
@nateiler Can you grab the latest on the `develop` branch and verify this is still needed? We made some other changes that will likely solve this issue.

(Just open up composer.json and change Craft’s version requirement to `"dev-develop as 3.0.0-beta.99"`, and then run `composer update`.)

brandonkelly:
Closing this due to inactivity.

PR #1704, opened 2017-05-13
Request doesn't have a function isActionRequest

Rias500:
It does have a function `getIsActionRequest()` though!


brandonkelly:
Thanks!

PR #1711, opened 2017-05-15
enable support for an array of ‘testToEmailAddress’ general config

nateiler:


takobell:
Thanks, Nate. Going to add this as a bug fix in the changelog since it looks like it's a regression from C2.

PR #1752, opened 2017-06-01
Fix default memcached config

### LINKED ISSUE #1751: Issue with default memcache config
### Description
When setting `'cacheMethod' => 'memcache'` in the `general.php` config file, the page fails with the error: 

```
Invalid argument supplied for foreach()
```

After digging around the code for a bit I found that [`setServers`](https://github.com/yiisoft/yii2/blob/master/framework/caching/MemCache.php#L255) expects an array of servers to be passed.

The current memcache default config that comes with craftcms3 is not an array of servers, but a single server declaration.

### Steps to reproduce

1. Set `'cacheMethod' => 'memcache'` in `general.php
2. Open a page

### Additional info

- Craft version: 3.0.0-beta.19
- PHP version: 7.0.19

kbrabrand:
This PR fixes #1751 by changing the default memcache config file from containing a single server to an array of servers.

### DIFF
--- a/src/config/MemCacheConfig.php
+++ b/src/config/MemCacheConfig.php
@@ -24,19 +24,21 @@ class MemCacheConfig extends Object
      * @var array An array of memcached servers to use.
      */
     public $servers = [
-        // A memcached server hostname or IP address.
-        'host' => 'localhost',
-        // Whether or not to use a persistent connection.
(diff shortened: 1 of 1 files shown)


takobell:
Thanks, @kbrabrand!

PR #1774, opened 2017-06-14
allow more flexibility when using database urls and postgresql

rkingon:


src/config/DbConfig.php
```diff
@@ -121,7 +121,12 @@ public function init()
         if ($this->url !== null) {
             $url = parse_url($this->url);
             if (isset($url['scheme'])) {
-                $this->driver = $url['scheme'];
+                $scheme = strtolower($url['scheme']);
+                if ($scheme === "pgsql" || strpos($scheme, "postgres") !== false) {
+                    $this->driver = self::DRIVER_PGSQL;
+                } else {
```
rkingon: potentially better off with a `else if ($scheme === "mysql")`

brandonkelly:
What’s an example DB connection URL that requires the added flexibility?

rkingon:
heroku gives the following:
`postgres://user:password@host/database`

brandonkelly:
👍 Thanks

rkingon:
the standard for postgres url's is actually `postgresql://user:password@host/database` so i used strpos so it handles `postgres|postgresql`

PR #1776, opened 2017-06-15
Improved unclear German translation

rthbrst:
German translation for permission 'Perform Craft CMS and plugin updates' was unclear as it read like this permission was required in order to access the control panel. It has been updated in order to make it clear that this permission refers to updates only.


takobell:
Thanks, @rthbrst!

PR #1782, opened 2017-06-20
returning wrong yesterday date

ejsoft:


takobell:
Can you provide a valid input value where this is failing for you? I'm not able to see any issues with the current implementation.

PR #1784, opened 2017-06-21
Add hyphen to German phrase "Dev-Modus"

dennisfrank:
It’s common in Germany to join connected words with a hyphen.


takobell:
Thanks!

PR #1808, opened 2017-06-30
Fix deprecation error

mikefrancis:
In my install of `3.0.0-beta.19` I get the following deprecation warning:

```DateTime::uiTimestamp() is deprecated. Use the |timestamp('short') filter instead.	/home/vagrant/Code/pod-point/intranet/vendor/craftcms/cms/src/helpers/Template.php (267)```

So this should fix.

### DIFF
--- a/src/helpers/Template.php
+++ b/src/helpers/Template.php
@@ -260,7 +260,12 @@ class Template
             }
         }
 
-        $key = "DateTime::{$item}()";
+        if ($item === 'uiTimestamp') {
+            $key = timestamp('short');
+        } else {
+            $key = "DateTime::{$item}()";
+        }
+        
         /** @noinspection PhpUndefinedVariableInspection */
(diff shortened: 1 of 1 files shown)


brandonkelly:
The line you changed is just the Template helper preparing a key for the deprecation log; it is not the actual offender – which is more likely something in your templates (or Craft's or a plugin's templates).

I just made some changes to the Deprecator service that should make it more clear which template is actually at fault (233c7c6a7fd9bacea167f9606a4fac19d0ec91b9) - it hasn't been too reliable to date.

mikefrancis:
@brandonkelly Ah ok. Just FYI this was on a fresh install of the beta with no templates or plugins

brandonkelly:
@mikefrancis Yeah not ruling out that it could be within Craft's templates. The next update should make it easier to identify where.

PR #1816, opened 2017-07-06
Update Task.php

LAV45:
In place of having to create `NestedSetsTrait` you can simply specify `NestedSetsBehavior` as an `@mixin`

### DIFF
--- a/src/records/Task.php
+++ b/src/records/Task.php
@@ -9,7 +9,6 @@ namespace craft\records;
 
 use Craft;
 use craft\db\ActiveRecord;
-use craft\db\NestedSetsTrait;
 use craft\db\TaskQuery;
 use creocoder\nestedsets\NestedSetsBehavior;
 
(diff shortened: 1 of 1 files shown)


brandonkelly:
No way! That's awesome. Can you please make the same changes to src/records/StructureElement.php, and delete src/db/NestedSetsTrait.php as well?

LAV45:
This issue is present with these files:
* [FieldLayoutTrait](https://github.com/craftcms/cms/blob/master/src/behaviors/FieldLayoutTrait.php)
* [NestedSetsQueryTrait](https://github.com/craftcms/cms/blob/master/src/db/NestedSetsQueryTrait.php)
* [NestedSetsTrait](https://github.com/craftcms/cms/blob/master/src/db/NestedSetsTrait.php)

Sorry, but I just wanted to leave a recommendation.
It's not a critical error and speed is not affected.

brandonkelly:
OK thanks anyway. There were a few other traits that could also be discarded. Resolved via c2c83bafc1e62a5eff233e614b62be9a376c923e.

PR #1856, opened 2017-07-21
Init events

markhuot:
This is copied from Laravel (fair warning). I use this feature constantly in Laravel when adding application specific or domain specific methods to core classes. For example, Laravel ships with a `Collection` class that will do almost anything, but if I want to sort an array of users by a business specific mapping, I can add the following,

```php
Collection::macro('myWeirdBusinessSpecificSort', function() {…
```

With that, I can call `$users->myWeirdBusinessSpecificSort()` and get back my sorted collection.

The alternative is to extend the base `Collection` with my own subclass, but too many internal items create Collections that I'd be constantly chasing `new Collection` calls to force my custom Collection.

One area I think this would be very helpful in Craft is with the out-of-the-box fields.

I'm currently working on a GraphQL plugin for Craft and it would be great if a field could `$field->getGraphQLDefinition()`. This way Craft would not have to ship with this method (and I would never expect it to) but my plugin can monkey patch it in. The implementation would look like this (inside my `Plugin::Init` method,

```php
\craft\fields\RichText::macro('getGraphQLDefinition', function() {
  return [
    'type' => Type::string(),
    'resolve' => function ($root, $args) {
      if (!empty($args['page'])) {
        return $root->{$this->handle}->getPage($args['page']);
      }

      return $root->{$this->handle}->getParsedContent();
    }
  ];
});
}
```

I could see this used in a number of similar ways:

- a plugin that wants to add additional config to a field could use `$field->getCustomConfig()`
- a plugin that wants to use the built in RichText field, but provide additional formatting options could use `$field->stripAllTagsExceptLinks()`

Finally, because this trait overrides the `__call` magic method all existing calls should continue to work as is. Additionally, if Craft decides to bring a macro in house, at some point, the class' native method will be called before the monkey patched version will.

Note: this potentially overwrites the `yii\base\Component::__call` method. I haven't noticed any issues in my testing, but am still looking for unwanted side effects.


michaelrog:
I'm a big fan of this idea. Would love to be able to monkey-patch methods on Element Queries as well.

brandonkelly:
Interesting… kindof similar to Yii [Behaviors](http://www.yiiframework.com/doc-2.0/guide-concept-behaviors.html) (which are essentially traits that are injected at runtime), except registered at the class level.

Before we go down this road, I’d like to see what the Yii guys think about introducing class-level behaviors, which would accomplish the same thing.

markhuot:
Ah, I should have figured their was a Yii native way to do this.

brandonkelly:
@michaelrog Yeah I didn’t bring that up because it’s happening at the instance level rather than the class level, so won’t help if you don’t already have a reference to the field/query/whatever.

michaelrog:
Yes, but could Craft expose the component instantiation process to plugins somehow? Maybe:
- Applicable component classes trigger an `afterComponentInit` event in their `init()` method, and pass `$this` as a param.
- I listen to see if the thing being instantiated is an instance of anything that I care about, and if so, now I have a reference to dynamically attach a behavior.

markhuot:
So I gave this some thought and was curious how a `Behavior` based approach would work. I borrowed the idea from @michaelrog and added an event to `\craft\base\Component` so that plugins could hook in to the `init()` method as necessary. This is now the only change necessary to support my use case above. This is a fairly major change for how short the diff is, so I understand any trepidation with it. Full disclosure, I have not tested the performance of this.

With this in place it is now possible in for a plugin to hook in to Craft's component lifecycle. It makes the implementation of custom behaviors feel very similar to how plugin routing is working in Craft 3.

For example, to add a GraphQL specific behavior to a `\craft\fields\RichText` instance I would do the following,

```php
use craft\fields\RichText;

class Plugin extends BasePlugin {
  function init() {
    // e.g., routing event hook
    Event::on(UrlManager::className(), UrlManager::EVENT_REGISTER_SITE_URL_RULES, function (RegisterUrlRulesEvent $event) use ($uri) {
      $event->rules["..."] = '...';
    });

    // hooking into the new component event, add a behavior
    Event::on(RichText::className(), RichText::EVENT_AFTER_INIT, function ($event) {
      $event->sender->attachBehavior('richTextBehavior', MyRichTextBehavior::class);
    });
  }
}
```

After the above, Yii's native `Behavior` functionality takes over from there and allows me to, later in my code, call `$richTextInstance->someBehaviorMethod();`.

brandonkelly:
Issue with this approach is that it’s limited to Component + subclasses, so wouldn't help if you wanted to add new Query methods, etc.. Wonder if the Yii folks would be interested in adding this event to yii\base\Object…

michaelrog:
Adding an event to _craft\db\Query_ would get the job done for now. But yeah, I think some standard lifecycle events on Yii's base class would be lovely.

markhuot:
I opened a PR for some discussion on adding this as a core Yii feature over here: https://github.com/yiisoft/yii2/pull/14545

markhuot:
My PR was closed as a duplicate of https://github.com/yiisoft/yii2/pull/8133, which avoids lifecycle events due to the performance concerns and, I have to assume, potential for abuse.

That said, it seems that Yii widgets [support an `init()` event](https://github.com/yiisoft/yii2/commit/7c1693479fbc8441d1565958942aeb7490872cb6) so… maybe the suggestion is to _not_ do it at the (global) component level. Instead, make targeted decisions, at an application level, about what interfaces should offer an `init()` event.

I suppose it's up to the Craft team if this is something they want to support, then.

brandonkelly:
I like it. Couple things that should change along with this:

- `craft\base\ApplicationTrait::_init()` fires an `afterInit` event. Probably should be renamed to `init` for consistency.

- `craft\web\twig\variables\CraftVariable::init()` currently fires `defineComponents` and `defineBehaviors` events. Thinking those should go away now in favor of an `init` event, so registering new components/behaviors is consistant across the Application, Component, and CraftVariable classes.

I’ll merge this into a feature branch to work on those changes.

PR #1860, opened 2017-07-21
Send detailed json error on ajax registration

S1lentium:
allows to client parse more error data on public registration


brandonkelly:
Thanks!

PR #1873, opened 2017-07-26
data-type attributes for matrix blocks and fields

swey:
It's nearly impossible for plugins to style matrix blocks and fields. For that reason `data-type` was added to the those blocks and all fields. (The matrix buttons already have a `data-type` attribute but the added blocks have no reference to that type/handle)

For example the change enables stylings like:
```
.matrixblock[data-type="quote"] .field[data-type="position"] {
	background: lightgreen;
	padding: 10px;
}
```

### DIFF
--- a/src/templates/_components/fieldtypes/Matrix/input.html
+++ b/src/templates/_components/fieldtypes/Matrix/input.html
@@ -10,7 +10,7 @@
                 {% set blockId = 'new'~totalNewBlocks %}
             {% endif %}
 
-            <div class="matrixblock{% if not block.enabled %} disabled{% endif %}" data-id="{{ blockId }}"{% if block.collapsed %} data-collapsed{% endif %}>
+            <div class="matrixblock{% if not block.enabled %} disabled{% endif %}" data-id="{{ blockId }}"{% if block.collapsed %} data-collapsed{% endif %} data-type="{{ block.getType().handle }}">
                 {% if not static %}
                     <input type="hidden" name="{{ name }}[{{ blockId }}][type]" value="{{ block.getType().handle }}">
                     <input type="hidden" name="{{ name }}[{{ blockId }}][enabled]" value="{% if block.enabled %}1{% endif %}">
--- a/src/templates/_includes/field.html
+++ b/src/templates/_includes/field.html
@@ -22,6 +22,7 @@
         instructions: instructions|e,
         id: field.handle,
         errors: errors,
-        input: input
+        input: input,
+        handle: field.handle
     } only %}
 {% endif %}
--- a/src/templates/_includes/field.html
+++ b/src/templates/_includes/field.html
@@ -22,7 +22,6 @@
         instructions: instructions|e,
         id: field.handle,
         errors: errors,
-        input: input,
-        handle: field.handle
+        input: input
     } only %}
 {% endif %}
--- a/src/templates/_includes/forms/field.html
+++ b/src/templates/_includes/forms/field.html
@@ -15,7 +15,7 @@
     (fieldClass is defined and fieldClass ? fieldClass : null)
 ]|filter|join(' ') -%}
 
-<div class="{{ fieldClass }}"{% if fieldId %} id="{{ fieldId }}"{% endif %}>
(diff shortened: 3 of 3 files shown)


brandonkelly:
For fields, why not just use the `#field-id`, which is already there and based on the `id` variable?

swey:
Hi @brandonkelly 
For *fields* there is a workaround for few cases. The example above could look like:
```
.field[id^="fields-articleBody-"][id$="-position-field"]
```
But two problems with that solution:
1. Because that `id` is always different, the selector is quite dirty
2. No way to address only `position` inside `quote`. In the demo's `articleBody` it would also address `position` inside `text` and `image`


If `data-type` will be added to the *block types*, the selector would look like:
```
.matrixblock[data-type="quote"] .field[id^="fields-articleBody-"][id$="-position-field"]
```
That would work, but is still quite dirty.
Thank you!

brandonkelly:
Maybe change that to `{% if fieldId %} data-id="{{ fieldId }}"{% endif %}` then.

swey:
But that doesn't help. To stay with the example from above.

Your solution would result in:
bildschirmfoto 2017-07-28 um 06 10 21

My solution results in:
bildschirmfoto 2017-07-28 um 06 10 57


If you don't like it here, I can make a new PR only for the matrix block types? That part is ok for you?

brandonkelly:
Are you sure about that? Just did a test and my code resulted in 

```html
<div class="field" id="fields-articleBody-205-fields-position-field" data-id="position-field">
```

brandonkelly:
(The `id` attribute will get namespaced, but `data-id` won't.)

swey:
Well, I have that version running on my local machine and it's working as expected.

Also check the definition of the used `id`:
https://github.com/craftcms/cms/blob/develop/src/templates/_includes/field.html#L23

brandonkelly:
"working as expected" meaning you got the namespaced result, or what I pasted?

swey:
Because `id` ist set to `handle`, only the `handle` is in my data attribute:
https://github.com/craftcms/cms/blob/develop/src/templates/_includes/field.html#L23

bildschirmfoto 2017-07-28 um 06 10 57

Have you tested it with my branch?

brandonkelly:
I don’t want to call it `data-type` because the ID isn’t a type. Is your only issue that `fieldId` gets suffixed with `-field`? `fieldId` is the correct variable to use here, as `id` refers to the ID of the actual input that is passed into the field template, but here we are putting this on the field container, not the input inside it.

swey:
- I called it `data-type` because the attribute is `data-type` in similar cases. But ok, the similar case is in the context of block types
- An `id` is always unique, like the attribute `id`. It makes no sense to have an `id` and a `data-id` with two different values.
- Probably `data-handle` is the best name. But so far we did not discuss the naming of the attribute, we discussed the reason for that extra attribute
- As said before: The main problem is that there is NO WAY to have a *block type* selector. The thing we are discussing is an additional change. So that's the smaller problem. But again: Even this change makes sense because at the moment you can't have a concrete selector and have to play again with `$=`

swey:
Once again: There is no possible way to have a CSS selector for **block types**. But I will stay with my fork or add the needed attributes with JS.

brandonkelly:
> I called it `data-type` because the attribute is `data-type` in similar cases. But ok, the similar case is in the context of block types

I think I see where you’re coming from now. From my perspective, this is not a similar case though, because this is a generic template that is used for pretty much every field in the Control Panel (not just custom fields), so `id` is rarely going to be a field handle. Maybe a better way to go would be to allow things that are rendering that `_includes/forms/field` template to define custom `data` attributes. Then we can take advantage of that when loading the template to display a custom field, and give it a `data-handle` attribute set to the field’s handle.

> An `id` is always unique, like the attribute `id`. It makes no sense to have an `id` and a `data-id` with two different values.

Technically, the `id` attribute will become unique thanks to namespacing that is applied to the template render output, but the `data-id` attribute would not. So they are different values, and `data-id` would give you the CSS hook you are looking for. But from a semantics perspective I agree with you here – it is awkward at the least.

> As said before: The main problem is that there is NO WAY to have a **block type** selector. The thing we are discussing is an additional change. So that's the smaller problem. But again: Even this change makes sense because at the moment you can't have a concrete selector and have to play again with `$=`

To be clear, I’m fine with `data-type` attribute you proposed for the Matrix block template. Only debating the `data-type` attribute on the general field template.

PR #1895, opened 2017-08-02
Render text body prior to markdown conversion

timkelty:
When using `craft\mail\Mailer::composeFromKey`, suppose you have a `SystemMessage` with a body like so:

```
{% for entry in entries %}
- {{ entry.link }}
{% endfor %}
```
`craft\mail\Mailer::send` takes care of parsing that as markdown for your HTML email, however, it does this _before_ any Twig parsing takes place.

This change would pre-parse the text template, so that when it parses as Markdown, the input would be as expected.


brandonkelly:
Took this in a different direction, but the end result should work for you.

timkelty:
👍  Thanks for listening!

PR #1915, opened 2017-08-14
added data-type attributes for matrix blocks

swey:
It's nearly impossible for plugins to style matrix blocks. For that reason `data-type` was added for those blocks.

**Why that naming?**
The buttons to add matrix blocks have the same `data-type` attribute. It's a consistent naming.

**Why not using an `id`?**
It's a possible way to go even it requires more complicated selectors in CSS/JS. But the current block ids used in `data-id` are numbers and have no reference to the block's type/handle. The whole naming scheme needed to be changed.

**Any other alternatives?**
My favorite solution is to add an `data-handle` for matrix block types, the buttons for them and all other fields as well. That's clean, enables simple CSS/JS selectors and is a reasonable usage of data attributes. But the last time similar changes weren't welcome.

### DIFF
--- a/src/templates/_components/fieldtypes/Matrix/input.html
+++ b/src/templates/_components/fieldtypes/Matrix/input.html
@@ -10,7 +10,7 @@
                 {% set blockId = 'new'~totalNewBlocks %}
             {% endif %}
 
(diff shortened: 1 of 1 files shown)


brandonkelly:
Thanks!

PR #1919, opened 2017-08-15
Avoid null as return for getMimeType() -> TypeError

MDXDave:
Make sure getMimeType does not return null for some filetypes, because this throws an TypeError.
For example "img" extension returns null.

> Return value of craft\elements\Asset::getMimeType() must be of the type string, null returned

### DIFF
--- a/src/elements/Asset.php
+++ b/src/elements/Asset.php
@@ -717,7 +717,8 @@ class Asset extends Element
     {
         // todo: maybe we should be passing this off to volume types
         // so Local volumes can call FileHelper::getMimeType() (uses magic file instead of ext)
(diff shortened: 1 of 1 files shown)


brandonkelly:
Seems like it would be better to just remove the `:string` return type declaration here, no? Why assume that it's `application/octet-stream` ?

MDXDave:
Mh it depends on the use case. I use the getMimeType-function for a download plugin, so `application/octet-stream` is the correct content type for downloading files with unkown extension. If the function can return ``null``, you have to check the return type by yourself.

brandonkelly:
This SE answer backs that up: https://stackoverflow.com/a/1176031/1688568

Thoughts @takobell @andris-sevcenko ?

PR #1925, opened 2017-08-17
added the data-type attribute for matrixblocks created with JS

swey:
The same `data-type` attribute as in #1915 was added for matrixblocks created with JS.

**Reason**
Newly added matrixblocks behaved different because they didn't have the same attributes.
Without this addition a custom styled matrixblock works perfect, but that styling doesn't take any effect for newly added blocks until they are saved.


brandonkelly:
Whoops, didn’t even think about that!

PR #1937, opened 2017-08-22
allowing more granular configurations for 'appType'

nateiler:
For example, this would allow a console app to specify it's own configuration independently of web app or rest app (ahem).

### DIFF
--- a/bootstrap/bootstrap.php
+++ b/bootstrap/bootstrap.php
@@ -209,7 +209,8 @@ $config = ArrayHelper::merge(
     ],
     require "{$srcPath}/config/app/main.php",
     require "{$srcPath}/config/app/{$appType}.php",
-    $configService->getConfigFromFile('app')
+    $configService->getConfigFromFile('app'),
+    $configService->getConfigFromFile(empty($appConfig) ? $appType : $appConfig)
 );
 
 if (defined('CRAFT_SITE') || defined('CRAFT_LOCALE')) {


nateiler:
Wanted to get some thoughts around this.  The idea is`config/app.php` is a global configuration override yet allow for more granular overrides.  This is particularly helpful when configuring a REST app which commonly requires different component configurations than the main Craft web app.

brandonkelly:
I don’t really want to support alternative `appType` values, since they are meant to represent the two ways PHP could be invoked (via web server or CLI); not necessarily how the request should behave.

That said, I know you have a legit reason for needing to provide REST request-specific configs and you need to use a different Application class, so I’ve just made that possible with 27fa2e7f09ba7d551033946c4a76271162952fed.

You will still need to define your main application config via `config/app.php`, but that file could itself just load a more specific config file depending on request type. For example if you have a REST-specific entrypoint PHP file, you could define a PHP constant from it:

```php
define('REST_REQUEST', 1);
```

And then from `config/app.php`:

```php
if (defined('REST_REQUEST')) {
    return require(__DIR__.'/rest.php');
}

// ...
```

And now, that config file could override the actual Application class like this:

```php
return [
    'class' => custom\rest\Application::class,
    // ...
];
```

narration-sd:
❤️ for approach, 👍 for method...

nateiler:
Thanks Brandon.  I think this will work perfectly...although we’re not needing to define a separate application any longer, it’s nice to know it’s available.

PR #1940, opened 2017-08-23
Fixed and cleaned up maxLevel check

swey:
E.g. the following was `true` for all structures even if `maxLevels` is set to `1` because the value is a `string`.
```
if (
    $section->type === Section::TYPE_STRUCTURE &&
    $section->maxLevels !== 1
)
```
The problem of wrong types seems not to be new because in other cases the string values were handled before - see changes.

Maybe the better solution is a cast inside the models, but all other numeric model values like `id`s are strings as well. Maybe a Yii/PDO problem, but the database table columns have the right type (`int`, `smallint`) (MySQL)

### DIFF
--- a/src/controllers/CategoriesController.php
+++ b/src/controllers/CategoriesController.php
@@ -270,7 +270,7 @@ class CategoriesController extends Controller
         // Parent Category selector variables
         // ---------------------------------------------------------------------
 
-        if ($variables['group']->maxLevels != 1) {
+        if ((int)$variables['group']->maxLevels !== 1) {
             $variables['elementType'] = Category::class;
 
             // Define the parent options criteria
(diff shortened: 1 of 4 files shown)


src/services/Sections.php
```diff
@@ -410,7 +410,7 @@ public function saveSection(Section $section, bool $runValidation = true): bool
                 }
 
                 // If they've set maxLevels to 0 (don't ask why), then pretend like there are none.
-                if ($section->maxLevels === 0 || $section->maxLevels === '0') {
```
swey: In this case the string value was handled before

src/controllers/CategoriesController.php
```diff
@@ -270,7 +270,7 @@ public function actionEditCategory(string $groupHandle, int $categoryId = null,
         // Parent Category selector variables
         // ---------------------------------------------------------------------
 
-        if ($variables['group']->maxLevels != 1) {
```
swey: In thise case the non-strict operator made the comparison work

brandonkelly:
Thanks!

PR #1947, opened 2017-08-24
Add block.getType().handle to matrixblock class

leevigraham:
Adds a new class as a styling hook for third party addons.

It would be great if this was back ported to v2.

### DIFF
--- a/src/templates/_components/fieldtypes/Matrix/input.html
+++ b/src/templates/_components/fieldtypes/Matrix/input.html
@@ -10,7 +10,7 @@
                 {% set blockId = 'new'~totalNewBlocks %}
             {% endif %}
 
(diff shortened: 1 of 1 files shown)


brandonkelly:
This is a duplicate of #1915.

leevigraham:
@brandonkelly, Any chance of getting this backported to v2? I guess we can target styling via an attribute selector. A class would be more obvious tho.

brandonkelly:
Yeah can backport to v2. Seems a little sketchy to add a class based on a user-defined handle, even if it had a `type-` prefix or something, so I think I prefer the `data-type` attribute.

brandonkelly:
Done for the next v2 release: ca9499251f94c6d81bf64f6203a30706828fdb69

PR #1965, opened 2017-09-04
data-type attributes for fields

swey:
Because there is no way to style field inputs by `type` at all, a `data-type` attribute was added for those.  With this addition it's possible to have selectors like `.field[data-type$="RadioButtons"]`

Any improvement suggestions?

### DIFF
--- a/src/base/Field.php
+++ b/src/base/Field.php
@@ -139,6 +139,8 @@ abstract class Field extends SavableComponent implements FieldInterface
         if ($this->translationMethod !== self::TRANSLATION_METHOD_CUSTOM) {
             $this->translationKeyFormat = null;
         }
+
+        $this->type = get_class($this);
     }
 
     /**
--- a/src/base/FieldTrait.php
+++ b/src/base/FieldTrait.php
@@ -54,6 +54,11 @@ trait FieldTrait
     public $translationKeyFormat;
 
     /**
+     * @var string|null The field’s type
+     */
+    public $type;
+
+    /**
      * @var string|null The field’s previous handle
      */
     public $oldHandle;
--- a/src/templates/_includes/field.html
+++ b/src/templates/_includes/field.html
@@ -21,6 +21,7 @@
         required: (not static ? required : false),
         instructions: instructions|e,
         id: field.handle,
+        type: field.type,
         errors: errors,
         input: input
     } only %}
--- a/src/templates/_includes/forms/field.html
+++ b/src/templates/_includes/forms/field.html
@@ -15,7 +15,7 @@
     (fieldClass is defined and fieldClass ? fieldClass : null)
 ]|filter|join(' ') -%}
 
-<div class="{{ fieldClass }}"{% if fieldId %} id="{{ fieldId }}"{% endif %}>
+<div class="{{ fieldClass }}"{% if fieldId %} id="{{ fieldId }}"{% endif %}{% if type is defined %} data-type="{{ type }}"{% endif %}>
     {% if label or instructions %}
         <div class="heading">
             {% if label %}


brandonkelly:
What’s the use case? It’s already possible to target field types by giving their input HTML a wrapper `div` or something – so the only place I could see this maybe being helpful is if you need to style a field type input that you don’t control (e.g. radio buttons in your example). But why?

swey:
We wan't to style it, because we want to improve the user experience in the control panel.

Radio buttons are a good example, because we want them side by side:
```
.field[data-type$="RadioButtons"] {
		.input {
			display: flex;
			flex-wrap: wrap;

			> div {
				margin-right: 10px;
			}
		}
	}
```

Also we want to have more fields side by side, but keep the full width for some types:
```
.matrixblock > .fields,
#tab1 {
	display: flex;
	flex-wrap: wrap;
	margin-right: -10px;
	margin-top: -24px;

	.field {
		flex: 1 1 250px;
		width: calc(50% - 10px);
		margin-right: 10px;
		margin-top: 24px !important;
	}

	.field[data-type$="RichText"],
	.field[data-type$="Matrix"] {
		flex: 0 0 auto;
		width: calc(100% - 10px);
	}
}
```

(Please note that the CSS above is SCSS.)

At the moment the `type` which is saved in the database or outputted in the field's type select on the "settings page" is not available at all in the "input pages".

brandonkelly:
Gotcha. I’m not a fan of the specific implementation, because that `_includes/forms/field` template is used for all fields throughout the Control Panel – not just custom fields. It would be weird if custom radio group fields had a `data-type="RadioButtons"` attribute, but any hard-coded radio groups didn’t get one. So instead, I’ve just given _all_ radio groups a `div.radio-group` container element (and `div.checkbox-group` for checkbox groups).

swey:
Don't get it. You select that `type` in a select dropdown, you save it in the database, but you don't want to output that value. With a `.radio-group`/`.checkbox-group` you can only handle one single case, but many other fields still have no class at all and can't be controlled via CSS.

(But Craft also doesn't output important information in other areas. For example the page for creating entries doesn't have a information of the section the entry is created in.)

Craft is a great CMS and thanks for your great work! But it's just not possible to customize/improve the user interface...

swey:
We really love the concepts of Craft and thank you for your great work! But we want to improve the user interface and have a hard time doing it because the most html tags have no classes and important information is not available at all.

One example is the missing field `type`, which is selected by the user in the type dropdown and also saved in the database. Another example is the missing section `handle` on the entries page. Would be really helpful to have data attributes with the handles and the types wherever possible.

With a `.radio-group`/`.checkbox-group` you can only handle one single case, but many other fields still have no class at all and can't be controlled via CSS.

> It would be weird if custom radio group fields had a data-type="RadioButtons" attribute, but any hard-coded radio groups didn’t get one. 

Would be perfect, because `data-type` is exactly the field type which is selected and saved in the database. I expect `data-type="craft\fields\RadioButtons`  only for fields with that type and not for other fields which may have radio buttons included.

So I don't want to control radio buttons or checkboxes, I want to control fields by type.

swey:
Screenshot of the `type` inside the option values:
bildschirmfoto 2017-09-07 um 12 52 11

Screenshot of the `type` saved in the database:
bildschirmfoto 2017-09-07 um 12 54 08

swey:
@brandonkelly Any feedback on my response?

brandonkelly:
OK, I get where you’re coming from. Just committed 2abba7399030e9b790b5a2cbc43bbfbb09e2ce46 which should resolve this for you - a little differently than you proposed but should still work.

swey:
@brandonkelly Thank you for your commit 2abba73. It's not exactly the same because the data attribute it not on the outer node, but that I can do that now with JS. It helps a lot the `type` is rendered now 👍

brandonkelly:
Yeah I’m not really interested in changing the outer `.field` element, as that is meant to be an agnostic wrapper for any sort of field. Doesn’t really make sense to give it type-specific attributes IMO.

PR #1970, opened 2017-09-06
Divide By Zero Bug

markhuot:
I'm not sure if this is actually a bug but wanted to propose it. When I use the `{% paginate` tag or try to call `\craft\helpers\Template::paginateCriteria` manually with a query that returns 0 rows I get a "divide by zero" error. This seems to be because when the limit is not set it is assigned the total.

This all means if no limit is set, and the query returns zero results you'll end up with the following,

```php
$totalPages = (int)ceil(0 / 0);
```

This PR adds a check early on that avoids this and bails out with an empty pagination object if the total results are 0.

### DIFF
--- a/src/helpers/Template.php
+++ b/src/helpers/Template.php
@@ -89,7 +89,7 @@ class Template
 
         // Get the total result count, without applying the limit
         $query->limit = null;
-        $total = $query->count();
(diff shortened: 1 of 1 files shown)


brandonkelly:
Thanks!

PR #1973, opened 2017-09-08
Plugin Installer

markhuot:
Currently my CI needs to install a plugin to run unit tests. I have a cludgy script to do that from the console,

https://github.com/markhuot/craftql/blob/c11c8689bbe83468003a82cc95df8b8f11a36644/bin/setup/install.php#L1-L12

This abomination gets called with `php path/to/awful.php`

It'd be nice if this script could be dropped for a native CLI plugin installer. This does just that.


brandonkelly:
Good idea, thanks!

PR #1981, opened 2017-09-11
added a possibility to define the HTML input type of plain texts

swey:
Added a possibility to define the HTML input type of plain texts because of:
- Support of browser validation for the entered values
- Improved keyboard layouts based on input type on touch devices

## Settings screenshot:
bildschirmfoto 2017-09-11 um 19 46 17

## Settings screenshot with opened dropdown:
bildschirmfoto 2017-09-11 um 19 46 21

## Additional information:
The other HTML5 input types (like date, time or color) are missing because they have their own Craft field types.

### DIFF
--- a/src/fields/PlainText.php
+++ b/src/fields/PlainText.php
@@ -57,6 +57,11 @@ class PlainText extends Field implements PreviewableFieldInterface
     public $charLimit;
 
     /**
+     * @var string The HTML5 input type (text, email, tel, url)
+     */
+    public $inputType;
+
+    /**
      * @var string The type of database column the field should have in the content table
      */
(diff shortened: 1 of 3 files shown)


brandonkelly:
Good idea, but decided that having dedicated field types would be better, so they can have back-end validation, and so their values on element indexes can be hyperlinked.

I didn’t add a Phone Number field type, since that would require something like [this library](https://github.com/giggsey/libphonenumber-for-php) to do the back-end validation, and at that point I think it would be better to do it as a plugin, so only people who actually need to store phone numbers need to add the additional dependencies.

swey:
Ok, thanks for the adding and also for your explanation!

PR #1985, opened 2017-09-14
Check if the parentId from the POST request is an array

kbrabrand:
Checks if the `parentId` request param value is an array and uses the first value if it is. Seems like it always is an array of one element anyway.


brandonkelly:
Decided to fix it slightly differently, and it also had to fix a related issue with categories. So already taken care of. Thanks though!!

kbrabrand:
Cool. Just glad to be able to let the customer that the fix will be put soon. Do you have a plan for beta-27, @brandonkelly?

brandonkelly:
@kbrabrand It's out now :)

PR #1986, opened 2017-09-15
Feature/field type refactoring

swey:
Thanks again for your commit 2abba73, @brandonkelly!

Maybe even better is to wrap the whole field with the `data-type` attribute. This allows full control of the fields by `handle` and by `type`.

Useful for cases where you want to improve the styling for whole fields, e.g. highlight the headline of a certain field type or hide the headline because the field input has several ones.

Of course I don't want to make changes of Craft's core just for our use cases, but I have to because the fields miss important information and there is no other way to add them. But I think this is also useful for many other developers creating custom fields or heavily customizing the control panel interface.

### DIFF
--- a/src/templates/_includes/field.html
+++ b/src/templates/_includes/field.html
@@ -14,14 +14,16 @@
 {% endif %}
 
 {% if instructions or input %}
-    {% include "_includes/forms/field" with {
-        label: field.name|t('site')|e,
-        translatable: translatable,
-        siteId: siteId,
-        required: (not static ? required : false),
-        instructions: instructions|e,
-        id: field.handle,
-        errors: errors,
-        input: '<div data-type="'~className(field)~'">'~input~'</div>'
-    } only %}
+    <div class="field-wrapper" data-type="{{ className(field) }}">
+        {% include "_includes/forms/field" with {
+            label: field.name|t('site')|e,
+            translatable: translatable,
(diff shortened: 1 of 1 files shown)


brandonkelly:
Guessing this will break Live Preview. Have you tested that?

swey:
Have not tested the live preview so far. Sorry. 
Will do it later the day.

brandonkelly:
Just thought of a way to output the exact HTML you were hoping for in #1965, but in a way that I’m happy with: At the `_includes/forms/field` level, it only cares about a generic `data` object that may be defined. If it is, it will create `data-x` attributes on the `<div class="field">` based on its keys/values. So it’s still agnostic to what those attributes might be for, which is important to me, but gives the `_includes/field` template a way to add a `data-type` attribute where you want it to go.

swey:
Yeah, understand what you mean.
So you call the partial with
```
additionalDataAttributes: {
 type:  className(field)
}
```

and inside `includes/forms/_fields`
you have something like
```
<div id=""{{additionalDataAttributes|dataAttributes}}>
```

Right?

brandonkelly:
Already done: 4c9b714fd4f7334d642d9b3b1f3204ffcbc7dc6f

swey:
Awesome, you're the best, @brandonkelly!

PR #2020, opened 2017-10-05
Stub getIsLivePreview for console requests

drifteaur:
I'm importing data into a Craft 3 site via a console command. I'm populating matrix fields, and they check getIsLivePreview(), but that method isn't available on the Console Request. This stub method helps.

`yii\base\UnknownMethodException: Calling unknown method: craft\console\Request::getIsLivePreview() yii\base\Component->__call()  (../vendor/yiisoft/yii2/base/Component.php line 290).`

### DIFF
--- a/src/console/Request.php
+++ b/src/console/Request.php
@@ -54,4 +54,14 @@ class Request extends \yii\console\Request
     {
         return false;
     }
+	
+    /**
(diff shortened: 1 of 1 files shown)


brandonkelly:
Thanks!

PR #2023, opened 2017-10-06
Add support for PGPASSFILE with pg_dump

khalwat:
Add support for PGPASSFILE with pg_dump


takobell:
Went with https://github.com/craftcms/cms/commit/242d6cd4564550ff3e36388f6972795937167580 which accounts for Windows as well.

PR #2027, opened 2017-10-10
Return the errors in the warning message

SteveEdson:
Rather than just logging that the validation failed, show the validation errors, to help with debugging.


brandonkelly:
Thanks!

PR #2033, opened 2017-10-12
Bringing LICENSE.md into accordance with markdownlint

rzumstein:
Changed all headings to use ATX style consistently, per MD003.  While the `<a>` tags do technically violate MD033 for being inline html, there is not a pure markdown way to open links in new tabs/windows, so I left them.


brandonkelly:
Thanks! (Though that license isn’t long for this world…)

PR #2042, opened 2017-10-18
Fix crash when URL rules are added by Yii modules.

jcherniak:
Most UrlRules will be an instance of \craft\web\UrlRules, but if rules are inserted via a plugin (module), they might be a \yii\web\UrlRule, which doesn't have the params property, causing a crash.

### DIFF
--- a/src/web/UrlManager.php
+++ b/src/web/UrlManager.php
@@ -348,7 +348,7 @@ class UrlManager extends \yii\web\UrlManager
             }
 
             if ($route !== false) {
-                if ($rule->params) {
+                if ($rule instanceof \craft\web\UrlRule && $rule->params) {
                     $this->setRouteParams($rule->params);
                 }
 


jcherniak:
For more backstory, if you want to be able to access the debug module directly (say to debug a plugin while not logged into Craft), this will be triggered.

Steps to reproduce:
1. Add the following to web/index.php to get the debug panel while anonymous (after `$app` is defined, before calling `$app->run()`):
```
$app = require $root.'/vendor/craftcms/cms/bootstrap/web.php';

if (YII_DEBUG) {
    $session = $app->getSession();
    $session->set('enableDebugToolbarForSite', true);
    $session->set('enableDebugToolbarForCp', true);
}

$app->run();
```
2. Visit /debug and see the following error:

> Unknown Property – yii\base\UnknownPropertyException
> Getting unknown property: yii\web\UrlRule::params

brandonkelly:
Thanks!

PR #2066, opened 2017-10-27
Better handling of template UnknownMethodExceptions

mildlygeeky:
Right now if you encounter an UnknownMethodException in a template, you are only given the compiled template as the source of the error, and it is difficult to trace back to the source of the original issue.

This would throw a Twig_Error_Runtime instead (as it looks like it is supposed to?), which will give the developer better information of the source of the exception.

There may be a bette way to handle this, or issues I'm not foreseeing that this causes, but this has helped us on debugging template issues.

### DIFF
--- a/src/helpers/Template.php
+++ b/src/helpers/Template.php
@@ -15,6 +15,7 @@ use craft\elements\db\ElementQueryInterface;
 use craft\i18n\Locale;
 use craft\web\twig\variables\Paginate;
 use yii\base\Object;
+use Twig_Error_Runtime;
 
 /**
  * Class Template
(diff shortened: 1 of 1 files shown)


brandonkelly:
Thanks! Implemented manually to make a slight change – now only actually throwing the exception when Dev Mode is enabled, which is in-line with how Twig will react if a non-Yii-Object-subclass’s `__call()` method is invoked and throws a BadMethodCallException.

mildlygeeky:
Very nice! Thanks for implementing - we were converting over a Nunjucks-based static site we had over to C3, and so there were lots of these issues as we hadn’t yet fleshed out the fields, and made finding the source in Twig pretty rough. Should be a good help with twig dev.

PR #2076, opened 2017-10-31
reused GetAssetUrlEvent for thumbs - resolves #2073

### LINKED ISSUE #2073: Feature Request: Possibility to overwrite Thumb Url
I like the solution to overwrite asset urls via `GetAssetUrlEvent`, but miss the possibility to do that for thumbnails. Would be really awesome to have it here as well.

See the asset url solution: https://github.com/craftcms/cms/blob/develop/src/services/Assets.php#L547-L557

Thank you!

swey:
Reused the `GetAssetUrlEvent` for thumbs, which adds the possibility to overwrite the thumb urls as well. See #2073


brandonkelly:
Thanks for submitting this! However now that I look at it, it’s clear that a separate event will be necessary, probably named `GetAssetThumbUrlEvent`, so that it can have the same properties as the `getThumbUrl()` method arguments (`$asset`, `$size`, `$generate`).

Issue #1588, opened 2017-04-01
Change focal point without re-saving image


carlcs:
It would be nice to set or change the focal point without re-saving image and consequently losing quality by adding new compression artifacts.

swey:
@andris-sevcenko Still no totally solved. The image files still will be replaced if only a focal point changes.

Why this is a problem to us:
-We have very big files, the saving is delayed even the image itself is not changed
- The file timestamps change
- We use Cloudinary as storage, which means we need unnecessary operations and the "changed" image has to be re-calculcated in many transformations

andris-sevcenko:
Fixed via  aee8f07

PR #2089, opened 2017-11-04
Optimized saving of images where just the focal point was changed

swey:
If only the focal point did change, the file does not be to be replaced.
Fully solved #1588 

Future improvement requests:
- `actionSaveImage` should be fully refactored and `imageChanged` should be checked before `$imageCopy` is created and all possible changes are checked one by one
- The UI could offer a simple "Save" button as long the image itself wasn't changed. No reason to offer the buttons "Replace Asset" and "Save as New Asset"


brandonkelly:
Duplicate of #1588

brandonkelly:
> The UI could offer a simple "Save" button as long the image itself wasn't changed. No reason to offer the buttons "Replace Asset" and "Save as New Asset"

Agree; just renamed it to “Save” regardless of whether the image was modified, as that works in all cases. (df1910b1cbbe10b98b9f7230eca294984dbf99c3)

andris-sevcenko:
Thanks, @swey!

PR #2101, opened 2017-11-10
update some japanese translation.

torounit:


narration-sd:
@torounit Hiroshi, I think Andris forgets his politeness on this one, though I think he is just being humorful about the depth of knowledge to understand Japanese. 

What I am sure of is that we all appreciate your adding to Craft's usefulness this way.  And there sre German and Canadian developers I know who can use it, so not only from your islands...

takobell:
Thank you, @torounit!

PR #2113, opened 2017-11-17
Modifying Assets phpdoc to properly explain upload location properties

rzumstein:
I noticed these two properties are not int when I was writing a content migration.


brandonkelly:
@andris-sevcenko are these edits accurate?

andris-sevcenko:
Yes indeed. We changed the way element sources are referred to in Craft 3.

PR #2178, opened 2017-12-07
The docsurl property is removed

johanzandstra:
This property is removed from the src/templates/_layouts/cp.html template and useless now.


brandonkelly:
Thanks!

PR #2255, opened 2018-01-01
Add new GetEntryEditTemplate event

narration-sd:
This is essential for a plugin functionality you are aware of.

It allows a plugin to decide a substitute template to render, when Craft is editing an Entry.

If the triggered event doesn't change the template Craft specifies, that template will be used for rendering as normal.

### DIFF
--- a/src/controllers/EntriesController.php
+++ b/src/controllers/EntriesController.php
@@ -13,6 +13,7 @@ use craft\base\Field;
 use craft\elements\Entry;
 use craft\elements\User;
 use craft\errors\InvalidElementException;
+use craft\events\GetEntryEditTemplateEvent;
 use craft\helpers\DateTimeHelper;
 use craft\helpers\Json;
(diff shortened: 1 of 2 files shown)


brandonkelly:
That method isn’t for rendering an entry’s _edit_ template – it’s for the entry’s front-end user-defined template. So at the least the name of the event needs to change.

I’m not sure a plugin should be able to edit the template though. Why not just instruct admins to change the section’s entry template from the settings?

narration-sd:
Hmm. Defiitely see your point and will come up with a better name, but in my case the first reason it was put in there, some while ago, does have to do with edits. There are by now other reasons, closer to your thought...

I'll write you an explanation just now on support, and thanks.

narration-sd:
Closing this in favor of Add new ShowEntryTemplateChoice event #2259, which answers the issue of naming vs. intent. Further explanation why this ability is needed is covered in a related support conversation.

PR #2256, opened 2018-01-01
Add new GetRequestRoute event

narration-sd:
This is essential for a plugin functionality you are aware of.

It allows a plugin to provide its own routing for a given Request.

In the plugin this is managed by a nice UX including an enable checkbox per route, convenient to the use cases. 

This event allows bypassing the order of Craft's normal routing decisions, which does not suit plugin abilities -- in particular, the forced routing based on element type.

CP routes are protected from engaging the new Event, as this isn't required and could be confusing.

If the triggered event doesn't specify a required route, Craft's normal routing behaviour will be followed.


brandonkelly:
> This event allows bypassing the order of Craft's normal routing decisions, which does not suit plugin abilities -- in particular, the forced routing based on element type.

Elements have a [EVENT_SET_ROUTE](https://docs.craftcms.com/api/v3/craft-base-element.html#EVENT_SET_ROUTE-detail) event, that will let you override the matched element’s route. Guessing that would work for you?

```php
use yii\base\Event;
use craft\elements\Entry;
use craft\base\Element;
use craft\events\SetElementRouteEvent;

Event::on(Entry::class, Element::EVENT_SET_ROUTE, function(SetElementRouteEvent $e) {
    /** @var Entry $entry */
    $entry = $e->sender;

    $e->route = // ...
});
```

narration-sd:
Hmm, sounds that it would work, and if so not sure how I missed it, having tried events I did find.

Same story on getting the test rig back together, and will let you know how this one goes.

Thanks!

narration-sd:
First look after the tester back together -- it doesn't  appear this event is getting fired. I left my pr event in, and that does continue to work.

Have to leave it for now, but I guess I'll have to go back later and  trace through from the bootstrap to discover the path being taken for a domain-only request. 

Perhaps this unexpected miss occurs because there's not a home page?? Will look into that also.

Really, there are so many cases for route decisioning,  hence my heading some of them off with this original pull request.

narration-sd:
Ok, that's the problem.  Short note to you on Support.

brandonkelly:
That event will be fired when an element’s URI is requested, right before the element type decides how to route the request.

narration-sd:
Well, the report on Support??  Because it isn't actually fired in times I need it, explained there...

narration-sd:
Ok, want to close this one cleanly out.

Per discussions on support thread, a combination of two events solves the actual need, without the original PR here, validated.

Thanks to @brandonkelly and @angrybrad for patience in discussing some ins and outs.

PR #2259, opened 2018-01-03
Add new ShowEntryTemplateChoice event

narration-sd:
This is to replace Add new GetEntryEditTemplate event #2255, which will be closed, as I can't  mutate it.

This naming aligns with the usual intents of the routine it alters in web/UrlManager, 

Comments have also been improved.

This event allows a plugin to decide a substitute template to render, when Craft is showing an Entry, as is required for a somewhat unusual but useful Craft 3 plugin near completion and offering -- explanation is in a related support conversation.

If the triggered event doesn't change the template that Crafthas specified, that template will be used for rendering as normal.


brandonkelly:
So we just talked this over, and decided to make the event a bit more generic. It now gets fired from `actionPreviewEntry()` rather than `_showEntry()`, so it doesn’t also get fired from `entries/view-shared-entry` requests.

Here’s how you can use the new event to customize the entry template:

```php
use yii\base\Event;
use craft\controllers\EntriesController;
use craft\events\ElementEvent;

Event::on(EntriesController::class, EntriesController::EVENT_PREVIEW_ENTRY, function(ElementEvent $e) {
    /** @var Entry $entry */
    $entry = $e->element;
    $settings = $entry->getSection()->getSiteSettings()[$entry->siteId] ?? null;
    if ($settings) {
        $settings->template  = // ...
    }
});
```

narration-sd:
Hey Brandon, had a quick look, and seems that should work a treat  -- thanks!  Do as you know quite appreciate the desire to corral extent of effects, as I've tried to on the partner PR of this one.

I've got the build system for what uses this torn apart, for the better verified late last night, and as soon as it's rebuilt, will give this a verifcation.

Def appreciated

narration-sd:
Ok, as promised, on a dark and stormy night, got this in and tested, I think.

In fact once code warped around to fit my context, it almost works too well :)

After a lot of other things past days which were adequately twisty....

I'll have further looking, if I can think of any other way to make it seem less solid, but looking pretty good on this one, Brandon.  

A good idea...

brandonkelly:
Thanks for the confirmation!

PR #2271, opened 2018-01-04
Feature/multiselect styles

alexmarchant:
Just adds a few lines of css to use the same font-size, border, border-radius, and focus styles as input fields.

Before:
image

After:
image

After (focus):
image


alexmarchant:
One note: can't seem to style the padding on <options> in safari, not sure why, still looks better there tho too even without the padding IMO:

image

brandonkelly:
Looks great, thanks!

PR #2277, opened 2018-01-06
Fixed ColorData.php bug

simondriesen:
Fixed small bug that causes `{{ myColor.blue }}`to return wrong value.


brandonkelly:
Thanks!

Issue #892, opened 2017-01-31
Tabs to separate user profile content


takobell:
> Created by: Chase Giunta (me@chasegiunta.com) on 2015/03/30 19:07:54 +0000
> Votes at time of UserVoice import: 15

-----------------------------------

Just like entries, there needs to be tabs in the user profile to separate content. This is especially useful for community driven sites, This also is somewhat of a workaround for the lack of locale-specific fields.

johanneslamers:
Yes please. Strange this isn't core yet.

lindseydiloreto:
Working on a project where Users will have around 30 profile fields... Splitting this into tabs would be _hugely_ helpful.

takobell:
Users are currently the only place where field layouts are used that already have some pre-existing tabs (Account/Profile (where you can an avatar by default, but other custom fields can show up)/Permissions).

So for this to be addressed, it needs to be part of the larger field layout refactoring we want to do.

yellostudio:
We've needed this from day one :)

bertoost:
Still nothing on this. This is so needed now with more than 30 fields per user :$

brandonkelly:
@bertoost Interesting idea. Related to #947 but at a tab level rather than field level. Can you please post as a new issue?

bertoost:
@brandonkelly I have easily managed to get custom tabs for users. I will make a PR for that.
I will create a new issue for the permission idea 👍

bertoost:
Created the update for this issue. Branched from the 2.6.2992 branche. How to merge now? Can't merge on develop, since this is 3.0 already.
https://github.com/bertoost/cms/tree/feature/custom-user-field-tabs

PR #2297, opened 2018-01-11
#892 Ability to have user profile fields separated by custom tabs in field layout

bertoost:
And included fallback to default as it was before. For those who are satisfied with the default tructure ;-)

### DIFF
--- a/src/templates/users/_edit.html
+++ b/src/templates/users/_edit.html
@@ -157,25 +157,38 @@
 				</div>
 
 				{% if not isNewAccount or CraftEdition == CraftPro %}
-					<div id="profile" class="hidden">
-						{% if not isNewAccount %}
-							{{ forms.field({
-								label: "Photo"|t
-							}, photoInput) }}
+					{% set customLayout = craft.fields.getLayoutByType('User') %}
+					{% for tab in customLayout.tabs %}
+						<div id="custom-{{ tab.id }}" class="hidden">
 
-							<hr>
-						{% endif %}
+							{# show profile image on first tab #}
+							{% if not isNewAccount and loop.first %}
+								{{ forms.field({ label: "Photo"|t }, photoInput) }}
+								<hr>
+							{% endif %}
 
-						{% if CraftEdition == CraftPro %}
 							{% include "_includes/fields" with {
-								fields: craft.fields.getLayoutByType('User').getFields(),
+								fields: tab.getFields(),
 								element: account
 							} only %}
+						</div>
+					{% else %}
+						<div id="profile" class="hidden">
+							{% if not isNewAccount %}
+								{{ forms.field({ label: "Photo"|t }, photoInput) }}
+								<hr>
+							{% endif %}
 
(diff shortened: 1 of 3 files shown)


src/templates/users/_edit.html
```diff
@@ -157,25 +157,38 @@
...
-						{% endif %}
+					{% set customLayout = craft.fields.getLayoutByType('User') %}
+					{% for tab in customLayout.tabs %}
+						<div id="custom-{{ tab.id }}" class="hidden">
+
+							{# show profile image on first tab #}
+							{% if not isNewAccount and loop.first %}
+								{{ forms.field({ label: "Photo"|t }, photoInput) }}
+								<hr>
+							{% endif %}
```
brandonkelly: By putting the Photo field here, it’s only going to show up if the user field layout actually has any tabs (it might not). In Craft 3 we moved the Photo field [over to the Account tab](https://github.com/craftcms/cms/blob/f2617ae5a6218329ceb42db2586500814437eafc/src/templates/users/_edit.html#L60-L64), after the First/Last Name fields. Can you move it there for this PR as well? (The following Photo field starting on line 177 can be deleted as well, in that case.)

bertoost: No. Look at the `{% else %}` part on line 174 below. That's the fallback part where the photo url is still there if you have no tabs configured. When you do, it's added on the first tab you added.

bertoost: Photo url either is part of the account. It's a additional thing to the user (in my opinion)

brandonkelly: Ah I get it, I assumed the `{% else %}` was for the `{% if %}` on line 159, not the `{% for %}`.

Still, I would prefer to have this be consistent with Craft 3, rather than create a Profile tab if there are no custom field layout tabs.

bertoost: I do understand that. But maybe you should reconsider placing first and last name fields too. Since they are not necessary for the account to login. The account-details are username/email and password and the others are profile-data :-) but if you really really really want that, I will change it.

brandonkelly: Eventually we will be making the placement of _all_ user fields (including Username, First/Last Name, Photo, custom fields, etc.) customizable from the Field Layout Manager, and at that point it will be up to the admin to decide how they want to organize everything. Until then, I’m OK with it the "Account" tab label being slightly un-semantic.

bertoost: Okay. I will move the photo field then and remove the support for the profile tab when you not have any tabs configured. Will update it a.s.a.p.
Btw. permissions for tabs and fields would be cool too then :) and different setup for different types of users and usergroups.

bertoost: @brandonkelly done! :-)

brandonkelly: And you tested right?

bertoost: Of course I did. Check screenshots:

image

image

PR #2298, opened 2018-01-11
Added 'email.onBeforeSendMailer' event just before the email is actually being send

bertoost:
Brandon,

Since Yii (or Craft) doesn't support priorities for events, hooking in on the email system to send emails via another provider is not that easy in Craft 2.
For multiple customer projects I had to make something to work with Mandrill API. I have managed this, but with a downside of the 'email.onBeforeSendEmail' event being fired twice (to catch all plugins their possible additions).
More details on the projects README;
https://bitbucket.org/bertoost/craft-cms-mandrill-service#take-over-control-of-all-outgoing-emails

That's why I added this event, just before the `$email->Send()` is being called. This way the email service could collect all additional data on the previous event, and another plugin can hook-in just before sending the actual email.

A little extra addition to make this happen.

### DIFF
--- a/src/services/EmailService.php
+++ b/src/services/EmailService.php
@@ -220,6 +220,16 @@ class EmailService extends BaseApplicationComponent
 	}
 
 	/**
+	 * Fires an 'onBeforeSendMailer' event.
+	 *
+	 * @param Event $event
+	 */
+	public function onBeforeSendMailer(Event $event)
+	{
+		$this->raiseEvent('onBeforeSendMailer', $event);
+	}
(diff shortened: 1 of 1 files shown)


bertoost:
Thanks Brandon for your answer. Totally agree and I created a solution for v2. Since upgrading is not that easy for custom projects with a lot of custom features. This was only a work around for sending all system emails via Mandrill. Better than invoking “onBeforeSendEmail” twice.
And of course I like the v3 approach with transporters :-)

PR #2306, opened 2018-01-14
Fixed using CRAFT_FRAMEWORK_PATH for including framework files in console file

bertoost:
For dev installations with vendor files etc.

This way you can create a "console" PHP file in your dev installation root, the same way as for the web index.php file, setting some constants.

```php
#!/usr/bin/env php
<?php

define('CRAFT_BASE_PATH', __DIR__ . '/craft/');
define('CRAFT_VENDOR_PATH', dirname(__DIR__) . '/craft_git/vendor/');
define('CRAFT_FRAMEWORK_PATH', CRAFT_VENDOR_PATH . 'yiisoft/yii/framework/');

require CRAFT_BASE_PATH . 'app/etc/console/yiic.php';
```

### DIFF
--- a/src/etc/console/yiic.php
+++ b/src/etc/console/yiic.php
@@ -59,7 +59,7 @@ ini_set('display_errors', 1);
 defined('YII_DEBUG') || define('YII_DEBUG', true);
 defined('YII_TRACE_LEVEL') || define('YII_TRACE_LEVEL', 3);
 
-require_once dirname(__FILE__).'/../../framework/yii.php';
+require_once CRAFT_FRAMEWORK_PATH.'yii.php';
 require_once CRAFT_APP_PATH.'Craft.php';
 require_once CRAFT_APP_PATH.'Info.php';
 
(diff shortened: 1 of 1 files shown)


brandonkelly:
Thanks!

bertoost:
Hmm now trying (for the first time) to use migration command, getting this error;

```
Fatal error: Class 'MigrateCommand' not found in /path/to/my/craft/checkout/src/consolecommands/MigrateCommand.php on line 50
```

Probably he doesn't know about Yii's Migrate Command where Craft's version extends from. Any idea?

brandonkelly:
@bertoost Is Craft symlinked in your project?

bertoost:
Yes it is. Since my develop installation uses the Git checkout (for cases if I need to change core stuff ;-))

brandonkelly:
Yeah that’s probably the issue. You may need to set additional constants to help Craft bootstrap itself. Been a while since I dealt with it though so don’t remember exactly what to do.

bertoost:
Did not yet figured it out.. Decided to change the app folder again from a symlinked-git version to a normal one and everything works again. Hopefully you can remember it some day ;-)

PR #2342, opened 2018-01-24
Throw error if calling migrate with type='plugin' but not plugin handle

jcherniak:
It's not inherintly clear from the docs or errors provided that you need to specify a plugin handle when calling "craft migrate --type=plugin".  This exception clarifies that (as I would think it would iterate all plugins if --plugin is blank).

### DIFF
--- a/src/console/controllers/MigrateController.php
+++ b/src/console/controllers/MigrateController.php
@@ -131,6 +131,8 @@ class MigrateController extends BaseMigrateController
                     throw new Exception('Invalid plugin handle: '.$this->plugin);
                 }
                 $this->plugin = $plugin;
+            } else {
+                throw new Exception('You must pass --plugin when --type="plugin"');
             }
         }
 


brandonkelly:
Thanks for pointing that out. Took a different approach to the fix, but it's better now:

screen shot 2018-01-24 at 11 43 42 am

jcherniak:
Is there any way to update all plugins at once using the console?  We are run on multiple environments with continuous integration, so EVERYTHING is handled by migrations (learning how to create new Matrices has been a _fun_ challenge).  We got ```craft migrate --type=app``` to do the app migrations and ```craft migrate ``` to do content migrations.  

So, following up on how I stumbled on this bug, is there an easy way to migrate all plugins at once?

brandonkelly:
Yep, run `migrate/all`. That will run app, content, and all plugin migrations.

PR #2356, opened 2018-01-26
Added getFocalPointPercentage

sjelfull:
Added a way to easily get the focal point as percentages.

Unsure if this should be `getFocalPoint($percentage = true/false)`, `getFocalPointPercentage()` or `return [ 'x' => 0.5, 'y' => 0.5, 'xPercentage' => 50, 'yPercentage' => 50 ]` on `getFocalPoint()`

### DIFF
--- a/src/elements/Asset.php
+++ b/src/elements/Asset.php
@@ -923,20 +923,20 @@ class Asset extends Element
      *
      * @return array|null
      */
-    public function getFocalPointPercentage ($includePercentSign = false)
+    public function getFocalPointPercentage($includePercentSign = false)
     {
         $focalPoint = $this->getFocalPoint();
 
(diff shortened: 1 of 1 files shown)


sjelfull:
Done.

brandonkelly:
After doing some refactoring to make `focalPoint` and `getFocalPoint()` consistent (ba6cafa7), I ended up just adding a `$asCss` argument to `getFocalPoint()` which returns the focal point in the CSS syntax. Think that is good enough – if you don’t need _exactly_ that, it’s easy enough to just do whatever formatting you need from the raw `x`/`y` values.

brandonkelly:
Yeah, they are converted to floats in setFocalPoint().

PR #2366, opened 2018-01-27
Better queue logs

ostark:
Before only "GLOBALS" and sql queries got logged to `queue.log` - I guess that was not intended. 

Now we log something that is very similar to the `craft queue/run -v` output.
If `devMode` is enabled sql queries are logged as well, but "GLOBALS" are still suppressed.

### DIFF
--- a/src/controllers/QueueController.php
+++ b/src/controllers/QueueController.php
@@ -10,7 +10,6 @@ namespace craft\controllers;
 use Craft;
 use craft\queue\QueueInterface;
 use craft\web\Controller;
-use yii\log\FileTarget;
 use yii\web\BadRequestHttpException;
 use yii\web\Response;
 use yii\web\ServerErrorHttpException;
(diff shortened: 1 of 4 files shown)


brandonkelly:
@ostark The current behavior should be that it logs anything sent with `Craft::info()`, `Craft::warning()`, and `Craft::error()` to queue.log, if it’s a queue request. Are you saying that’s not the case?

ostark:
It does. But the job description, start, end (+duration) is missing.

brandonkelly:
Gotcha, yeah agree the `[profile]` logs shouldn’t be in there, and it’s nice to have the console output.

ostark:
Thanks @brandonkelly.  I've applied your suggestions. Plus:

* All possible log targets (e.g. StdErr) have the same verbosity
* All system infos are excluded now
* Events work with `controllers/QueueController` as well (before it didn't)

PR #2373, opened 2018-01-30
Update broken links

hankchizljaw:
Firstly, thanks so much for Craft. It's such a good CMS 🙂

I imagine this will be a temporary change, but when I went to go through the upgrade docs, I hit a 404. I've just replaced the `master` segment with `v3` so the links work again.


hankchizljaw:
Not a problem! Thanks for the continued hard work 🙂

PR #2380, opened 2018-01-31
Update Dutch translation for "asset"

nilsenpaul:
Throughout Craft, the Dutch word 'Materiaal' is used wherever the English noun 'Asset' is shown. 'Materiaal' is a word that, although it CAN mean 'the material needed to fill a site', doesn't quite fit here. In my opinion, 'Bestand' would be better, and the plural 'Bestanden' is already used in the navigation sidebar. 'Bestand' literally means 'File'.


brandonkelly:
Thanks!

PR #2412, opened 2018-02-07
Fixed users save button on every tab (related to #892)

bertoost:
Discovered that the save button is not appearing on the custom user field tabs. Changed the button location from inside a tab-content-div to below those divs and before the closing form tag.


brandonkelly:
Doh, I missed this PR and fixed the same bug yesterday (541c102b). Thanks anyway!

bertoost:
Cool! haha

PR #2436, opened 2018-02-14
Use a better referrer strategy

Zae:
Use the `origin-when-cross-origin` referrer strategy

This keeps referrers working internally (for redirecting back for example), but only sends the main domain to external sites.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referrer-Policy


Zae:
The no-referer strategy broke some stuff in our site because we use the referrer to redirect users back to the page after our work was done.

brandonkelly:
Ah yeah that would work well. The main reason we added a referrer policy is so 3rd party sites can’t find out what your CP trigger is.

brandonkelly:
Doh – I totally overlooked that this was a pull request. Thanks!

Zae:
Yeah, that's what I thought which is a good idea!

Thanks for fixing so quickly, when do you expect the next tag will be?

brandonkelly:
Tuesday

PR #2469, opened 2018-02-21
Add `id="photo-field"` to container div

lindseydiloreto:
This would make it much easier for custom CSS (or JS) to target the photo div.

    <div class="field" id="photo-field">

Useful in scenarios where the developer may want to hide the photo field.

### DIFF
--- a/src/templates/users/_edit.html
+++ b/src/templates/users/_edit.html
@@ -58,7 +58,8 @@
 
             {% if not isNewUser %}
                 {{ forms.field({
-                    label: "Photo"|t('app')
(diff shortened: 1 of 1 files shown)


ryanmasuga:
Upvoted this because we would be that developer. On a Craft 2 site, we're using a normal Asset field for a user profile photo (so we can focus point, etc) and have no need for the default. Hiding it would be easier with this id. 

We'll also want to hide the little default profile photo thumbnails next to the email addresses in the user lists, but that's a different problem and I think we can do that with the styles that are there already.

PR #2484, opened 2018-02-23
Fixed bug where getFieldParamNamespace returns wrong namespace

andrelopez:
@brandonkelly I found a bug when  `getFieldParamNamespace()` is called from a custom field and the field is inside a Matrix field. Returns  an extra dot inside the namespace 
``` php
public function normalizeValue($value, ElementInterface $element = null)
    {
        $namespace = $element->getFieldParamNamespace();
        ...
        ...
```
Best


brandonkelly:
Thanks!

Issue #2496, opened 2018-02-26
EmailVerifyUrl and PasswordResetUrl generated from current site?


milanjandric:
### Description
Should URLs generated for EmailVerify email and PasswordReset email be to the current site, or to a primary site? 

I have been expecting to be returned to the same site I have used to fill in a "Forgotten password" form when clicked on a link in an email, and not to the "Primary" one.

### Additional info
- Craft version:Craft Pro 3.0.0-beta.29
- PHP version: 7.0.27

brandonkelly:
Currently it is expected that the URLs would be for the primary site. If you have a use case where that isn’t desirable, can you post it as a new feature request, and explain the reason?

PR #2497, opened 2018-02-26
Use current site as a return url from emails.

milanjandric:
fixes: #2496


brandonkelly:
This would be a breaking change for all sites that are relying on the current behavior.

If we make it possible to point these URLs to the current site, there needs to be a FR issue about it first explaining the use case, and it will need to be implemented with a config setting so existing sites don’t break.

PR #2505, opened 2018-02-27
Improve german translation for 'Eingabe'

rostockahoi:
There has been a wrong translation of entry to 'Eingabe' where it actually should be 'Eintrag' in multiple places. 'Eingabe' is the noun of 'eingeben' which means enter sth. into sth.


brandonkelly:
Thanks!

PR #2514, opened 2018-03-01
Renamed the urlModel to UrlModel

roelvanhintum:
Caused a Class 'Craft\UrlModel' not found (FeedsService.php:108)


brandonkelly:
Thank you!

Issue #2567, opened 2018-03-13
PGPASSWORD needs quotes around it


alexmarchant:
### Description

I'm getting an error when trying to run database backups:

```
2018-03-13 14:43:59 [71.167.189.10][1][-][error][yii\base\Exception] yii\base\Exception: Could not create backup: The shell command "PGPASSWORD=******* pg_dump --dbname=photoism_staging --host=localhost --port=5432 --username=craft --if-exists --clean --file="/srv/www/photoism/releases/20180312220800/site/storage/backups/photoism_180313_214359_2k54a8huxa_v3.0.0-rc14.sql" --schema=public --exclude-table-data 'public.assetindexdata' --exclude-table-data 'public.assettransformindex' --exclude-table-data 'public.cache' --exclude-table-data 'public.sessions' --exclude-table-data 'public.templatecaches' --exclude-table-data 'public.templatecachecriteria' --exclude-table-data 'public.templatecacheelements'" failed with exit code 1: pg_dump: [archiver (db)] connection to database "photoism_staging" failed: FATAL:  password authentication failed for user "craft"
FATAL:  password authentication failed for user "craft" in /srv/www/photoism/releases/20180312220800/site/vendor/craftcms/cms/src/controllers/UtilitiesController.php:380
Stack trace:
#0 [internal function]: craft\controllers\UtilitiesController->actionDbBackupPerformAction()
#1 /srv/www/photoism/releases/20180312220800/site/vendor/yiisoft/yii2/base/InlineAction.php(57): call_user_func_array(Array, Array)
#2 /srv/www/photoism/releases/20180312220800/site/vendor/yiisoft/yii2/base/Controller.php(157): yii\base\InlineAction->runWithParams(Array)
#3 /srv/www/photoism/releases/20180312220800/site/vendor/craftcms/cms/src/web/Controller.php(74): yii\base\Controller->runAction('db-backup-perfo...', Array)
#4 /srv/www/photoism/releases/20180312220800/site/vendor/yiisoft/yii2/base/Module.php(528): craft\web\Controller->runAction('db-backup-perfo...', Array)
#5 /srv/www/photoism/releases/20180312220800/site/vendor/craftcms/cms/src/web/Application.php(237): yii\base\Module->runAction('utilities/db-ba...', Array)
#6 /srv/www/photoism/releases/20180312220800/site/vendor/craftcms/cms/src/web/Application.php(445): craft\web\Application->runAction('utilities/db-ba...', Array)
[...]

PR #2568, opened 2018-03-13
Add quotes to PGPASSWORD for database backups

alexmarchant:
Fixes https://github.com/craftcms/cms/issues/2567


angrybrad:
Looks like we had a similar fix for MySQL back in RC12 that we didn't port to the PostgreSQL schema. https://github.com/craftcms/cms/commit/99fc27155a98382b8ff5db7b0892fb317880d261

Would probably want to do something similar in case someone actually has a `"` in their password, no?

alexmarchant:
Smart... hows that?

angrybrad:
Perfect... thanks, @alexmarchant!

PR #2580, opened 2018-03-15
Document disabled color

carlcs:


brandonkelly:
While that technically works, disabled isn’t a color, and if the status is disabled then there’s no need to set `color` in the first place.

carlcs:
I thought it would be nice to have all options documented. `Element::STATUS_DISABLED` is not the the only status where a white or transparent color can make sense.

brandonkelly:
In that case we should give it an alias, like `white`.

PR #2614, opened 2018-03-21
Sort plugins by name

carlcs:
They are currently sorted by ID in the sidebar and settings page.


carlcs:
Should the names be translated first?

Issue #2617, opened 2018-03-21
EVENT_REGISTER_SITE_TEMPLATE_ROOTS doesn't work on site requests


BenParizek:
### Description

It seems `EVENT_REGISTER_SITE_TEMPLATE_ROOTS` currently makes certain assumptions that apply to the CP that are different on the front-end.

Example implementation code:

```php
Event::on(View::class, View::EVENT_REGISTER_SITE_TEMPLATE_ROOTS, function(RegisterTemplateRootsEvent $event) {
    $myTemplateRoot = Craft::getAlias('@barrelstrength/sproutforms/templates/_special/templates');

    $event->roots = [
        'sprout-forms' => $formTemplateRoot
    ];
});
```

### Steps to reproduce

1. Add a `EVENT_REGISTER_SITE_TEMPLATE_ROOTS` to a plugin 
2. Try to load a page with a template that only exists in your plugin

### Additional info

- Craft version: RC15
- PHP version: 7.1
- Database driver & version: MySQL
- Plugins & versions: Lots under development

BenParizek:
I've proposed a pull request above, that can at least illustrate what I think the issue may be.

It may be worth considering if the same syntax is desirable for front-end and backend requests.

It appears back-end requests require an array with the `plugin-handle` to help determine the sub-templates.

```
$event->roots = [
    'sprout-forms' => $basepath
];
```

However, on the front-end this may be meaningless info, as we can't assume that any particular folder structure is being respected, and a simple array may be a more expected pattern:

```
$event->roots[] = $basepath;
```

The pull request addresses the existing problem that the current syntax doesn't seem to work. However, it may make sense to update to this new syntax at some point.

brandonkelly:
> 2. Try to load a page with a template that only exists in your plugin

How are you trying to load that page? Going with your example, Craft will only check your template root if the page URI begins with `sprout-form/...`.

BenParizek:
Hmm, maybe I just need documentation here. I've actually decided on another route for this at the moment. After I got deeper into the potential direction, other potential issues revealed themselves. So maybe it's best to back away from this request until I understand what was supposed to be happening and see proper docs.

To give you an idea of what I was thinking, I was trying to make another template folder accessible on all requests as a fallback, and worry about avoiding conflicts in the plugin itself. So, from what you are saying, I think I wanted something like:

```
$event->roots = [
    '' => $basepath
];
```

The idea was to have something like this in the template:

```
[...]

PR #2618, opened 2018-03-21
Fixes bug where EVENT_REGISTER_SITE_TEMPLATE_ROOTS did not work on site requests #2617

BenParizek:


brandonkelly:
Site requests are the _only_ place that event will work, and it already works as intended in our testing. Can you clarify what the bug is, and how you’re solving it?

PR #2623, opened 2018-03-21
Color field default value validation

### LINKED ISSUE #2588: Colour Field type default value validation
### Description

It appears as though you cannot set a default colour that contains letters. I am able to save #000000 for example, but not #FFFFFF. I can however save whatever valid hex I like in the entry where the field is being used.

### Steps to reproduce

1. Create a colour field
2. Set a default hex that contains letters
3. Try to save the field

### Additional info

- Craft version: 3.0.0-RC15

Radabaugh:
Fixes #2588


brandonkelly:
There were other inconsistencies between Default Color setting validation and normal Color field inputs. So I moved all normalization logic to `ColorValidator::normalizeColor()`, and now `craft\fields\Color::normalizeValue()` _and_ `ColorValidator::validateAttribute()` run the value through that.

PR #2625, opened 2018-03-21
Fixed #2612

### LINKED ISSUE #2612: No validation on edit of Number field
### Description
Number field validation does not work when _editing_ a Number field.

### Steps to reproduce

1. Create a Number field and save it.
2. Edit the newly created Number field and enter a string in 'Min' - it causes a Yii error: `Argument 1 passed to craft\helpers\Db::getNumericalColumnType() must be of the type integer or null, string given, called in .../vendor/craftcms/cms/src/fields/Number.php on line 124`

*note* validation while creating a new Number field works as expected. It's only when you edit the field and try to save an invalid value where it fails to validate.

### Additional info

- Craft version: 3.0.0-RC15
- PHP version: PHP 7.1.14
- Database driver & version:
- Plugins & versions:

Radabaugh:
Fixes #2612

### DIFF
--- a/src/fields/Number.php
+++ b/src/fields/Number.php
@@ -121,6 +121,9 @@ class Number extends Field implements PreviewableFieldInterface
      */
     public function getContentColumnType(): string
     {
+        if (!is_numeric($this->min) || !is_numeric($this->max) || !is_numeric($this->decimals)) {
+            return false;
+        }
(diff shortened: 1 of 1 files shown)


brandonkelly:
Fixed via 0ec7014f

PR #2636, opened 2018-03-26
Monospace setting for plain text fields

carlcs:
Would be nice to have this as a setting, or merge the Simple Text plugin into core and use it in places where multiline code input is used, e.g. Email Messages.


brandonkelly:
Thanks!

PR #2637, opened 2018-03-26
Add beforeSend event (Fix #2634)

### LINKED ISSUE #2634: Add before send email message event
We have a scenario where we'd like to check a blacklist prior to sending to a recipient.  We have identified bounced email addresses and would like to suppress subsequent sends... and all we need is an event prior to send which will:

1) Not send if event is not valid.
2) Access to the Message object (in order to remove one of many recipients).

We could replace the entire component, but this might be useful for others a well.

Rias500:
My attempt at adding a beforeSend event to the mailer

### DIFF
--- a/src/events/MailMessageEvent.php
+++ b/src/events/MailMessageEvent.php
@@ -1,0 +1,27 @@
+<?php
+/**
+ * @link https://craftcms.com/
+ * @copyright Copyright (c) Pixel & Tonic, Inc.
+ * @license https://craftcms.github.io/license/
+ */
+
+namespace craft\events;
+
+use craft\mail\Message;
+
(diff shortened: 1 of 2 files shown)


brandonkelly:
There’s already a `beforeSend` event on the `mailer` component: https://www.yiiframework.com/doc/api/2.0/yii-mail-basemailer#EVENT_BEFORE_SEND-detail

```php
use yii\base\Event;
use yii\mail\BaseMailer;
use yii\mail\MailEvent;

Event::on(BaseMailer::class, BaseMailer::EVENT_BEFORE_SEND, function(MailEvent $e) {
    // ...
});
```

PR #2649, opened 2018-03-28
The class is defined as urlModel

rbroen:
The class is defined as urlModel, not UrlModel.

### DIFF
--- a/src/services/FeedsService.php
+++ b/src/services/FeedsService.php
@@ -105,7 +105,7 @@ class FeedsService extends BaseApplicationComponent
 
 			if ($permalink)
 			{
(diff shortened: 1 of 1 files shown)


brandonkelly:
We fixed the class name in Craft 2.6.3013, via 6031fbda24bbd53fe30266c1bd000b02930da873. Have you updated to that version yet?

rbroen:
The change was intended for "craftcms:master-v2", the change is intended for FeedsService.php. 

Should I create a new pull request targeted at "craftcms:master-v2"?

brandonkelly:
Ahhh... `master-v1` is just out of date. We’ll fix that.

PR #2673, opened 2018-04-04
Correct 'Singles' Dutch translation

Wiejeben:
After discussing this change in the official Dutch Craft Slack channel, we decided 'Eenmalig' doesn't fit the context and can be found confusing.


angrybrad:
Thanks!

PR #2688, opened 2018-04-05
Added an event when saving in an ElementEditor

joshangell:
This allows any other cp JavaScript to listen to this event and therefore react when an element is saved from an ElementEditor.

I have made this change against Craft 2, but it should work on both.

### DIFF
--- a/src/resources/js/craft.js
+++ b/src/resources/js/craft.js
@@ -10361,6 +10361,9 @@ Craft.ElementEditor = Garnish.Base.extend(
 					{
 						Craft.livePreview.updateIframe(true);
 					}
+					
+					// Trigger an event so that other js can react to this action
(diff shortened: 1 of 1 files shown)


brandonkelly:
That’s not the right way to go about it

- change would need to be made on ElementEditor.js, while `grunt watch` is running
- pretty sure you can’t have `.`s in event names, and attempting to listen to `elementEditor.saveElement` would really be listening to `elementEditor` events, with the listener getting namespaced to `saveElement`.

But I’m open to adding something similar. Can you explain the use case?

brandonkelly:
Nevermind, just saw #2689. Will follow up there.

joshangell:
Ah sorry, I couldn't get the local build working properly, which to be fair means this should never have been a PR ... sorry about that!

PR #2791, opened 2018-04-23
Add `tablePrefix` to `SetupController.php`, allowing it to be used as a command-line parameter

engram-design:
We're doing the following in our own command-line scaffolding, and it works great! However, it seems like `tablePrefix` is the only option not able to be set via the command line. For instance, we're doing:

```
$dbOptions = [
    'driver' => 'mysql',
    'user' => 'root',
    'server' => 'localhost',
    'port' => '3306',
    'password' => 'somepassword',
    'database' => 'craft_database',
    'tablePrefix' => 'craft_',
];

Craft::$app->runAction('setup/db-creds', $dbOptions);
```

Which throws an error `Error: Unknown option: --tablePrefix`. Adding this fixes the issue.

### DIFF
--- a/src/console/controllers/SetupController.php
+++ b/src/console/controllers/SetupController.php
@@ -83,6 +83,7 @@ class SetupController extends Controller
             $options[] = 'password';
             $options[] = 'database';
             $options[] = 'schema';
+            $options[] = 'tablePrefix';
         }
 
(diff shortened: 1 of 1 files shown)


brandonkelly:
Thanks!

PR #2799, opened 2018-04-24
Prevent Flash JS when deleting a Matrix block on the frontend

tomwelch:
When deleting matrix blocks from the front end of the site a number of cpresources are injected into the page at reload adding additional styling and a number of errors in the console.

screen shot 2018-04-24 at 13 35 16

The afterDelete method in src/elements/MatrixBlock.php is injecting the cpresources to clean up the browser's Local Storage of collapsed matrix blocks. As we're deleting them from the frontend, we don't need this functionality so I've updated the if statement to check if the request is a site request as well as a console request, and if so, skip including the cpresources.

This issue has also come up on the Craft CMS StackExchange:
https://craftcms.stackexchange.com/questions/25637/craft-3-resaving-matrix-from-frontend-causing-issues-craft-is-not-defined

### DIFF
--- a/src/elements/MatrixBlock.php
+++ b/src/elements/MatrixBlock.php
@@ -400,7 +400,7 @@ class MatrixBlock extends Element
      */
     public function afterDelete()
     {
-        if (!Craft::$app->getRequest()->getIsConsoleRequest()) {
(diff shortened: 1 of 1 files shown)


brandonkelly:
Thanks for pointing that out!

PR #2837, opened 2018-05-02
When checking mime type of SVG files, be forgiving if they lack an XML declaration

jcherniak:
When trying to read SVG files to pass through the svg() TWIG function, if they lack a XML declaration, they are not read as `FileHelper::isSvg` returns false.

Given the comment in `FileHelper::getMimeType`, this seemed like the appropriate place to make the change.  Otherwise, you could allow 'image/svg' in the FileHelper::isSvg.

Either way, this allows you to reference local SVG files that lack an XML declaration.

To reproduce:

1. Create a local asset volume.
2. Create an Asset field linking to it.
3. Upload this SVG into the folder and pick it in the asset picker:  [alarm-clock.svg.txt](https://github.com/craftcms/cms/files/1968530/alarm-clock.svg.txt) (Rename it to alarm-clock.svg, renamed due to Github not allowing SVG uploads).
4. In your template, add the following twig code:
`{{ svg(icon.folder.volume.path ~ "/" ~ icon.filename) }}`
5. Notice the function returns an empty string rather than the content of the file.

Note: this is an icon from Font-Awesome Pro, so adding XML declarations to all of the 800+ files is not practical.  Given the comment in FileHelper::getMimeType, it seems that was supposed to cover SVG files anyway.

### DIFF
--- a/src/helpers/FileHelper.php
+++ b/src/helpers/FileHelper.php
@@ -259,7 +259,7 @@ class FileHelper extends \yii\helpers\FileHelper
         $mimeType = parent::getMimeType($file, $magicFile, $checkExtension);
 
         // Be forgiving of SVG files, etc., that don't have an XML declaration
-        if ($checkExtension && in_array($mimeType, ['text/plain', 'text/html'])) {
+        if ($checkExtension && in_array($mimeType, ['text/plain', 'text/html', 'image/svg'])) {
             return static::getMimeTypeByExtension($file, $magicFile) ?? $mimeType;
         }
 


brandonkelly:
Strange - I believe the only valid SVG MIME type is `image/svg+xml`, so something may be messed up with your system’s magic file (the file that provides patterns that help PHP identify what a given file’s MIME type is, based on its contents). In my case, `alarm-clock.svg` did correctly come back as `image/svg+xml`.

That said, I’ve fixed this for the next release by updating `FileHelper::isSvg()` to just make sure that the MIME type starts with `image/svg`, rather than specifically checking for `image/svg+xml`.

PR #2838, opened 2018-05-02
Allow resource handles to be passed into svg() twig function

jcherniak:
This allows you to take SVG-based assets and include them inline in TWIG.

`{{ svg(asset.stream) }}`

Combined with my [plugin](https://github.com/pinfirestudios/craft3-dataurifilter), you can then include it sanitized and inline as a data uri like:

`{{ asset.title }}`

### DIFF
--- a/src/web/twig/Extension.php
+++ b/src/web/twig/Extension.php
@@ -806,14 +806,19 @@ class Extension extends \Twig_Extension implements \Twig_Extension_GlobalsInterf
     /**
      * Returns the (sanitized) contents of a given SVG file, namespacing any of its IDs in the process.
      *
-     * @param string $svg The SVG file path or contents
+     * @param string|resource $svg The SVG file path or contents
      * @param bool $sanitize Whether the file should be sanitized first
      * @return \Twig_Markup|string
      */
-    public function svgFunction(string $svg, bool $sanitize = true)
+    public function svgFunction($svg, bool $sanitize = true)
     {
+        // If it's a resource, read it until the end.
+        if (is_resource($svg))
+        {
+            $svg = stream_get_contents($svg);
+        }
         // If we can't find an <svg> tag, it's probably a file path
-        if (stripos($svg, '<svg') === false) {
+        elseif (stripos($svg, '<svg') === false) {
             $svg = Craft::getAlias($svg);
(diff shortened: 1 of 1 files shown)


brandonkelly:
Couldn’t you just give the asset a `__toString()` method? PHP will automatically convert it to a string when it’s passed to that function, if so.

jcherniak:
This is happening in TWIG and asset is a craft\elements\Asset object.  Asset::__toString returns the title, not the contents of the file.

jcherniak:
I could see adding Asset::getFileContents which contains `function getFileContents()
{
    return stream_get_contents($this->getStream);
}`

as another way to fix this.

brandonkelly:
Ahh, so your plugin is adding the `getStream()` method to assets. Yeah, in that case, a separate `getContents()` method would do the trick.

jcherniak:
My plugin isn't doing anything to the asset, that's the built-in Asset class.  My plugin is simply taking the resulting output from the SVG() function and forming it into a data-uri.

To be clear, the bug here is that this doesn't work:

1. Create a new asset source.
2. Upload a SVG into it.
3. You want inline SVG in your twig template based on the file, so...
3a. Create an asset field called "asset"
3b. In TWIG do something like the following:

`{{ svg(entry.asset.stream) }}`

I'd expect that to output inline SVG, but instead without this patch, I get a crash: 
image

Passing the asset itself:

`{{ svg(entry.asset) }}`

produces empty output.

The only way I could get this to work w/o a patch is:
`{{ svg(entry.asset.volume.path ~ "/" ~ entry.asset.fileName) }}`

That works fine in this case, but if you have a non-local folder, this will not work.  Hence my suggestion to allow svg() to accept streams (or add `\craft\elements\Asset::getFileContents()`).  That way you could read SVG files from remote volumes.

jcherniak:
To be clear, when I mention ```getStream```, I'm referring to this: https://github.com/craftcms/cms/blob/develop/src/elements/Asset.php#L900

brandonkelly:
Doh, I didn’t even realize we added that function! 🙃

I just added a check for actual Asset objects in the `svg()` method, so you can just do this: `{{ svg(asset) }}`.

Thanks for bearing with me.

Issue #927, opened 2017-01-31
Add 'default' option to the number field


angrybrad:
> Created by: Neal Stammers (neal@netgorilla.co.uk) on 2015/03/10 18:02:16 +0000
> Votes at time of UserVoice import: 10

-----------------------------------

When adding a number field you can currently set the min and max, but it would be useful to also be able to set a default value.

Thanks, Neal.

Visiluna:
Yes, please, please, please!

brandonkelly:
@Visiluna feel free to add a 👍 reaction to the original post ;)

Visiluna:
Not sure how to do that

narration-sd:
@Visiluna  he just means to use the little emoji button on the upper right corner of the original posting. When you click it, you get some choices, including  👍 , which means +1. 

These are counted, so yours will be added fairly and visibly. And this way also, not a string of comments added to do it. Take care.

brandonkelly:
https://github.com/blog/2119-add-reactions-to-pull-requests-issues-and-comments

Visiluna:
I was view it on my phone, where apparently adding reactions is not an option. I've added it now, thanks.

benface:
I believe this can be closed? :P

PR #2886, opened 2018-05-15
Improve number field: allow empty/null, add defaultValue

### LINKED ISSUE #988: Leave "Number" field blank by default
> Created by: Lindsey DiLoreto (lindsey@doublesecretagency.com) on 2015/03/11 04:48:18 +0000
> Votes at time of UserVoice import: 6

I totally get why a Number field would default to zero *when you save it*, but it doesn't really make sense for it to be pre-set to zero when creating a new entry.

It's peculiar that I need to delete the existing value before I can enter in the new value. I can't think of any other fields which exhibit that behavior.

timkelty:
Fixes #988, #927

The number field is much more useful when we allow empty/null values (instead of converting to `0`). Disallowing this, should then be done by making the field required, like other fields.

Also added is an optional default value for fresh elements.

/cc @lindseydiloreto

### DIFF
--- a/src/fields/Number.php
+++ b/src/fields/Number.php
@@ -38,6 +38,11 @@ class Number extends Field implements PreviewableFieldInterface
     // =========================================================================
 
     /**
+     * @var int|float|null The default value for new elements
+     */
+    public $defaultValue;
+
+    /**
      * @var int|float The minimum allowed number
      */
     public $min = 0;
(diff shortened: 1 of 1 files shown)


brandonkelly:
Merged, thanks! Though in my testing, Number fields were already able to save as `null` when left blank.

PR #2897, opened 2018-05-17
Fix MigrationHelper::dropAllForeignKeysToTable

ackwell:
Seems like `dropAllForeignKeysToTable` was silently failing, which was causing some fails in `dropTable` due to FK constraints.

The extra `updateType` and `deleteType` columns weren't being excluded from the column list, and it seems like it was collecting _all_ the columns, across both FKs and tables, while trying to remove them.

Noticed there's already a helper to cull out the non-column data from a FK array, so I've just swapped that into place.


brandonkelly:
MigrationHelper is a little complex. Can you send us a database and test code that demonstrates the issue? support@craftcms.com

ackwell:
Thanks for taking a look, should be a mail landing shortly.

Issue #2903, opened 2018-05-18
BUG: Template cache memory exhausted.


blasvicco:
### Description
Memory usage can reach over 250MB caching template.
The function in `craftcms/cms/src/services/TemplateCaches.php` called `includeElementQueryInTemplateCaches` can be optimized reducing the memory usage to less than 15 Mb.

### Suggestion
```PHP
    /**
     * Includes an element criteria in any active caches.
     *
     * @param Event $event The 'afterPrepare' element query event
     */
    public function includeElementQueryInTemplateCaches(Event $event)
    {
        // Make sure template caching is enabled
        if ($this->_isTemplateCachingEnabled() === false) {
            return;
        }

        if (!empty($this->_cachedQueries)) {
            /** @var ElementQuery $elementQuery */
            $elementQuery = $event->sender;
            $query = $elementQuery->query;
            $subQuery = $elementQuery->subQuery;
            $customFields = $elementQuery->customFields;
            $elementQuery->query = null;
            $elementQuery->subQuery = null;
            $elementQuery->customFields = null;
            // We need to base64-encode the string so db\Connection::quoteSql() doesn't tweak any of the table/columns names
            $serialized = base64_encode(serialize($elementQuery));
            $elementQuery->query = $query;
            $elementQuery->subQuery = $subQuery;
            $elementQuery->customFields = $customFields;
            $hash = md5($serialized);

            gc_disable(); // disable automatic garbage collector
            foreach ($this->_cachedQueries as &$queries) {
                $queries[$hash] = [
                    $elementQuery->elementType,
                    $serialized
                ];
                gc_collect_cycles();
            }
            $queries = null; // release memory
            unset($queries); // break the binding between variable name and variable content
            gc_enable(); // enabling automatic garbage collector
        }
    }
```

### Steps to reproduce

1. Mesure the memory usage of a non cached request with the `enableTemplateCaching` = true.

### Additional info

- Craft version: 3
- PHP version: 7.0.12

PR #2904, opened 2018-05-18
BUGFIX: Template cache memory exhausted. #2903

blasvicco:
### Issue:
https://github.com/craftcms/cms/issues/2903

### Detail:
-. We disable the automatic garbage collector in order to do it manually.
-. We add `NULL` assignation to free the memory before break the binding between variable name and variable content with the `unset`.

### DIFF
--- a/src/services/TemplateCaches.php
+++ b/src/services/TemplateCaches.php
@@ -230,13 +230,17 @@ class TemplateCaches extends Component
             $elementQuery->customFields = $customFields;
             $hash = md5($serialized);
 
+            gc_disable();
             foreach ($this->_cachedQueries as &$queries) {
                 $queries[$hash] = [
                     $elementQuery->elementType,
                     $serialized
                 ];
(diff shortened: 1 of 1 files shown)


src/services/TemplateCaches.php
```diff
@@ -230,13 +230,17 @@ public function includeElementQueryInTemplateCaches(Event $event)
...
 
+            gc_disable();
             foreach ($this->_cachedQueries as &$queries) {
                 $queries[$hash] = [
                     $elementQuery->elementType,
                     $serialized
                 ];
+                gc_collect_cycles();
             }
+            $queries = null;
```
brandonkelly: @blasvicco What’s the reason for this line, considering the `unset($queries)` on the next line?

brandonkelly: Thanks for the explanation! That was news to me.

blasvicco: ;)

PR #2917, opened 2018-05-23
Added sent parameter to 'email.onBeforeSendEmail' event (Craft 2)

bertoost:
To give listeners the ability to report a success state when they take over control of sending the email (performAction: false).

## The problem

When you have a listener (a plugin, or whatever), which is taking over control of sending emails, it will set `$event->performAction = false;` to stop Craft from sending the email.
When that one is set to false, it will always return boolean false to wherever the call is coming from.

For example; when you run "Test email" from the CP in the Email settings, it will always show you a failure message since the return value was false.

## The solution

The solution is simple, give the plugins the ability to determine and set the success state from the `email.onBeforeSendEmail` event and use that when the performAction is set to false by the listener.

With this fix, it is possible!

### DIFF
--- a/src/services/EmailService.php
+++ b/src/services/EmailService.php
@@ -267,10 +267,12 @@ class EmailService extends BaseApplicationComponent
 		$event = new Event($this, array(
 			'user' => $user,
 			'emailModel' => $emailModel,
-			'variables'	 => $variables
+			'variables'	 => $variables,
+			'success'    => false,
 		));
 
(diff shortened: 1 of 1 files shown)


brandonkelly:
Not opposed to this, but can you rename `success` to `sent`?

bertoost:
Of course @brandonkelly !
I am glad you're not opposed to this ;-)

bertoost:
Cool!

PR #2930, opened 2018-05-30
Fix japanese translation typo.

Ippey:
I found some typo in Japanese. I fixed bellow,

- "Last name" : Fix "ラストネーム" to "姓".  "姓" is more natural.
- "First name" : Fix "ファイル名" to "名".  "ファイル名" means "File name".
- Fix "こんにちわ" to "こんにちは" in mail body.  "こんにちは" is more formal.


brandonkelly:
Thanks!

PR #2989, opened 2018-06-13
Sort fields secondarily by handle to fix order flakiness

boboldehampsink:
When two or more fields have the same name (but a different handle of course), their sort order becomes unreliable (at least with PostgreSQL). This PR fixes that by secondarily sorting by handle.


brandonkelly:
Thanks!

Issue #2987, opened 2018-06-12
PHP Info Utility Local vs Master Values


sbossarte:
### Description

I have encountered a case where a web host provides configurations for PHP in a way that override local configuration values for PHP, but leave the master values at a system default that cannot be changed. While the correct values are displayed in a regular `phpinfo()` call, Craft's PHP Info utility prefers displaying the master value if the two differ.

Understandably, due to changes that Yii or other modules may make via `ini_set`, displaying the local value may not be wholly accurate for all requests. It threw me off seeing an unexpected value reported by the utility.

### Additional info

- Craft version: 3.0.10.3
- PHP version: 7.1.18
- Database driver & version: MySQL 5.5.56
- Plugins & versions: None

brandonkelly:
The PHP Info utility actually gets its info from a normal `phpinfo()` call:

https://github.com/craftcms/cms/blob/e43b4dcd9c6924d4da6b120a9416e82b181a2c77/src/utilities/PhpInfo.php#L66

So not sure why your results would differ from the utilities. Any ideas @angrybrad?

sbossarte:
What I mean to say is that the local value (first column) would be a different value from the master value (second column), and that the master value is what's displayed, instead of the local:

https://github.com/craftcms/cms/blob/e43b4dcd9c6924d4da6b120a9416e82b181a2c77/src/utilities/PhpInfo.php#L110

brandonkelly:
Ahh, yeah OK. Feel free to PR it :)

sbossarte:
Alrighty, figured it was best to open the issue first to ask about the behavior.

https://github.com/craftcms/cms/pull/2990

angrybrad:
Merged... thx!

PR #2990, opened 2018-06-13
PHP Info displays overridden values

sbossarte:
https://github.com/craftcms/cms/issues/2987

This changes behavior from preferring to display master values, as provided by any core PHP config, to instead display local values, which may be overridden (secondary settings from a host, `.htaccess` configurations, changes made with `ini_set`).

### DIFF
--- a/src/utilities/PhpInfo.php
+++ b/src/utilities/PhpInfo.php
@@ -104,11 +104,7 @@ class PhpInfo extends Utility
                         continue;
                     }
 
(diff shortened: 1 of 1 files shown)


angrybrad:
Any chance you can send the HTML output of a call to `phpinfo()` from a PHP script on a box you're seeing this on to support@craftcms.com?

sbossarte:
I've sent a copy of the output along.

angrybrad:
Thank you!

Issue #2992, opened 2018-06-13
FR: Service API Additions


sbossarte:
Presently, with some Craft services, there exist functions such as the following:

```
\craft\services\Fields::deleteFieldById()
\craft\services\Fields::deleteField()
```

These functions allow objects managed by the service to be deleted via providing an ID, or providing the object itself, respectively.  However, there are a few cases where there is only a `ById` form of these functions, such as `\craft\services\Categories::deleteGroupById()`, and `\craft\services\Structures::deleteStructureById()`, among others.

For consistency, it would be nice to see equivalent functions added for these services. If this sounds like a good idea, I could create a PR with these changes.

---

Additionally, there are two instances of function names that seem inconsistent with the rest of the definitions:

`\craft\services\Tags::deleteTagGroup()` and `\craft\services\Tags::deleteTagGroupById()` specify their group types, whereas the other services that manage some kind of 'group' object leave off the type name, and simply use `deleteGroup` and similar. This is the case with Categories, Fields, Sites, and UserGroups.

`\craft\services\AssetTransforms::deleteTransform()` Takes an ID instead of a transform itself, which it then uses to find an appropriate transform.

brandonkelly:
If you have the time we’d certainly welcome a PR :)

As far as fixing naming inconsistencies goes, we’ll need to maintain all current methods for backwards compatibility, but  they can be deprecated and just wrap the new method.

PR #3000, opened 2018-06-16
Service api changes

sbossarte:
This is my PR incorporating some changes mentioned in https://github.com/craftcms/cms/issues/2992. A detailed outline of my changes are at the bottom of this comment.

Most all of the group-related functions in the `Tags` service have `tagGroup` in their name, opposed to the other services that have 'group', and I figured asking if deprecating all of those would be a good idea in order to rename them, or if leaving it alone would be better for anyone maintaining things. If a change is preferred, I wouldn't mind making that change and adding it to this PR, or a future one if need be. For now, my change to the `Tags` service has followed its own naming convention instead of the other services.

Also, in the initial issue, I pointed out the `Structures` service and its lack of a `deleteStructure()` function. Though it's worth mentioning that after looking through the source, I now understand why that service is built the way it is, and how my previous assumption doesn't apply in this instance.

Changes within this pull request were broken into three commits:

- Updated variable name in Tags Controller to better reflect actual use.
	- `/craft/controllers/TagsController`
		- Modified `actionDeleteTagGroup()` to better name `sectionId` to `groupId`.
- Added some service functions to handle deleting objects by a reference or by id.
	- `/craft/services/Categories`
		- Added `deleteGroup()`, which now contains the bulk of the logic that `deleteGroupById()` had.
		- Modified `deleteGroupById()` to forward to `deleteGroup()` after retrieving and verifying the group.
	- `/craft/services/Tags`
		- Added `deleteTagGroup()`, which now contains the bulk of the logic that `deleteTagGroupById()` had.
		- Modified `deleteTagGroupById()` to forward to `deleteTagGroup()` after retrieving and verifying the group.
	- `/craft/services/UserGroups`
		- Added `deleteGroup()`, which now contains the bulk of the logic that `deleteGroupById()` had.
		- Modified `deleteGroupById()` to forward to `deleteGroup()` after retrieving and verifying the group.
- Updated AssetTransforms service to match the 'delete' and 'deleteById' convention.
	- `/craft/services/AssetTransforms`
		- Added `deleteTransformById()` which will accept an ID, and forward to `deleteTransform()` after retrieving and verifying the transform.
		- Modified `deleteTransform()` to accept an `AssetTransform` object.
		- Modified `deleteTransform()` to raise deprecation warnings when called with an integer ID.
	- `/craft/controllers/AssetTransformsController`
		- Modified `actionDeleteTransform()` to call the updated `deleteTransformById()` function of the `AssetTransforms` service.


brandonkelly:
Thanks a lot!

> I wouldn't mind making that change and adding it to this PR, or a future one if need be. For now, my change to the `Tags` service has followed its own naming convention instead of the other services.

We’d welcome the PR, but yeah it would be a bigger ordeal. There may be other classes that have similar inconsistancies as well.

PR #3022, opened 2018-06-23
Ensure volumes delete their associated field layouts when deleted.

sbossarte:
### Issue

It seems after deleting a volume, its associated fieldlayout remains in the database. This occurs both with manual creation and deletion, as well as programmatically.

### Example test code

```php
echo 'Creating volume' . "\n";

$volume = \Craft::$app->getVolumes()->createVolume([
	'type' => \craft\volumes\Local::class,
	'name' => 'Test Assets',
	'handle' => 'testAssets',
	'hasUrls' => true,
	'url' => '@web/test-assets',
	'settings' => [
		'path' => '@webroot/test-assets'
	]
]);
$fieldlayout = new \craft\models\FieldLayout([
	'type' => \craft\elements\Asset::class,
]);
$volume->setFieldLayout($fieldlayout);
if (!\Craft::$app->getVolumes()->saveVolume($volume)) {
	throw new \Exception('Failed to create volume!');
}

$volumeId = $volume->id;
$fieldlayoutId = $fieldlayout->id;

echo 'Volume has ID: ' . $volumeId . "\n";
echo 'Volume field layout has ID: ' . $fieldlayoutId . "\n";

echo 'Deleting volume' . "\n";
\Craft::$app->getVolumes()->deleteVolume($volume);

echo 'Does volume still exist? ' . (\craft\records\Volume::find()->where(['id' => $volumeId])->count() ? 'yes' : 'no') . "\n";
echo 'Does field layout still exist? ' . (\craft\records\FieldLayout::find()->where(['id' => $fieldlayoutId])->count() ? 'yes' : 'no') . "\n";
```

### DIFF
--- a/src/services/Volumes.php
+++ b/src/services/Volumes.php
@@ -634,6 +634,9 @@ class Volumes extends Component
                 Craft::$app->getElements()->deleteElement($asset);
             }
 
+            // Delete the field layout.
+            Craft::$app->getFields()->deleteLayoutById($volume->fieldLayoutId);
+
             // Nuke the asset volume.
             $db->createCommand()
                 ->delete('{{%volumes}}', ['id' => $volume->id])


brandonkelly:
Thank you!

PR #3033, opened 2018-06-26
Add file_exists check in front of filesize lookup in Asset validation

efatsi:
Fixes the issue discussed with @lukeholder over at https://github.com/craftcms/commerce/issues/355.


brandonkelly:
Thanks!

PR #3042, opened 2018-06-28
Updated norwegian translations

kristianj:
More consistent naming and changed some translations which were wrong or hard to understand.


angrybrad:
Thank you!

PR #3050, opened 2018-07-01
NL translation improvements

bertoost:
"Algemeen" (= General) is not a good translation for "Globals".
I suggest "Globalen". Since I mostly create a "General" field-layout inside globals.

Also "Singles" is not "Eenmalig" (which means "one-off").
I suggest "Losstaand" which means more like "individual" or "self-contained".


brandonkelly:
Thanks!

PR #3061, opened 2018-07-04
Update base.html

watarutmnh:
Fixed a bug where the `<html lang>` attribute was hard-set to `en-US` rather than the current application language. (Refer: https://github.com/craftcms/cms/pull/3053)

### DIFF
--- a/src/templates/_layouts/base.html
+++ b/src/templates/_layouts/base.html
@@ -5,7 +5,7 @@
 {% set bodyClass = (bodyClass is defined ? bodyClass~' ' : '') ~ craft.i18n.getLocaleData().getOrientation() -%}
 
 <!DOCTYPE html>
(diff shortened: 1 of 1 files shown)


lukeyouell:
@watarutmnh this has already been changed as per your request, no?

https://github.com/craftcms/cms/pull/3053

watarutmnh:
@lukeyouell You are right, but I wanted to change to Craft2 too.

brandonkelly:
Fix is going to be a little more complicated in Craft 2 since `craft.locale` will return  locale IDs in the format `en_gb` rather than `en-GB`.

watarutmnh:
@brandonkelly I see the point. it's much more complicate than I thought. For example, en_bw the locale ID, should not convert to en-BW. It should be "BW" to along with ISO 3166 format.
I'm grad that Craft 3 has been solved this.

PR #3067, opened 2018-07-04
Add `typeChange` event to `Craft.EntryTypeSwitcher`

joshangell:
This allows plugins to listen to the event and do things when the entry type is switched e.g. `Garnish.on(Craft.EntryTypeSwitcher, 'typeChange', function() {…});`

I have tested this locally by running the grunt build and it works well.


narration-sd:
Just made me curious by how you described this, Josh. 

I'm not aware of a plugin system for Garnish, save for generic means as this: https://craftcms.stackexchange.com/questions/1386/can-garnish-be-extended-to-allow-plugins-to-use-custom-keyboard-shortcuts

So if not, I guess you;'re talking about a Craft plugin 'listening to an event', and then I wonder why not pick up the change in Craft itself -- pr'ing an event if there isn't an appropriate one there?

brandonkelly:
@narration-sd `Garnish.on()` is a way to add class-level events to classes that extend Garnish.Base (e.g. Craft.EntryTypeSwitcher). So you don’t have to worry about finding the actual instance of the class. Similar to Yii’s [Event::on()](https://www.yiiframework.com/doc/api/2.0/yii-base-event#on()-detail).

joshangell:
Thanks @brandonkelly!

khalwat:
@joshangell I could kiss you for this. It's exactly what I need to fix a weird edge-case issue. Thank you!

joshangell:
@khalwat you're welcome, just make sure you wear a mask ...

PR #3087, opened 2018-07-10
Added extra Twig template file extension (html.twig)

bertoost:
To be compatible with IDE's and other frameworks etc.

When you also use other systems like Symfony, you'd probably used to create templates like `template.html.twig` and probably you've configured your IDE to recognize `.html.twig` as Twig templates (in terms of syntax highlighting).

I always set these in the general config. But I think this should be default. So, here's my PR :-)

### DIFF
--- a/src/config/GeneralConfig.php
+++ b/src/config/GeneralConfig.php
@@ -203,7 +203,7 @@ class GeneralConfig extends BaseObject
     /**
      * @var string[] The template file extensions Craft will look for when matching a template path to a file on the front end.
      */
-    public $defaultTemplateExtensions = ['html', 'twig'];
+    public $defaultTemplateExtensions = ['html', 'twig', 'html.twig'];
     /**
      * @var mixed The default amount of time tokens can be used before expiring.
      *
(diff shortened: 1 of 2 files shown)


brandonkelly:
This would add even more filesystem lookups when resolving templates, so probably not a good idea. Better to just keep overriding the config setting to just the extension(s) you want from your own config file.

bertoost:
Hmm then the inclusion code isn't right? once found one (in order of the definition) it should stop?
So for all templates that already exists with the original extension, it will not trigger an extra filesystem lookup?

brandonkelly:
It does stop as soon as it finds a match, but consider that when loading a template called `blog/_entry`, for example, these are the places Craft must look: (bold ones would be added if we merge this PR)

- blog/_entry.html
- blog/_entry.twig
- **blog/entry.html.twig**
- blog/_entry/index.html
- blog/_entry/index.twig
- **blog/_entry/index.html.twig**

So with this PR, if you actually go with `.html.twig` templates, then Craft will need to check either 3 or 6 files before it finds your actual template file. It may not seem like much, but if a single request ends uploading 10 templates, that’s 20 or 40 extra file lookups for that request, which will have an impact on page load times.

So it would be better to just override the config setting, and set it to just `['twig.html']`, so Craft only has 0 or 1 extra template file lookups it needs to make  before finding  yours.

PR #3135, opened 2018-07-23
Add Norwegian Nynorsk (nn) translation

semonje:
**Adds support for Norwegian nynorsk 🎉**
Whole thing translated manually by a native user, so it should be pretty decent. 

*(Also noticed the Norwegian bokmål (nb) version seemed somewhat sloppy here and there, some Google translate vibes – for any Norwegians seeing this)*


semonje:
I now realize I've forgot to add `config/locales/nn.php`. Will add shortly

brandonkelly:
Thanks for providing this!!

Those `config/locales/` files are auto-generated, so we'll take care of that.

PR #3153, opened 2018-07-25
Use setError for unsuccessful password update

davist11:
If a password was unable to be updated, a session error should be sent instead of a notice


brandonkelly:
Good call, thanks!

PR #3155, opened 2018-07-25
ö != ő, probably fixed all wrong usages

Saboteur777:


Saboteur777:
This and https://github.com/craftcms/cms/pull/3154 should be merged into one.

brandonkelly:
Thanks! FFR GitHub will let you push additional commits into an existing branch that has an active pull request, and they will get included in the PR.

So in this case, you could have made these changes on your [patch-3](https://github.com/Saboteur777/cms/tree/patch-3) branch, and they would have been included in #3154.

Saboteur777:
Yeah, I realised that after creating the PR. I was ninja-editing through the web editor so I couldn't revert it after that.

PR #3172, opened 2018-07-31
Updated docblock example in saveElement()

carlcs:


brandonkelly:
Yep that’s no longer needed. Thanks!

PR #3180, opened 2018-08-03
'Asset'-related strings updated, missing translation added

Saboteur777:
`Save as a new asset` had no translation - added. 
Asset had different translation - unified them.


brandonkelly:
Thanks!

PR #3189, opened 2018-08-10
Typo-/ortographic corrections and uniformity

8yard:
Changed é to e in imperative verbs.
Changed " " to “ ” (opening and closing quotes) – the correct characters in Norwegian are actually « », but they might look out of place on screen for most people. Thoughts?
Changed \'en to -en for binding the suffix to definite forms of abbreviations and acronyms.
Changed ... (three dots) to … (ellipsis) like the English version, also added spacing between the complete word and the ellipsis since it is the sentence and not the word that is truncated.
Changed the translation of Live from Direkte to Publisert, since it's referring to entry status.


brandonkelly:
Thanks you!!

PR #3200, opened 2018-08-14
Missing string from HU translation

Saboteur777:
Hungarian translation for "This field is translated for each site." added.


brandonkelly:
Thanks!

PR #3217, opened 2018-08-21
Update Matrix.php

engram-design:
Fixed an error that could occur if a Matrix block set to collapsed by a queue job.


brandonkelly:
Thanks Josh!

Issue #1524, opened 2017-03-13
Element content refactoring 


brandonkelly:
- All fields should be stored globally (remove the `context` column from `craft_fields`)
- Each field layout should correlate to a “field set”. Each field set has its own content table (`craft_content_X`).
- Field sets can be localizable or not, determining whether their content table has a `siteId` column, and whether that should be factored into element queries.
- Field sets can override Name, Handle, and Instructions settings, and possibly some field type-specific settings (e.g. the "Add button label" in relational fields).
- Element Query classes should be able to specify the field set IDs that will be relevant to a given query.
- Each element should store its field set ID, so Element Queries know which field set the element is associated with (which custom field columns' values to care about) just by looking at its row data. 
- Element titles should be moved to the craft_elements_i18n table, next to Slugs.
- Field Layout Designers should have inline field creation, to make up for the fact that Matrix field configurators will just have users *select* fields for their block types, rather than creating them.

j-greig:
Excited for this one!

PR #3225, opened 2018-08-21
Add 'author' to reserved field handle list

missmatsuko:


angrybrad:
The issue with this is that currently field handles are validated on a global basis.  "author" might be reserved in the context of entries, but it's perfectly valid in the context of tags or categories, for example.

This will be resolved once https://github.com/craftcms/cms/issues/1524 get implemented, though.

PR #3246, opened 2018-08-29
Fixed ElementQuery count() method always returns an integer.

bertoost:
Now there are two situations, when calling the parent, it returns a string instead of an integer.

### DIFF
--- a/src/elements/db/ElementQuery.php
+++ b/src/elements/db/ElementQuery.php
@@ -1153,7 +1153,7 @@ class ElementQuery extends Query implements ElementQueryInterface
             return count($cachedResult);
         }
 
-        return parent::count($q, $db) ?: 0;
+        return (int)parent::count($q, $db) ?: 0;
     }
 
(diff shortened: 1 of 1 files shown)


bertoost:
Wauw. Never thought about that. I even didn't know it exists.
The max int kinda large for such counts. But okay.
This was addressed to me cause I'd like to compare strict mode and my code wasn't executing cause it returns a string while an integer expected.

bertoost:
And therefor.. when you reach that number in your query count, you definitely should think about what you're doing ;-)

brandonkelly:
Yeah most likely no one will ever run into it, but I had considered making this change in the past, and after realizing that Yii had a reason to not mess with the data type here, I decided to leave it alone.

bertoost:
Yeah I understand that Brandon. Therefor I'd like to develop strictly ;-) like `->count() === 1`.. but I will cast in inline then ;-)

PR #3289, opened 2018-09-11
Add a Twig function to create yii\db\Expression objects

carlcs:
Sometimes it’s not possible to pass a db expression as a string, becaus Yii escapes it and generates invalid SQL. https://github.com/yiisoft/yii2/issues/553

Example use of the function:

```twig
{% set ids = [1639, 2079] %}

{% set orderByParam = ids ?
    dbExpression('FIELD(elements.id, ' ~ ids|reverse|join(', ') ~ ') desc, title asc') :
    'title asc' %}

{% set query = craft.entries
    .limit(50)
    .orderBy(orderByParam) %}
```

### DIFF
--- a/src/web/twig/Extension.php
+++ b/src/web/twig/Extension.php
@@ -713,7 +713,7 @@ class Extension extends \Twig_Extension implements \Twig_Extension_GlobalsInterf
             new \Twig_SimpleFunction('className', 'get_class'),
             new \Twig_SimpleFunction('clone', [$this, 'cloneFunction']),
             new \Twig_SimpleFunction('csrfInput', [$this, 'csrfInputFunction']),
-            new \Twig_SimpleFunction('dbExpression', [$this, 'dbExpressionFunction']),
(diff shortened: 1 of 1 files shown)


brandonkelly:
I like this. Can you rename the function to `expression()` instead of `dbExpression()` though?

brandonkelly:
(Worth noting that your particular example could also be achieved using the [fixedOrder](https://docs.craftcms.com/v3/dev/element-queries/entry-queries.html#fixedorder) param though.)

carlcs:
@brandonkelly `fixedOrder` only works when using it in combination with `ids()` though. In this example the `ids` variable is an array of element IDs for entries that should show up first, but there are more entries returned.

carlcs:
```twig
{% set ids = blog.featuredBlogEntries.ids() %}
```
This makes the example clearer.

brandonkelly:
Ah gotcha, thanks for clarification! Though I realize expressions could be used for a variety of things so wouldn't have declined the PR anyway :)

PR #3298, opened 2018-09-12
Use Craft queue over Yii queue.

putyourlightson:
I believe this was an incorrect reference to the Queue.


brandonkelly:
Yep, thanks!

PR #3310, opened 2018-09-18
Anonymise logs

lukeyouell:
Due to GDPR it would be really handy if we were able to anonymise the log prefixes. 

This PR allows for log anonymisation to be enabled from the general config file.

### DIFF
--- a/src/config/GeneralConfig.php
+++ b/src/config/GeneralConfig.php
@@ -73,6 +73,10 @@ class GeneralConfig extends BaseObject
      */
     public $allowUppercaseInSlug = false;
     /**
+     * @var bool Whether Craft logs should be anonymized.
+     */
+    public $anonymizeLogs = false;
+    /**
      * @var bool Whether users should automatically be logged in after activating their account.
      */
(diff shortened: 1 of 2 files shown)


brandonkelly:
@lukeyouell you mainly just don't want the IP in there right?

lukeyouell:
That's correct, but I thought it would make sense to remove the user & session ID as well. 

I can alter the code to keep these if you'd prefer?

brandonkelly:
Took it in a slightly different direction, because I’d rather avoid creating a new config setting for this. So now `craft\log\FileTarget` has a new `$includeUserIp` property, which is `false` by default.

If you really need to show IP addresses for debugging purposes or whatever, you can enable it from `config/app.php`:

```php
return [
    'components' => [
        'log' => function() {
            $config = craft\helpers\App::logConfig();
            if ($config) {
                $config['targets'][0]['includeUserIp'] = true;
                return Craft::createObject($config);
            }
            return null;
        },
    ],
];
```

lukeyouell:
Sounds good to me! Cheers Brandon.

PR #3342, opened 2018-10-03
make sure $settings is an array

bragle:


brandonkelly:
Have you run into a case where it isn’t an array?

bragle:
Yeah, sorry for the lack of a description. It might be a bit difficult to replicate but if it affects me, it also might affect other people. I've updated craft 2.6 to 3 with the Analytics (dukt) plugin, and I get this error when I load the dashboard: array_merge(): Argument #2 is not an array. I dumped the variable and from what i understand, Analytics is trying to pass (string) "null" as the settings value. There are multiple ways of fixing this, but i believe checking if $settings is an array is the best solution because it would prevent anyone from running into this issue again. Alternatively, checking if settings is "null" or fixing the plugin itself would work, but if there are any other cases like this, they too will get this error.

brandonkelly:
Makes sense, thanks!

PR #3356, opened 2018-10-08
 Add template hooks to global set editor pages

monachilada:
This mimics the hooks found in the entry edit page templates.


brandonkelly:
Good call, thanks!

PR #3374, opened 2018-10-12
Update German translation for stack trace

MoritzKn:
In German there is no translation for stack trace. People just use the English term. I'm a native German and software developer but never heard the term "Stapelspur". Searching in Google gives 492 results.


brandonkelly:
Thanks!

MoritzKn:
Could it be that the translations are maintained somewhere else? I'm asking because this PR was basically reverted in the commit: b0240ba2

angrybrad:
@MoritzKn sorry about that. We use a 3rd party translation service and they get a little overzealous when they see an `'English' => 'English'` string sometimes. Yours will be back in the next release.

PR #3375, opened 2018-10-12
Adds more events to Matrix fields and the Entry Type switcher

joshangell:
`Craft.MatrixInput` gets `afterInit` which fires at the very end of the init method, and `blockAdded` which fires just after a new block has been added to the DOM.

The `blockAdded` event also provides the new block html via the `event.data` property e.g.:

```js
Garnish.on(Craft.MatrixInput, 'afterInit', function(ev){
    var $newBlock = $(ev.data.block);
    // ...
});
```

`Craft.EntryTypeSwitcher` gets `beforeTypeChange` which fires just before any changes to the DOM are made.


joshangell:
Thanks Brandon!

Issue #3392, opened 2018-10-23
trashy needs


narration-sd:
### Description

Hmm. For what-you-know, I need to create elements on the fly and soon destroy them. With 3.1, you have the new trashable/recoverable/garbage collected elements.

I've dug down as far as to seem to see no interface for setting an element and all its possible children to be trashed, or to extend deleteElementById() to assure the same.

I'd prefer deleteElementById() to have a defaulted false argument to do this, as it's clean and also won't involve prior resaves of elements and children created by duplicate(), as one example I use.

Thanks...

### Steps to reproduce

1. create element
2. delete it within same session
3. for proximate forever, until gc, it will still be there. My database is filling up....!

### Additional info

- Craft version: 3.1x

narration-sd:
p.s. I didn't add any of the soft delete aspects for these elements, because I don't want that. But the code in Element::deleteElementByID still seems to check and silently fail to delete the master record or its children. 

Another way to look at the problem...that in fact soft delete/trash-gc isn't actually optional??

brandonkelly:
I don’t think we’ll add the ability to immediately hard-delete elements from the service, but your code can just do this after calling `Elements::deleteElement()`:

```php
Craft::$app->db->createCommand()
    ->delete('{{%elements}}', ['id' => $theEntryId])
    ->execute();
```

narration-sd:
Ok, and will this assure child elements get deleted also?

I'm particularly concerned with plugins like Neo...some trouble with that.

brandonkelly:
It would only partially delete Neo blocks, assuming its table has a cascade-delete foreign key to the `elements` table. Some Neo block data would hang around until it got garbage-collected after 30 days, but it won’t be visible or show up in queries or anything, so that shouldn’t matter.

narration-sd:
Well, let's try another way of stating the steps. It's really simple.

1) bring up Entries, with several Sections in their Groups available (e.g. Singles, Channels, Structures)
2) view the sections
3) type something unfindable in the Search field
4) note all of the sections disappear
5) now do a natural thing: click on one of  the Groups: e.g. Channels
6) you don't see any sections come into view
7) but the Search has gone blank again, so you have no clue why not

I'm suggesting not to correct that the Search went blank, but to let it blank and auto-refresh for the selected Group. 

[...]

PR #3403, opened 2018-10-25
3.1 element full delete

narration-sd:
This is to implement changes as requested in https://github.com/craftcms/cms/issues/3392, making 3.1's Soft Delete optional. 

It's implemented by selecting the new soft delete or the previous hard delete code exactly. 

The choice is made simply, via a flag field added to the current ElementEvent.

The plugin in need, Live Vue, simply sets this flag when deleting its hidden Live Preview records. The process with potential flag state is protected by a try/catch/finally code path, when deletion occurs  via Elements service deleteById().  

Since the operation occurs in Elements::deleteElement(), it will be available for any type of element deletion. At the same time, it's defaulted so that unless an EVENT_BEFORE_DELETE_ELEMENT is taken by the using program, soft delete will always be used as expected in 3.1.


brandonkelly:
Thanks, making this an event property works for me.

`$doFullDelete` has been renamed to `$hardDelete`, and moved to a new event class, `craft\events\DeleteElementEvent`.

PR #3438, opened 2018-11-03
Update table-fields.md

nfq:
`.all()` shouldn't be invoked for Table Fields


brandonkelly:
Table fields just store their data as a serialized array, so looping through it doesn’t require a database query, so no `.all()` is necessary in this case. You’d end up just getting an error because arrays don’t have `.all()` methods.

PR #3454, opened 2018-11-08
updated error message translations

hendrikeng:
made UserName Validator error message translatable
added de translation for {attribute} cannot contain spaces.


brandonkelly:
Thanks!

PR #3485, opened 2018-11-21
Migration documentation

TonyDeStefano:
I've tested these examples on my own craft project after digging around for instructions on how to properly do content migrations. I thought it might be useful for someone else and save them a little time.

### DIFF
--- a/docs/extend/migrations.md
+++ b/docs/extend/migrations.md
@@ -298,7 +298,7 @@ public function safeUp()
             ]);
             
             // Build section
-            $section = new Section([
+            $section = new \craft\models\Section([
                 'name' => 'My Section',
                 'handle' => 'mySection',
                 'type' => 'channel'
@@ -323,7 +323,7 @@ public function safeUp()
         if (!Craft::$app->getSections()->getEntryTypesByHandle('myEntryType')) {
             
             // Build Entry Type
-            $entryType = new EntryType([
+            $entryType = new \craft\models\EntryType([
                 'name' => 'My Entry Typw',
                 'handle' => 'myEntryType',
                 'hasTitleField' => true,
--- a/docs/extend/migrations.md
+++ b/docs/extend/migrations.md
@@ -147,3 +147,303 @@ When a plugin has an Install migration, its `safeUp()` method will be called whe
 ::: tip
 It is *not* a plugin’s responsibility to manage its row in the `plugins` database table. Craft takes care of that for you.
 :::
+
+## Examples
+
+**Fields**
+
+Values required to build each field type can be found at `craftcms/cms/src/fields`
+
+```php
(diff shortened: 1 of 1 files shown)


brandonkelly:
Thanks for submitting that! Craft 3.1 (coming in December) is adding a new feature called “Project Config” which helps sync things like fields, sections, and entry types, across multiple environments.

You can read more about it here: https://craftcms.com/blog/craft-3-1-dev-preview-is-here

And here’s the in-progress docs for it: https://github.com/craftcms/cms/blob/3.1/docs/project-config.md

Considering that, there won’t really be a need to create these things in content migrations, though.

ryanpcmcquen:
@brandonkelly, is it worth having this in the docs with a big deprecation warning (for historical purposes)?

brandonkelly:
Well nothing is getting _deprecated_ here; content migrations and these APIs will still exist, there will just be less of a need to use them together. That said, it could make sense for us to write [guides](https://craftcms.com/docs) for this stuff, e.g. “How to Programatically Create a Matrix Field”.

ryanpcmcquen:
@brandonkelly, to that end then, this documentation seems to add value (it is already helping me).

ryanpcmcquen:
@brandonkelly, we have added some more to this documentation, would you be willing to give it another look and provide us some feedback on whether or not this would be viable to merge now?
https://github.com/OriginalFunko/cms/blob/migration-documentation/docs/extend/migrations.md

brandonkelly:
@ryanpcmcquen by “guides” I mean guides at https://craftcms.com/guides, unrelated to the migration documentation.

We just released Craft 3.1 Beta [yesterday](https://craftcms.com/blog/craft-3-1-beta), with the new [Project Config](https://github.com/craftcms/cms/blob/3.1/docs/project-config.md) feature. With that I can’t imagine anyone is going to want to write content migrations for adding fields or sections.

ryanpcmcquen:
@brandonkelly, gotcha ... unless you aren’t running a beta and are writing migrations right now (like we are). Any idea when 3.1 will be out of beta?

Also, since this functionality isn’t getting deprecated as much as it is being superseded, is it going to remain undocumented? If so, I guess we will just maintain our own documentation until we start using project configs.

brandonkelly:
It’s all partially documented in the [class reference](https://docs.craftcms.com/api/v3/)

- [saveSection()](https://docs.craftcms.com/api/v3/craft-services-sections.html#public-methods)
- [saveField()](https://docs.craftcms.com/api/v3/craft-services-matrix.html#public-methods) (needs example)
- [saveLayout()](https://docs.craftcms.com/api/v3/craft-services-fields.html#public-methods)

That is all pulled from the doc blocks on the actual PHP methods in the codebase. If you want to submit PRs that improve those examples, we’d be happy to review/accept them.

PR #3494, opened 2018-11-23
Include showSiteMenu option in elementSelect field

monachilada:
I'm not sure if there's a reason why this isn't included, but it would be nice if showing the `showSiteMenu` could be configured in cases where the `Craft.BaseElementSelectInput` is being included in for example a custom field type.

### DIFF
--- a/src/templates/_includes/forms/elementSelect.html
+++ b/src/templates/_includes/forms/elementSelect.html
@@ -33,6 +33,7 @@
     sourceElementId: sourceElementId,
     viewMode: viewMode,
     limit: limit ?? null,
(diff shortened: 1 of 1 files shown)


brandonkelly:
It’s not included because Craft doesn’t have the ability to create relationships or reference tags for an element on a specific site. Being able to choose the site when selecting an element would give you the impression it was going to record the site you selected as part of the relation.

Do you have a use case where the selected site will actually be recorded?

brandonkelly:
Thanks for the explanation! Just merged for tomorrow’s release.

PR #3512, opened 2018-12-03
@web in Valet causes issues

philipboomy:
Hi

English is not my primary language so please change according. Also another user used variables instead of alias so perhaps that's better.

When I first ran into this issue it took me like 5 hours to figure out why I couldn't run multisite in Valet. I know a few other users have had this issue as I helped them out with it. So I thought it was about time I proposed a change in the docs to include a tip.

### DIFF
--- a/docs/sites.md
+++ b/docs/sites.md
@@ -6,6 +6,10 @@ You can define one or more sites at different domains, using a different set of 
 
 The multi-site feature in Craft is for sites with the same publishing team. You manage the multi-site content at the entry level, with the ability to enable Sections you want included in a site.
 
+::: tip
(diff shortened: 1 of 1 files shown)


brandonkelly:
Thanks for the suggestion. Your issue probably had nothing to do with Valet; if all of your sites start with `@web`, Craft won’t have any idea which of those is actually being requested, because `@web` always just gets replaced by whatever the currently-requested host name is.

philipboomy:
Thanks for quick response.

It was only my primary site that used @web but it "overwrote" my other sites. So my test2.com and test3.com showed the content etc from my primary site. When I changed @web to an alias or test.com then it all worked.

brandonkelly:
Right, because `@web` always resolves to the currently requested host name. So as far as Craft could tell, every request was really requesting the primary site.

Issue #3428, opened 2018-04-04
Use other session handler


boboldehampsink:
There is no documentation on how to use a different session handler then `yii\web\Session`. Here's an example of how to use both the session handler and cache of `yiisoft/yii2-redis` with Craft 3:

```php
    'components' => [
        'redis' => [
            'class' => yii\redis\Connection::class,
            'hostname' => $redis['host'],
            'port' => $redis['port'],
            'password' => $redis['pass'],
            'database' => 0,
        ],
        'session' => [
            'class' => yii\redis\Session::class,
            'as session' => [
                'class' => craft\behaviors\SessionBehavior::class,
            ],
        ],
        'cache' => [
            'class' => yii\redis\Cache::class,
            'defaultDuration' => 86400,
        ],
    ],
```

PR #3515, opened 2018-12-03
Document session component config

timkelty:
There seems to be a lot of confusion on how to do this, so I thought an example would help.
I also update the example to configure the redis component separately, which seemed better to me.

I also tried to track down the Yii docs for `as xxxxxx` behavior stuff, but can't seem to find anything.

Pretty sure @boboldehampsink found them before. I think we discussed [here](https://github.com/craftcms/cms/issues/3428), but the page seems dead now.

Any idea where this documentation is?

### DIFF
--- a/docs/config/app.md
+++ b/docs/config/app.md
@@ -78,7 +78,7 @@ return [
 
 ## Session Component
 
-In a load-balanced environment, you may want to override the default session component to store them in a centralized store (e.g. Redis):
+In a load-balanced environment, you may want to override the default session component to store them in a centralized location (e.g. Redis):
 
 ```php
 <?php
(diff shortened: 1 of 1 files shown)


brandonkelly:
> Pretty sure @boboldehampsink found them before. I think we discussed [here](https://github.com/craftcms/cms/issues/3428), but the page seems dead now.
> 
> Any idea where this documentation is?

Last code block on in this section: https://www.yiiframework.com/doc/guide/2.0/en/concept-behaviors#attaching-behaviors

brandonkelly:
Merged, will show up after tomorrow’s 3.0 release. Thanks!

PR #3516, opened 2018-12-03
Updates DbConfig.php to support pdo_mysql.default_socket path

ldennison-bsd:
Updates DbConfig.php to support pdo_mysql.default_socket path to support Unix socket usage in MySQL DB connections

### DIFF
--- a/src/config/DbConfig.php
+++ b/src/config/DbConfig.php
@@ -227,10 +227,14 @@ class DbConfig extends BaseObject
      */
     public function updateDsn()
     {
+        $unixSocketDefaultPath = ini_get('pdo_mysql.default_socket');
+
         if (!$this->database) {
             $this->dsn = null;
         } else if ($this->driver === self::DRIVER_MYSQL && $this->unixSocket) {
             $this->dsn = "{$this->driver}:unix_socket={$this->unixSocket};dbname={$this->database};";
+        } else if ($this->driver === self::DRIVER_MYSQL && $unixSocketDefaultPath) {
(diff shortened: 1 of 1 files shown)


ldennison-bsd:
This is related to #3517 .

ldennison-bsd:
It's important to note that this change may be destructive (i.e. cause issues) if the application's webserver is configured with a default socket path that's incorrect. There's a larger potential solution here where the socket path is tested for validity instead of being assumed to be correct.

brandonkelly:
I don’t see how we could make this change (or #3517) without breaking several Craft installs, and testing DB config settings on every request isn’t a great solution. Craft should be able to assume that DB connection settings are correct.

brandonkelly:
You could manually support this on a per-install basis by adding this line to `config/db.php`:

```php
'unixSocket' => ini_get('pdo_mysql.default_socket') ?: null,
```

ldennison-bsd:
Indeed, testing multiple DB connections isn't the greatest, but the test could be simply that the socket file exists. For example, the default MySQL location on many installs is /var/lib/mysql/mysql.sock. If that reference doesn't exist (which should be a trivial file_exists()), then it could be assumed the socket doesn't exist. So the code change here could be updated with a file_exists on that socket path to likely address the concern of broken installs for misconfigured server environments.

brandonkelly:
Just because the file exists doesn’t necessarily mean that it’s pointing to the same DB that the Craft site should be talking to.

brandonkelly:
We talked about this internally and decided Craft shouldn’t be making an assumption that just because `pdo_mysql.default_socket` is set to an existing file path, that it should be using it.

PR #3517, opened 2018-12-03
Updates Schema.php to support pdo_mysql.default_socket

ldennison-bsd:
Updates Schema.php to support pdo_mysql.default_socket for Unix socket usage when using mysqldump. The current code does not look for the default PDO socket path, which results in mysqldump failing to connect to the socket if mysqldump itself is not explicitly told (via MySQL configuration or similar) where the socket exists. For server installations that use the standard pdo_mysql.default_socket INI setting, this is unnecessary since that default socket location can be looked up and attempted. That's what this code change does.

Code style may need a bit of adjustment. GitHub interface is showing me different things from page to page.

### DIFF
--- a/src/db/mysql/Schema.php
+++ b/src/db/mysql/Schema.php
@@ -323,9 +323,12 @@ SQL;
             'host=' . $dbConfig->server . PHP_EOL .
             'port=' . $dbConfig->port;
 
+        $unixSocketDefaultPath = ini_get('pdo_mysql.default_socket');
         if ($dbConfig->unixSocket) {
             $contents .= PHP_EOL . 'socket=' . $dbConfig->unixSocket;
-        }
+        } else if ($unixSocketDefaultPath) {
+            $contents .= PHP_EOL . 'socket=' . $unixSocketDefaultPath;
+	    }
 
         FileHelper::writeToFile($filePath, $contents);
 


ldennison-bsd:
This is related to #3516 .

ldennison-bsd:
It's important to note that this change may be destructive (i.e. cause issues) if the application's webserver is configured with a default socket path that's incorrect. There's a larger potential solution here where the socket path is tested for validity instead of being assumed to be correct.

brandonkelly:
If #3516 is implemented, this change wouldn’t be necessary, no?

ldennison-bsd:
From my understanding, this issue (#3517) deals with the constructed configuration that gets passed to mysqldump. The other issue (#3516) deals with Craft's connection to the database at the application layer for normal requests (i.e. not mysqldump for DB backups). So it would seem that one change does not affect the other. This change would actually fix DB backups for server environments that rely on the PDO default socket configuration, whereas the other change is just a nice additive to consider the default socket before TCP connections.

I did some light testing that would seem to confirm that understanding. If you think that's in error, please let me know and I can construct a more thorough test case for your consideration.

brandonkelly:
If #3516 were implemented, then `$dbConfig->unixSocket` would be set from there, so this condition would pass:

https://github.com/craftcms/cms/pull/3517/files#diff-d02e7ff2a111baeeff47694c413fabd4R327

brandonkelly:
Closing because this wouldn’t be necessary if the `unixSocket` DB config setting is set, and that can be set to `pdo_mysql.default_socket ` from `config/db.php` per https://github.com/craftcms/cms/pull/3516#issuecomment-443898640.

PR #3526, opened 2018-12-05
Include solution for pagination

philipboomy:
I had problems caching my loop with pagination until I saw this post: https://craftcms.stackexchange.com/questions/26950/cache-not-invalidating-when-using-pagination-with-cache?rq=1

Before that my cache was not invalidated when saving a new entry.

Feel free to use it and to change the text and placement. I just know it could be useful to have there as other than me has run into this "issue".

### DIFF
--- a/docs/dev/tags/cache.md
+++ b/docs/dev/tags/cache.md
@@ -41,6 +41,14 @@ You can combine this parameter with [globally](#globally) to cache templates on 
 :::
 
 ::: warning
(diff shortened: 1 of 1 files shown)


brandonkelly:
This must have been an old copy of the Craft repo… we moved the docs over to [craftcms/docs](https://github.com/craftcms/docs/) a while ago. Can you submit the PR over there?

PR #3527, opened 2018-12-05
use separate namespace and module name examples

KatieMFritz:
The current docs use `foo` as the namespace and the module name, making it hard to tell from code examples which value should go in which place. I replaced all namespace examples with `bar`.


brandonkelly:
Works for me, thanks!

PR #3533, opened 2018-12-07
Fixed problem with path not being preserved

bragle:
Had some trouble getting Craft to work with a forwarded request, so I added an option to preserve URI.

Due to a host to host forward, the access url for craft had to be on a specific path. This seems to fix the issue.

### DIFF
--- a/src/config/GeneralConfig.php
+++ b/src/config/GeneralConfig.php
@@ -395,6 +395,10 @@ class GeneralConfig extends BaseObject
      */
     public $omitScriptNameInUrls = false;
     /**
+     * @var bool Whether Craft should keep the path as is.
+     */
+    public $alwaysPreserveURI = false;
+    /**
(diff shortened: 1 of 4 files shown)


bragle:
I also found a problem with a plugin (Sprout SEO) trying to pass a string instead of an int to getElementById. I don't think this should be necessary, and preferably fixed in the plugin, but if there are more cases like this, it might be better to prevent it by checking the var.

bragle:
My problem is that when an asset is retrieved through craft (ex `index.php?p=actions/assets/generate-transform&transformId=3`), craft will remove the path from the URL that i set in my config. Craft lives on a separate server and can only be accessed through a URL with a specific path (ex `example.com/craft/`). The solution I provided works for me, but I understand that it might not be the best solution to this problem. I should probably have posted an issue instead, so sorry about that.

brandonkelly:
If you post an issue about it and explain more about the environment and what the current/expected behaviors are, we can try to work on a solution for you.

PR #3555, opened 2018-12-13
Update UsersController.php

bertoost:
The CP login page shows me this error

```
Fatal error: Redefinition of parameter $variables in /var/www/html/craft/app/controllers/UsersController.php on line 506
```

Removing the double definition is removing the error and make the login work as expected again


brandonkelly:
Doh! Thanks.

bertoost:
👍

Issue #3566, opened 2018-12-17
Feature Request: `actionInput` template helper function


AugustMiller:
### Description

It'd be lovely to be able to generate the hidden `input` element for form actions… We already have shortcuts for `csrfInput()` and `redirectInput()`, so this feels like a handy addition.

Effectively, thinking something like this:

```php
    // new \Twig_SimpleFunction('actionInput', [$this, 'actionInputFunction'])

    // ...

    /**
     * Returns an action input wrapped in a \Twig_Markup object.
     *
     * @return \Twig_Markup|null
     */
    public function actionInputFunction($actionPath)
    {
        return TemplateHelper::raw('<input type="hidden" name="action" value="' . $actionPath . '">');
    }
```

(I wasn't sure whether the name of the field depends on the `actionTrigger` setting?)

Happy to issue a PR—but wanted to make sure I wasn't missing something that's already available!

brandonkelly:
Sure, we’d accept the PR. [docs/dev/functions.md](https://github.com/craftcms/cms/blob/develop/docs/dev/functions.md) would need to be updated as well.

> (I wasn't sure whether the name of the field depends on the actionTrigger setting?)

Nope, that only applies to query string `action` params.

PR #3573, opened 2018-12-17
Add `actionInput` Twig function

AugustMiller:
Should address #3566.

English docs have been updated, but I'm no use in translating! 😢


brandonkelly:
Thanks!

PR #3581, opened 2018-12-20
Added `json_decode` Twig filter

lindseydiloreto:
Bringing a little balance to the force. Adds a `json_decode` filter to counter the native `json_encode` filter.

No idea why it's not part of Twig's core, but now it doesn't matter! ¯\\\_(ツ)_/¯

### DIFF
--- a/src/web/twig/Extension.php
+++ b/src/web/twig/Extension.php
@@ -211,6 +211,7 @@ class Extension extends \Twig_Extension implements \Twig_Extension_GlobalsInterf
             new \Twig_SimpleFilter('index', [ArrayHelper::class, 'index']),
             new \Twig_SimpleFilter('indexOf', [$this, 'indexOfFilter']),
             new \Twig_SimpleFilter('intersect', 'array_intersect'),
+            new \Twig_SimpleFilter('json_decode', [$this, 'jsonDecodeFilter']),
             new \Twig_SimpleFilter('json_encode', [$this, 'jsonEncodeFilter']),
             new \Twig_SimpleFilter('kebab', [$this, 'kebabFilter']),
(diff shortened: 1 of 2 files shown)


src/web/twig/Extension.php
```diff
@@ -211,6 +211,7 @@ public function getFilters(): array
             new \Twig_SimpleFilter('index', [ArrayHelper::class, 'index']),
             new \Twig_SimpleFilter('indexOf', [$this, 'indexOfFilter']),
             new \Twig_SimpleFilter('intersect', 'array_intersect'),
+            new \Twig_SimpleFilter('json_decode', [$this, 'jsonDecodeFilter']),
```
brandonkelly: Pretty sure that this would work just as well –
```suggestion
            new \Twig_SimpleFilter('json_decode', 'json_decode'),
```

brandonkelly: Ah I had missed that `$assoc = true`.

Maybe this then?

```php
new \Twig_SimpleFilter('json_decode', [Json::class, 'decode']),
```

(and import `craft\helpers\Json` at the top)

lindseydiloreto: Good call, that sounds perfect! Ironically, I just discovered `Json::decode` last night.

I'll try to make that change within the next 24 hours and resubmit. 👍

lukeholder:
Thanks for the submission @lindseydiloreto, this PR has been replaced by #3678

lindseydiloreto:
Great, thanks @lukeholder! Sorry I never circled back on this... life gets crazy sometimes! 🙃

marionnewlevant:
twigperversion has had this for awhile

brandonkelly:
@marionnewlevant I think Commerce had a need for it, so made sense to add it in core.

Issue #3582, opened 2018-12-20
Feature Request: UID Routing Token


AugustMiller:
All records have a `uid` column—sometimes, it's helpful to generate obfuscated URLs, or at least hide the sequential nature of a record. It appears as though they follow a pretty standard `8-4-4-4-12` pattern.

It'd be awesome to have a new routing “token” compatible with the UIDs provided by Yii, say, `{uid}`:

<img width="642" alt="screen shot 2018-12-20 at 1 14 38 am" src="https://user-images.githubusercontent.com/1895522/50275637-ae23dc00-03f4-11e9-9acc-e67fd619f9d8.png">

In `routes.php`, it'd look like:

```php
return [
    'account/subscriptions/<subscriptionUid:{uid}>' => ['template' => '_account/subscriptions/subscription'
];
```

The Regex is a little cumbersome to include in a routes file (and isn't obvious what it's for, at first blush), so this feels like a worthwhile shortcut…

```
[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}
```

Happy to issue a PR if y'all are amenable to it! I'll be sure to update the [relevant docs](https://docs.craftcms.com/v3/routing.html#advanced-routing-with-url-rules), as well.

💞

brandonkelly:
Sure, we’ll take a PR for that.

AugustMiller:
This was added a long while back, so I'm closing!

PR #3583, opened 2018-12-20
Add UID route token

AugustMiller:
Addresses #3582.

Sorry if this is too presumptuous—had the changes queued locally, so I figured it was worth committing!

✌️


brandonkelly:
@AugustMiller Have you tested both of these regexes with capital letters?

@angrybrad if UUIDs are only including letters a-f, we should probably update StringHelper::isUUID() to match this regex. (Also not sure why the `u` modifier is in there… a-f definitely won’t require that.)

AugustMiller:
In its current state, it won't match capital letters… all the references I've seen to the pattern assume lower-case `a-f`.

My intent was to match whatever is being output by `StringHelper::UUID()`, not necessarily cover outside world implementations.

Up to this point, I hadn't realized the “audit columns” were a Craft ActiveRecord extension, and not something that Yii provided!

angrybrad:
@AugustMiller can you check if the functionality still works as you'd expect with those changes?

AugustMiller:
Unfortunately not—probably due to the subset of regular expressions that can actually be injected into a route?

For example, the final route (example taken from our app) ends up as this, in the Yii route debugger:

```
account/membership/history/<subscriptionUid:/^[A-Z0-9]{8}-[A-Z0-9]{4}-4[A-Z0-9]{3}-[89AB][A-Z0-9]{3}-[A-Z0-9]{12}$/i>
```

…which is skipped by: `account/membership/history/1a8e179d-3fc5-4fc5-ba52-85ccc286f5ad`.

It seems like we might need an additional route-compatible pattern?

angrybrad:
@AugustMiller yeah, you're correct.  This should do it though (actually tested it works with a UID route this time): https://github.com/craftcms/cms/commit/cbd3abef05cfce4efbebe00da58e826c6b287681

AugustMiller:
Works like a charm. 👌

PR #3590, opened 2019-01-03
Fix drag'n'drop asset when using elementSelect

nstCactus:
This allows passing the asset field id to the elementSelect form component, which is needed to determine where to store assets uploaded via drag'n'drop.

Without this fix, dropping an image on an elementSelect component defined like this would result in an error:

```twig
{{ forms.elementSelect({
    name: 'desktopImage',
    elements: value.desktopImage.all(),
    sources: asset.sources,
    criteria: { kind: asset.allowedKinds },
    jsClass: 'Craft.AssetSelectInput',
    elementType: 'craft\\elements\\Asset',
}) }}
```

**Error:** 
> The field provided is not an Assets field


andris-sevcenko:
Thanks for this!

Issue #3592, opened 2019-01-03
craft\migrations\m180520_173000_matrix_context_to_uids fails for upgrade to 3.1.0-beta.5.1


ryanpcmcquen:
### Description

**One of Craft CMS’s migrations failed.**:

```
yii\base\ErrorException: Undefined offset: 37
Migration: craft\migrations\m180520_173000_matrix_context_to_uids
Output:
> update in {{%fields}} ... done (time: 0.006s)
> update in {{%fields}} ... done (time: 0.001s)
> update in {{%fields}} ... done (time: 0.001s)
> update in {{%fields}} ... done (time: 0.001s)
> update in {{%fields}} ... done (time: 0.001s)
> update in {{%fields}} ... done (time: 0.001s)
> update in {{%fields}} ... done (time: 0.001s)
> update in {{%fields}} ... done (time: 0.001s)
> update in {{%fields}} ... done (time: 0.001s)
> update in {{%fields}} ... done (time: 0.001s)
> update in {{%fields}} ... done (time: 0.001s)
> update in {{%fields}} ... done (time: 0.001s)
> update in {{%fields}} ... done (time: 0.001s)
> update in {{%fields}} ... done (time: 0.001s)
> update in {{%fields}} ... done (time: 0.001s)
> update in {{%fields}} ... done (time: 0.001s)
> update in {{%fields}} ... done (time: 0.001s)
> update in {{%fields}} ... done (time: 0.001s)
> update in {{%fields}} ... done (time: 0.002s)
> update in {{%fields}} ... done (time: 0.001s)
> update in {{%fields}} ... done (time: 0.001s)
> update in {{%fields}} ... done (time: 0.001s)
> update in {{%fields}} ... done (time: 0.001s)
> update in {{%fields}} ... done (time: 0.001s)
> update in {{%fields}} ... done (time: 0.001s)
> update in {{%fields}} ... done (time: 0.001s)
> update in {{%fields}} ... done (time: 0.001s)
> update in {{%fields}} ... done (time: 0.001s)
> update in {{%fields}} ... done (time: 0.001s)
> update in {{%fields}} ... done (time: 0.001s)
> update in {{%fields}} ... done (time: 0.001s)
> update in {{%fields}} ... done (time: 0.001s)
> update in {{%fields}} ... done (time: 0.001s)
> update in {{%fields}} ... done (time: 0.001s)
> update in {{%fields}} ... done (time: 0.001s)
[...]

PR #3593, opened 2019-01-03
Check for `dateDeleted` column before adding it.

ryanpcmcquen:
Fixes https://github.com/craftcms/cms/issues/3592.


brandonkelly:
There are a ton of things that could go wrong if you try to update Craft over a partially updated database, where the previous attempt failed. We’d prefer that you just revert all the changes and start fresh.

ryanpcmcquen:
@brandonkelly, it is not a partially updated database, see my comment here: https://github.com/craftcms/cms/issues/3592#issuecomment-451286726

Issue #2980, opened 2018-06-11
Feature Request: Add Asset Indexing as a cli action


sergeifilippov:
It would be great to be able to run Asset Indexing via the `craft cli`.

timkelty:
Converting the indexing process to a Yii queue is pretty critical for any site with a sizeable amount of assets.

We've run into all sort of problems with indexing (latest being https://github.com/craftcms/cms/issues/2876#issuecomment-434434075), which makes indexing basically a "cross your fingers" operation, that's nearly impossible to debug.

ryanpcmcquen:
Is there a way to index assets from a migration script?

brandonkelly:
Not currently.

ryanpcmcquen:
This works for me inside of a migration script for indexing assets on a specific volume:

```php
        try {
            if ($volume = Craft::$app->getVolumes()->getVolumeByHandle('VOLUME_HANDLE')) {
                foreach (Craft::$app->getAssetIndexer()->getIndexListOnVolume($volume) as $item) {
                    Craft::$app->getAssetIndexer()->indexFile(
                        $volume,
                        $item['path'],
                        '',
                        false
                    );
                }

            } else {
                return false;
            }
            return true;

        } catch (\Throwable $e) {
            echo $e->getMessage();
            return false;
        }
```

PR #3595, opened 2019-01-05
Add asset indexing to Craft cli.

ryanpcmcquen:
Closes https://github.com/craftcms/cms/issues/2980.


ryanpcmcquen:
I actually think the default should be `false` (`0`) on `$cacheImages` here, but since the GUI default is `true`, I made the default `true` here as well. I'm guessing a lot of people don't want to fill their database with an image they are already storing somewhere else (like on an S3 bucket).

brandonkelly:
They get stored in the `storage/` folder, not your database, and it’s a prerequisite to generating new image transforms, so a reasonable argument could be made either way.

andris-sevcenko:
Sorry for the delayed response - thinks have been a tad hectic lately.

Can you, please, reduce the same indexing session id for one indexing session instead of generating a new one for every file that is getting indexed?

ryanpcmcquen:
@andris-sevcenko, thank you for reviewing this! How would you recommend doing that? Can it just use a timestamp?

andris-sevcenko:
@ryanpcmcquen you should call the `Craft::$app->getAssetIndexer()->getIndexingSessionId()` method once per session and use the returned value for all actual indexing calls.

Also, the docblock above the class definition does not look like anything that Craft has. ([starting here](https://github.com/craftcms/cms/pull/3595/files#diff-32af748b77d508e466862f9008ecd96fR18))

ryanpcmcquen:
@andris-sevcenko, isn't this similar?
https://github.com/craftcms/cms/blob/3.1/src/console/controllers/MigrateController.php#L24

I changed it to use the session id as requested. 🎉

ryanpcmcquen:
@andris-sevcenko, I will restore it then! Anything else?

ryanpcmcquen:
@andris-sevcenko, I restored the docblock, any other comments on the PR?

andris-sevcenko:
@ryanpcmcquen ah sorry, totally forgot about this one today - it was on my list, too :/

I'll take care of this tomorrow my time, sorry for the delay!

andris-sevcenko:
Thank you for your patience, Ryan. You have been more than awesome!

ryanpcmcquen:
Thank you Andris!

khalwat:
Wow, thank you @ryanpcmcquen -- saved me from writing it!

ryanpcmcquen:
@brandonkelly, should I have set the `@author` in the docblock? Wasn't sure what the policy was here ...

PR #3613, opened 2019-01-10
A warning about checkboxes limitation

MattWilcox:
If we'd known this before building a rather huge project, we would not be relying on a search parameter looking through a 700,000 record search index table in order to get our results...

### DIFF
--- a/docs/checkboxes-fields.md
+++ b/docs/checkboxes-fields.md
@@ -67,3 +67,7 @@ Checkboxes fields have the following settings:
     {% endfor %}
 </ul>
 ```
+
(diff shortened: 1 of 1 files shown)


brandonkelly:
You don’t have to use the `search` param to filter entries by a Checkboxes field value; you can use the field’s own element query param.

I just added a new [Querying Elements with Checkboxes Fields](https://docs.craftcms.com/v3/checkboxes-fields.html#querying-elements-with-checkboxes-fields) to the Checkboxes field docs which demonstrates how to use it.

PR #3615, opened 2019-01-10
Allow filename modification post-sanitation

### LINKED ISSUE #3608: Overzealous filename sanitation stripping valid starting and ending characters
### Description
Files uploaded to Craft should preserve their original filenames as closely as possible. I'm not sure if there's a historical reason for it, but FileHelper.php is stripping out some perfectly valid characters from the start and end of filenames: `-_.`. In particular, we're running into pain where a client is trying to upload several hundred detailed product specifications using a format like: `123-4.5-__.pdf` to represent products in the 123-4.5 range. When an asset like this is uploaded, it is sanitized to `123-4.5.pdf` which is incorrect in this use case. I believe you could make a case for stripping out periods; however, I see no reason to force users to avoid dashes and underscores at the start and end of filenames.

I would propose at the very least, adding a configuration option to specify blacklisted starting and trailing characters.

### Additional info

- Craft version: 3.0.37
- PHP version: N/A
- Plugins & versions: N/A
- Host Platform: Amazon Linux/Mac OS

Mosnar:
This change resolves an edge-case issue where a developer may need to modify a filename after it has been sanitized by Craft. It adds a new property to the SetAssetFilenameEvent class for `originalFilename` that contains the filename prior to Craft's sanitation. This property is purely for developers to reference when making changes to the filename property. Additionally, the extension of the file has been added to the event to give developers additional context to the filename being sanitized should they want to apply different rules to different extensions or allow them to make adjustments to the extension.

Closes #3608


andris-sevcenko:
Thank you!

PR #3617, opened 2019-01-10
Update app.php with missing entry 'and'

luke-nehemedia:
Added the entry 'and' with the correct German translation. This fixes an issue with the |duration twig filter where the conjugation 'and' is not properly translated.

Should probably added to '3.1' too.


brandonkelly:
Thanks! I’ll merge into `3.1`.

luke-nehemedia:
Thanks!

PR #3625, opened 2019-01-14
Idea: Allow error/notice message with the redirect tag

ostark:
I stumbled about this requirement today 
https://craftcms.stackexchange.com/questions/29094/how-do-you-set-a-flash-message-in-the-template-in-craft-3

Since a flash massage is usually bound to a redirect, it might be a good addition to the `{% redirect %}` tag.

This PR adds the option to set a `error` or `notice` message with the following syntax:

```
{% redirect '/some-url' with notice 'Something good happend.' %}
{% redirect '/some-url' with error 'Something bad happend.' %}
```


brandonkelly:
Good idea – merged for Craft 3.1!

PR #3646, opened 2019-01-16
Update GeneralConfig.php

carlcs:


brandonkelly:
Thanks! It was originally called `disableAdminFunctions` so the wording was accurate when first written :)

Issue #3717, opened 2019-01-25
Error: Return value of craft\helpers\App::phpConfigValueInBytes() must be of the type integer, float returned


BrandonJamesBorders:
### Description
Installing plugins from the CMS no longer work. 

### Steps to reproduce

1. Go to the plugin store page within Craft CMS install.
2. Find a plugin to install and hit click "Try" or "Install"
3. See error message.

### Additional info

- Craft version: 3.1.4
- PHP version: 7.2.8
- Database driver & version: MySQL 5.5.5
- Operating System: Windows 10.0.17763.0

BrandonJamesBorders:
If I'm assuming correctly that this is related to 3.1.4 and checking of php memory_limit, this is what I get when I run the phpConfigValueInBytes function:
```
$var = 'memory_limit';
$value = trim(ini_get($var));
$unit = strtolower(substr($value, -1, 1));
$value = (int)$value;
switch ($unit) {
    case 'g':
        $value *= 1024;
    // no break (cumulative multiplier)
    case 'm':
        $value *= 1024;
    // no break (cumulative multiplier)
    case 'k':
        $value *= 1024;
}
var_dump($value)
```
`$value` returns `float(2147483648)`

If I dump out `$value * 1024` I get `int(2097152)`, but for somereason when that happens inside the switch statement it converts to a float.

BrandonJamesBorders:
So this is a 32-bit PHP error combined with having a memory limit >= 2048M as the result would be `2,147,483,648` and the max integer size on 32-bit PHP is `2,147,483,647`.

BrandonJamesBorders:
#3721 

Proposed fix to change expected response as a float, which suppor

brandonkelly:
Thanks for the bug report and investigating it! Fixed for the next release. In the meantime, you can get around it by changing your `memory_limit` config setting to `-1`.

PR #3721, opened 2019-01-25
Update phpConfigValueInBytes to return float

BrandonJamesBorders:
High memory limits (>= 2048M) may result in a float being returned on 32bit versions of PHP.

### DIFF
--- a/src/helpers/App.php
+++ b/src/helpers/App.php
@@ -185,11 +185,11 @@ class App
      * @param string $var The PHP config setting to retrieve.
      * @return int The value normalized into bytes.
      */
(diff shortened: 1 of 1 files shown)


BrandonJamesBorders:
#3717

brandonkelly:
Thanks for submitting. Don’t want to force a `float` response though, and there isn’t any reason to force the initial `$value` (line 192 in your PR) to a float, as at that point it should just be `2048` at that point. PHP can autoconvert the int to a float if it ends up getting too high further down the function, as it’s doing in your case.

PR #3722, opened 2019-01-26
Fix `json_decode` Twig filter

AugustMiller:
Looks like the docs for this were correct (that it called `Json::decode(...)`), but the actual method name was still just the raw PHP function name.

See the original PR: #3678


brandonkelly:
Doh, thanks!

boboldehampsink:
Ah, just wanted to create an issue. I'm seeing

```
yii\base\ErrorException: call_user_func_array() expects parameter 1 to be a valid callback, class 'craft\helpers\Json' does not have a method 'json_decode' in
```

But I guess this fixes it

PR #3771, opened 2019-02-01
Suppress exceptions on Twig getEnvSuggestions

### LINKED ISSUE #3769: Dealing with aliases and environment variables containing special characters ($, @)
### Description

We store our database credentials, among other things, in the environment and use auto-generated passwords. This has caused an issue with the move to autosuggest fields and the use of environment variables/aliases in the CP since frequently, our passwords contain or even start with "@" or "$". This causes a pretty nasty error in the CP displaying the password in question:

image

I'm not sure what the best way around this is, but disallowing these characters in passwords/forcing people to change passwords to less secure ones doesn't feel like the right approach. 

### Steps to reproduce

1. Use a password in env with an "@" or "$"
2. Try to use an autosuggest field in the CP

### Additional info

- Craft version: 3.1.7
- PHP version: 7.2.13
- Plugins & versions: N/A

Mosnar:
Suppresses exceptions thrown by Twig's autosuggest field when an alias's value starts with "@"

Fixes #3769

### DIFF
--- a/src/web/twig/variables/Cp.php
+++ b/src/web/twig/variables/Cp.php
@@ -333,7 +333,7 @@ class Cp extends Component
         foreach (array_keys($_ENV) as $var) {
             $envSuggestions[] = [
                 'name' => '$' . $var,
-                'hint' => $security->redactIfSensitive($var, Craft::getAlias(getenv($var)))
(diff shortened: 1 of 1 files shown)


brandonkelly:
Whoops, missed this and already fixed :) Thanks anyway!

PR #3774, opened 2019-02-02
Update Norgwegian translation

halvorsanden:
I have worked through the entire translation after updating to 3.1. There were maybe just a couple of things I would call direct mistakes, but rather some issues with consistency and unnatural syntax that sticks out in the long run. Also improved a lot of punctuation that is different from English.


brandonkelly:
Thanks so much for doing that!

PR #3798, opened 2019-02-07
Don’t warn about manually set @web

carlcs:


brandonkelly:
Good idea. Might do the same for `@webroot` as well.

PR #3805, opened 2019-02-07
Skip assets when indexing

TonyDeStefano:
I ran into an issue where the indexing failed after about 15,000 assets. I added this code to allow you to skip the assets that had already been indexed (which is a huge time-saver)


andris-sevcenko:
Makes sense. Thanks!

PR #3831, opened 2019-02-12
Update README.md

drewtownchi:
The @ sign in front of the assetBaseUrl and assetBasePath prevents the alias from working correctly.  In my config when I removed the @ sign I was then able to properly utilize the aliases @assetBaseUrl and @assetBasePath


brandonkelly:
It shouldn’t matter either way… Craft adds in the `@` if it’s missing.

PR #3835, opened 2019-02-13
FIx Japanese translation: First Name, Last Name

watarutmnh:
"First Name" translation had been mixed up with "File Name".


brandonkelly:
Thanks!

PR #3838, opened 2019-02-13
Move the opcache invalidation to the File Helper

coudenysj:
This way, all users of the method will benefit from the invalidation.

E.g. https://github.com/enupal/translate/blob/6e706885e61ba9460b18ef565ed88a05781e6f2a/src/services/Translate.php#L125


brandonkelly:
Good idea!

PR #3839, opened 2019-02-13
Update app.md

mattstein:
Update method of getting mail settings now that the previous variant is deprecated.


brandonkelly:
Thanks!

PR #3850, opened 2019-02-15
Updated empty array check in saveRelations.

extensibleseth:
I ran into a funky if statement in saveRelations(), services/RelationsService.php:55, triggered by a Sprout Forms submit. The $targetIds array is passed with one empty string element, [0 => ''], so ($targetIds) evaluates to true, but that borks the insert because it wants something that evaluates to an integer.

Array_filter checks the truthiness of the values and returns the true ones. This might be more appropriately fixed further upstream, but this works for me.


brandonkelly:
Applied the fix slightly differently, as yours wouldn’t have helped if `$targetIds` contained some non-empty values in addition to an empty value.

PR #3857, opened 2019-02-18
Revert an erroneous German translation change

ul8:
Current translation added with 3.1 ('Deprecation Warnings' => 'Wertminderungswarnung') is translated as if in context of finance/tax deprecation. The previous translation was correct.


brandonkelly:
Thanks!

PR #3860, opened 2019-02-18
Return new CSRF values in JSON response for login/logout.

### LINKED ISSUE #3858: FR: Return new CSRF token in response of any controller that consumes CSRF
Currently, if you have an SPA or even just ajax requests to controllers that require a CSRF, you can pass it along, probably [getting it from your template](https://craftcms.com/guides/enabling-csrf-protection#updating-your-javascript)

Great – except now that CSRF has been used, so you need a new one, but if your page hasn't reloaded, you don't have one. Making a subsequent request just to get a CSRF is wasteful, and Craft doesn't really have a controller for that anyway (although you could use `users/get-remaining-session-time`).

What would be ideal is any controller that consumed a CSRF token, also created a new one and returned it in the json response.

Related: https://craftcms.stackexchange.com/questions/28735/csrf-in-headless-applications from @AugustMiller

timkelty:
Fixes #3858

### DIFF
--- a/src/controllers/UsersController.php
+++ b/src/controllers/UsersController.php
@@ -262,13 +262,14 @@ class UsersController extends Controller
         // Passing false here for reasons.
         Craft::$app->getUser()->logout(false);
 
-        if (Craft::$app->getRequest()->getAcceptsJson()) {
+        $request = Craft::$app->getRequest();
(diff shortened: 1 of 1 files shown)


brandonkelly:
Thanks for the PR!

PR #3878, opened 2019-02-21
Fix query column casing

### LINKED ISSUE #3877: Localize Relations job fails on PostgreSQL
### Description

The Localize Relations job fails on PostgreSQL. This is because PostgreSQL is case sensitive. See this line https://github.com/craftcms/cms/blob/develop/src/queue/jobs/LocalizeRelations.php#L70

### Additional info

- Craft version: 3.1.12
- PHP version: 7.3.2
- Database driver & version: PostgreSQL 10.6
- Plugins & versions: N/A

boboldehampsink:
Fixes #3877


brandonkelly:
Doh, didn’t see this. Fixed already.

PR #3898, opened 2019-02-25
Typo in doc

SayChi:
The documentation about relation fields has a typo in the code which made it reference to a class that doesn't exist. Also there is a link that is supposed to target the doc for BaseRelationField, but that doc doesn't exist


brandonkelly:
Thanks for the typo fix.

Links that start with `api:` get swapped out with class reference links. (See https://docs.craftcms.com/v3/extend/element-types.html#relation-field for an example.)

brandonkelly:
Hm looks like 
https://docs.craftcms.com/api/v3/craft-fields-baserelationfield.html is broken for some reason. Probably a bug in the class reference generator though; the Craft docs itself are not the problem here.

PR #3901, opened 2019-02-25
Check also Sites locales when getting user preferred language

### LINKED ISSUE #3900: Users with a preferred language set receive system messages in incorrect language.
### Description

In a multisite installation users have a preferred language that is non existent in the App Locales. This makes these users receiving System messages emails in the primary language instead.

The issue can be fixed by updating the `getPreferredLanguage` method:
https://github.com/craftcms/cms/blob/develop/src/elements/User.php#L1152
```
    public function getPreferredLanguage()
    {
        $language = $this->getPreference('language');

        // Make sure it's valid
        if ($language !== null &&
            in_array($language, Craft::$app->getI18n()->getAppLocaleIds(), true) ||
            in_array($language, Craft::$app->getI18n()->getSiteLocaleIds(), true)) {
            return $language;
        }

        return null;
    }
```

### Additional info

- Craft version: Craft Pro 3.0.40.1
- PHP version: 7.2.14

bzin:
This should fix #3900

### DIFF
--- a/src/elements/User.php
+++ b/src/elements/User.php
@@ -1154,7 +1154,9 @@ class User extends Element implements IdentityInterface
         $language = $this->getPreference('language');
 
         // Make sure it's valid
-        if ($language !== null && in_array($language, Craft::$app->getI18n()->getAppLocaleIds(), true)) {
+        if ($language !== null && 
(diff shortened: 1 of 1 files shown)


brandonkelly:
Pretty sure #3900 isn’t a real issue if the `extraAppLocales` config setting is being used correctly.

Issue #3874, opened 2019-02-20
Plugin listing ignores bottom padding in FireFox


JayBox325:
### Description
Nothing major at all. But when viewing the list of plugins in my project in FireFox the bottom padding is overlapped by the table within.

Could just add a bottom margin to the table#plugins element?

![screenshot 2019-02-20 at 20 19 35](https://user-images.githubusercontent.com/13233809/53122101-ac3ddc80-354d-11e9-820b-ad0c33873d78.png)

I notice it doesn't overlap on Mobile/Tablet sizes, only desktop.

### Additional info

- Craft version: 3.1.12
- Firefox version: 65.0.1

brandonkelly:
Well that’s awkward. Seems like a Firefox bug though – nothing is telling it to ignore the bottom padding on `#content`.

magnessjo:
@brandonkelly, 

  I took a stab at fixing this while I had some down time : [PR](https://github.com/craftcms/cms/pull/3905).  

  It’s an issue on every page that has the #content element (not just the plugin page).  You can reproduce it by shrinking your window to ~500px.  

  It might not the ideal solution but it will hopefully point you guys in the right direction.

brandonkelly:
This has been fixed for Craft 3.4. (See https://github.com/craftcms/cms/issues/2883#issuecomment-554751414 for more details.)

JayBox325:
I've just installed it and this is much better. Thanks. I just upped it to 48px and it looks a bit neater giving the final panel some more breathing room away from the bottom of the screen.

The page ended with a matrix where I viewed this and it did seem a bit tucked away and benefited from a little bit extra to bring it up into the page.

But this is obviously just my opinion and not a change request. Happy with it how it is! Great job.

PR #3905, opened 2019-02-25
Fixes: #3874 : Modify how padding is applied to #content element

magnessjo:
Fixes issue with Firefox not supporting padding on element with parent of max-height 0 and flex.

The issue is caused by min-height: 0  on the #main-content element.  I am assuming that they design intention was to keep all of the columns (header, sidebar and content) equally aligned with an overflow scroll y being set.  

I did QA this solution on a local website on mine at height of 500px to reproduce the issue.  This solution did fix the problem on all pages. I didn’t see any negative effects on main nav pages (dashboard, entires, globals, etc) or plugins (SEOmatic, or my own).  I also tested in Chrome and Safari and didn’t see any issues with this fix.  

Note: I consider this a band-aid fix.  In my opinion, a long term solution would be a restructuring on the HTML to avoid nesting of multiple layers of flex box.  Another solution, would to be modify the CSS to use grid.  Both of those would take more time than it might currently be worth.  

In hindsight, I think the wildcard selector should be set with the last-child attribute.  I don't want to edit the PR to avoid confusion (and I haven't QA'ed it) but *:last-child could help to avoid any possibility of multiple child elements.  Applying the padding specifically to the last child element.

### DIFF
--- a/src/web/assets/cp/dist/css/_cp.scss
+++ b/src/web/assets/cp/dist/css/_cp.scss
@@ -737,7 +737,7 @@ $sidebarLinkSecondaryColor: hsl($hue, 5%, 75%);
           @include border-right(1px solid $hairlineColorAlpha);
         }
 
-        & > * {
+        & > *:last-child {
           padding-bottom: 48px;
         }
       }
--- a/src/web/assets/cp/dist/css/_cp.scss
+++ b/src/web/assets/cp/dist/css/_cp.scss
@@ -732,9 +732,13 @@ $sidebarLinkSecondaryColor: hsl($hue, 5%, 75%);
           flex: 1;
           background: #fff;
           background-clip: padding-box;
-          padding: 24px 24px 48px;
+          padding: 24px 24px 0;
           word-wrap: break-word;
           @include border-right(1px solid $hairlineColorAlpha);
+        }
+
+        & > * {
+          padding-bottom: 48px;
         }
       }
 


JayBox325:
Upgrading the CP to use CSS Grid would be a positive change, but that would be a big task. Craft 3.2 perhaps?

magnessjo:
@JayBox325, 

  What I've heard before (years ago) from Brandon on a Podcast was that switching from Garnish (Pixel & Tonic jQuery library) would be a big burden.  I would expect that some of the HTML semantic structure is in place due to Garnish.  Switching to Grid would require some tweaks to the code I walked through today (in my opinion).  I wouldn't expect this type of change for Craft 3.*.  It might be on the roadmap for Craft 4 though.  In general,  the UI works really well so I would argue it is probably low priority.

JayBox325:
@magnessjo That makes sense. Flexbox is more than adequate. Just a shame FireFox renders it differently on desktop. But as you say, it still works, so it's definitely a low priority.

brandonkelly:
I’m not really following the fix… as you can see here, `#content` will always be the last child of `#content-container`:

https://github.com/craftcms/cms/blob/54c81a370c2b557e863fb768c800406571123473/src/templates/_layouts/cp.html#L252-L264

So setting `padding: 24px 24px 48px;` on `#content` should be identical (as far as `padding-bottom` goes) as setting `padding-bottom: 48px;` on `#content-container > *:last-child`. Are you just finding that Firefox is doing a better job of applying the correct styles when you set it this way?

Curious, have you tried this on the `#content` styles?

```css
padding: 24px 24px 0;
padding-bottom: 48px;
```

> Another solution, would to be modify the CSS to use grid.

Not sure how Grid would help here? There’s nothing really grid-y about the main CP layout. (You can see how the CP is structured in the comment at the top of [src/templates/_layouts/cp.html](https://github.com/craftcms/cms/blob/develop/src/templates/_layouts/cp.html).)

Or are you recommending a complete redesign that would be able to make use of Grid? (If that’s what you’re getting at, it’s coming in Craft 4 or 5.)

magnessjo:
@brandonkelly 


###### PRIORITY ######

Just to be clear: this should be consider a low level issue.  The UI is 100% usable on every browser I tested (Chrome, Safari, Firefox).  I just have some down time from switching jobs and this looked like an annoying bug so I took a stab at it.  

## My Understanding of the Issue ##

The root cause is min-height : 0 being set on the  element `#content-container` and it also being set on that element's parent `#main-content`.  Firefox is collapsing the padding bottom based on min-height setting.  Removing the min-height from `#main-content` will produce two difference experiences with the overflow-y: scroll in Chrome and Firefox.  See attached.  My assumption is that Firefox is collapsing the padding bottom because it doesn't have an explicit height to follow on either parent element.  

Note: I did try a number of other solutions and I did spend a fair amount of time researching the issue and reading the Flexbox spec.  I did not find any similar situation with multiple elements nested with min-height : 0 causing the padding to collapse online.  So I went with my assumption.  I could absolutely be wrong.

chrome
firefox



## Fixes ##

### Option A: Band-Aid -- Suggested Fix ###

This is what I would consider an equivalent to a clear fix for this problem.  It forces the last child element of the `#content-container` to be explicit with it bottom padding.  This causes the parent to calculate the height based on the children's height (height + padding).  The addition of the :last-child attribute was just to future proof it.  I didn't see an instance when there were multiple children inside `#content-container`, but just to be safe.

Note: Modifying the padding bottom like you suggested would not make a difference.

### Option B: Restructure HTML -- Probably not a good idea at this point ###

There are a million ways use HTML & CSS.  Everyone takes a slightly different approach.  My approach would not have been to include multiple layers of min-height : 0 on Flexbox children.  To me, the automatic minimum sizing feels like a loop hole, even thought it is part of the [spec](https://drafts.csswg.org/css-flexbox/#min-size-auto).

Note: I think the UI for control panel is great!  I'm not suggesting it needs to change to fix this issue, but the issue does stem from multiple layers of min-height being set so that Firefox will overflow-y : scroll properly.  

### Option C: CSS Grids -- Craft 4 or 5  ###

CSS Grids would not be a quick solution to this problem.  It would require a significant restructure to the control panel CSS (and most likely HTML).  I could see [recoding on the control panel down the line](https://github.com/craftcms/cms/pull/3905#issuecomment-467187370) but I think that is a much bigger request than this bug requires.  

Note: I would be happy to help when you guys do decided to go back to update Garnish and rework the build process for control panel.

brandonkelly:
@magnessjo I just merged this in and updated the compiled CSS, however I’m not seeing any different, at least on the Settings > Plugins page as originally reported in #3874.

magnessjo:
@brandonkelly, 

  I don't see it working correctly either.  I am looking into it now.  I will get back with you shortly.

magnessjo:
My apologizes @brandonkelly , I screwed up.  I can certainly see why you were confused now.  

The element that the padding should be applied to is the last child within the `#content`.  I got confused going between my example and the fork.  

I have pushed an update if you would still like to use it.  What I've tested (and the goal is): 

```
#main-container #main #main-content #content-container #content > *:last-child { padding-bottom: 48px; }
```

I sincerely apologize for the confusion and any wasted time that you spent on it! I am willing to do a few hours of tedious & boring work to make up for it!

brandonkelly:
@magnessjo Haha, no worries, however I think you will need to create a new pull request as this one is now merged.

magnessjo:
@brandonkelly,

  Added a new [PR](https://github.com/craftcms/cms/pull/3916). 

Thank you!

PR #3921, opened 2019-02-27
Add new() Twig function to create PHP objects

khalwat:
Craft has already added the `clone()` function, and exposed all of `craft.app` in Twig templates. One missing piece of the puzzle is the ability to create new PHP objects via Twig.

Many plugins exist just to expose this functionality, and this would eliminate the need. For example, a recent discussion came up as whether to use my [Cookies](https://github.com/nystudio107/craft-cookies) plugin or use functionality already exposed by Craft.

From the Yii2 docs:

> Yii represents each cookie as an object of yii\web\Cookie. Both yii\web\Request and yii\web\Response maintain a collection of cookies via the property named cookies. The cookie collection in the former represents the cookies submitted in a request, while the cookie collection in the latter represents the cookies that are to be sent to the user.

We can get most of the way there via:

```twig
{%  set requestCookies = craft.app.request.cookies %}
{%  set responseCookies = craft.app.response.cookies %}
```

...which will give us a [CookieCollection](https://www.yiiframework.com/doc/api/2.0/yii-web-cookiecollection) that we can use to access or remove cookies... but we can't add new cookies, because `::add()` needs a `yii\web\Cookie` object passed to it.

This is just one example where a plugin could be eliminated, and the core functionality of Craft used to accomplish some common things.

### DIFF
--- a/src/web/twig/Extension.php
+++ b/src/web/twig/Extension.php
@@ -721,6 +721,7 @@ class Extension extends \Twig_Extension implements \Twig_Extension_GlobalsInterf
             new \Twig_SimpleFunction('expression', [$this, 'expressionFunction']),
             new \Twig_SimpleFunction('floor', 'floor'),
             new \Twig_SimpleFunction('getenv', 'getenv'),
+            new \Twig_SimpleFunction('new', [$this, 'newFunction']),
             new \Twig_SimpleFunction('parseEnv', [Craft::class, 'parseEnv']),
             new \Twig_SimpleFunction('plugin', [$this, 'pluginFunction']),
             new \Twig_SimpleFunction('redirectInput', [$this, 'redirectInputFunction']),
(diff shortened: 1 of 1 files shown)


brandonkelly:
Let’s call it `create()` instead of `new()`, and point it at `Craft::createObject()` rather than an inline function in the Twig extension class.

That way it won’t only work for classes with `$config` as the first constructor argument, and you’ll get a couple different supported syntaxes.

```twig
{% set obj = create('my\\class\\name') %}

{% set obj = create({
    class: 'my\\class\\name',
    configOpt1: 'foo',
    configOpt2: 'bar'
}) %}

{% set obj = create('my\\class\\name', [
    'constructArg1',
    'constructArg2'
]) %}
```

khalwat:
Works for me. You want to do it, or should I?

brandonkelly:
I’m on it…

PR #3935, opened 2019-03-01
Update to reflect the move to Discord

mikewink:
This commit just updates the move from Slack to Discord.


brandonkelly:
Thanks for pointing that out! We had come up with a whole new list of popular resources for the repo READMEs, and just forgot to update the docs index with it. Just did that, so this PR is no longer necessary.

PR #3956, opened 2019-03-07
deprication warning (.all) in a template

SayChi:


brandonkelly:
`elements` is supposed to already be an array. Can you post the stack trace you’re seeing for the deprecation warning?

SayChi:
You're right, looks like I was just using it wrong. I was feeding it the query without doing .all on my end, oops

PR #3966, opened 2019-03-08
Display up to ten documentation search results

mattstein:
Because five may not be enough, and fifty is too many.


lindseydiloreto:
Yes, please! Five is definitely not enough. 🙂

brandonkelly:
Thanks! Cherry-picked this rather than merged, so we can publish the change right away. (f09cf8fe8dc6302fb8ac1569b222533eb01259f9)

mattstein:
This setting apparently isn't honored by Algolia-powered search. I'll figure out what needs doing and report back with a new PR.

Issue #3967, opened 2019-03-08
Tokens documentation is inaccurate


AugustMiller:
It looks as though the `craft\services\Tokens::createToken` method documentation describes a `routes.php`-style object…

```php
['action' => "controller/action", 'params' => ['foo' => 'bar']]
```

…when in reality, the shape is more like this:

```php
['controller/action', ['foo' => 'bar']
```

At least… this is what I ended up needing to do?

That brought up a question: can you route a token to a template, or does it have to map to a controller action? (This is also a bit unclear, as the comment suggests “…*if you want* [it] to be routed to a controller action…”)

AugustMiller:
The latter format is used elsewhere, i.e: https://github.com/craftcms/cms/blob/develop/src/controllers/LivePreviewController.php#L60-L65

brandonkelly:
Thanks for pointing that out. Fixed now. https://docs.craftcms.com/api/v3/craft-services-tokens.html#method-createtoken

PR #3968, opened 2019-03-08
Adjust documentation to match token routing implementation

AugustMiller:
Opening to address the discrepancy noted in #3967!

I wasn't sure what language should be used to describe the "matching request…" 🤔


brandonkelly:
Decided it's worth showing more complete usage examples, that illustrate the different supported `$route` param syntaxes (84943329e35f1e24b0d5a0ab29375bbadc83b653).

PR #3985, opened 2019-03-12
Really display up to ten search results

mattstein:
This is just like [my last PR](https://github.com/craftcms/cms/pull/3966), except that it'll work. Algolia results don't honor `searchMaxSuggestions` (previous PR), but that's safe to leave in place since it impacts the native VuePress search. 

I made sure this is legit before submitting. See the [Snipcart docs](https://snipcart.docs.workingconcept.com) for a demo of the exact behavior this would enable.


angrybrad:
> This is just like my last PR, except that it'll work.
😂

brandonkelly:
Thanks!

PR #3992, opened 2019-03-13
Translatable fields

jan-dh:
Updated the documentation on the different translation options for fields.


brandonkelly:
Good idea, thanks!

PR #4000, opened 2019-03-15
Update changes-in-craft-3.md

lealea:
Change instances of "the following" to be more clear about referencing "older Craft 2." When using the "the following" phrase, it references _everything_ underneath it. But, it's actually related to Craft 2 items only. Not the _entire_ list of functions or config settings, etc

e.g. 
Original: 
"The following config settings have been deprecated in Craft 3, and will be completely removed in Craft 4"

See how it can make it seem like _all_ the config settings will be removed in Craft 4?

My suggestion:
"Older Craft 2 config settings have been deprecated in Craft 3, and will be completely removed in Craft 4"


brandonkelly:
Each of those lines precedes a table of deprecated config settings, all of which are included in “the following”. Not sure I’m _following_ how “the following” could be misleading in those contexts?

lealea:
Heh. To me, when I first read this, it looked like the *new* configs in Craft 3 would _also_ be removed. 

The config settings below include tables showcasing both deprecated *and* new. “The following configs” when I first read it made it look like *all* the settings listed below would be gone in Craft 4. Which didn’t make sense to me because there was no explanation as to why the Craft 3 settings would disappear. But theoretically is possible considering Craft 2 configs did, too.

brandonkelly:
Ahhh, I see what you mean now.

narration-sd:
good catch, @lealea  ;)

PR #4026, opened 2019-03-21
Update user-profile-form.md

rauwebieten:
added enctype="multipart/form-data" to form tag
Otherwise file upload does not work


andris-sevcenko:
Thanks!

Issue #4048, opened 2019-03-26
For \craft\mail\Message setTo($to) doesn't work if $to is User[]


bighoho123:
### Description
When calling `setTo($to)`, if $to is of User[], it will throw "Array to string conversion" error

### Steps to reproduce

```php
$message = new \craft\mail\Message();
$to = User::findAll();
$message->setTo($to);
$message->setSubject("Test");
$message->setHtmlBody("Hello World");
```

### Additional info

- Craft version: Craft Pro 3.0.26.1
- PHP version: 7.2.5
- Database driver & version:
- Plugins & versions:

I think the reason being the Swiftmailer will expect $to to be `['email1@email.com'=>'Person 1', 'email2@email.com'=>'Person 2']` instead of `[['email1@email.com'=>'Person 1'],[ 'email2@email.com'=>'Person 2']]`

brandonkelly:
Thanks for pointing that out, it’s been fixed for the next release.

PR #4049, opened 2019-03-26
Fix _normalizeEmails($emails)

bighoho123:
Fix an issue preventing passing User[] as $emails. This is related to issue #4048

Because _$normalizeEmail() can return both array or string, therefore the logic in _normalizedEmails() should cater for that.


brandonkelly:
Thanks for the PR! Ended up taking the fix in a different direction, though.

PR #4051, opened 2019-03-26
extendable element index / handle index context

nateiler:
It would be nice to use/extend the base element index controller; however it's basically useless with all private properties/methods.  

This PR converts private into protected declarations.  In addition, it adds a new method used to determine whether actions should be included or not ref: `includeActions` method vs multiple `* === 'index'` conditionals.

Index lists are really useful.  I'd like to see configurable table layouts per index.  I went down this rabbit whole a while back and it would require a few more PR's... but this is a start.


brandonkelly:
Looks good.

PR #4068, opened 2019-03-28
Argument 1 passed to craft\services\Structures::getStructureById() must be of the type int, string given

thenotsoft:


brandonkelly:
PHP will automatically convert numeric strings to ints when they’re passed to a function argument that has an `int` type declaration. So there’s probably a deeper issue here. I’d check what the values are before getting passed to `getStructureById()`.

thenotsoft:
the value is obtained from the database and its string format and `getStructureId` expects a int

`public function getStructureById(int $structureId, bool $withTrashed = false)`

brandonkelly:
> the value is obtained from the database and its string format and getStructureId expects a int

Yeah but again, if the string is numeric, then PHP will auto convert it to an int for you in this case. So if you are getting this error, then the value is most likely not a numeric string.

If you have Xdebug, you could try creating a conditional breakpoint here:

https://github.com/craftcms/cms/blob/07dcd7f593c2faf56303f3348c061204270ae32d/src/base/Element.php#L417

with the following condition:

```php
(string)(int)$source['structureId'] !== (string)$source['structureId']
```

That will break right before you’d get the PHP error you quoted in your title.

thenotsoft:
Maybe it should be converted, but in this case it is not converted, and it gives an error.

PHP Version 7.3.2

brandonkelly:
Try creating a PHP file on your server with this:

```php
<?php
function test(int $foo) {
    var_dump($foo);
}
test('5');
```

Then access it with your browser. You will see that even though you passed in a string, PHP automatically typecasted it to an integer.

Now change the last line to

```php
test('foo');
```

And you will see that you now get the “Argument 1 passed to test() must be of the type int, string given” error, similar to the one this PR is about.

That’s what I mean by “numeric string” (`'5'`). My point is that there’s no bug in Craft here, PHP is fully capable of handling this situation whiteout pre-typecasting the string to an int, and if you’re getting this error, then you probably have a data problem.

I would suggest you update to Craft 3.1.20+, and then run the following command:

```bash
./craft project-config/rebuild
```

And see if that fixes the issue. If not, please send your `composer.json` and `composer.lock` files, and a database backup, over to support@craftcms.com and we can help your look into it from there.

PR #4108, opened 2019-04-08
Added hint to relation field

putyourlightson:
Added hint to registering the relation field with the Fields service.

### DIFF
--- a/docs/extend/element-types.md
+++ b/docs/extend/element-types.md
@@ -701,6 +701,8 @@ class Products extends BaseRelationField
 }
 ```
 
+If you want your relation field to be available as a custom field then you’ll need to [register it with the Fields service](https://docs.craftcms.com/v3/extend/field-types.html#registering-custom-field-types), so Craft will know about it when populating the list of available field types.
+
 ## Reference Tags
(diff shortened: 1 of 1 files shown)


brandonkelly:
Appreciate the PR but this seems a little redundant? We’re already linking to https://docs.craftcms.com/v3/extend/field-types.html right above it, which has a code example on how to register the field type. (And there’s hardly any point in creating a field type if you don’t register it, so the wording is a little awkward, too.)

putyourlightson:
Yeah, that's true. I was tutoring someone recently and we read the docs together and both missed the Field Types link, probably because we just grabbed the code sample for the relation field. I have a few custom field types in my plugins that are used in the plugin's CP pages only and are not registered, so I think there is a use case for that. 

Perhaps just a bit more of a push in the right direction would help, for example:

> You can give your element its own relation field by creating a new field type that extends craft\fields\BaseRelationField _and optionally [registering it](https://docs.craftcms.com/v3/extend/field-types.html#registering-custom-field-types) with the Fields service_.

No worries if you find it unnecessary.

PR #4112, opened 2019-04-09
Include the ability for Craft CMS Fields to add their own PHPDocType

### LINKED ISSUE #3894: Include the correct type into the property of the ContentBehavior
### Description
First of all: The way you handle the ContentBehavior is absolutely awesome and I appreciate your idea really much. 
I'm not sure if you have ever thought about including a more specific type into the PHPDoc when you create this class? I know this is far away from being a first world problem but including a property like

```PHP
public static $phpDocType = 'mixed';
```
into `craft\base\Field` that can overwritten by any class that extends it. For example the `BaseRelationField` could have 

```PHP
public static $phpDocType = 'craft\elements\db\ElementQuery';
```
would be really awesome.

You could then fetch the type of the field together with the handle in your query and check if the specific class exists to grab the static attribute

```PHP
$docType = class_exists($fieldType)? $fieldType::$phpDocType : 'mixed';
//------
//------
/**
 * @var {$docType} Value for field with the handle “{$handle}”.
 */
public \${$handle};
```

If you think that might make Craft better but don't have the time for it I'm happy to create a pull request.

This way: when we type `$element->user->` PHPStorm (or others) won't complain about missmatches and suggest the functions of `ElementQuery`

Anubarak:
Include an function into `craft\base\FieldInterface::getPHPDocType` to add the correct PHPDoc for the `ContentBehavior` and `ElementQueryBehavior` to autocomplete their field values

Fixes #3894

I tested it with Matrix and Supertable fields -> a nested setup so one field type can have a bunch of different values separated by `|`. The result could be something like this
```PHP
    /**
     * @var mixed|\craft\elements\db\ElementQuery|\DateTime Value for field with the handle “test2”.
     */
    public $test2;

    * @method self test2(mixed|\craft\elements\db\ElementQuery|\DateTime $value) Sets the [[test2]] property
```
This field appear in global context, inside two matrix fields and inside a supertable field so there should be no conflicts for multiple field handles with the same or differnt types.

I implemented it for
- BaseRelationField
- BaseOptionField
- Date

The rest can be done as you like.
Please rename my variables if you want, I'm usually uncreative when it comes to that.

**Edit:**
I'm not really sure why there are merge conflicts ¯\\_(ツ)_/¯
Actually I forked your repo and replaced
```PHP
$fieldHandles = (new Query())
    ->distinct(true)
    ->from([Table::FIELDS])
    ->select([$column, 'type'])
    ->all();
```
with
```PHP
// .... some comments
$fieldHandles = [];
$fieldsWithTypes = [];
$fields = (new Query())
    //->distinct(true)
    ->from([Table::FIELDS])
    ->select([$column, 'type'])
    ->all();
// index it properly
foreach ($fields as $field){
    $index = $field['handle'];
    $fieldClass = $field['type'];
    $docType = $fieldClass::getPHPDocType();
    if(isset($fieldsWithTypes[$index]) && in_array($docType, $fieldsWithTypes[$index], true)){
        // we already have this field type in the array
        continue;
    }
    $fieldsWithTypes[$index][] = $docType;
}
// now that we have the unique field types for all contexts build the PHPDoc string
foreach ($fieldsWithTypes as $handle => $types){
    $fieldHandles[$handle] = implode('|', $types);
}
```


brandonkelly:
This is all merged into the 3.2 branch. Thanks! (Made several changes though—see 1c84c12)

Anubarak:
Love it, thank you very much. Sorry for forgetting about the missing component class that might cause exceptions🙈 I hope it didn't consume too much time.

khalwat:
OMG Yes.

brandonkelly:
Also worth noting that this only covers the field values;  I left the element query param types alone because those are almost always going to be different types (and generally a little more open-ended so I don’t think it’s worth actually implementing that method).

PR #4117, opened 2019-04-09
Fix setPasswordSuccessPath config name in docs

missmatsuko:
Set password form docs: <https://docs.craftcms.com/v2/templating/set-password-form.html>
Config docs: <https://docs.craftcms.com/v2/config-settings.html#setpasswordsuccesspath>


brandonkelly:
Thanks!

PR #4122, opened 2019-04-10
Exclude node_modules + hiddens for getTemplateSuggestions

ricricucit:
Skip hidden files and directories, and `node_modules` folder, if exists, since it OOM the PHP runtime using a large `node_modules` folder (AKA always). @brandonkelly maybe you know of better filtering

### DIFF
--- a/src/web/twig/variables/Cp.php
+++ b/src/web/twig/variables/Cp.php
@@ -434,7 +434,19 @@ class Cp extends Component
             return [];
         }
 
-        $iterator = new \RecursiveIteratorIterator(new \RecursiveDirectoryIterator($root));
+        $directory = new \RecursiveDirectoryIterator($root);
+
+        $filter = new \RecursiveCallbackFilterIterator($directory, function ($current, $key, $iterator) {
+          // Skip hidden files and directories,
+          // and `node_modules` folder, if exists, since it does OOM the PHP runtime
+          //(TODO: @brandonkelly maybe you know of better filtering)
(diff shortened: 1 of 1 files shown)


brandonkelly:
Why do you have a `node_modules` folder inside your `templates` folder?

ricricucit:
Well, that's the root of the my front-end. That's where I have my webpack config, .babelrc, other dotfiles, etc.

brandonkelly:
Your `templates` folder should not be inside your web root (nor should `node_modules`), and it should contain nothing but Twig templates.

ricricucit:
Fair enough, I read that in the docs, but then where do I store my development files for the front-end?

`/templates` is for twig only
`/web` is for assets (as in "public" folder I suppose)

Where would you suggest to put files like `webpack.config.js` or `.babelrc` or any other front-end dev resource that is _not_ a public one?

I used the `templates` folder because it seemed the most natural place. Having templating files together with front-end dev files seems like a good idea, as well as also having a _README.md_ that documents how to install front-end dependencies and run the front-end dev environment.

brandonkelly:
Generally you’d put those files at the root of your project, alongside `templates/`, `web/`, `composer.json`, etc.

ricricucit:
Generally, but what about this? https://github.com/craftcms/commerce/tree/develop/templates/shop

brandonkelly:
This came up again (#5991) and I guess it’s harmless, so just went ahead and merged it in. Thanks!

brandonkelly:
Craft 3.4.18 is out now with this change.

PR #4127, opened 2019-04-11
Abstract logging logic to trait for non-FileTarget logging

timkelty:
This abstracts the non-FileTarget specific logic to a trait, and uses the init methods of `FileTarget` and `LogTrait` to set things dynamically (e.g. from Craft config).

This lets someone make a new logger with all the Craft-added stuff (message formatting, ip logging, etc) as simple as:

```
class MyTarget extends \yii\log\SyslogTarget
{
    use \craft\base\LogTrait;
}
```

### DIFF
--- a/src/base/LogTrait.php
+++ b/src/base/LogTrait.php
@@ -1,0 +1,90 @@
+<?php
+/**
+ * @link https://craftcms.com/
+ * @copyright Copyright (c) Pixel & Tonic, Inc.
+ * @license https://craftcms.github.io/license/
+ */
+
+namespace craft\base;
+
+use Craft;
+use craft\helpers\ArrayHelper;
+use yii\helpers\VarDumper;
+use yii\web\Request;
+
+/**
+ * LogTrait implements the common methods and properties for log target classes.
+ *
+ * @author Pixel & Tonic, Inc. <support@pixelandtonic.com>
+ * @since 3.2
+ */
+trait LogTrait
+{
+    /**
+     * @var bool Whether the user IP should be included in the default log prefix.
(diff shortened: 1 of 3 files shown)


brandonkelly:
I don’t like the idea of FileTarget configuring itself. 

If you want to take advantage of the changes `craft\log\FileTarget` make over `yii\Log\FileTarget`, then just extend it.

timkelty:
> I don’t like the idea of FileTarget configuring itself.

Fair enough, the LogTrait part was really the important part. I reverted the init parts if that is more agreeable.

> If you want to take advantage of the changes craft\log\FileTarget make over yii\Log\FileTarget, then just extend it.

I can't, `craft\log\FileTarget` already extends `yii\Log\FileTarget`, but I don't want FileTarget as my base.
What I'm after is logging using `\yii\log\SyslogTarget` or `codemix\streamlog\Target`, but with the Craft niceties that are currently stuck in `craft\log\FileTarget`, even though they aren't really specific to file logging.

timkelty:
@brandonkelly I rebased my branch on latest `develop` and removed the `init` configuring: https://github.com/craftcms/cms/compare/develop...timkelty:log-trait

Apparently PRs don't get updated with new commits once they're closed.

If you still don't like this, let me know if you have other ideas (other than just copying `craft\log\FileTarget` wholesale into my own logger).

brandonkelly:
Ah, makes sense now. Just cleaned up some stuff and merged your `log-trait` branch. (fc54c0e04830d2def92d051949b76dc902088d10)

timkelty:
Awesome, I appreciate it!

PR #4131, opened 2019-04-12
CP: Disable (gif) animation *before* resizing

ostark:
Hey P&T,

Correct me if I'm wrong; I assume it's your intention to  prevent resizing all layers of an animated gif for CP thumbnails. Unfortunatly animations are disabled after the actual resize happens.


andris-sevcenko:
Thank you kind sir!

PR #4156, opened 2019-04-18
Allow widget icon to be an SVG string (not just a path)

Tam:
Just so I can have a _dynamic_ icon for one of my widgets. Very important. Totally not staggeringly unnecessary.

### DIFF
--- a/src/controllers/DashboardController.php
+++ b/src/controllers/DashboardController.php
@@ -557,6 +557,9 @@ class DashboardController extends Controller
         if ($iconPath === null) {
             return $this->_getDefaultWidgetIconSvg($widget);
         }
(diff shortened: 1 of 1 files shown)


brandonkelly:
I’m open to the idea but `iconPath()` would need to be renamed to `icon()` as well. I’ll go ahead and make that change for 3.2. In the meantime you would need to save out the icon into the `storage/` directory and return its path.

```php
use craft\helpers\FileHelper;

$path = Craft::$app->path->getStoragePath() . "/<your-widget-name>/{$this->id}.svg";
FileHelper::writeToFile($path, $svgContents);
return $path;
```

PR #4163, opened 2019-04-20
Schema.php chmod the DumpConfigFile to 0644

millerf:
Chmod the file. On some distribution, the php user write files too 'open', which generate a :
`[Warning] World-writable config file 'my.cnf' is ignored`
Therefore preventing the dump as it uses incorrect settings

### DIFF
--- a/src/db/mysql/Schema.php
+++ b/src/db/mysql/Schema.php
@@ -323,7 +323,10 @@ SQL;
         }
 
         FileHelper::writeToFile($filePath, $contents);
-
+        // Chmod the file. On some distribution, the php user write files too 'open', which generate a :
+        // [Warning] World-writable config file 'my.cnf' is ignored
+        // Therefore preventing the dump as it uses incorrect settings
+        chmod($filePath, 0644); 
         return $filePath;
     }
 }


brandonkelly:
This is what the [defaultFileMode](https://docs.craftcms.com/v3/config/config-settings.html#defaultfilemode) config setting is for. (See also [defaultDirMode](https://docs.craftcms.com/v3/config/config-settings.html#defaultdirmode).)

millerf:
@brandonkelly  I know, but `defaultFileMode` is not used in `writeTofile()`.
https://github.com/millerf/cms/blob/patch-1/src/helpers/FileHelper.php#L344

And I didn't want to create a big mess by modifying a "core" function such as `writeToFile()`.

What should I do?  Set the `defaultFileMode` instead of the `0644` here? Again, this is for a very specific reason  and only to access the `db` for dumping. The file is deleted at the end of the operation anyway....

brandonkelly:
Sorry, I misunderstood. All clear now.

brandonkelly:
Thanks for the PR!

millerf:
Thanks for taking the time to read it!!

PR #4176, opened 2019-04-24
add an example of saving an element with afterSave

sbatson5:
## What's in this PR?

I've been working on my first Craft Plugin with no prior experience, and there have been a couple hiccups.  For Elements, it wasn't very apparent to me immediately when `afterSave()` came into play and whether it was something I needed to handle myself.  I think having one quick blurb about how to save an element along with the explanation of `afterSave()` could be useful.  Otherwise, a new dev doesn't quite get when this comes into play and they've done a lot of setup with knowing how to execute it.

### DIFF
--- a/docs/extend/element-types.md
+++ b/docs/extend/element-types.md
@@ -110,31 +110,9 @@ public function afterSave(bool $isNew)
 }
 ```
 
-These methods will be called automatically when calling <api:craft\services\Elements::saveElement()> from a Service or Controller.  Here is an example of a method that would create our Element and save it within our ProductService:
-
-```php
-public static function saveNewProduct($model) {
-  $product = new JobElement([
-    'price' => $model->price,
-    'currency' => $model->currency
-  ]);
-
-  $db = \Craft::$app->getDb();
-  $transaction = $db->beginTransaction();
-  try {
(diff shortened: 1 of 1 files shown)


brandonkelly:
@sbatson5 Thanks for the PR! Does GitHub give you the ability to allow me to edit the PR code?

sbatson5:
@brandonkelly I could probably set you as an owner of the forked repo, but that workflow would seem a bit odd. I'm happy to make edits if you want to leave a review on the line or point me in the right direction.

brandonkelly:
@sbatson5 I know when you first create a PR, GitHub will give you the option, just for the PR branch. Just not sure if it’s possible to change that setting once the PR has already been created. If not, can you close the PR and recreate, with that option enabled?

sbatson5:
This has been checked since I opened the PR @brandonkelly.  Are you not able to edit?

Can I ask what changes you are looking to make?

Screen Shot 2019-04-26 at 8 30 47 PM

brandonkelly:
Sorry, not sure why the edit button wasn't showing up in the pull request page like it usually is, but I was able to edit it directly on your repo. All merged now! Thanks again for the PR

PR #4177, opened 2019-04-24
remove documentation link since it 404s

sbatson5:
## What's in this PR?

It looks like https://github.com/craftcms/docs doesn't exist and we get a 404.  I tried searching for a docs repo but couldn't find one.
It looks like most of the documentation lives in this repo, and the live doc site links to this repo for requested changes.


brandonkelly:
Thanks for pointing that out! Went with a different edit that points people to the `docs/` folder, plus explains how to get the VuePress dev sever up and running.

PR #4180, opened 2019-04-25
Add email and url types to table field

timkelty:
Adds the remaining simple string core field types, email and url, to the table field.


brandonkelly:
Merged for 3.2!

PR #4182, opened 2019-04-26
Update README.md with download info

fantasticmachine:
Add information about downloading pre-built Craft 2 version from CDN (lifted directly from https://craftcms.com/guides/downloading-previous-craft-versions)

Some people still need to be able to get Craft 2 and wind up at the Github page where they are faced with the need to run a build script. Since the previous versions are available from the CDN, it makes sense to include the instructions on how to get those within the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,3 +76,14 @@ The build script supports the following options:
 * Use HTML5 "placeholder" attributes for input and text fields.
 * Make sure to use <label> tags.
 
+## Download a pre-built version Craft 2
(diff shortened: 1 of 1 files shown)


brandonkelly:
We already link to the latest version of Craft 2 in the [installation instructions](https://docs.craftcms.com/v2/installing.html#step-1-upload-the-files) in the Craft 2 docs. This readme file was written well before Craft became a public repo and not meant to act as general documentation.

PR #4187, opened 2019-04-27
CRAFT_STORAGE_PATH target needs to exist

alexwybraniec:
Make it clearer that Craft performs a realpath() check on CRAFT_STORAGE_PATH and returns false if the directory does not exist. 

Currently, if the target of CRAFT_STORAGE_PATH doesn't exist, Craft just looks like it is ignoring CRAFT_STORAGE_PATH.


brandonkelly:
Thanks for the suggestion!

PR #4197, opened 2019-04-30
Throw a NotFoundHttpException when sendFile fails to find the file

johnnynotsolucky:
When `Response::sendFile` can't find the file on the system, it throws
an `ErrorException` which comes through to the client as a HTTP 500
error.

This can be verified when attempting to load a file in cpresources which
doesn't exist:

```bash
$ curl -sSL -D - http://craft.local/cpresources/eb39576b/does-not-exist.js -o /dev/null

HTTP/1.1 500 Internal Server Error
Date: Tue, 30 Apr 2019 09:10:50 GMT
Server: Apache/2.4.29 (Ubuntu)
X-Powered-By: Craft Commerce
Connection: close
Transfer-Encoding: chunked
Content-Type: text/html; charset=UTF-8
```

This commit adds code to catch the exception, determine whether it is
actually a file not found error, and throws the `NotFoundHttpException`,
otherwise, forwards the original exception on.

After this change:

```bash
$ curl -sSL -D - http://craft.local/cpresources/eb39576b/does-not-exist.js -o /dev/null

HTTP/1.1 404 Not Found
Date: Tue, 30 Apr 2019 09:14:28 GMT
Server: Apache/2.4.29 (Ubuntu)
X-Powered-By: Craft Commerce
Transfer-Encoding: chunked
Content-Type: text/html; charset=UTF-8
```

### DIFF
--- a/src/web/Response.php
+++ b/src/web/Response.php
@@ -9,6 +9,7 @@ namespace craft\web;
 
 use Craft;
 use yii\web\HttpException;
+use yii\web\NotFoundHttpException;
 
 /**
  * @inheritdoc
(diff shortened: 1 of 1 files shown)


brandonkelly:
We can’t safely assume that the file being sent by Response::sendFile() was directly requested by the user, so a 404 may not be appropriate there.

Application::_processResourceRequest() now explicitly checks if the requested resource file exists and if not, returns a 404, which resolves this.

PR #4208, opened 2019-05-02
Update Paginator.php

wei2go:
This Query class doesn't seem to be used in this file, and is causing this weird intermittent error: "PHP Compile Error – yii\base\ErrorException, Cannot use yii\db\Query as Query because the name is already in use." Error happens when using {% paginate %} in Twig template.

Since this class is not really in use in this file, can you consider removing it?

### DIFF
--- a/src/db/Paginator.php
+++ b/src/db/Paginator.php
@@ -10,7 +10,7 @@ namespace craft\db;
 use craft\helpers\ArrayHelper;
 use yii\base\BaseObject;
 use yii\db\Connection as YiiConnection;
-use yii\db\Query;
+// use yii\db\Query;
 use yii\db\QueryInterface;
 use yii\di\Instance;
 


brandonkelly:
Thanks for pointing that out! The class was being referenced, but only by a PHP comment, so I’ve namespace it instead.

Strange that it would only intermittently cause an issue. Are you running different versions of PHP on different servers perhaps?

wei2go:
Hi Brandon,

Thanks for fixing this.

The reason I said it was intermittent is because I was trying to debug it, and then it fixed by itself after I rolled back all my debugging efforts. So I guess there could be some cache thing? But I don't cache the pagination part of template though. Weird. Anyway...

brandonkelly:
Hm yeah maybe an OPcache thing or something, not sure!

PR #4211, opened 2019-05-03
add getLinkText method to element types

sbatson5:
## What's in this PR?

I'm currently learning craft while building a plugin for my company.  Working with elements, there have been a few things I've run into that haven't seemed immediately obvious.  When editing elements via the provided template:
```twig
{% extends '_layouts/elementindex' %}
{% set title = 'Products' %}
{% set elementType = 'ns\\prefix\\elements\\Product' %}
```

I noticed the first column wasn't showing the id as I expected.  It wasn't until I stumbled on https://github.com/craftcms/cms/commit/fef75e196b9a96c4fb514e491c4f4ce0b89bb1fe#diff-0cf13e6e9757a569110e508f362ad1a9R482
that I realized I could update `__toString()` to return what I wanted.

This felt non-obvious and I felt like I was potentially overwriting a method that could be used elsewhere.  So, I thought it would be nice to have a convenience method `getLinkText()` (happy to rename it if people feel this name is no good).

This method would just return `$this->__toString()` by default, so it shouldn't break any existing plugins that overwrite `__toString()` already to accomplish this.  But it would allow us to update just that text without overwriting the method we get from the PHP base class.


src/base/Element.php
```diff
@@ -1263,14 +1263,22 @@ public function getLink()
...
 
         return null;
     }
 
+    /**
+     * @inheritdoc
+     */
+    public function getLinkText()
+    {
+        return $this->__toString();
```
brandonkelly: Typecasting the element to a string is better than calling `__toString()` directly.
```suggestion
        return (string)$this;
```

src/base/ElementInterface.php
```diff
@@ -433,6 +433,13 @@ public function getUrl();
      */
     public function getLink();
 
+    /**
+     * Returns a string to be used in for links to edit this element.
+     *
+     * @return string|null
+     */
+    public function getLinkText();
```
brandonkelly: Should be renamed to `getName()`, and `craft\elements\User::getName()` should switch over to an `@inheritdoc` comment.

PR #4213, opened 2019-05-04
Allow plugins and modules to change the filename of replaced assets

benface:
This PR adds the ability for code hooking into the `EVENT_BEFORE_REPLACE_ASSET` event to change the filename of an asset that's being replaced. This is useful when using a service like Imgix whose remote cache takes some time to get purged, to make sure a fresh version of the asset is fetched right after saving changes made in the image editor (which triggers that event) or replacing the image manually.


andris-sevcenko:
Just add a changelog entry and we'll be good to go!

benface:
@andris-sevcenko Done!

andris-sevcenko:
Thank you!

sgtpenguin:
Apologies if this is not the right place to ask this question, but it seems relevant!

I'm in a situation now where a client wants replaced assets to use the original file's name, regardless of the fact that they are uploading a file with a different name, so that the url remains the same, to preserve a "permalink" to the file.  This behavior exists in the Wordpress Enable Media Replace plugin.

I was curious about creating a plugin/module to do this.  It seems to me like this PR could open up that functionality, right?  In this event, would we have access to the old file name?

Thanks for any insight you have!

brandonkelly:
@sgtpenguin Yep, that’s doable! Here’s the code to get that working, which you could put in a custom module:

```php
use yii\base\Event;
use craft\services\Assets;
use craft\events\ReplaceAssetEvent;

Event::on(
    Assets::class,
    Assets::EVENT_BEFORE_REPLACE_ASSET,
    function(ReplaceAssetEvent $e) {
        $e->filename = $e->asset->filename;
    }
);
```

sgtpenguin:
Amazing.  Thank you @brandonkelly.

PR #4214, opened 2019-05-04
Fix bug in migration 3.2.0-alpha.5

putyourlightson:
This pull request fixes the following exception that is thrown if user groups exist in project config without any permissions set in 3.2.0-alpha.5.

```
yii\base\ErrorException: in_array() expects parameter 2 to be array, null given

Migration: craft\migrations\m190417_085010_add_image_editor_permissions

Output:

Exception: in_array() expects parameter 2 to be array, null given (/vendor/craftcms/cms/src/migrations/m190417_085010_add_image_editor_permissions.php:58)
#0 /vendor/craftcms/cms/src/migrations/m190417_085010_add_image_editor_permissions.php(58): ::in_array('???', '???', '???')
```


brandonkelly:
Thanks for the PR!

PR #4219, opened 2019-05-06
Optimise Job Progress

putyourlightson:
The current progress is never updated, resulting in the database being updated regardless of whether the value of `$this->_progress` has changed or not.

Additionally, `$this->_progress` should be stored as a rounded integer to avoid updating the database more times than necessary. For example, if there are 1,000 steps then the database will be update 1,000 times instead of the expected 100.


narration-sd:
Great catch, Ben @putyourlightson

brandonkelly:
Thanks!

PR #4237, opened 2019-05-09
Add htmlTagAttributes filter to Twig

timkelty:
…because `yii\helpers\BaseHtml::renderTagAttributes` is awesome!

Back in the day I wrote a Craft 2 plugin for this: https://github.com/timkelty/htmlattributes-craft

Now that it can be added to Craft with a couple lines of code, seems like a big win.

Example usage:

```twig
{% set attributes = {
      class: ['one', 'two'],
      disabled: true,
      readonly: false,
      data: {
        baz: 'Escape this "',
        qux: {
          some: ['data', '"quoted"']
        }
      },
      style: {
        'background-color': 'red',
        'font-size': '20px'
      },
} %}

<div {{ attributes|htmlTagAttributes }}></div>
```

### DIFF
--- a/src/web/twig/Extension.php
+++ b/src/web/twig/Extension.php
@@ -16,6 +16,7 @@ use craft\helpers\ArrayHelper;
 use craft\helpers\DateTimeHelper;
 use craft\helpers\Db;
 use craft\helpers\FileHelper;
+use craft\helpers\Html;
 use craft\helpers\Json;
 use craft\helpers\Sequence;
 use craft\helpers\StringHelper;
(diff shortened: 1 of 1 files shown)


brandonkelly:
Couple things:

- The PR should target the `3.2` branch instead of `develop`.
- It should probably be a Twig _function_ rather than filter, since it’s not really modifying the array, as much as generating something completely new using the array for configuration.
- It should probably just be called `attr()`.

timkelty:
All set, @brandonkelly!

- [x] The PR should target the 3.2 branch instead of develop.
- [x] It should probably be a Twig function rather than filter, since it’s not really modifying the array, as much as generating something completely new using the array for configuration.
- [x] It should probably just be called attr().

brandonkelly:
Looks great, thanks!

PR #4240, opened 2019-05-09
Fix @return doc error for getElementUriForSite()

chasegiunta:
Also can return `false`
https://github.com/yiisoft/yii2/blob/master/framework/db/Command.php#L419-L420


brandonkelly:
Thanks Chase!

Issue #3931, opened 2019-02-28
Allow element fields to filter "sources with URLs"


timkelty:
It would be great if there were an added field setting checkbox for "Only Sources with URLs".

This could in the "Advanced" settings perhaps, so you could check "All" + "Only Sources with URLs" to allow any sources with URLs.

PR #4242, opened 2019-05-10
Abstract structure logic away from categories field, allow structure selections for entries

### LINKED ISSUE #2632: Option in Entries field settings to make it relate ancestors
Basically as a first step for https://github.com/craftcms/cms/issues/1949, “Make Categories = Structures”.

timkelty:
Fixes #2632 

Just after I pushed I saw that you already had #2632 on the 3.2 roadmap, but oh well. Take my codez if you like :)

The approach I took was to abstract all the structure-based stuff from `craft\fields\Categories`, to `BaseStructureRelationField`, and then add a new "Structure Entries" fieldtype.

Now `craft\fields\Categories` and `craft\fields\StructureEntries` can extend a common interface…which also means people can supplant Categories with Structure + Structure Entries fields to get the best of both worlds. 👯‍♀️

---

Alternatively, you could add options to the existing Entries field, which is how I started. I ended up going with a separate base class and field as it seemed cleaner – I didn't have to introduce attributes like `relateParents`, `allowRelateParents`, `structure`, etc. Also, since `relateParents` only really makes sense with a single input source that is a structure, figuring out how to toggle all that on the entries field setting template seemed needless complex.


brandonkelly:
I’ll take a look at this when I’m ready to look into #2632, and see if there’s anything we can use.

timkelty:
@brandonkelly sounds good.

Ultimately I think it would be better with a `relateParents` setting on the existing Entries field (which is [where I started](https://github.com/craftcms/cms/pull/4242/commits/630a6c598f0a5e8a29422d3b33e8827b319b1136)), but that seemed more invasive to the codebase, so I figured I'd leave that to you.

UI-wise, this might be done with a checkbox toggle, similar to the "Restrict uploads to a single folder?" on Assets fields.

It also might be nice to consider other features (https://github.com/craftcms/cms/issues/3931) where there is an initial selection that changes the available sources. If there were an initial dropdown that toggled the sources below with options like "All Entries", "Entries with URLs", "Structures", etc. that might work well.,

PR #4255, opened 2019-05-13
Allow conditionally empty uri patterns

### LINKED ISSUE #4254: Allow Entry URI Format to be conditionally blank (no URL)
Several times we've wanted a _fully_ dynamic "Entry URI Format", where conditionally, some entries in a section (usually by depth), don't have URLs.

Making this "Entry URI Format" something like this has a peculiar effect:
```
{% if level == 1 %}projects/{slug}{% endif %}
```

I want entries with `level` 2 to have _no_ url, but instead they seem to get a homepage URL (even if there is already a homepage single). Subsequent level 2 entries will fail because of the occupied URL.

timkelty:
Fixes #4254

This allows the URI format to render to empty, so that entries in the same section may optionally omit URLs.

Screen Shot 2019-05-13 at 10 40 00 AM

Screen Shot 2019-05-13 at 10 38 34 AM

### DIFF
--- a/src/helpers/ElementHelper.php
+++ b/src/helpers/ElementHelper.php
@@ -54,9 +54,11 @@ class ElementHelper
     {
         /** @var Element $element */
         $uriFormat = $element->getUriFormat();
+        $testUri = self::_renderUriFormat($uriFormat, $element);
 
         // No URL format, no URI.
-        if ($uriFormat === null) {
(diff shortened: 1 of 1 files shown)


brandonkelly:
Thanks for the PR! Took the commit in a _slightly_ different direction though.

PR #4260, opened 2019-05-14
Add doesTableExist MigrationHelper function

johnnynotsolucky:
When uninstalling a plugin, sometimes we might try dropping foreign keys on a
table that doesn't exist yet. This function lets us check whether or not
that table exists yet before dropping foreign keys.

For example, in `Install::safeDown()` we could have this which would throw if the table hadn't been created yet:

```php
MigrationHelper::dropAllForeignKeysToTable('{{%table_does_not_exist}}', $this);
$this->dropTableIfExists('{{%table_does_not_exist}}');
```

With this, we can check if the table exists first:

```php
if (MigrationHelper::doesTableExist('{{%table_does_not_exist}}', $this) {
    MigrationHelper::dropAllForeignKeysToTable('{{%table_does_not_exist}}', $this);
    $this->dropTable('{{%table_does_not_exist}}');
}
```

### DIFF
--- a/src/helpers/MigrationHelper.php
+++ b/src/helpers/MigrationHelper.php
@@ -59,6 +59,24 @@ class MigrationHelper
     }
 
     /**
+     * Returns whether a table exists.
+     *
+     * @param string $tableName
+     * @return bool
+     */
+    public static function doesTableExist(string $tableName, Migration $migration = null)
+    {
+        $db = $migration ? $migration->db : Craft::$app->getDb();
+        $schema = $db->getSchema();
+        $schema->refresh();
+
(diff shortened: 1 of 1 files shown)


brandonkelly:
You could already use [craft\db\Connection::tableExists()](https://docs.craftcms.com/api/v3/craft-db-connection.html#method-tableexists) for that:

```php
if ($this->db->tableExists('{{%table_does_not_exist}}')) {
    MigrationHelper::dropAllForeignKeysToTable('{{%table_does_not_exist}}', $this);
    $this->dropTable('{{%table_does_not_exist}}');
}
```

PR #4264, opened 2019-05-15
Update installation.md

lasseklein:
As a first time Craft user and first time installer it took me quite a while to figure out where to best place the Craft files. I have proposed a way of making it clearer to other first time users. Please feel free to use, change it or disregard it.


lasseklein:
I changed from a comment under terminal to a general tip since this applies both to manual installation and composer installation.

brandonkelly:
Welcome to Craft, and thanks for the PR!

There’s already a section explaining the “Directory Structure” section just below the “Downloading an Archive File Manually” instructions, so I put a tip there that explains what `web/` is and how you can rename it, which seemed like a better place to put it.

lasseklein:
Sounds great!

On my second day of getting into Craft just now, and I’m blown away. This is such a great product! Thank you!

Lasse

PR #4273, opened 2019-05-16
Improved Project Config behaviour when dealing with deferred events.

andris-sevcenko:


src/services/ProjectConfig.php
```diff
@@ -987,6 +972,70 @@ private function _applyChanges(array $changes)
...
+                // a) The counter for at least one path decreases
+                // b) A path is eliminated
+                foreach ($startingDeferCounts as $path => $deferCount) {
+                    if (empty($deferCounts[$path]) || $deferCounts[$path] < $deferCount) {
+                        $movingAlong = true;
+                    }
+                }
+
+                // It's not progressing, so increase the defer counter
+                if (!$movingAlong && ++$deferCounter < $this->maxDefers) {
```
brandonkelly: Shouldn’t this be `>` instead of `<` ?

src/services/ProjectConfig.php
```diff
@@ -987,6 +972,70 @@ private function _applyChanges(array $changes)
...
+                    throw new OperationAbortedException($message);
+                }
+            }
+
+            // This will either set this for the first time or update with the new values
+            $startingDeferCounts = $deferCounts;
+
+            // Try to process all events
+            // Make a copy, so we can modify the original array
+            $deferredEvents = $this->_deferredEvents;
```
brandonkelly: This won’t actually make a copy of the array; just a new reference to it. To create an actual copy of the array, use `array_merge($this->_deferredEvents)`. Though it’s not clear to me why you want to make a copy anyway though? Seems like the condition on the `while` loop on line 985 is expecting `$this->_deferredEvents` to be updated.

andris-sevcenko: > This won’t actually make a copy of the array; just a new reference to it. 
Right, but as soon as the modification happens on line 1028, a copy will be created, because it's an array not an object. I think.

> Though it’s not clear to me why you want to make a copy anyway though? Seems like the condition on the while loop on line 985 is expecting $this->_deferredEvents to be updated.

It's so that I'm able to modify the original array on line 1028 without disrupting the `foreach` loop.

src/services/ProjectConfig.php
andris-sevcenko: Yeah, looks like it.

src/services/ProjectConfig.php
brandonkelly: Nah, a modification will still just affect the one array, and both references (`$deferredEvents` and `$this->_deferredEvents` will reflect that change.

andris-sevcenko:
Closed in favor of https://github.com/craftcms/cms/commit/54a1c4f52a44f55c66775e3317604d8d406572bb

PR #4282, opened 2019-05-17
Add a new user permission, "customizeSources"

abryrath:
Non-admin clients would like the ability to show/hide certain table columns/sources. This patch allows non-admin users, with this new permission enabled, to see the cog wheel in an element index page and save their preferences.


brandonkelly:
Thanks for the PR! Merged for the next release.

PR #4284, opened 2019-05-19
Prevent exception in console request

putyourlightson:
When attempting to render a template that contains `{% cache %}` tags in a console request, an exception is thrown because the `craft\console\Request::getPathInfo` method does not exists.

```
Exception 'Twig\Error\RuntimeError' with message 'An exception has been thrown during the rendering of a template ("Calling unknown method: craft\console\Request::getPathInfo()").'
```

This pull request prevents the `getPathInfo` method from being called (and therefore the exception being thrown) if `$global` is false and the request is a console request.

### DIFF
--- a/src/services/TemplateCaches.php
+++ b/src/services/TemplateCaches.php
@@ -119,6 +119,11 @@ class TemplateCaches extends Component
             return null;
         }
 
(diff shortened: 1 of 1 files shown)


brandonkelly:
Thanks Ben!

PR #4289, opened 2019-05-20
Add closing bracket

ademers:


brandonkelly:
Thanks Andrea!

That file is actually generated from AssetQuery.php, so I just edited there and pushed the fix myself.

PR #4304, opened 2019-05-23
Enforce requirement of authenticated Craft session

putyourlightson:
This pull request enforces the requirement of an authenticated Craft session by removing anonymous access to the controller actions, as specified in the comment on line 25. If this was intentional then the change should be reverted and line 25 should be updated accordingly.

> Note that all actions in the controller require an authenticated Craft session via [[allowAnonymous]].


brandonkelly:
The `export/export` action requires a token so it’s safe to allow anonymous access.

PR #4309, opened 2019-05-24
ArrayValidator clean up

timkelty:
I was extending ArrayValidator and this threw me for a loop…I certainly could be missing something, but I couldn't figure out why it was written this way.

- The validation logic was duplicated in both validateAttribute and validateValue
- validateValue was actually never called, as it is `protected` and not called from `validateAttribute` (unless something else extended ArrayValidator)
- the `!is_string($value)` check in `validateAttribute` seemed backward

### DIFF
--- a/src/validators/ArrayValidator.php
+++ b/src/validators/ArrayValidator.php
@@ -102,36 +102,12 @@ class ArrayValidator extends Validator
     /**
      * @inheritdoc
      */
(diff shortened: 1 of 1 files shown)


brandonkelly:
Thanks! No idea what I was thinking there with `validateValue()` 🤷‍♂️

PR #4311, opened 2019-05-24
Add optional callback for ArrayValidator.

timkelty:
This may be a stretch, as I'm not sure there is an immediate use for it, but I found it helpful authoring a plugin.

If not, at least I can extend ArrayValidator and add it now that #4309 is fixed.

Example usage (validating an attribute is an array of `MyModel`s):

```
    public function rules()
    {
        return [
            ['myModels', ArrayValidator::class, 'callback' => function ($value) {
                return $value instanceof MyModel;
            }],
        ];
    }
```

### DIFF
--- a/src/validators/ArrayValidator.php
+++ b/src/validators/ArrayValidator.php
@@ -49,6 +49,13 @@ class ArrayValidator extends Validator
     public $min;
 
     /**
+     * @var callable|null callback validation for each element in the array
+     * @see invalidElements for the customized message for invalid elements.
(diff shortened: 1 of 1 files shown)


brandonkelly:
Hm yeah I don’t think there’s any precedence for this. You could always just check for errors on the model attribute after validate and do whatever you need to do with them from there, or write your own validation method.

timkelty:
Yep, sounds good 🤠.

PR #4323, opened 2019-05-28
Allow Titles to be published from draft

linchpinstudios:
**Currently**
If you set a `$draft->title` on when it is published it gets replaced with the `getSection()->name` on a `Section::TYPE_SINGLE` entry. 

**Changes**
Checks for a `$draft->title` on `Section::TYPE_SINGLE` and only replaces it wit the `getSection()->name` if there isn't a title provided.


brandonkelly:
Thanks for pointing this out! That code actually hasn’t been necessary since before Craft 2.2, when Single entries gained the ability to have actual titles. Just removed it for the next release.

PR #4327, opened 2019-05-29
Added fields translation method explanation - Update sites.md

giacomocarra:
Translation method integration is reported as explained by @Anubarak 
Thank you.


brandonkelly:
Thanks for the suggestion! Added similar info to the [Fields](https://docs.craftcms.com/v3/fields.html#translation-methods) page where it belongs, and now the [Sites](https://docs.craftcms.com/v3/sites.html#step-4-define-translation-method-of-fields) page links to that.

PR #4337, opened 2019-05-30
Add method for element tabs to use on elementsindex template

sbatson5:
## What's in this PR?

This PR adds a method for `getElementEditTabs` to elements, allowing plugin developers to easily show responsive tabs for their custom elements.  Previously, if I had a custom element with multiple fields/tabs, I would have to manually write something to return those tabs when rendering my edit template and ensure they had a `label`, `url` and `class`.

I had the assumption that I could just use `$element->getTabs()` but that has to be cleaned before the template will properly render it.

With this, you can easily do this for an edit template
```
return $this->renderTemplate(
    'plugin-name/element/_edit',
    [
        'element' => $element,
        'id' => $id,
        'tabs' => $element->getElementEditTabs(), // get the tabs pre-formatted
        'title'              => 'Edit Element',
        'continueEditingUrl' => 'plugin-name/element/{handle}'
    ]
);
```

This also reduces some duplicated logic on the Entries and Categories controllers, which do this manual tab creation.

### DIFF
--- a/src/base/Element.php
+++ b/src/base/Element.php
@@ -2357,6 +2357,37 @@ abstract class Element extends Component implements ElementInterface
         }
     }
 
+    /**
+     * @inheritdoc
+     */
+    public function getElementEditTabs(): array
+    {
+        $tabs = $this->getFieldLayout()->getTabs();
+
+        return array_map(function($tab) {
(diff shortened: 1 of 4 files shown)


sbatson5:
@brandonkelly I added a quick example of how to use this. Let me know if you think this a good approach for easily adding tabs to element edit page

PR #4339, opened 2019-05-30
add doc example of actions with messaging

sbatson5:
## What's in this PR?

This is a slight documentation update to add an example of defining actions with confirmation and success messages.  I only discovered this was possible by looking at the source code of another plugin and I feel most bulk actions would want a confirmation message.

### DIFF
--- a/docs/extend/element-types.md
+++ b/docs/extend/element-types.md
@@ -487,6 +487,29 @@ protected static function defineActions(string $source = null): array
 }
 ```
 
+If you would like warning messages before any action, you can specify them within `defineActions()` rather than just returning a flat array.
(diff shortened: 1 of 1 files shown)


brandonkelly:
`craft\base\ElementAction` doesn’t have `confirmationMessage` or `successMessage` properties; `craft\base\ElementActionInterface` just has [getConfirmationMessage()](https://docs.craftcms.com/api/v3/craft-base-elementactioninterface.html#method-getconfirmationmessage) and [getMessage()](https://docs.craftcms.com/api/v3/craft-base-elementactioninterface.html#method-getmessage) (already documented at those URLs), and how your element action chooses to implement them (if at all) is up to you.

PR #4348, opened 2019-06-03
Adding ability to override get modal action and index settings

nateiler:
This PR adds the ability to override the action endpoint to retrieve the modal body and allows overriding index settings.

Note: There isn't a compressed version of this change.  I figured ya'll could run it through your automated process.


brandonkelly:
Sorry for the delay on this! Now merged for the next Craft 3 release.

brandonkelly:
Craft 3.7.38 is out now with that change.

PR #4350, opened 2019-06-03
Handle disabled element index ids as string (as well as array)

nateiler:
When sending a lot (thousands) of disabled element Ids via array, it's possible to exceed the `max_input_vars` limits.  Sending as comma delimited string alleviates this.


brandonkelly:
@nateiler Sorry, this fell through the cracks. Looks like the PR doesn’t include any JS changes though?

PR #4356, opened 2019-06-04
Add cacheCleanupDuration config setting

olivierbon:


brandonkelly:
Decided it makes more sense to delete expired template caches as part of garbage collection, rather than the template caches service needing to keep track of when to do it.

PR #4396, opened 2019-06-12
Make basic Craft email work better with dark modes

mildlygeeky:
I use dark mode on the Mac, and found that (at least for Spark Mail), having a hardcoded HTML email body text color was causing HTML emails coming from Craft to be difficult to read - could become more of an issue when iOS 13 and its dark mode come:

image

Thinking that leaving it up the email client will provide the best compatibility for coping with dark mode settings, or any other manual overrides a user might have (the same might go for setting pixel-specific font size and line-heights, but will leave that up to you).


brandonkelly:
Thanks for the PR!

PR #4400, opened 2019-06-13
Ignore yii\i18n\PhpMessageSource error and warning log messages. 

### LINKED ISSUE #4109: Missing translation categories should not be logged as Error
If you are in the CP using a non-english lang, your logs will be littered with tons of errors:

```
2019-04-08 14:21:33 [-][1][-][error][yii\i18n\PhpMessageSource::loadFallbackMessages] The message file for category 'relabel' does not exist: /app/vendor/composer/anubarak/craft-relabel/src/translations/fr-CA/relabel.php Fallback file does not exist as well: /app/vendor/composer/anubarak/craft-relabel/src/translations/fr/relabel.php
```

…which can be annoying when you're looking for actual errors. Seems like falling back is the expected behavior, and shouldn't be logged as an error.

gtettelaar:
Resolves #4109.


brandonkelly:
Thanks!

PR #4401, opened 2019-06-13
Add the userImpersonation setting

### LINKED ISSUE #1186: Add impersonation user permission
> Created by: Anonymous (gijs@itmundi.nl) on 2015/03/16 11:59:50 +0000
> Votes at time of UserVoice import: 2

Currently only super admins are able to use impersonate, but its also useful to other users.

gtettelaar:
Resolves #1186 

I didn't add a warning message - albeit It probably would be appropriate. Not really sure if they have to be translated first...


gtettelaar:
Think i made a mistake in git. I'll re-open this one with a 3.2 base.

Issue #3501, opened 2018-11-28
Feature request: make 'login as user' a permission (per user group)


nilsenpaul:
### Description
I would like one of my editors to be able to login as another user, but I do not want to make him an admin. Seems like this is now impossible. Would be great if you could set a permission for this, would be even better if this could be done per user group.

brandonkelly:
An “Impersonate users” permission has just been added for the next 3.2 Beta release.

It doesn’t take user groups into account, however even if a user has that permission, a “Login as user” option will only show up for users with equal or fewer permissions assigned to them, so there’s no possibility of user impersonation being used as a permission escalation vector.

moreguppy:
Hi @brandonkelly, I was going to create a new issue but this seems related? 

### Description

- I have 2 User Groups — Editor and Artist
- The Editor has permissions set to 'Impersonate User'
- I log in as the Editor and try to impersonate an Artist *but I can't see the option in the dropdown menu on the User page in the backend*
- However, as an Admin I __can__ impersonate other users
- The 'Editor' User Group has more permissions checked than the 'Artist' User Group

### Additional info

- Craft version: Craft Pro 3.3.14
- PHP version: 7.3.10
- Database driver & version: MySQL 5.7.26
- Plugins & versions: Amazon S3 (1.2.5), Bulk Edit (1.1.1), FeedMe (4.1.2), Redactor (2.4.0)

brandonkelly:
@moreguppy if the artist has _any_ permissions that the editor doesn’t have (either directly or via the user group), then impersonation wan’t be allowed. If you’re sure that’s the case, please send in your `composer.json` and `composer.lock` files, and a database backup, over to support@craftcms.com and we can look into it from there.

moreguppy:
@brandonkelly thanks for the help, that worked.

Perhaps under 'Impersonate Users' checkbox, there could be sub-checkboxes to pick which user groups they can impersonate?

brandonkelly:
@moreguppy Part of the reason we didn’t give user group-specific permissions is, there’s still the possibility that a user has some permissions set directly on their account, so even if they are in the permitted user group, you may want them to be excluded from impersonation. Which would just add another layer of confusion like you had earlier. (“User A has permission to ‘Impersonate users’ *and* ‘Impersonate users in the Artists group’, and yet they *still* can’t impersonate my Artist user - WTF!”)

[...]

PR #4403, opened 2019-06-13
Add the impersonateUsers User permission

### LINKED ISSUE #1186: Add impersonation user permission

gtettelaar:
Doh! Of all things, I didn't test #4401 via the CP - I changed the permission check in `actionEditUser()` to reflect the new permission so that the `Login as {user}` dropdown option is visible. Sorry bout that! 

Of what I can see `craft\web\User` has two references to the Impersonate functionality: `Line 306` & `Line 406` but they both deal with "If a user is using Craft whilst impersonating someone" not "Is this user allowed to impersonate other users" the latter is where this permission makes changes.

Resolves #1186

### DIFF
--- a/CHANGELOG-v3.md
+++ b/CHANGELOG-v3.md
@@ -6,6 +6,8 @@
 - Added `craft\base\ElementInterface::getIsDraft()`.
 - Added `craft\base\ElementInterface::getIsRevision()`.
 - Added the `impersonateUsers` user permission.
+- Added `craft\services\UserPermissions::doesFirstUserHaveMorePermissions()`
+- Added `craft\services\UserPermissions::isImpersonationAllowed()`
 
 ### Fixed
 - Fixed a bug where entry drafts weren’t getting updated slug values once their initial slug had been saved, if their entry type had a custom title format. ([#4373](https://github.com/craftcms/cms/issues/4373))
--- a/CHANGELOG-v3.md
+++ b/CHANGELOG-v3.md
@@ -3,11 +3,10 @@
 ## Unreleased (3.2)
 
 ### Added
+- Added the “Impersonate users” permission. ([#3501](https://github.com/craftcms/cms/issues/3501))
 - Added `craft\base\ElementInterface::getIsDraft()`.
 - Added `craft\base\ElementInterface::getIsRevision()`.
-- Added the `impersonateUsers` user permission.
-- Added `craft\services\UserPermissions::doesFirstUserHaveMorePermissions()`
-- Added `craft\services\UserPermissions::isImpersonationAllowed()`
+- Added `craft\services\Users::canImpersonate()`.
 
 ### Fixed
 - Fixed a bug where entry drafts weren’t getting updated slug values once their initial slug had been saved, if their entry type had a custom title format. ([#4373](https://github.com/craftcms/cms/issues/4373))
--- a/CHANGELOG-v3.md
+++ b/CHANGELOG-v3.md
@@ -5,6 +5,7 @@
 ### Added
 - Added `craft\base\ElementInterface::getIsDraft()`.
 - Added `craft\base\ElementInterface::getIsRevision()`.
+- Added the `impersonateUsers` user permission.
 
 ### Fixed
 - Fixed a bug where entry drafts weren’t getting updated slug values once their initial slug had been saved, if their entry type had a custom title format. ([#4373](https://github.com/craftcms/cms/issues/4373))
--- a/src/controllers/UsersController.php
+++ b/src/controllers/UsersController.php
@@ -180,7 +180,7 @@ class UsersController extends Controller
     public function actionImpersonate()
     {
         $this->requireLogin();
-        $this->requireAdmin(false);
+        $this->requirePermission('impersonateUsers');
         $this->requirePostRequest();
 
         $userSession = Craft::$app->getUser();
@@ -697,7 +697,7 @@ class UsersController extends Controller
             }
 
             if (!$isCurrentUser) {
-                if ($userSession->getIsAdmin()) {
+                if ($userSession->checkPermission('impersonateUsers')) {
                     $sessionActions[] = [
                         'action' => 'users/impersonate',
                         'label' => Craft::t('app', 'Login as {user}', ['user' => $user->getName()])
--- a/src/controllers/UsersController.php
+++ b/src/controllers/UsersController.php
@@ -176,6 +176,7 @@ class UsersController extends Controller
      * Logs a user in for impersonation. Requires you to be an administrator.
      *
      * @return Response|null
+     * @throws ForbiddenHttpException
      */
     public function actionImpersonate()
     {
@@ -188,15 +189,11 @@ class UsersController extends Controller
 
         $userId = $request->getBodyParam('userId');
 
-        // Are they actually allowed to impersonate this user?
-        $currentUser = $userSession->getIdentity();
-
-        // Fast track people who are admins - so we dont have to call any DB.
-        if (!$currentUser->admin) {
-            $desiredImpersonationSubject = Craft::$app->getUsers()->getUserById($userId);
-            if (!Craft::$app->getUserPermissions()->isImpersonationAllowed($currentUser, $desiredImpersonationSubject)) {
-                throw new ForbiddenHttpException('You do not have suffecient priveleges to impersonate this user');
-            }
+        // Make sure they're allowed to impersonate this user
+        $usersService = Craft::$app->getUsers();
+        $impersonatee = $usersService->getUserById($userId);
+        if (!$usersService->canImpersonate($userSession->getIdentity(), $impersonatee)) {
+            throw new ForbiddenHttpException('You do not have sufficient permissions to impersonate this user');
         }
 
         // Save the original user ID to the session now so User::findIdentity()
@@ -707,7 +704,7 @@ class UsersController extends Controller
             }
 
             if (!$isCurrentUser) {
-                if (Craft::$app->getUserPermissions()->isImpersonationAllowed($currentUser, $user)) {
+                if (Craft::$app->getUsers()->canImpersonate($currentUser, $user)) {
                     $sessionActions[] = [
                         'action' => 'users/impersonate',
                         'label' => Craft::t('app', 'Login as {user}', ['user' => $user->getName()])
--- a/src/controllers/UsersController.php
+++ b/src/controllers/UsersController.php
@@ -180,7 +180,6 @@ class UsersController extends Controller
     public function actionImpersonate()
     {
         $this->requireLogin();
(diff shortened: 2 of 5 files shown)


brandonkelly:
We should also limit user impersonation to users with less permissions than the current user, OR once you’re logged in as that user, factor in the real user’s permissions when doing permission checks. Otherwise this could introduce a permission escalation vector. (All it would take is being granted this one permission, and you basically become an admin, because you could impersonate an admin, and then grant yourself admin status, or start doing things you’re generally not permitted to do, etc.)

gtettelaar:
> We should also limit user impersonation to users with less permissions

+1.
 
In `craft\elements\User::can()` I can implement an additional check to ensure the user is not punching above their weight when doing something through an impersonated session. 

It's probably also worth forcing admins who assign this permission to limit the assigned impersonation rights to users in specific user groups. So that the default is not _all_ users. (I.E. An editor can only impersonate people in the `authors` group e.t.c.). Just to restrict it a little bit further.
 
I'll finish this over the weekend for you to have a look at and drop a message once it's done. I'll label the PR as [WIP]. Thanks! @brandonkelly

brandonkelly:
> It's probably also worth forcing admins who assign this permission to limit the assigned impersonation rights to users in specific user groups. So that the default is not all users. (I.E. An editor can only impersonate people in the authors group e.t.c.). Just to restrict it a little bit further.

I don’t think we need to worry about user _groups_ at all, just permissions.

When determining if User A should be allowed to impersonate User B, we should check the following:

- If User A is an admin, then we can immediately determine that they’re allowed and skip other checks.
- If User B is an admin (and User A isn’t), they are disallowed.
- If User A doesn’t have the `impersonateUsers` permission, they are disallowed.
- If User A doesn’t have all the same permissions as User B (whether granted directly or via user groups), they are disallowed.
- Otherwise, they are allowed.

> Coincidentally this would also resolve #3501

Thanks for bringing that one up, I didn’t realize there were two issues asking for basically the same thing. Just closed #1186 in favor of #3501. (And the user groups thing was just a suggestion in #3501. If we implement this based on permissions only, it will still be resolved.)

gtettelaar:
That's a fair point. Removing the user groups from the equation makes this easier for me to implement. The permission check will resolve any "security" concerns. 

That's not really why I liked the user group based assignment though- I found the linking to user group an elegant way to **force** admins to be explicit when assigning this permission. Albeit the same result can be achieved by adding a warning in the `UserPermissions` class under the `impersonateUsers` permission explaining what this permission does (Similar to how `administrateUsers` does it). I was more focused on awareness of what this permission does.

As said, I have an idea to get this implemented into the 3.2 base. Will finish that today or tomorrow and explain it once done.

gtettelaar:
Looks like it's done.

`craft\services\UserPermissions::isImpersonationAllowed()` is used by `UsersController` in `actionEditUser()` and `actionImpersonate()` as a gatekeeper method. It checks all the above rules. 

Furthermore `craft\web\User` is edited so that `checkPermission()` takes into account the permissions of the impersonating user (if there is one). Also i removed a `->admin()` call in d187ca7a97295129b59860a091e9ac4e9267a4aa - this additional query parameter did not make sense if we are allowing non-admin impersonation. 

`craft\services\UserPermissions::doesFirstUserHaveMorePermissions()` is the check to ensure that the _impersonatee_ **cannot** have more permissions than the _impersonator_ but the _impersonator_ **can** have more permissions than the _impersonatee_.

I tested it locally and it looks to work as expected with the permission matching being enforced as expected. Can you confirm this? @brandonkelly

brandonkelly:
> Furthermore `craft\web\User` is edited so that `checkPermission()` takes into account the permissions of the impersonating user (if there is one).

I know I suggested that as one of the possibilities to prevent permission escalation, but we shouldn’t do it. When you are impersonating a user, the whole experience should be identical to the other user’s, including permission enforcement.

gtettelaar:
> I know I suggested that as one of the possibilities to prevent permission escalation, but we shouldn’t do it. When you are impersonating a user, the whole experience should be identical to the other user’s, including permission enforcement.

That change wouldn't make that much difference to the experience - as the permissions should overlap based on earlier checks. With that in mind, there's no real reason to keep it either. Removed it.

brandonkelly:
Ah I misunderstood what you meant; I thought it was only checking the original user in case of impersonation, rather than as an additional check. But yeah, no real reason  to do it, we can trust the session.

brandonkelly:
@gtettelaar so is this PR good to go, or still working on it?

gtettelaar:
@brandonkelly It's ready If you're happy with it. I removed the two chunks in the `craft\web\User` as you asked. 

Took a little more time to get across the line but thanks for the input!

michaelrog:
> If User A doesn’t have all the same permissions as User B (whether granted directly or via user groups), they are disallowed.

Is it necessary to make sure A has _all_ of B's permissions?

What if `craft\web\User::checkPermission()` could check whether the user is being "impersonated" and, if so, return positively only if _both_ both the user and impersonator have the given permission?

gtettelaar:
@michaelrog That was added in fc6eeac7d808f3bdd1a5192f53c062d491db4d79 but then removed in d36a1d9aaa6b50c8ce513e091e41b8156de20d69 because of Brandon's [suggestion](https://github.com/craftcms/cms/pull/4403#issuecomment-502193210). Now regardless of whether we keep it in or not. It is probably imperative to ensure that `User A` has all the permissions that `User B` has even if we were to check the permissions in `checkPermission()`. That "having all permissions" check needs to happen before _any_ impersonation session is started.

If we remove the check in `UsersController` that would mean any user who has the `impersonateUsers` permission can impersonate any other non-admin user. Then it would be determined at execution time if the user can actually execute a specific system function like, editing an entry or accessing a plugin. 

That would mean that the experience of a logged in user (`User B`) and an impersonator (`User A`) could vary because `User A` does not have all the permissions that `User B` has - but he can still impersonate him. An example: `User B` would see an `utilities` tab but `User A` would not because he doesn't have any `utility` based permission. 

My guess is that the end result is, certain actions would be visible to `User B` but not to the impersonating `User A` - OR - the impersonating `User B` would see a bunch of `An unknown problem has occurred` messages because of permission problems.

Hope this makes sense?

brandonkelly:
> Is it necessary to make sure A has _all_ of B's permissions?

Yes because if User B can do something User A isn’t allowed to do, you wouldn’t want to give User A the ability to do that thing simply by impersonating User B.

> What if `craft\web\User::checkPermission()` could check whether the user is being "impersonated" and, if so, return positively only if both both the user and impersonator have the given permission?

That’s exactly what d36a1d9aaa6b50c8ce513e091e41b8156de20d69 just reverted. But then you’re not getting a true representation of User B’s experience. Better to just not give User A the option to impersonate User B in the first place.

brandonkelly:
Doh, you beat me to it @gtettelaar :)

michaelrog:
Right, I get the reasons behind the implementation. I'm asking if providing a "true experience" is the point of impersonation, vs the ability to do mutually authorized things on another User's behalf.

I can see it going either way. For example, I've just recently implemented a system where I have low-permissions Support Team users who can impersonate other users on the front-end purely for the purpose of placing Commerce orders on their behalf. In that case, the impersonated user might theoretically have plenty of permissions that the Support Team user doesn't, and that would be moot to the task one wants to perform on behalf of the other.

To be clear, I'm not strongly opinionated about this. I definitely see the benefit of providing a "true experience" of the impersonated user's permissions [or nothing at all]. Just kind of posing it as a gut check. 

In practice, I suspect it would be extremely rare for an impersonating user to not have at least the permissions of the user they are impersonating, so... ¯\_(ツ)_/¯

gtettelaar:
> In practice, I suspect it would be extremely rare for an impersonating user to not have at least the permissions of the user they are impersonating, so... ¯_(ツ)_/¯

Hitting the hammer on the head with that one. 

My inner pragmatist says we should default to the current implementation, BUT. A config setting could be available (defaulting to `false`) which if set to `true ` disables the gatekeeper check in the `UsersController` and enables a point-by-point check done in `checkPermission()`. This would ensure that you can achieve what you want to achieve. Albeit it is an edge case...

Inherently this could cause a quote-on-quote "worse" impersonation experience but it would have to be explicitly enabled and assigned to a user by an admin. That double fail-safe should provide enough insurance in the event there is a need for this feature in a particular "configuration" of Craft.

P.S. 
Im not entirely sold on a global config setting for `impersonateUsers`. @michaelrog Would it be valid to determine this on a user-by-user (or group) case?

brandonkelly:
YAGNI

gtettelaar:
> YAGNI

[He would](https://github.com/craftcms/cms/pull/4403#issuecomment-502254665). 

Regardless, Let's just merge the PR as it is now for `BETA-3` and then if a bunch more people start asking for more fine-grained control over the impersonation permission it can always be added later one way or another.

michaelrog:
> [He would](https://github.com/craftcms/cms/pull/4403#issuecomment-502254665).

Eh, it's a corner's corner case. (I agree with Brandon; YAGNI.)

Switching up flavors of impersonation via a relatively obscure config option is probably not a good experience for the vast majority of customers.

gtettelaar:
> Switching up flavors of impersonation via a relatively obscure config option is probably not a good experience for the vast majority of customers.

+1.
I added the point in response to the low-level support team example - as an idea to support that. The config setting wasn't the best idea  (as the P.S noted). A per user setting could be an option if a bunch of people ask for it but it's not necessary to do it now. 

> @gtettelaar so is this PR good to go, or still working on it?

@brandonkelly It's ready🚀

brandonkelly:
All merged. Thanks @gtettelaar!

hubertott:
Is there an ability to disable **impersonateUsers** at a global level. ie. override the permission at the config level so even Admins of the system cannot impersonate another user? [That is without modifying the codebase]

In the Craft Commerce sense we feel impersonating a user may leads to the ability for someone to make purchases on someone else's behalf. For example if a payment gateway allows you to store an authorisation token for future purchases.

brandonkelly:
You could remove it using the []() event, from a custom module:

```php
use yii\base\Event;
use craft\controllers\UsersController;
use craft\events\RegisterUserActionsEvent;

// ...

Event::on(
    UsersController::class,
    UsersController::EVENT_REGISTER_USER_ACTIONS,
    function(RegisterUserActionsEvent $e) {
        foreach ($e->sessionActions as $i => $action) {
            if ($action['action'] === 'users/impersonate') {
                unset($e->sessionActions[$i]);
            }
        }
    });
```

hubertott:
Thank you for the prompt response.

PR #4407, opened 2019-06-14
Adds a console command for testing email settings

### LINKED ISSUE #4020: FR: terminal command to send test emails
It would be nice to have a terminal command to send test emails for the instances where we're overriding email settings in app.php with different configurations for environments etc.

At the moment to send a test email from the dashboard you need to fill in settings and it uses those, even if they're being overridden. 

A command in terminal could solve this, something like;

`./craft email/test [user@example.com] [environment]`

gtettelaar:
Resolves #4020

Not entirely sure what the original FR meant with the `multi-env` setup in `app.php` point. Will double check there - PR is still WIP for now. 

For now, this command allows you to test the default Craft email settings OR test custom settings based on any custom user input. You can select custom transport adapters and configure those to send a test email through the console


gtettelaar:
Should be ready now. 

There is one issue I can't figure out but I'm sure you will know the answer quickly. 

On `Line 100` I can't seem to extract the used `MailSettings` and `TransportAdapter` from a `craft\mail\Mailer` object. Without this, it's not possible to create the mail settings string as required by `_renderMailSettingsString`. Any ideas?

@brandonkelly

angrybrad:
Cleaned up a bit and merged... thanks!

Issue #3489, opened 2018-11-22
Add a detail button to see failed queue job error


riasvdv:
### Description

It would be great to have a "detail" button between the "retry" and "cancel" buttons on a queue job to see what went wrong.

brandonkelly:
We’ve added a new Queue Manager utility for Craft 3.4, and jobs have a details view that reveals the error that caused a job to fail:

![A failed job’s details view within the Queue Manager utility in Craft 3.4)](https://user-images.githubusercontent.com/47792/71935184-08f0b880-315b-11ea-9084-cc360e0fa0b0.png)

Issue #2753, opened 2018-04-16
Request: Task Manager


andrewfairlie:
### Description
As per a topic [on Twitter](https://twitter.com/simons1984/status/985887449456041985) I think it would be valuable if the utility section of the CMS allowed you to cancel tasks rather than having to delete them from the Database table.

narration-sd:
Putting in a vote for this by thumb above, as it seems much a good idea.

There used to be a plugin for it on Craft 2, but in fact I'd gotten fingers pretty adept at clearing the tasks table manually, while bringing up something with kind of involved use of Task (Jobs now, converted).

As a Utilities function, you'd gain not just for development. Admins could clear a stuck site immediately, when/if this happens, do troubleshooting later or on staging. In fact, I build such UI abilities into the plugin, on top of a considering circumstances but automatic stuck-killer. Because you integrate...another frame of use.

If implementing, would suggest a kill-them-all ability as well as individual take-out. This would give the ability to  skip task which Craft knows are important to keep (or auto-restart them?). The same mechanism could intervene if you try to manually delete a special-handling task....

brandonkelly:
We’ve added a new Queue Manager utility for Craft 3.4! This has completely replaced the queue HUD in the global sidebar – clicking on the active job there will just link you over to the utility (provided you have permission to access it).

![The Queue Manager utility in Craft 3.4](https://user-images.githubusercontent.com/47792/71934820-1c4f5400-315a-11ea-939e-2c1704739487.png)

![A job’s details view within the Queue Manager utility in Craft 3.4](https://user-images.githubusercontent.com/47792/71934931-63d5e000-315a-11ea-8c28-8dd6fced9c6e.png)

Special thanks to @gtettelaar for kickstarting this.

bencroker:
> This has completely replaced the queue HUD in the global sidebar – clicking on the active job there will just link you over to the utility (provided you have permission to access it).

I'm loving the Task Manager so far, but does the queue HUD have to go completely? I realise that the actions may require permissions, but for (especially non-technical) users it is very helpful to see the current job summary (especially with the recent UI improvements) and what jobs are pending.

<img width="303" alt="Screenshot 2020-01-08 at 01 11 48" src="https://user-images.githubusercontent.com/57572400/71939821-3a49a280-31b4-11ea-9b68-434802a53d1b.png">

brandonkelly:
[...]

PR #4419, opened 2019-06-19
Queue management improvements

### LINKED ISSUE #4040: Feature Request: Queue permissions
I would like to have a few extra permissions for queue related actions. Having permissions to control who can view, retry and release queue jobs would be great. I use jobs for importing/exporting data and our customers should not be able to manage those jobs.

Thanks in advance!

gtettelaar:
**Proposal** to resolve #4040, #2753, #3489, #4777.

This PR contains:

A Queue manager utility #2753 powered by Vue which comprises of an Index page showing a max of 50 jobs (see note below) and a "Details" page where the specific job information is displayed (#3489). The component allows releasing of jobs and retrying of specific jobs. You can also Release all and Retry all from within the Web utility. A short polling service is provided so that you can view the progress of the jobs in real-time (ish).

The `craft\queue\QueueInterface` now requires a `getJobDetails`, `retryAll` and `releaseAll` method. This was required to allow the Queue utility to also support non `craft\queue\Queue` versions of the Queue and to allow retrieval of more detailed Queue information which `getJobInfo` doesn't provide.

This PR adds a `release` option to the Queue console. Both releasing a specific job and releasing all queue jobs is supported through the console now.

Because of the way Craft's utilities are structured a `utility:queue-manager` permission is implicitly defined (#4040). Specific `craft\controllers\QueueController` actions now require this permission. Because of this the release and retry functionality has been removed from the sidenav widget (3b46e27452327e5c0185a32139bb9ddfa6a309f6) as this was no longer needed and would break with permission issues. The side nav now acts as an indicator of queue status and progress only. 

### Two things: 
- Currently, _some_ ES6 based JS is used (I.E Arrow functions). Seeing as arrow functions are Chrome 45 and up and Craft's minimum is Chrome 29 and up some transpilation will be required here. I'm not familiar with how the `gruntfile.js` works so i didn't submit the built/minified versions of the JS - need some help here.
- The limit of the number of Jobs is hardcoded to 50. I don't know if there is ever a scenario where you need to see more than 50 Jobs in a single time that can't wait until some jobs are done but we could implement this feature through a config setting, input field saving to local storage e.t.c. For now, it's hard coded. 

WIP Because I need to do some final edge case testing, the ES 6 thing and one last run of cleanup.


gtettelaar:
This one should be all ready for review now. Tested it with various jobs and run-configs - looks like it's working well on my end. 

The only thing still to-do is the ES6 point.

dsmrt:
I'd love to see some performance improvements with large queues in the control panel. We utilize the queue heavily. Recently, I loaded up a queue with 200k jobs and it really slowed down the control panel which I thought was odd. On this install, we have separate worker containers for the queue (separate from the web app containers, so they are isolated) and `runQueueAutomatically` turned off. My hunch is it's either the queue list on the side nav or the queue job info call checking for updates. Just something to think about.

gtettelaar:
@dsmrt This PR _currently_ doesn't address it but it's something that is probably worth adding in before anything gets merged. 

The side nav queue list doesn't define a limit to the number of jobs so in your case that queue list has to get all 200K jobs (`queue/get-job-info`), loop through them, and send them back to your CP. This probably slows the whole things down a notch or two...

Just tested this locally with 20,000 jobs - my computer isn't a fan. 

There are probably two ways to improve the CP jobs index:

- With the new management panel, you don't even need the CP index - if you have some other way to figure out if Jobs failed - so a setting to just disable it all together _could_ be something.
- A limit to the number of jobs the CP gets on every page via `GeneralConfig`. 

Personally, I think option two seems most useful. The nice thing about the CP jobs bar is that you can quickly see if jobs failed and respond accordingly - with 50 Jobs the overhead should be non-existent. Setting it to 0 would also effectively mean disabling it making option 1 moot...

Perhaps there is an existing way around this that I haven't found yet (there is a `limit` param on `queue/get-job-info` so....). I'll revisit this one the initial review of the PR is done and let P&T have their say on it as well before making any drastic changes.

dsmrt:
@gtettelaar, having those 2 improvements would be great. I'll add an issue for option 2 cause it'd be great to that out before the queue manager is baked. Thanks for your help!

brandonkelly:
This has (finally) been merged in! Thanks for the kickstart @gtettelaar!

PR #4434, opened 2019-06-22
Allow hard-deletion of trashed elements through the CP

### LINKED ISSUE #4420: FR: Ability to permanently delete soft-deleted elements from the CP
### Description
We ran into an issue today where we were testing some functionality involving users. We would delete the users and try to re-signup a user; however, that won't work as expected if the email addresses are the same. This is because the user is just in a trashed state. Ideally, one could select a particular trashed element and delete it permanently.

**Edit**
Looks like there's something else going on with the user registration issue, so that may be unrelated. I think a module was malfunctioning and created some duplicate data. The feature request should still be valid though.

### Additional info

- Craft version: 3.1.30
- PHP version: N/A
- Plugins & versions: N/A

gtettelaar:
Resolves #4420 

Adds a `craft\elements\actions\HardDelete` class for hard-deletion. Only trashed elements will actually be deleted by this class (non-trashed are ignored). Preventing any loop-holes in the existing workflow. 

The `craft\controllers\ElementIndexesController::availableActions()` is adjusted accordingly.

### DIFF
--- a/CHANGELOG-v3.2.md
+++ b/CHANGELOG-v3.2.md
@@ -42,6 +42,7 @@
 - It’s now possible for plugins and modules to define custom actions on console controllers.
 - Added a testing framework for Craft and plugins, powered by Codeception. ([#3382](https://github.com/craftcms/cms/pull/3382), [#1485](https://github.com/craftcms/cms/issues/1485), [#944](https://github.com/craftcms/cms/issues/944))
 - Added `craft\base\BlockElementInterface`.
+- Added `craft\elements\actions\HardDelete`.
 - Added `craft\base\Element::EVENT_AFTER_PROPAGATE`.
 - Added `craft\base\Element::EVENT_REGISTER_PREVIEW_TARGETS`.
 - Added `craft\base\Element::previewTargets()`.
--- a/CHANGELOG-v3.md
+++ b/CHANGELOG-v3.md
@@ -4,6 +4,7 @@
 
 ### Added
 - Added the `{% dd %}` Twig tag. ([#4399](https://github.com/craftcms/cms/issues/4399))
+- Added `craft\elements\actions\HardDelete`.
 
 ### Changed
 - `craft\web\twig\variables\CraftVariable` no longer triggers the `defineComponents` event. ([#4416](https://github.com/craftcms/cms/issues/4416))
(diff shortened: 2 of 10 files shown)


brandonkelly:
If we do this, the hard-delete action should only be available to already-soft-deleted elements, similar to the Restore action.

gtettelaar:
@brandonkelly 

The [ElementIndexesController](https://github.com/craftcms/cms/blob/5d38cff498688e2b7f8cd8ca4143b75c4102a27d/src/controllers/ElementIndexesController.php#L486) handles the `HardDelete` class exactly the same as the `Restore` class. The [HardDelete](https://github.com/craftcms/cms/blob/5d38cff498688e2b7f8cd8ca4143b75c4102a27d/src/elements/actions/HardDelete.php#L86) class will not delete any elements if they are not trashed already (We skip these elements). Did I miss a check in some place?

brandonkelly:
Sorry you’re right, I read the code wrong.

putyourlightson:
It looks like this PR may have been closed without being merged, or am I missing something @brandonkelly?

brandonkelly:
Hm I think that was probably accidental, as I don’t remember closing this and there’s no explanation. Looks like @gtettelaar deleted the branch though…

gtettelaar:
@brandonkelly I didn't. This was closed because the `3.2` branch was deleted when `3.2` was released - [my branch](https://github.com/gtettelaar/cms/tree/feature/hard-delete-action) still exists. Looks like the same happened to #4499, #4463, #4481, #4442, and #4419 as these were all closed on the same date, without explanation - which is uncharacteristic.

Merging them will probably require a base change to `develop` (plus resolving some merge conflicts 
😂).

putyourlightson:
Ouch

brandonkelly:
Ah ok, well in any case GitHub won’t let me reopen the PR – probably a permissions issue. Can you do that please? I can handle changing the base, etc.

gtettelaar:
Yea, Github doesn't allow me to open them either. Probably something to do with a missing base branch, and I can't change the base until the PR is open.

Unless anyone has any idea's It's probably best to just create identical PRs as copies of the original right?

brandonkelly:
Aha! I just republished a 3.2 branch, and now I was able to open this, along with your other PRs. So that explains it – simply by deleting the 3.2 branch GitHub automatically closed any PRs pointed at it. Thanks!

gtettelaar:
Cool! I saw this one opened and was wondering how you got them to work. Github _should_ probably handle branch deletions a bit better IMO.

PR #4469, opened 2019-06-30
Helper class for getting CSRF token name & value

lindseydiloreto:
Getting a CSRF token name & value the traditional way is clunky and difficult to remember...

```php
$csrfTokenName  = Craft::$app->getConfig()->getGeneral()->csrfTokenName;
$csrfTokenValue = Craft::$app->getRequest()->getCsrfToken();
```

I find it to be nearly impossible to remember where those two pieces of data live.

**I propose this simple helper class:**

```php
$csrfTokenName  = Csrf::tokenName();
$csrfTokenValue = Csrf::tokenValue();
```

This would make it _much_ easier to register CSRF tokens via PHP...

```php
use craft\helpers\Csrf;

$view->registerJs('
    window.csrfTokenName  = "'.Csrf::tokenName().'";
    window.csrfTokenValue = "'.Csrf::tokenValue().'";
', $view::POS_END);
```

If possible, I'd love to make it just as accessible via Twig tags. But I wasn't sure the best way to approach the Twig stuff.

Let me know if you have any suggestions!

### DIFF
--- a/src/helpers/Csrf.php
+++ b/src/helpers/Csrf.php
@@ -1,0 +1,39 @@
+<?php
+/**
+ * @link https://craftcms.com/
+ * @copyright Copyright (c) Pixel & Tonic, Inc.
+ * @license https://craftcms.github.io/license/
+ */
+
+namespace craft\helpers;
+
+use Craft;
+
+/**
+ * Class Csrf
+ *
+ * @author Pixel & Tonic, Inc. <support@pixelandtonic.com>
+ * @since 3.1
+ */
+class Csrf
+{
+    /**
+     * Get the CSRF token name.
+     *
+     * @return string
+     */
+    public static function tokenName(): string
+    {
+        return Craft::$app->getConfig()->getGeneral()->csrfTokenName;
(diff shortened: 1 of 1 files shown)


brandonkelly:
It just occurred to me that there’s no reason that any application code should ever need to get the CSRF token param name from the general config. That should only be checked once, when the `request` component is first being initialized:

https://github.com/craftcms/cms/blob/ddd2e97db78c27ddaeffd405b58fea107935b6ae/src/helpers/App.php#L626

Everywhere, the param name can (and should) be fetched from the `request` component directly, in the same place that the CSRF token value is pulled from.

```php
$csrfParam = Craft::$app->request->csrfParam;
$csrfToken = Craft::$app->request->csrfToken; // or getCsrfToken()
```

I’ve just updated the docs and application code to use that `Request::$csrfParam` property rather than `GeneralConfig::$csrfTokenName`. That should be easier to remember.

lindseydiloreto:
That works too! Thanks @brandonkelly 🍺

PR #4470, opened 2019-06-30
Document translation message placeholders

sa3dany:
This would have saved me the time I spent:

- Reading the yiiframework documentation about the feature
- Digging through craft source code to find how the translation filter is invoked and the parameter names used

Please assume this is a draft, I just wanted to open a pull request while everything was still fresh in my mind.
I don't have much experience writing docs, so a review or taking over the pull request is welcome.


brandonkelly:
Thanks for the PR!

PR #4481, opened 2019-07-03
Enable template paths by SiteGroup handle

### LINKED ISSUE #3101: [FR] Site group-based templates
### Description

If there is only translation between localized site, we probably dont need to add new templates
````
templates/
├── account.twig      --> default site A account template, and works for both EN, DE, FR localization
````
But with multiple sites + localization, 
Site A, Site B, Site C can have a large difference in page layout.
it becomes:

````
templates/
├── account.twig      --> default site A account template,  works for all EN, DE, FR localization
└── siteBen /
    └── account.twig  --> siteB EN account template
└── siteBde /
    └── account.twig  --> siteB DE account template
└── siteBfr /
    └── account.twig  --> siteB FR account template
└── siteCen /
    └── account.twig  --> siteC EN account template
└── siteCde /
    └── account.twig  --> siteC DE account template
└── siteCfr /
    └── account.twig  --> siteC FR account template
````

Can we make use of siteGroup concept, and workout something like this:
````
templates/
├── account.twig      --> all sites default account template,  works for all EN, DE, FR localization
└── siteGroupB /
    └── account.twig  --> siteGroupB account default template,  works for all EN, DE, FR localization
└── siteGroupC /
    └── account.twig  --> siteGroupC account default template,  works for all EN, DE, FR localization
````

gtettelaar:
Resolves #3101

This PR adds template resolution based on the handle of a SiteGroup and adds a handle to the `SiteGroup` Model.

### DIFF
--- a/CHANGELOG-v3.2.md
+++ b/CHANGELOG-v3.2.md
@@ -106,6 +106,7 @@
 - Added the `Craft.Preview` JavaScript class.
 
 ### Changed
+- Craft will now also attempt to find templates by the SiteGroup `handle` property ([#3101](https://github.com/craftcms/cms/issues/3101))
 - Relational fields are now capable of selecting elements from multiple sites, if they haven’t been locked down to only related elements from a single site. ([#3584](https://github.com/craftcms/cms/issues/3584))
 - Reference tags can now specify the site to load the element from. ([#2956](https://github.com/craftcms/cms/issues/2956))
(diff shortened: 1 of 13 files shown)


Mosnar:
As a note, Giel and I were discussing edge-cases this could introduce and basically came to this possible scenario: 
**If you have a site that is intended to resolve templates from the base template directory and any generated site group handle matches that site handle, it is possible for templates to be inadvertently loaded from this site group's template folder rather than the root directory.**

I think this would be incredibly rare and unlikely, but it is a breaking change.

PR #4488, opened 2019-07-05
Fix example code for displaying Matrix block contents

j-greig:
Need to loop through `matrixBlocks` not `blocks`


brandonkelly:
Thanks!

Issue #2944, opened 2018-06-01
Copy Asset URLs (feature request)


proimage:
Can there be either a menu item for copying the URL(s) of the selected asset(s) (see below), or if "Preview file" isn't able to handle a filetype, include a direct link to the file in the preview window (or both)? Something like, "Preview not available. You can try to [open this file in a new tab](#)."

![image](https://user-images.githubusercontent.com/1027523/40860793-59a9d254-65ef-11e8-91cd-c374afc9c2c6.png)

brandonkelly:
@proimage Thanks for posting these. Please keep each issue limited to a single feature request though, so we can track progress on them individually. (You can edit this one to just include one of the FR’s, and then post two new issues for the others.)

proimage:
Whoops, of course! Done.

jsunsawyer:
Yes! It seems that **Preview File** replaced **View Asset** when asset previews got added. Definitely confusing for users coming from Craft 2.

@proimage For now, you can enable the **Link** option to quickly view an asset's source.

proimage:
> For now, you can enable the Link option

What is this "Link option" of which you speak?

brandonkelly:
Doh, forgot about that – so this is already implemented (or something close enough, anyway). Just click the gear icon below the volume list on the Assets page, and check the box next to “Link”, for whichever volumes you want to be able to access files in directly.

bstein-clever:
I ran into this today, as well. It's extremely intuitive coming from Craft 2, especially since I need to add the link column to *all* of my asset folders, which is a little frustrating.

Add one more vote to having the ability to open a link to the asset if "preview" doesn't work for the file type.

brandonkelly:
Fair enough, I’ll reopen.

johandouma:
As mentioned above, I am aware that we can add the Link as a column, but there are already quite a few options in the cog drop down that relate to copying, eg previewing, download, I think it would make sense to add a Copy URL in that drop down as well.

ryanmasuga:
It never occurred to me to add the Link column in the list view, but this feels a bit like a cheat considering how many other options are already in the gear dropdown. Also, I pretty much have to go through 30-40 asset volumes and remember to add that to the view. I'd assume that so long as "Assets in this volume have public URLs" is checked, there could be a way to get that from the gear in almost exactly the same way as "Copy reference tag" works.

khalwat:
[...]

PR #4499, opened 2019-07-08
Add CopyAssetUrl element action

putyourlightson:
This pull request adds a "Copy URL" element action to the assets index page, addressing feature request https://github.com/craftcms/cms/issues/2944.

The action is only available to assets that belong to a volume that has public URLs enabled.

Screenshot 2019-07-08 at 13 40 09


src/elements/Asset.php
```diff
@@ -265,6 +266,9 @@ protected static function defineActions(string $source = null): array
                 ]
             );
 
+            // Copy URL
+            $actions[] = CopyAssetUrl::class;
```
gtettelaar: Wouldn't it be better to do: 
`
if ($volume->hasUrls) {
    $actions[] = CopyAssetUrl::class;
}
`
This would make sure if `hasUrls` is disabled on the volume you don't see the unneeded `Copy URL` button.

putyourlightson: Ah yes, I somehow missed that `$volume` is available in the static method, thanks!

khalwat:
Related issue: https://github.com/craftcms/cms/issues/2944

brandonkelly:
All merged in for 3.5. Thanks @putyourlightson!

bencroker:
Thank you!

PR #4539, opened 2019-07-12
Set test env

boboldehampsink:
`Config::$env` is `null` when running tests. This makes it use the `*` environmental config values, and you can't set specific `test` config values.

This fixes that.

However I also feel that maybe https://github.com/craftcms/cms/blob/develop/src/test/TestSetup.php#L214-L218 could be merged with https://github.com/craftcms/cms/blob/develop/src/test/TestSetup.php#L247-L250 somehow (which already has the `test` env)

### DIFF
--- a/src/test/TestSetup.php
+++ b/src/test/TestSetup.php
@@ -213,6 +213,7 @@ class TestSetup
                 'components' => [
                     'config' => [
                         'class' => Config::class,
+                        'env' => 'test',
                         'configDir' => CRAFT_CONFIG_PATH,
                         'appDefaultsDir' => $srcPath . '/config/defaults',
                     ],


gtettelaar:
> However I also feel that maybe https://github.com/craftcms/cms/blob/develop/src/test/TestSetup.php#L214-L218 could be merged with https://github.com/craftcms/cms/blob/develop/src/test/TestSetup.php#L247-L250 somehow (which already has the test env)

Agree. You can probably set the `components.config` key to the locally available `$configService`.

angrybrad:
Fixed here, which should have the same effect and cleans up the duplication a bit as well: https://github.com/craftcms/cms/commit/acd876eb9d9463e2dc99897c7d9aa16ddb41c55d

PR #4540, opened 2019-07-12
Add GQL 

andris-sevcenko:
Todo list (also in `services/Gql.php`)

1) Obviously, docs for everything and every method (including exceptions thrown)
  + services/gql
  + helpers/gql
  + gql/*
  + Description for *every* field on GQL types. For hard-coded types this goes in type definition files. For actual Craft fields we can reuse the `instructions` property.
2) Event for registering interfaces (because they get eager-loaded if `prebuildSchema` is set to `true` when building schema.)
3) Eager-loading currently ignores the passed arguments. Some tests are borked because of that. This includes the `limit` argument to relation fields currently.
4) Maybe move out token permissions to a separate service?
5) Speaking of, maybe put the token settings somewhere else on the Craft settings page and definitely add an icon
6) Figure out and add query complexity costs per GQL type.
7) Configurable limits on both query complexity and nesting level.
8) Sprinkle some cache on top (see if able to integrate with Craft template caching, as similar principles should apply - element changes invalidate a cached response and structural changes (when project config gets updated) invalidates cached responses. Cached by token by query.
9) Pretty sure the test coverage went down with the last dev iteration - circle back and check.
10) Tests for disabled and expired tokens.
11) Tests for matrix blocks returning correct union types.
12) Expose site structure over GQL as well
13) Once (12) is in place, get rid of, say, `sectionId` and `sectionUid` fields and replace with a `section` type that has `uid` and `id` fields.
14) MAYBE change all closures to static method and make schema itself cacheable per token per query.


src/gql/interfaces/elements/User.php
```diff
@@ -0,0 +1,93 @@
...
+            'firstName' => Type::string(),
+            'lastName' => Type::string(),
+            'email' => Type::string(),
+            'admin' => Type::boolean(),
+            'locked' => Type::boolean(),
+            'suspended' => Type::boolean(),
+            'pending' => Type::boolean(),
+            'lastLoginDate' => DateTime::getType(),
+            'lastInvalidLoginDate' => DateTime::getType(),
+            'invalidLoginCounr' => Type::int(),
```
monachilada: `invalidLoginCounr` should be `invalidLoginCount`?

andris-sevcenko: .... maybe?

monachilada:
I was playing around with this today and it's amazing. I was probing to see if the magic preview tokens could be combined with it but both of the following throw 404 `Template not found: actions/gql` errors.

```
curl --header "Authorization: Bearer ABC123" http://domain.local/actions/gql/?token=XYZ987&query=%7BqueryEntries%7Btitle%7D%7D

curl --header "Authorization: Bearer ABC123" --header "X-Craft-Token: XYZ987" http://domain.local/actions/gql/?query=%7BqueryEntries%7Btitle%7D%7D
```

Omitting any token value works just fine. I guess this means the request is getting routed as a preview before the query gets a chance to execute? Would this be expected at this point?

andris-sevcenko:
@monachilada you can accomplish that by (and that is now the recommended way) setting up a route that points to the `gql` action. For example, if your `config/routes.php` file looks like
```php
return [
    'api' => 'gql',
    // ...
];
```
Then `curl --header "Authorization: Bearer ABC123" http://domain.local/api?token=XYZ987&query=%7BqueryEntries%7Btitle%7D%7D` should work.

Issue #3830, opened 2019-02-12
Why can't a section both have hasUrls set to false and a template configured?


pieterbeulque:
https://github.com/craftcms/cms/blob/e97588fb197fc88a9ea497e0d0c0a29874ef2250/src/services/Sections.php#L649-L655

I'm having trouble using `Craft::$app->sections->saveSection($section)` to create a section with settings configured like this:

```php
new Section_SiteSettings([
  'siteId' => $site->id,
  'hasUrls' => false,
  'uriFormat' => 'IGNORE', // This is neccessary to skip validation errors, somehow
  'template' => 'myTemplate',
]);
```

Am I trying to do something that I shouldn't try to do or is this a bug?

pieterbeulque:
It worked with these `Section_SiteSettings`.

```php
new Section_SiteSettings([
  'siteId' => $site->id,
  'hasUrls' => true,
  'uriFormat' => '__home__', // This is a magic keyword, apparently
  'template' => 'myTemplate',
]);
```

brandonkelly:
If you don’t want the section to have URLs, then you should just leave the `uriFormat` set to `null` (or don't include the line at all).

pieterbeulque:
@brandonkelly, that's what I thought, but that threw a validator error saying `uriFormat` cannot be blank.

brandonkelly:
@pieterbeulque The only time `uriFormat` is required is if the section type is a Single, or if `hasUrls` is `true`.

pieterbeulque:
OK, then it threw the validation error because it was a single.

Thanks for clarifying this, @brandonkelly!

Issue #1662, opened 2017-04-24
Unified “Content” view


brandonkelly:
The Control Panel’s current way of organizing your content isn’t always ideal, because it doesn’t expose the actual information architecture of the site.

Common example is having a Single section for the landing page of a Structure section. You have to click on “Singles” to find the landing page, but the Structure section’s name to find the rest of the section. It gets even worse if the landing page is something besides an entry section, like a category group.

To fix, Craft should have an optional “Content” view, which gives authors a unified look at all the site’s content (regardless of element type), arranged according to the site‘s information architecture. It would be up to the admin to decide which content should be present, and how it should be displayed in relation to other content.

This would require a new interface that things that want to participate in the page could implement.

Would resolve #831, #990, #1032, and #1490.

narration-sd:
copying this in, as it gives a hint towards how a solution here might be arranged, also a current work-around.

07:57] 
eric.chantigny @theskyfloor That’s why I mostly never use singles, I use entry types in a Structure section (exception is Homepage).  That way, I have control over the IA.  Of course, that doesn’t apply to channel content, but the landing page would be in the structure, and the entry type would be something like News Landing, which uses a template based on that entry type to pull all the news.

[07:58] 
theskyfloor @eric.chantigny for sure - i think they definitely need to address these needs better in the cp views

timkelty:
FWIW, I regularly have client requests for more parity between element types, specifically with revisions/drafts.

Namely, clients want (and don't understand why they don't) have revisions and drafts for _entries_, but not for **categories** or **commerce products**.

timkelty:
@steverowling made #2021 to track that request

timkelty:
> Namely, clients tend to want (and expect) revisions and drafts for categories and commerce products, like they do for entries.

Along the same lines, I've also had the need to extend the concept of "Entry Types" to other elements (namely categories), so I could have different field layouts.

pogorzelskip:
I guess that it would be also usefull for breadcrumb generating - right now, if i want to make breadcrumb from different sections i have to set it up manually in each template, system has no udea that page "about" is subpage of "home".

Kashkin:
I would love the ability to associate a Channel with a Structure entry. I really miss this from my ExpressEngine days where we used the Structure plugin.

I think Statamic's implementation is even better, and would go a long way to solving the issues mentioned here.

[...]

Issue #1490, opened 2017-03-11
Allow structures to have a "homepage"


leevigraham:
### Description

Let's say I have a section of my site called "Services" with urls that look like:

* /services
* /services/web-development
* /services/web-development/craft-cms
* /services/web-design

and I want to limit editing of these entries to a particular user / group.

There's two ways I could go about this:

Option 1. Create a new structure

I could create a new structure called "Services" with the `Entry URL Format` set to `{slug}`. A parent page would need to be created called `Services` with the `uri`: `/services`. This would list all the services and probably be a different entry type `Services Index`. Then authors could publish sub pages below 'Services' with a different entry type `Service Detail`.

The issue with this approach is that an author could publish an entry with any URL becuase they are not limited to just `/services` as defined by the 'Entry URL format' structure setting.

Option 2: Structure + Single

An alternative is to create a single for 'Services' with a locked down url and a structure for the Service details. This allows for fine grained permissions on both the index and the structure. The only downside is that the author needs to look in Singles to change the index, and 'Services' structure to change the entries. Using [Expanded Singles](https://github.com/engram-design/ExpandedSingles) makes this a bit easier but then there are two 'Services' links in the sidebar.

Option 3: Allow an entry to be the 'homepage' for a structure.

I'm thinking adding an option to structure entries for 'This is for the *structure* homepage' could work here. If the checkbox was clicked the uri for the structure would be set to `/` instead of the slug.

I'm already hacking something similar with a 'Site Structure' by setting the slug for one entry to `__home__`.

![image](https://cloud.githubusercontent.com/assets/25124/23821327/0bd76c20-0683-11e7-81ac-6de103ddbec5.png)

PR #4567, opened 2019-07-15
Allow singles to have no URL's

### LINKED ISSUE #3883: Allow singles to have no URL
When Craft is used as a headless CMS, you might still have one off entry types for certain content. Can we please have a “Entry doesn’t have a URL” option for single sections, which disables the route and Live Preview.

I guess you could benefit from it yourself! 😉
https://github.com/pixelandtonic/craftnet/blob/master/src/controllers/api/v1/PluginStoreController.php#L106

gtettelaar:
Proposal to resolve #3883 and the validation issue that partly caused #3830. 

Basically, we add a `hasUrl` checkbox on the singles site settings table which if turned off ensures singles have no URL. The `__home__` functionality is retained if you toggle this option to on and don't enter a URI format.

I'm quite curious if this is entirely the correct implementation - I tested this and it seems to work as expected but I'm a tad suspicious at how easy this was to implement - not sure if I missed something somewhere.


carlcs:
Can we make the URI config more consistant with channel and structure sections? Have the toggle for all group types maybe?

gtettelaar:
> Have the toggle for all group types maybe?

I'm assuming by group types you mean single, channel and structure? 

> Can we make the URI config more consistent with channel and structure sections?

There probably is a way - but I'm not entirely sure if it has a point.

The problem with having the `hasUrl` toggle for channels and structures is that it would be pointless - but for singles, it's a requirement. 

If you leave the URI format empty on channels/structures you implicitly tell Craft `I don't want this section to have URLs`. 

However, with a single, leaving the URI format empty tells Craft `I want this section to be the homepage. `Craft then set's your section's URI format to `__home__`. We thus need to explicitly tell Craft to disable URLs rather than implicitly. 

We could have the toggle for channels and structures as well but It wouldn't serve a real purpose - other than UI consistency. I'm not sold on whether that is worth it. I'm fine with both implementations though and can see the added value of having this be consistent. Let's see what happens during the review. 

@carlcs

carlcs:
Yeah I don’t have the answer, but was thinking maybe “Leave blank if entries don’t have URLs, enter __home__ if this is the homepage“ could work for all group types.

gtettelaar:
> “Leave blank if entries don’t have URLs, enter home if this is the homepage“

That's tough because when you create a new entry in a Channel section how would Craft define unique URI's?

#1490 Will probably require a rework of how Craft defines the homepage. The FR asks that structure entries (and then probably channel entries as well) should allow one of the entries in a structure to be the Homepage (and have `__home__` as URI).

#1490 will be resolved through #1662 - which is part of the `4.0` roadmap. 

That being said I think this PR offers a decent bridging implementation to allow Craft to be able to do what #3883 requested in a way that may be a little awkward but still manageable (if that makes sense). 

An implementation where we remove this awkwardness can then be baked into `4.0`.

@carlcs

brandonkelly:
Here’s what I came up with:

Screenshot of Site Settings for a Single section

Issue #1335, opened 2017-02-02
Error finding plugin templates


wbrowar:
### Description
In trying to get a template to render from my plugin into the front-end, [using this pattern](https://github.com/craftcms/docs/blob/master/en/updating-plugins.md#rendering-plugin-templates-on-front-end-requests):
```
        use craft\web\View;

        $oldMode = Craft::$app->view->getTemplateMode();
        Craft::$app->view->setTemplateMode(View::TEMPLATE_MODE_CP);
        $html = Craft::$app->view->renderTemplate('adminbar/src/templates/bar');
        Craft::$app->view->setTemplateMode($oldMode);
        Craft::dd($html);
```
This results in this error:

Twig Template Loading Error – craft\web\twig\TemplateLoaderException
Unable to find the template “adminbar/src/templates/bar”.

*Please note: I've tried several variations of the path to my template and they all result in the same error*

### Steps to reproduce

1. Install Craft via composer
2. Symlink to my plugin
3. Create a plugin Variable and a plugin Service
4. Add call Service method within Variable function that's placed in my current page's template.

### Additional info

- Craft version: 3.0.0-beta.1
- PHP version: 7.0.13
- Database driver & version: MySQL 5.6.34
- Plugins & versions: No other plugins

carlcs:
This is what worked for me to render a "CP" template from my service to the front-end.

```PHP
<?php
namespace wbrowar\adminbar\services;

use wbrowar\adminbar\Plugin;

use Craft;
use craft\base\Component;

class AdminbarService extends Component
{
    public function render(): string
    {
        $view = Craft::$app->getView();
        $oldTemplatesPath = $view->getTemplatesPath();

        $view->setTemplatesPath(Plugin::getInstance()->getBasePath());
        $html = $view->renderTemplate('/bar');

        $view->setTemplatesPath($oldTemplatesPath);

        return $html;
    }
}
```

wbrowar:
@carlcs Thank you! I just had to add `/templates` to the path, but it worked:

```
        $view = Craft::$app->getView();
        $oldTemplatesPath = $view->getTemplatesPath();

        $view->setTemplatesPath(AdminBar::getInstance()->getBasePath());
        $html = $view->renderTemplate('/templates/bar');

        $view->setTemplatesPath($oldTemplatesPath);

        Craft::dd($html);
```
@brandonkelly From your perspective, is `getTemplateMode()` preferred over this method?

brandonkelly:
[...]

PR #4570, opened 2019-07-15
Add $templateMode parameter to View methods

putyourlightson:
This pull request adds an optional `$templateMode` parameter to the following methods in `craft\web\View`:

    public function renderTemplate(string $template, array $variables = [], $templateMode = null): string

    public function renderPageTemplate(string $template, array $variables = [], $templateMode = null): string

    public function renderTemplateMacro(string $template, string $macro, array $args = [], $templateMode = null): string

    public function doesTemplateExist(string $name, $templateMode = null): bool

The benefit of this is that rather than have to do this:

    // Set Craft to the site template mode
    $oldTemplateMode = $view->getTemplateMode();
    $view->setTemplateMode($view::TEMPLATE_MODE_SITE);

    $templateExists = $view->doesTemplateExist($template);

    // Restore the original template mode
    $view->setTemplateMode($oldTemplateMode);

We can now do this:

    // Check if template exists using the site template mode
    $templateExists = $view->doesTemplateExist($template, $view::TEMPLATE_MODE_SITE);

Besides being DRYer, it may help avoid confusion such as in https://github.com/craftcms/cms/issues/1335.


putyourlightson:
Yay!!

brandonkelly:
Thanks Ben!

PR #4573, opened 2019-07-15
Print full error for failed Project Config sync.

ryanpcmcquen:
It's quite difficult to track down errors with only the message.


ryanpcmcquen:
Nevermind, I see that it logs, just not to the console (which is where I would prefer it).

PR #4576, opened 2019-07-16
Update changes-in-craft-3.md

engram-design:
Here's a common controller action that isn't mentioned in the docs to update.

Also, as a side note - the old controller actions aren't in camel case - they probably should be right?


brandonkelly:
Thanks! And yeah, the old action names should probably be camelCased. Just made that change. (887d30b1f2ec51611ab698be581676429de92851)

PR #4585, opened 2019-07-16
Add InterceptTokenRouteEvent to craft\web\UrlManager::_getRequestRoute

narration-sd:
This event makes it possible for a plugin to override a token Route in further releases of 3.1, so that a route previously registered can be used in desired cases.

It's intended to be used for 3.1 share requests. If desired, we could further restrict this implementation to only apply on those cases.

This implementation also includes ability to override the token Route, still bypassing ordinary routing. It's an option not actually used for current plugin, so it would be fine to take this action out if undesirable.

Ref: offline conversation which this should put a close to.

### DIFF
--- a/src/events/InterceptTokenRouteEvent.php
+++ b/src/events/InterceptTokenRouteEvent.php
@@ -21,5 +21,6 @@ class InterceptTokenRouteEvent extends Event
     // =========================================================================
 
     public $useTokenRoute = true;
+    public $useReturnedRoute = false;
     public $route = null;
 }
--- a/src/events/InterceptTokenRouteEvent.php
+++ b/src/events/InterceptTokenRouteEvent.php
@@ -1,0 +1,25 @@
+<?php
+/**
+ * @link https://craftcms.com/
+ * @copyright Copyright (c) Pixel & Tonic, Inc.
+ * @license https://craftcms.github.io/license/
+ */
+
+namespace craft\events;
+
+use yii\base\Event;
+
+/**
+ * RegisterUrlRulesEvent class.
+ *
+ * @author Pixel & Tonic, Inc. <support@pixelandtonic.com>
+ * @since 3.0
+ */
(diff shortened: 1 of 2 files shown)


narration-sd:
n.b. I've added some functionality, if you approve. The first version was in thinking to limit just to needs, but  in considering how other evident projects will find the same needs to accomodate edit shares, the extra features may be useful. 

I've also got improved inline documentation, which might be done for the original version if that's the one you choose to approve -- thanks.

brandonkelly:
> It's intended to be used for 3.1 share requests. If desired, we could further restrict this implementation to only apply on those cases.

Did you mean 3.2? We’re not releasing any more 3.1 releases at this point.

narration-sd:
Well, I did mean 3.1, not that I don't get lost in the alphameric soup once in a while.

I'd let you know that I've considered at least once a day about the amount of extra effort I've been putting in to assure Live Vue with at least the late version of 3.1. 

To smooth the road at least for Gatsby/Gridsome sorts of users who may be reluctant to immediately go to 3.2.

If you feel strongly about it, maybe this is the way I can softly close the door: say they can try out the RC on 3.1.latest if preferred, but that this ability will go away on full release, as 'Craft support is going to 3.2, thus so for Live Vue also'. 

And for the RC/3.1, Previews only, not Shares - they'll get a Twig page call for those, whether they have one or not. A bit messy, but the logic conundrum that forces this on late 3.1 is why I made this PR.

narration-sd:
p.s. I'm presuming this travis ci build failure hasn't anything to do with the PR; didn't feel it could cause one.

brandonkelly:
> If you feel strongly about it, maybe this is the way I can softly close the door: say they can try out the RC on 3.1.latest if preferred, but that this ability will go away on full release, as 'Craft support is going to 3.2, thus so for Live Vue also'.

I think that’s completely reasonable.

PR #4592, opened 2019-07-17
Added translations NL (dutch)

KevinBeckers:
- Delete {type}
- Save Entry
- Save as a new {type}
- Save as a Draft
- Save {type}


brandonkelly:
Thanks! We are working on getting new translations ordered and added as well, but this helps :)

KevinBeckers:
No problem, when I come across some translations which are not jet translated I will try to add them.

Issue #4490, opened 2019-07-05
withOne eager-loading param


brandonkelly:
Element queries should have a `withOne` param, which would work similar to `with`, except that only the first related element would be loaded for each parent element, and those eager-loaded elements wouldn’t need to be accessed using the array syntax (so `entry.fieldHandle` rather than `entry.fieldHandle[0]`).

putyourlightson:
I submitted pull request https://github.com/craftcms/cms/pull/4598 as a proposed simplified solution. If you however decide that `withOne` should work similarly to `with` then I'd be happy to take a shot at that too.

brandonkelly:
Gotta hold off until 4.0 for this, as it’s going to introduce a breaking change for element types that have implemented custom eager-loading properties, and aren’t anticipating the possibility of `withOne = true`.

brandonkelly:
Though worth noting that as of Craft 3.3.1, it is possible to pass `limit: 1` to an eager loading criteria. Not as elegant as this idea, but at least provides a bit of a performance boost when only one element is needed.

```twig
{% set entries = craft.entries()
    .with([
        ['assets', {limit: 1}],
    ])
    .all() %}
```

PR #4598, opened 2019-07-17
Added $withOne parameter to ElementQuery

putyourlightson:
This is a simplified solution to https://github.com/craftcms/cms/issues/4490 in which `$withOne` is a boolean that defaults to `false`. Setting it to `true` will limit the results to a single related element rather than an array.

Usage example:
```
{% set entry = craft.entries.with('relatedField').withOne().one() %}

Related Element Title: {{ entry.relatedField.title }}
```

The disadvantage of this is that you cannot mix fetching single related elements with arrays of related elements, for example:
```
{% set entry = craft.entries.with('relatedField1').withOne('relatedField2').one() %}

Related 1 Element Titles: {{ entry.relatedField1[0].title }}, {{ entry.relatedField1[1].title }}, ...

Related 2 Element Title: {{ entry.relatedField2.title }}
```

Tha advantage of the proposed method is that there is no confusion as to whether the syntax `entry.relatedField` or `entry.relatedField[0]` should be used.


michaelrog:
> The disadvantage of this is that you cannot mix fetching single related elements with arrays of related elements....

Curiously, why not support that mixed syntax? As far as I can tell, there's not a _technical_ reason... Is it _just_ to avoid confusion about access syntax?

Issue #4610, opened 2019-07-18
Editing modals should only be closable by using the 'cancel' button.


ryanpcmcquen:
### Description

Currently clicking out of an editing modal will close it, even if there is unsaved work. This can be especially aggravating for users entering data who have opened multiple modals.

### Steps to reproduce

1. Double click an entry inside of another entry.
2. Click away from the modal dialog.
3. The modal closes and all unsaved work is lost.

### Additional info

- Craft version: 3.1.x

brandonkelly:
This is a duplicate of #2512

Issue #2512, opened 2018-03-01
Warn before switching sites in element editor HUDs


davidhellmann:
Feedback from two Clients yesterday during a Backend Workshop:

When you have fields on Assets und you edit them and you switch the site (dropdown) there is not warning that changes are lost like "You have changed stuff, please save before you switch" or stuff like this. Can you add a small text under the Dropdown when something changes?

For me as Dev it's fine but for clients it's not the best behaviour —  UX : )

![image](https://user-images.githubusercontent.com/764292/36839735-23047a1e-1d43-11e8-8326-08e626e83eac.png)

ryanpcmcquen:
For anyone else bitten by this issue, until an official fix is in place, this has been working well for my team:

```css
.hud-shade,
.modal-shade {
    position: static;
}
```

We added it using: https://plugins.craftcms.com/cp-css

brandonkelly:
This has been resolved for Craft 3.5, which will start warning authors before switching sites within an element editor HUD, if there are any unsaved content changes.

PR #4611, opened 2019-07-18
Only allow editing modals to close with the cancel button.

ryanpcmcquen:
Resolves: https://github.com/craftcms/cms/issues/4610

Signed-off-by: Ryan McQuen <rmcquen@funko.com>


brandonkelly:
This would have unintended consequences. The actual fix for #4610 / #2512 is going to be a bit more complicated.

ryanpcmcquen:
@brandonkelly, what unintended consequences? In my testing everything (including the sidebar) works as expected. Although @andris-sevcenko suggested using Garnish's `hideShadeOnClick` prop instead.

Issue #1713, opened 2017-05-17
Automatic width image transforms fail


drifteaur:
### Description

A scale transform with a set width and a blank (automatic) height works.
However if the width is blank and height is given, the transform fails:

Argument 1 passed to craft\image\Raster::scaleToFit() must be of the type integer, null given, called in /vagrant/vendor/craftcms/cms/src/services/AssetTransforms.php on line 1340

### Steps to reproduce

1. Set a scale to fit transform with a set pixel height and a blank width
2. Try to show the image on the front end, it will not generate the transform

### Additional info

- Craft version: Craft CMS 3.0.0-beta.17
- PHP version: 7.0.15-0ubuntu0.16.04.4
- Database driver & version: MySQL 5.7.18-0ubuntu0.16.04.1
- Plugins & versions: none

PR #4643, opened 2019-07-23
Fix SVG scaling bug when using transform with only one dimension value

auralon:
Patches an error (see below) that occurs when using an image transform that only has a defined height (or width) on an SVG asset.

Argument 1 passed to craft\image\Svg::scaleToFit() must be of the type integer, null given, called in ../vendor/craftcms/cms/src/services/AssetTransforms.php on line 1424

### DIFF
--- a/src/elements/Asset.php
+++ b/src/elements/Asset.php
@@ -749,7 +749,10 @@ class Asset extends Element
             return null;
         }
 
-        if ($this->getMimeType() === 'image/gif' && !Craft::$app->getConfig()->getGeneral()->transformGifs) {
+        if (
+            ($this->getMimeType() === 'image/gif' && !Craft::$app->getConfig()->getGeneral()->transformGifs) ||
+            ($this->getMimeType() === 'image/svg+xml')
+        ){
             return AssetsHelper::generateUrl($volume, $this);
         }
 


andris-sevcenko:
Eh, that's a little unclear about what it accomplishes and deals with the symptoms, not the cause.

andris-sevcenko:
Just pushed a commit that allows the width to be null, so that height can be used for dimension calculation as well. 

Apparently, I don't really learn [from past mistakes](https://github.com/craftcms/cms/issues/1713).

auralon:
Hi @andris-sevcenko your patch fixes this issue nicely.  Thank you!

andris-sevcenko:
Thanks for noticing it!

auralon:
@andris-sevcenko do you happen to know when|if  I'll be able to pull a fix for this on 3.1 any time soon?

andris-sevcenko:
It's probably not going to happen since updating to 3.2 is pretty seamless and shouldn't be a hurdle for most sites. If you really need this and can't update to 3.2, the safest path is probably fork Craft at the last 3.1 release and apply your own patch to that. That should be safe to use indefinitely since 3.1 is incredibly unlikely to change anymore.

PR #4654, opened 2019-07-25
Set UID to null so new UID is generated on duplication.

### LINKED ISSUE #4653: Duplicating Entry Doesn't Create a Unique UID
### Description

When you select an Entry and choose "Duplicate" from the action menu, it creates a new entry with the same UID.

### Steps to reproduce

1. Create an entry
2. Check the box next to the entry
3. Choose "Duplicate"

This query shows current duplicates in the database:

```sql
SELECT uid, count(*) AS `count` 
FROM elements 
GROUP BY uid 
HAVING `count` > 1
```

### Additional info

- Craft version: 3.1.34.2
- PHP version: 7.2.11

TonyDeStefano:
Closes #4653 for the 3.1 branch.

### DIFF
--- a/src/services/Elements.php
+++ b/src/services/Elements.php
@@ -597,6 +597,7 @@ class Elements extends Component
         $mainClone->setAttributes($newAttributes, false);
         $mainClone->duplicateOf = $element;
         $mainClone->id = null;
+        $mainClone->uid = null;
         $mainClone->contentId = null;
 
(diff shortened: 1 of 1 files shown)


brandonkelly:
We aren’t releasing additional 3.1 releases.

brandonkelly:
@ryanpcmcquen We made an exception in this case because it affects Craft 3.1 → 3.2 upgrades.

PR #4658, opened 2019-07-26
Update entry-queries.md

maddyBoy:
On https://docs.craftcms.com/v3/dev/element-queries/entry-queries.html#slug bracket was not closed.


brandonkelly:
Thanks for pointing that out! That file is autogenerated, so I just fixed it at the source.

Issue #4548, opened 2019-07-12
Long titles mess up the header layout on the edit Entry page


rungta:
![Messed up header on the edit Entry page](https://user-images.githubusercontent.com/663677/61132967-1bef3c80-a4da-11e9-88f6-cb1608b5512b.png)

### Steps to reproduce

1. Create an entry with 60 or more characters in its title field.

### Additional info

- Craft version: 3.2.1

brandonkelly:
Fixed this for the next release. Here’s a few ways the header will look depending on viewport width:

<img width="1085" alt="Edit Entry page header in a wide viewport" src="https://user-images.githubusercontent.com/47792/62239050-02c70680-b389-11e9-8519-04b9eb522ce0.png">

<img width="789" alt="Edit Entry page header in a medium-wide viewport" src="https://user-images.githubusercontent.com/47792/62239121-22f6c580-b389-11e9-84b7-df978889abee.png">

<img width="355" alt="Edit Entry page header on a mobile viewport" src="https://user-images.githubusercontent.com/47792/62239143-2c802d80-b389-11e9-834b-4b4a5183491a.png">

Thanks to @sebastian-lenz for the help with title truncation!

rungta:
Suggestion: what if we wrap the title instead of truncating it?

![Title wrap](https://user-images.githubusercontent.com/663677/62274790-34f66800-b45e-11e9-8877-ebc57391fb0f.png)

sebastian-lenz:
I prefer the truncated variant. The title should quickly indicate which page the user is on and the full title is always visible in the title field located on all tabs. Also, css cannot truncated multiline text (as far as I am aware) so with multiline display the header could grow in height to unexpected dimensions.

brandonkelly:
Also I like having the revision menu connected to the heading.

ChristopherDosin:
@brandonkelly Is it possible to refer to the commit with the fix, so i can take a look how you changed it?

PR #4681, opened 2019-07-31
Fix display of long titles

sebastian-lenz:
Resolves the issue described in #4548 by forcing the header to no wrap and apply the text overflow strategy as described here: https://css-tricks.com/flexbox-truncated-text/

Bildschirmfoto 2019-07-31 um 12 22 52

The entry edit header is still too wide on smaller screens, but at least it does not break apart. Maybe we could:
- Rename "Save entry" to "Save"
- Move "Save as draft" to dropdown in the save button
- Remove labels of "Preview" and "Share" Buttons, the icons are unique enough
- Move "Preview", "Share" and "Save as Draft" to a menu

### DIFF
--- a/src/templates/_layouts/cp.html
+++ b/src/templates/_layouts/cp.html
@@ -229,7 +229,7 @@
                         {% block header %}
                             {% block pageTitle %}
                                 {% if title is defined and title|length %}
-                                    <h1>{{ title }}</h1>
+                                    <h1><span>{{ title }}</span></h1>
                                 {% endif %}
                             {% endblock %}
                             {% block contextMenu %}{% endblock %}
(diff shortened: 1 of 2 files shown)


brandonkelly:
Thanks for the PR! Have you tested this change on mobile browsers?

sebastian-lenz:
Ah, mobile. I'll have a look...

sebastian-lenz:
Okay, with these changes in place it is easy to restore the old behavior for mobile using a media query, I've updated the PR.

However, the mobile appearance is not ideal either:

Bildschirmfoto 2019-07-31 um 17 26 09

KatieMFritz:
Still getting this on Craft Pro 3.2.9

Screen Shot 2019-08-09 at 5 38 48 PM

brandonkelly:
@KatieMFritz I just re-tested in Firefox, Chrome, and Safari, and long titles are getting truncated as expected. What browser are you using?

Did you get an error when updating? You could try running `composer install` for good measure. And maybe try emptying out your `web/cpresources/` folder, and clear your browser cache.

KatieMFritz:
I think it was my custom CSS. Sorry!

brandonkelly:
Haha, no worries :)

PR #4683, opened 2019-07-31
Remove expiryDate from actionCreateToken

jordikroon:
Shared URL's will automatically expire after 24 hours (`P1D`). 

This PR removed the `expiryDate` from the createToken method call. Doing so will change the expiration of the preview token to make it match with the `defaultTokenDuration` config setting. 

This defaultTokenDuration config value has a default value of 86400 / 1 day, thus only those who have altered this value will have a different behaviour.

### DIFF
--- a/src/controllers/LivePreviewController.php
+++ b/src/controllers/LivePreviewController.php
@@ -65,8 +65,7 @@ class LivePreviewController extends Controller
             ]
         ];
 
-        $expiryDate = (new \DateTime())->add(new \DateInterval('P1D'));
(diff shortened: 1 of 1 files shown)


brandonkelly:
The file you edited is the old Live Preview controller, which is only in used by categories and some third party element types, specifically for Live Preview, where I think a 24 hour limit was intentional and reasonable.

I’ve just made the same change to the new Preview controller though, which is used for previewing entries (both Live Preview *and* the Share button), so it makes more sense to have that one be configurable.

jordikroon:
I missed that. They looked similar. Thanks for adding!

PR #4707, opened 2019-08-05
Added condition to setSettings

### LINKED ISSUE #4706: Plugin setAttributes error due to getting/setting of settings behaviour
### Description

`Plugin::setSettings` is called where a plugin is initiated, however, the default behaviour is to call `Plugin::getSettings`, which can return `null`. This results in exceptions for plugins that do not have settings.

```
Error: Call to a member function setAttributes() on null
```

### Reference

https://github.com/craftcms/cms/blob/1239460455a61552510153fd249e48e3c325c27f/src/base/Plugin.php#L195-L206

https://github.com/craftcms/cms/blob/1239460455a61552510153fd249e48e3c325c27f/src/base/Plugin.php#L211-L214

joshuabaker:
Fixes #4706

### DIFF
--- a/src/base/Plugin.php
+++ b/src/base/Plugin.php
@@ -210,7 +210,11 @@ class Plugin extends Module implements PluginInterface
      */
     public function setSettings(array $settings)
     {
-        $this->getSettings()->setAttributes($settings, false);
+        $settingsModel = $this->getSettings();
+
+        if ($settingsModel) {
+            $settingsModel->setAttributes($settings, false);
+        }
     }
 
     /**


brandonkelly:
I don’t think we should just pretend nothing went wrong here.

joshuabaker:
Probably my naivety, but I’m not quite sure what went wrong. 😅

The `setSettings` method can definitely fail due to it dependence on `getSettings`, no?

brandonkelly:
It’s only a bug if the plugin somehow has settings stored in its `settings` column in the plugin table, which _should_ only be the case if the plugin actually has settings, so `getSettings()` returns a settings model, not `null`.

In Postmark’s case, it used to have settings and now it doesn’t. So the plugin itself should have been the one to clear out existing settings.

joshuabaker:
Thanks for the return and clarification, @brandonkelly. I appreciate that.

PR #4709, opened 2019-08-05
Add disabled class to disabled select fields

rynpsc:
Adds the disabled class when select fields are disabled.


brandonkelly:
Thanks!

PR #4711, opened 2019-08-06
Loading `project.yaml` and Fixtures

markhuot:
It is currently impossible to utilize the project.yaml _and_ fixtures in a test scenario. I'm not sure if this is something you'd like to support because it's non-trivial to add, but I figured I'd open a PR with some suggestions for your consideration.

First, the motivation for this is to implement TDD (or any testing) on a front-end only build of a website within Craft. E.g., imagine a developer working entirely in the template layer and the Craft admin. They may create several sections, add fields to those sections, configure some globals, etc. Additionally they may have some "boilerplate" entries that define things like a representative blog post or the scaffolding of the homepage.

Currently, the above developer would have to save everything out as a `project.yaml` file so they could sync configuration between production and staging. But for testing they would then have to redefine the section and entry type configuration in a second place, in fixtures, because they load too late in the test setup.

The attached fix moves the `project.yaml` setup out of the `_before` flow and moves it deeper in to the `loadFixutres` flow directly. This ensures it loads _before_ the fixtures but not before the app is initialized.

The _major_ downside to this approach is that the current `loadFixtures` method is `private` as defined in the Yii2 Codeception module. That means the attached won't actually work. If this seems feasible, though, I'm happy to open a PR over there for consideration of the upstream update.

If this doesn't make any sense and there's a better way to do this I'm all ears, but I couldn't find anything in my reading of the docs or my perusal of the source.

### DIFF
--- a/src/test/Craft.php
+++ b/src/test/Craft.php
@@ -166,6 +166,17 @@ class Craft extends Yii2
             return;
         }
 
+        $db = \Craft::createObject(
+            App::dbConfig(self::createDbConfig())
+        );
+
+        \Craft::$app->set('db', $db);
+    }
+
+    /**
+     * @inheritDoc
+     */
+    protected function loadFixtures($test) {
         // Re-apply project config
         if ($projectConfig = $this->_getConfig('projectConfig')) {
             // Tests just beginning. . Reset the project config to its original state.
@@ -189,11 +200,7 @@ class Craft extends Yii2
             }
         }
 
-        $db = \Craft::createObject(
-            App::dbConfig(self::createDbConfig())
-        );
-
-        \Craft::$app->set('db', $db);
+        return parent::loadFixtures($test);
     }
 
     /**


aaronbushnell:
Hey @markhuot! You _should_ be able to utilize both fixtures and project.yaml if I'm not mistaken? Here's how I have them setup in one of my tests:

**tests/fixtures/data/entries.php**
```php
<?php

return [
    [
        "sectionId" => $this->sectionIds['example'],
        "typeId" => $this->typeIds['example']['example'],
        "title" => "My Entry 1"
    ],

    [
        "sectionId" => $this->sectionIds['example'],
        "typeId" => $this->typeIds['example']['example'],
        "title" => "My Entry 2"
    ],

    [
        "sectionId" => $this->sectionIds['example'],
        "typeId" => $this->typeIds['example']['example'],
        "title" => "My Entry 3"
    ]
];
```

**tests/unit/ExampleTest.php**
```php
<?php

namespace myprojecttests;

use Codeception\Test\Unit;
use UnitTester;

use Craft;
use craft\elements\Entry;

use myprojecttests\fixtures\EntriesFixture;

class ExampleTest extends Unit
{
    // Public properties
    // =========================================================================

    /**
     * @var UnitTester
     */
    protected $tester;

    // Public methods
    // =========================================================================
    public function _fixtures(): array
    {
        return [
            'entries' => [
                'class' => EntriesFixture::class
            ]
        ];
    }

    // Tests
    // =========================================================================
    public function testTotalEntryFixtures()
    {
        // Confirm we have three entries
        $this->assertEquals(3, Entry::find()->section(['example'])->total());
    }
}
```

You'll notice I'm using 
```
"sectionId" => $this->sectionIds['example'],
"typeId" => $this->typeIds['example']['example'],
```
Where `example` is the handle of a "Section" and the handle of the "Entry Type".

markhuot:
@aaronbushnell, I'm not seeing the same thing. If you empty out your database and re-run the unit tests do you get a full install/setup for your tests? I can force it to work by updating `codeception.yml` to set `dbSetup: {clean: false, setupCraft: false}` and then the the tests don't affect my DB. But with those both set to true the fixtures and `project.yaml` seem to fight.

What's your `codeception.yml` look like?

aaronbushnell:
> If you empty out your database and re-run the unit tests do you get a full install/setup for your tests?

So I have my tests setup to purge my database and seed a new one (then it uses `project.yaml` to configure the schema my live site has)

> What's your codeception.yml look like?

Here ya go!

```yaml
actor: Tester
paths:
  tests: tests
  log: tests/_output
  data: tests/_data
  support: tests/_support
  envs: tests/_envs
settings:
  bootstrap: _bootstrap.php
params:
  - tests/.env
modules:
  config:
    \craft\test\Craft:
      configFile: 'tests/_craft/config/test.php'
      entryUrl: 'http://mysite.test/index.php'
      projectConfig: {file: 'config/project.yaml'}
      migrations: []
      plugins: []
      cleanup: true
      transaction: true
      dbSetup: {clean: true, setupCraft: true}

```

markhuot:
Hrm, I wonder why my experience is so different… I’ll try it on a fresh clone later and see what happens. Maybe my version is out of date.

gtettelaar:
@markhuot Just to check one thing. Do you have the [useProjectConfigFile](https://docs.craftcms.com/api/v3/craft-config-generalconfig.html#useprojectconfigfile) option enabled in `tests/_craft/config/general.php`

angrybrad:
@markhuot any luck using a fresh clone? I'm not seeing the behavior you're describing, either.

elivz:
In my case, adding a `tests/_craft/config/general.php` file and setting `useProjectConfigFile` fixed it. What was confusing is that without that, the project config actually *was* being loaded anyway, just after fixtures were run.

gtettelaar:
> In my case, adding a tests/_craft/config/general.php file and setting useProjectConfigFile fixed it. What was confusing is that without that, the project config actually was being loaded anyway, just after fixtures were run.

I'm assuming you mean the Project Config _data_ (i.e. Sections, Sites email settings e.t.c.) right?

elivz:
@gtettelaar Correct

gtettelaar:
**tl;dr**
#4804 will resolve the awkwardness that was causing confusion when mixing Project Config with Testing.

@elivz Yea this was weird behaviour. Here's a basic explanation:

If you had `useProjectConfigFile` enabled when running tests Craft would execute the following lines of code when setting up the testing DB: 
https://github.com/craftcms/cms/blob/82e14b2c1be3c9afeac9dbfb751f63da707153b9/src/migrations/Install.php#L1019-L1023

This executed _before_ any fixtures were applied. If you had `useProjectConfigFile` disabled Craft would not run this code so the data from `project.yml` would not be applied.  

After this Codeception/Yii load your fixtures. In your case, this was happening _without_ Project Config data. With for example @aaronbushnell the Project Config _was_ loaded. All because of what `useProjectConfigFile` was set to. 

Once that was done Codeception calls `craft\test\Craft::_before()`. Here because you specified a `project.yml` file in `codeception.yml` we call `Craft::$app->getProjectConfig()->applyConfigChanges()` with the Project Config data: 
https://github.com/craftcms/cms/blob/82e14b2c1be3c9afeac9dbfb751f63da707153b9/src/test/Craft.php#L170-L178

Now regardless of if `useProjectConfigFile` was enabled if you supplied a `project.yml` file we would parse that and apply it's config to the Testing environment. The end result was the fact that your project config data would be present in your tests but not in your fixtures (because `_before()` is called _after_ your fixtures are applied).

elivz:
Yes, that describes what I have been seeing perfectly. I just wasn't able to track down the root cause. Everything is working great since I added the `tests/_craft/config/general.php` file, and this additional warning should make things much clearer in the future. 

Thank you for all your hard work on this testing framework! It's really going to improve the whole plugin ecosystem in the long run.

angrybrad:
Assuming https://github.com/craftcms/cms/pull/4804 resolves this, so going to go ahead and close, but feel free to re-open if you're still having issues, Mark.

PR #4729, opened 2019-08-08
[Docs] Fix twig error

Jan10:
Twig Syntax Error – Twig\Error\SyntaxError
A hash key must be followed by a colon (:). Unexpected token "arrow function" of value "=>" ("punctuation" expected with value ":").


brandonkelly:
Thank you!

PR #4747, opened 2019-08-12
Added hardDelete param to deleteElementById

tim-pixeldeluxe:
Copied hardDelete paramter from the deleteElement function and added it to the deleteElementById function.


brandonkelly:
Good call, thanks!

PR #4750, opened 2019-08-13
Prevent the use of the actionTrigger in uriFormat

### LINKED ISSUE #4154: Prevent the use of '/actions' as a URI segment as it is reserved by the system
### Description
For our Craft site, a section was created named Actions, and the system automatically populated the "Entry URI Format" with `actions/{slug}` which lead to a great deal of confusion on my end when I tried to view the template I just created and was greeted with a 404 error.

After some research, I think that I have discovered that [`actions` is a reserved URI segment by the system](https://docs.craftcms.com/v3/config/config-settings.html#actiontrigger). If this is the case I would think that the system should prevent people from attempting to use it as part of their own URI formats.

### Steps to reproduce

1. Create a site section named "Actions"
2. Make sure the Entry URI Format is `actions/{slug}`
3. Try to access a template at that URI

### Additional info

- Craft version: Craft Pro 3.0.40
- PHP version: 7.3.1
- Database driver & version: MySQL 5.7.24
- Plugins & versions:

gtettelaar:
Resolves #4154. 

Prevents the use of the actionTrigger param in URI Format's.


brandonkelly:
Thanks Giel!

gtettelaar:
@brandonkelly You did most of the work😂. 

Just out of curiosity, what's the point of $disallowTriggers as a config setting? Is there a use case to entries having a URI that _does_ start with the action/cp trigger?

brandonkelly:
@gtettelaar Just best practice. It’s easy enough to have it configurable, and it’s easier to understand/document the behavior if it’s associated with a property that can be toggled.

gtettelaar:
@brandonkelly Hmm, fair enough!

PR #4757, opened 2019-08-14
[3.3]Handle based Global Set reference tag's

### LINKED ISSUE #4645: FR: Reference tags for globalset identified by handle instead of id
We're currently dealing with use case of 1 codebase for about 40 individual sites, so all the configuration and standardisation we can use is very welcome. 

I have however an issue with reference tags and globalset, it seems that globalset can only be identified using id and not their handle.  

The problem is that we're also heavily relying on migrations and project.yaml (which are blessings in the case) and the ids for globalset are not the same over the different environments, but the handles of course are.  

So now I'm using 
`{globalset:26:companyName} `
for instance, but that fails to work of course when the id of the globalset is not 26 for some reason

We would like to use 
`{globalset:agent:companyName} `
instead

This request is related to https://github.com/craftcms/cms/issues/1227

gtettelaar:
Resolves #4645

### DIFF
--- a/CHANGELOG-v3.3.md
+++ b/CHANGELOG-v3.3.md
@@ -1,0 +1,4 @@
+# Running Release Notes for Craft CMS 3.3
+
+## Changed
+- You can now use handles in Global set reference tags. ([#4645](https://github.com/craftcms/cms/issues/4645))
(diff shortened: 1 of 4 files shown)


brandonkelly:
Thanks Giel!

Issue #3937, opened 2019-03-02
[Feature Request] Add "title" argument to Twig svg() function


benface:
The `<title>` tag of an `<svg>` can be used to provide alternative text for accessibility reasons. It would be nice if we could add/replace the `<title>` tag when inlining SVGs with the `svg()` function in Twig. Sometimes the same icon can mean different things depending on the context, or on a multilingual site you may want to translate it.

michaelrog:
I could envision a more general `tags` param being useful. It could accept an array of tag names as keys, and markup as values — to allow adding/replacing `title` _or_ `desc`.

brandonkelly:
Closing this in favor of #4660

brandonkelly:
Sorry, just realized this is referring to a `<title>` HTML element, not the `title` attribute. Reopening.

brandonkelly:
Just added new `|append` and `|prepend` Twig filters for Craft 3.3, which append/prepend HTML elements to another element, including SVGs.

```twig
{{ svg('@webroot/icons/wave.svg')
    |append('<title>Waving Hand</title>')
    |append('<desc>A waving hand icon.</desc>') }}
```

If you think the SVG may already contain `<title>` / `<desc>` elements, then you can pass a second argument that determines how that should be handled to avoid duplicate elements:

- `'keep'` will leave the existing element alone and not add the new one
- `'replace'` will replace the existing element with the new one

```twig
{{ svg('@webroot/icons/wave.svg')
    |append('<title>Waving Hand</title>', 'keep')
    |append('<desc>A waving hand icon.</desc>', 'replace') }}
```

I’ve also [added](https://github.com/craftcms/cms/commit/a4104e119cbe2c27946ee8ae390b9e43f4291363) a new `tag()` Twig function, for rendering HTML tags using a declarative syntax:

```twig
{{ svg('@webroot/icons/wave.svg')
    |append(tag('title', {text: 'Waving Hand'}), 'keep')
    |append(tag('desc', {text: 'A waving hand icon.'}), 'replace') }}
```

You can combine these new filters with the [new `|attr` filter](https://github.com/craftcms/cms/issues/4660#issuecomment-522347729) as well:

```twig
{{ svg('@webroot/icons/wave.svg')
    |attr({role: 'img', 'aria-labelledby': 'title  desc'})
    |append(tag('title', {id: 'title', text: 'Waving Hand'}), 'replace')
[...]

PR #4762, opened 2019-08-15
[3.3] Allow custom attributes and elements in the `svg()` function

### LINKED ISSUE #4660: Add any attributes with svg()
### Description
We can already add classes to SVGs with this function, but it would be great if we could add any attribute.

It would make it a lot easier to follow [accessible SVG patterns](https://developer.paciellogroup.com/blog/2013/12/using-aria-enhance-svg-accessibility/), especially with #3937 and if there were some way to reference the title and/or desc elements added that way.

gtettelaar:
Proposal to resolve #4660 and #3937. 

Not sure if it's required for the old `string $class` argument to go through a deprecation phase until `4.0` so I can remove that if it's unnecessary.

### DIFF
--- a/src/web/twig/Extension.php
+++ b/src/web/twig/Extension.php
@@ -951,7 +951,7 @@ class Extension extends AbstractExtension implements GlobalsInterface
      */
     public function svgFunction($svg, bool $sanitize = null, bool $namespace = null, $attributes = [], array $elements = [], string $class = '')
     {
-        // Deprecate the $class argument
+        // Deprecate the $class argument. Allow both svg(class='class') and svg('', true, true, 'class')
         $oldClass = is_string($attributes) ? $attributes : $class;
         if ($oldClass) {
             $attributes = ['class' => $oldClass];
--- a/src/web/twig/Extension.php
+++ b/src/web/twig/Extension.php
@@ -1026,11 +1026,15 @@ class Extension extends AbstractExtension implements GlobalsInterface
         // Load up the attributes
         foreach ($attributes as $key => $value) {
             if (is_string($key) && is_string($value)) {
-                $svg = preg_replace('/(<svg\b[^>]+\b' . $key . '=([\'"])[^\'"]+)(\\2)/i', "$1 {$value}$3", $svg, 1, $count);
+                $encKey = Html::encode($key);
+                $encVal = Html::encode($value);
+
+                $svg = preg_replace('/(<svg\b[^>]+\b' . $encKey . '=([\'"])[^\'"]+)(\\2)/i', "$1 {$encVal}$3", $svg, 1, $count);
                 if ($count === 0) {
-                    $svg = preg_replace('/<svg\b/i', "$0 {$key}=\"{$value}\"", $svg, 1);
+                    $svg = preg_replace('/<svg\b/i', "$0 {$encKey}=\"{$encVal}\"", $svg, 1);
                 }
             }
+
         }
 
         return TemplateHelper::raw($svg);
--- a/src/web/twig/Extension.php
+++ b/src/web/twig/Extension.php
@@ -944,17 +944,15 @@ class Extension extends AbstractExtension implements GlobalsInterface
      * @param bool|null $namespace Whether class names and IDs within the SVG
      * should be namespaced to avoid conflicts with other elements in the DOM.
      * By default the SVG will only be namespaced if an asset or markup is passed in.
-     * @param string|null $class Deprecated - use $attributes instead.
-     * @param array|null $attr A list of attributes that should be added to the root `<svg>` element.
+     * @param string|null $class Deprecated - use |attr instead.
      * @param array|null $elements Any elements that should be added under the  root`svg` element
      * @return Markup|string
      */
-    public function svgFunction($svg, bool $sanitize = null, bool $namespace = null, string $class = null, array $attr = [], array $elements = [])
+    public function svgFunction($svg, bool $sanitize = null, bool $namespace = null, string $class = null, array $elements = [])
     {
-        // Deprecate the $class argument. Allow both svg(class='class') and svg('', true, true, 'class')
+        // Deprecate the $class argument.
         if ($class) {
-            $attr['class'] = isset($attr['class']) ? $attr['class'] . ' ' . $class : $class;
-            Craft::$app->getDeprecator()->log("svg(class=$class)", 'Passing a string $class argument is deprecated. Pass an array with a `class` key instead.');
+            Craft::$app->getDeprecator()->log("svg(class=$class)", 'Passing a string $class argument is deprecated. Use the `|attr` filter.');
         }
 
         if ($svg instanceof Asset) {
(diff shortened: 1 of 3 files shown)


src/web/twig/Extension.php
```diff
@@ -944,11 +944,20 @@ public function shuffleFunction($arr)
      * @param bool|null $namespace Whether class names and IDs within the SVG
      * should be namespaced to avoid conflicts with other elements in the DOM.
      * By default the SVG will only be namespaced if an asset or markup is passed in.
-     * @param string|null $class A CSS class name that should be added to the `<svg>` element.
+     * @param array|string|null $attributes A list of attributes that should be added to the root `<svg>` element.
```
brandonkelly: `$attributes` should be `$attr`, which should only allow `array|null`, not `string`, and be placed *after* `$class` for BC.

src/web/twig/Extension.php
```diff
@@ -944,11 +944,20 @@ public function shuffleFunction($arr)
...
+     * @param array|null $elements Any elements that should be added under the  root`svg` element
```
brandonkelly: This should be just be `$title`, which can be `string|null`. No need to support additional nested elements AFAIK. (And move this along with `$attr` to be after `$class`.)

src/web/twig/Extension.php
```diff
@@ -944,11 +944,20 @@ public function shuffleFunction($arr)
...
      * @return Markup|string
      */
-    public function svgFunction($svg, bool $sanitize = null, bool $namespace = null, string $class = null)
+    public function svgFunction($svg, bool $sanitize = null, bool $namespace = null, $attributes = [], array $elements = [], string $class = '')
     {
+        // Deprecate the $class argument. Allow both svg(class='class') and svg('', true, true, 'class')
+        $oldClass = is_string($attributes) ? $attributes : $class;
+        if ($oldClass) {
+            $attributes = ['class' => $oldClass];
+            Craft::$app->getDeprecator()->log('Extension::svgFunction()', 'Passing a string $class argument is deprecated. Pass an array with a `class` key instead.');
```
brandonkelly: The log key should be `'svg(class)'` – or something like that, to specifically refer to the `class` argument.

src/web/twig/Extension.php
```diff
@@ -1017,12 +1026,45 @@ public function svgFunction($svg, bool $sanitize = null, bool $namespace = null,
             }
         }
 
-        if ($class !== null) {
-            $svg = preg_replace('/(<svg\b[^>]+\bclass=([\'"])[^\'"]+)(\\2)/i', "$1 {$class}$3", $svg, 1, $count);
+        // Load up the attributes
+        foreach ($attributes as $key => $value) {
```
brandonkelly: You should be able to build the attribute HTML using `Html::renderTagAttributes()`, just like the `html()` function uses.

src/web/twig/Extension.php
gtettelaar: What about `<desc>` tags? The underlying topic of #4660 and #3937 was accessibility so wouldn't it be prudent to cover both elements - not just `<title>`?

src/web/twig/Extension.php
gtettelaar: Yea, the reason I added `string` was because `$attributes` took the position of `$class` - so BC meant it __could__ have taken in strings. Your solution is better.

src/web/twig/Extension.php
gtettelaar: image

src/web/twig/Extension.php
gtettelaar: Sure, but you still need the loop - I think. We need to check whether any attributes already exist in the SVG and deal with them. I.E. If someone specifies a `class` key in the `$attr` array, but the SVG already has a class. 

The rendered result would need both the class that already existed and the class defined in `$attr`. So we can take the same approach as in the `$elements` loop. Build the HTML with an `HtmlHelper::**` function and then add if afterwards but not sure if that's what you mean?

gtettelaar: A rough sketch of how that function _might_ look: 
https://gist.github.com/gtettelaar/389e1ad76740193c803c69b8b725dcdb

src/web/twig/Extension.php
brandonkelly: Maybe both `$title` and `$desc` then?

src/web/twig/Extension.php
brandonkelly: That doesn’t account for attribute names that are dynamically defined by `renderTagAttributes()`. For example if you pass `['data' => ['foo' => 'bar']]`, that will end up as `data-foo="bar"`.

Maybe best to just tack whatever attributes are defined here onto the end of the `<svg>` element, without changing any of the original attributes. If an element specifies the same attribute twice, browsers will just ignore the first one.

src/web/twig/Extension.php
gtettelaar: It's your call but personally, I'm not a fan. That means we end up with

- the duplication of running the same code twice or
- A loop using a hardcoded argument that we may as well make open to public use. 

It's a matter of preference so... Only thing I will say is if any accessibility tags are ever added in the future that will require another argument to the function.

gtettelaar: Also, the `<metadata>` tag could be supported by the `$elements` argument. Not sure what it's use is though?

src/web/twig/Extension.php
gtettelaar: > Maybe best to just tack whatever attributes are defined here onto the end of the <svg> element, without changing any of the original attributes. If an element specifies the same attribute twice, browsers will just ignore the first one.

What about [this](https://gist.github.com/gtettelaar/23fa9ba757889c60e78e8e37a029b114)?

brandonkelly: We can’t rely on the DOM extension for this, because that wasn’t built with SVGs in mind and errors out pretty easily.

I just added a new `|attr` filter that resolves #4660 (90c0eff97e3d4e55a53132e17217c2d40f92fd0b), so this change can be reverted completely now – except for the `$class` deprecation log.

gtettelaar: > I just added a new |attr filter that resolves #4660 (90c0eff), so this change can be reverted completely now – except for the $class deprecation log.

Nice! Reverted it

> We can’t rely on the DOM extension for this, because that wasn’t built with SVGs in mind and errors out pretty easily.

Just for my own curiosity do you have an article highlighting these flaws? I didn't come across this in any research, plus `enshrined/svg-sanitize` uses the DOM extension for SVG sanitation so Craft uses it indirectly even after this change was rolled back.

brandonkelly: To use it with SVGs you have to call `loadXML()` rather than `loadHTML`, but then it loses its ability to handle HTML syntax where it differs from strict XML rules (e.g. attributes that don’t have a value, and elements like `<br>` that aren’t self-closing but are void nonetheless). That’s fine for svg-sanitizer because SVGs are XML. But not reliable when embedding an SVG into an HTML document, at which point the SVG becomes HTML, and we need to be supporting HTML attribute syntax.

gtettelaar: Hmm, that's interesting! Thanks!

src/web/twig/Extension.php
brandonkelly: Ultimately I think you’re right that we shouldn’t limit the types of elements that can be added. But at that point it really doesn’t need to be a feature of the `svg()` function anymore. So I just [added](https://github.com/craftcms/cms/commit/5ee3d5f91fa90da2c36c8757224af767236efb1b) new generic `|append` and `|prepend` Twig filters, in the same vein as the new `|attr` filter, which can be used on SVGs or any other HTML element.

Full explanation here: https://github.com/craftcms/cms/issues/3937#issuecomment-522548611

gtettelaar: 2 birds one stone - nice one!

PR #4766, opened 2019-08-16
[3.3] Allow subnav's to have their own badge count

### LINKED ISSUE #4756: Plugin subnav menus to allow optional badge counts
### Feature Request!
Can we please have badge counts on subnav items? Noticed that we can have them on the main plugin item, and in sidebars, but not on subnavs. This could be an optional array key which is ignored if left blank.

https://github.com/craftcms/cms/blob/66e16f7f7a2bfaa0af97196c77c41a8346328db1/src/templates/_layouts/cp.html#L150

Updating the above code wrap the text in a span with class of label and then checking for item.badgeCount in a condition to add a span with class of badge would enable this. Then in our plugin where we set our subnav label and url we can easily just define a badgeCount also.

This would be a handy option to have where we are using subsections within plugins along with sidebars like so:

Screenshot 2019-08-14 at 16 18 20

The sidebar counters are already possible but the subnav ones aren't. Not sure if theres an explicit reason for this but could it please be added? Thank you.

gtettelaar:
Resolves #4756

### DIFF
--- a/CHANGELOG-v3.3.md
+++ b/CHANGELOG-v3.3.md
@@ -1,7 +1,7 @@
 # Running Release Notes for Craft CMS 3.3
 
 ## Added
-- Added the option for subnav's to have their own badge count. ([#4756](https://github.com/craftcms/cms/issues/4756))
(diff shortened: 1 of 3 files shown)


brandonkelly:
Thanks!

PR #4772, opened 2019-08-19
Testing: fix deprecation warning and error in PhpStorm

putyourlightson:
Fixes the deprecation warning (see http://phptest.club/t/bootstrap-deprecations-in-3-0/2196):

```
DEPRECATION: 'settings: bootstrap: _bootstrap.php' option is deprecated! Replace it with: 'bootstrap: _bootstrap.php' (not under settings section). See https://bit.ly/2YrRzVc 
```

And adds `ob_start();` to the main `_bootstrap.php` file to prevent `headers already sent` errors being thrown when running tests through PhpStorm (https://stackoverflow.com/questions/31175636/headers-already-sent-running-unit-tests-in-phpstorm/55040543#55040543).


codeception.yml
```diff
@@ -5,8 +5,8 @@ paths:
   data: tests/_data
   support: tests/_support
   envs: tests/_envs
+bootstrap: _bootstrap.php
```
gtettelaar: The old configuration option (placing `bootstrap` under `settings`) was deprecated as of [Codeception 3.1](https://github.com/Codeception/Codeception/blob/3.1/CHANGELOG-3.1.md) so if we want to go down this route the minimum composer version needs to be bumped aswell. Doing this in Codeception `3.0.X` would [break](https://travis-ci.com/craftcms/cms/builds/123672953) the tests because anything under 3.1 wouldn't understand the new `bootstrap: _bootstrap.php` syntax. 

Also, in line with this change, we can probably stop autoloading the fixture classes using `Codeception\Util\Autoload\Autoload` and start using Composer as the [blog post](http://phptest.club/t/bootstrap-deprecations-in-3-0/2196) advises.

@angrybrad Thoughts on bumping to `3.1`?

docs/testing/testing-craft/getting-started.md
```diff
@@ -108,6 +108,9 @@ define('CRAFT_TRANSLATIONS_PATH', __DIR__ . '/_craft/translations');
 define('CRAFT_VENDOR_PATH', dirname(__DIR__).'/vendor');
 
 TestSetup::configureCraft();
+
+// Prevent `headers already sent` error when running tests in PhpStorm
+ob_start ();
```
gtettelaar: Any reason we don't put this in `craft\test\TestSetup::configureCraft()`?

putyourlightson: You're right, I've moved it into `TestSetup`.

codeception.yml
putyourlightson: Thanks @gtettelaar!

putyourlightson: Strangely, the blog post title is "Bootstrap Deprecations in 3.0".

angrybrad: @gtettelaar Can't see any reason not to.

gtettelaar: Cool! 

I'll clean up the unrequired bootstrap files as well as the example suite for `./craft tests/setup` and then this will be ready to go.

gtettelaar:
@angrybrad Should be ready for you to have a look at

PR #4775, opened 2019-08-19
[3.3] User opt-in to show exception details even if `devMode` is off. 

### LINKED ISSUE #4765: Feature Request: devMode per user
You should be able to enable devMode per user and globally using the general config. You would enable devMode from the user preferences tab, the same way you enable the Yii debug bar.

This is especially useful when trying to find the cause of an error on your production env without having to enable devMode. Because sometimes you encounter errors on your production site which you cannot reproduce locally.

gtettelaar:
Resolves #4765


brandonkelly:
All merged, thanks!

src/web/ErrorHandler.php
```diff
@@ -129,25 +129,13 @@ public function isCoreFile($file)
      */
     protected function renderException($exception)
     {
-        // Treat UserExceptions like normal exceptions when Dev Mode is enabled
-        if (YII_DEBUG && $exception instanceof UserException) {
+        // Show the full exception view for all exceptions when Dev Mode is enabled (don't skip `UserException`s)
+        // or if the user is an admin and has indicated they want to see it
+        if ($this->_showExceptionView()) {
```
gtettelaar: Hmm... Really should have thought of doing it that way😅

PR #4776, opened 2019-08-19
Fix error in job queue example

medoingthings:
Noticed that the example task gets stuck, even thought it’s fairly simple. Turns out there was a typo.


brandonkelly:
Thanks!

PR #4781, opened 2019-08-20
Clear existing user groups and field layouts

olivierbon:
This avoids leaving orphaned/duplicated user groups in project config on `project-config/rebuild`

### DIFF
--- a/src/services/ProjectConfig.php
+++ b/src/services/ProjectConfig.php
@@ -889,7 +889,10 @@ class ProjectConfig extends Component
             'config' => $configData,
         ]);
         $this->trigger(self::EVENT_REBUILD, $event);
-
+        
+        // Remove any existing user groups and fieldlayouts from $currentConfig
+        unset($currentConfig['users']['groups'], $currentConfig['users']['fieldLayouts']);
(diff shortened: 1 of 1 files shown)


brandonkelly:
@olivierbon @andris-sevcenko Should this be expanded to all the things that `rebuild` is going to recreate?

olivierbon:
System routes and emails only exist here so no, and not really possible with plugins because of settings. The rest (fields, sections...) are already getting overwritten completely.

andris-sevcenko:
Yeah, the whole point of this PR is to fix an oversight in the rebuild process. When recovering data from the PC that only exists there before rebuild (plugin info, routes, etc), there was an oversight that the `user` key is sort of a hybrid. It contains stuff that only exists in the PC, but also contains stuff that exists in the DB. This PR unsets the stuff that can be rebuilt from the DB before merging the two arrays.

ryanpcmcquen:
Strange that this would get deployed when this didn't: https://github.com/craftcms/cms/pull/4654

Especially since the 3.1.x branch is unusable without that fix.

PR #4793, opened 2019-08-21
Clarify Project Config in testing docs

gtettelaar:
Based on #4711 it's probably smart to add a note stating that the `useProjectConfigFile` general config option must be enabled for project config to work.


angrybrad:
Thanks!

Issue #3698, opened 2019-01-23
Rebuilding search index takes a long time


chrislam:
### Description
Rebuilding the search index on my project is taking a very long time. At this point it seems to be about 10% done after about 3 hours.

It also seems that I need to keep the Search Indexes page open to continue rebuilding the search index and if it closes or my session ends I have to restart from the beginning.

It would be good if the rebuilding process could be more robust, not requiring the user to keep a window open and session active.
It could be a good idea to make this operation run as a job or multiple jobs working on batches of content.
Another idea could be to add a console command to rebuild the search index.

### Steps to reproduce

1. Click the Update search indexes button on the Search Indexes Utility screen.

### Additional info

- Craft version: Craft Pro 3.1.3
- PHP version: 7.2.5
- Database driver & version: MySQL 5.6.42
- Plugins & versions:
    - Contact Form 2.2.3
    - Contact Form Extensions 1.0.13
    - Contact Form Honeypot 1.0.2
    - Cookies 1.1.11
    - Element API 2.5.4
    - Queue Manager 1.1.0
    - Redactor 2.2.1
    - Scout 1.1.1
    - Super Table 2.1.5.1
    - Video Embedder 1.1.1

brandonkelly:
Craft 3.1.2 added an `index-assets` console command that you can use instead of the web-based utility. (#3595)

Sounds like that will work better for you :)

Run this to learn more about it:

```bash
./craft index-assets --help
```

chrislam:
That command seems to be solely for asset volumes, I am talking about the search index for content.

guyinpv:
I concur with this one.
I was just playing around and sort of accidentally clicked to rebuilt index for fun, not thinking it would be an all-day event.

Also found out the hard way I have to leave the page open.

I only have about 700 entries but this is going to take forever. 

I'm fine with it taking forever, but would be nice to not keep page open, or to at least have a progress bar or some kind of warning like "leave this page open!" and let me know when it's done, etc.

echantigny:
I'm also looking for a better way of doing this.  I have over 5000+ commerce products, 10000 users and a dozen pages.  It takes hours for me too of course.  A CLI command to trigger this would be great, with a progress visual if possible.  

And if that makes any sense, might be worth deleting/recreating the items one by one instead of clearing the whole table out before starting.  That way, any search functionality would still mostly work on the front end if anything needs to happen.

himynameisphil:
[...]

PR #4801, opened 2019-08-22
Remove references to search indexes

iparr:
This no longer exists, in settings or anywhere (?) according to https://github.com/craftcms/cms/issues/3698


brandonkelly:
Doh, thanks for pointing that out! It’s still possible to rebuild search indexes using the `resave` command, so I’ve updated the docs to point to that.

PR #4831, opened 2019-08-28
Add Markdown directive to GraphQL

AugustMiller:
This is my first time using GraphQL, and first time looking into its internals… so this is mostly a clone of the FormatDateTime directive.

It appears to work as intended, with arguments documented for auto-completion in the IDE.

Let me know if this is useful, or if it needs to be reworked!


andris-sevcenko:
@brandonkelly is [right](https://github.com/craftcms/cms/issues/4832#issuecomment-525940765) and Craft GraphQL API could benefit from small and simple directives. Nobody is forcing anyone to use them in environments where they are not supported.

PR #4833, opened 2019-08-28
Feature/graphql console command

khalwat:
Adds two console commands:

- `graphql/dump-schema`   Dump out a given GraphQL schema to a file
- `graphql/print-schema`  Print out a given GraphQL schema

Both operate on either the currently active schema, or fall back on the public schema. You can also pass a `token` parameter in to designate a schema by token:

```
--token: string
  The token to look up to determine the appropriate GraphQL schema
```

e.g.:

```
graphql/dump-schema --token=woof
```

This is especially useful for IDE auto-completion of GraphQL schemas.


lukeholder:
@khalwat general config file deleted?

PR #4834, opened 2019-08-28
Add Graphql console controller

khalwat:
Adds two console commands:

- `graphql/dump-schema`   Dump out a given GraphQL schema to a file
- `graphql/print-schema`  Print out a given GraphQL schema

Both operate on either the currently active schema, or fall back on the public schema. You can also pass a `token` parameter in to designate a schema by token:

```
--token: string
  The token to look up to determine the appropriate GraphQL schema
```

e.g.:

```
graphql/dump-schema --token=woof
```

This is especially useful for IDE auto-completion of GraphQL schemas.


pauloelias:
Awesome work @khalwat!

In addition to the 👍 I would also like to say this will be a huge help to our workflow since our headless CMS does not have `devMode` enabled. We still need to be able to test queries with the most up-to-date schema and I'd prefer not to have to fork this (now) huge database to a staging environment every time we update content models.

When we run headless we only spin up temporary review apps for each release but we don't need a long-running staging server in between releases because we don't need to worry about any frontend twig templates breaking.

andris-sevcenko:
You are a beautiful man and let noone tell you otherwise.

PR #4841, opened 2019-08-28
Add export events to modify exports

engram-design:
We'd like to modify the output of exports, including the original query. I'm wondering if the amount of information we're passing to the event is overkill, but figure you might as well pass as much as you can.

I've added two events, one to just modify the element query, the other to modify the output once the query has been fetched.

Our use-case is modifying Commerce Orders to export line-items as rows.

Thoughts?

### DIFF
--- a/src/controllers/ExportController.php
+++ b/src/controllers/ExportController.php
@@ -9,6 +9,7 @@ namespace craft\controllers;
 
 use Craft;
 use craft\base\ElementInterface;
+use craft\events\ExportEvent;
 use craft\helpers\ElementHelper;
 use craft\helpers\FileHelper;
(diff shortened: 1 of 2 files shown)


brandonkelly:
I’m not sure this is a good idea; it would back us into a corner in terms of how exporting works, and the current solution is not meant to be the long term solution.

engram-design:
Right, so no other way to modify the export?

Notably, there's actually conflicts with Commerce exports and the native export. Two different buttons produce two different results, which is extremely confusing - https://github.com/craftcms/commerce/issues/976

Otherwise, I'll hide these export buttons altogether and spin up by own export page.

Issue #4851, opened 2019-08-29
[FR] Flush revisions CLI command/Utility


olivierbon:
It would useful to have a CLI command/Utility to flush revisions when `maxRevisions` is adjusted mid-project.

As an example, if you decide to set `'maxRevisions' => 10` on a project which didn't have a limit before, all revisions beyond the limit will only get flushed when the entry itself is resaved in the Control Panel.

gtettelaar:
Wouldnt `resave/*` clear the revisions? Admittedly it's not purpose-built and so a little slower but it should work fine?

olivierbon:
Hey @gtettelaar, no it doesn't. Revisions get flushed [when a revision is created](https://github.com/craftcms/cms/blob/develop/src/services/Revisions.php#L188).

gtettelaar:
@olivierbon Doh! Didn't look at the code but just assumed this would happen during saving in `->getElements()->saveElement()`.

michtio:
+1 on this tho, deleting entry revisions should be possible or have a dedicated command line exec.

brandonkelly:
Just added a `utils/prune-revisions` command for Craft 3.4.

PR #4852, opened 2019-08-29
Get rid of excess revisions on resave

gtettelaar:
Propsal for #4851

Not sure if bundling the purging of revisions to the resave process is the direction you want to go in however seeing as the [search index updating](https://github.com/craftcms/cms/issues/3698#issuecomment-503317402) is now also bundled to resaving it seemed appropriate.

### DIFF
--- a/CHANGELOG-v3.md
+++ b/CHANGELOG-v3.md
@@ -6,8 +6,10 @@
 - Added the `enableGql` config setting. ([#4836](https://github.com/craftcms/cms/issues/4836))
 - Added the `children` field to the `EntryInterface` and `CategoryInterface` GraphQL types. ([#4843](https://github.com/craftcms/cms/issues/4843))
 - Added the `markdown` GraphQL directive. ([#4832](https://github.com/craftcms/cms/issues/4832))
+- Added the `craft\services\Revisions::pruneExcessRevisions()`.
 
 ### Changed
+- Craft will now prune excess revisions when resaving elements. ([#4851](https://github.com/craftcms/cms/issues/4851))
 - Preview target URIs can now be set to environment variables (e.g. `$NEWS_INDEX`) or URLs that begin with an alias (e.g. `@rootUrl/news` or `@rootUrl/news/{slug}`).
 - Templates passed to `craft\web\View::renderString()` and `renderObjectTemplate()` can now incude front-end templates.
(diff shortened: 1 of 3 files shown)


brandonkelly:
I don’t think it’s worth the extra effort each time an element is resaved, since the only time this is an issue is when the `maxRevisions` value changes.

gtettelaar:
@brandonkelly Correct me if I'm wrong but the main reason a dev, or Craft, would call `resave()` is to ensure that the selected Elements are as up-to-date with the Project Configuration as possible and to replace any old data (i.e. URI's in `{{%elements_sites}}`) with new data. Ensuring there are not any excess revisions seems part of that scope IMO. Does it add to the execution time of the function? No doubt, but so does calling `saveContent()` when the only thing you want to achieve is updating the element URI's or title's right?

For the same reason, you now use `resave/` if you want to update the search indexes I would say It's logical to have the purging of revisions be a part `resave/` aswell. I altered the [code to be less opinionated](https://github.com/gtettelaar/cms/blob/feature/prune-on-resave/src/console/controllers/ResaveController.php#L294-L326) and only purge revisions when `resave/` is called, not globally on `craft\services\Elements::resave()`. Not sure if there is another alternative other than creating a purpose-built `purge-excess-revisions` console command, but if you go down that route then an `update-search indexes` command would also make sense.

Not sure if I'm missing something in the code about this but from my perspective as a user this made sense. Honestly, I actually assumed [purging of revisions was already done by](https://github.com/craftcms/cms/issues/4851#issuecomment-526194958) `resave/`😂. 

P.s
Another solution might be to just store the revision limit on the individual sections and add a `getRevisionLimit()` method to `craft\base\Element`. That would add more granular control on how many revisions to store on a section-by-section basis with the side benefit that we would only have to purge `maxRevisions` once the Project Config value changes. Might be something for a different FR albeit I have never had a use for this setting so not sure _how_ much it would make sense.

brandonkelly:
Updating search indexes makes sense because there’s a chance that a field type was added/removed/updated that could cause the search keywords to differ from the last time elements were saved.

My point here is that in general the only time the `maxRevisions` setting needs to be enforced is when a new revision is added, and the one and only exception to that is if the setting value changes to a something smaller. Considering that, a dedicated CLI command seems like it would make more sense.

gtettelaar:
Yeah alright, that's fair enough!

PR #4860, opened 2019-08-29
Allow changing template that will be loaded in "before" web View events

joe-hx:
Enhancement to web View events `EVENT_BEFORE_RENDER_TEMPLATE` and `EVENT_BEFORE_RENDER_PAGE_TEMPLATE`, allowing the developer to replace the template that will be rendered. This allows a variety of creative solutions where routing may not be appropriate.

Before this, you cannot change the template about to be rendered, requiring the template to process and then overwriting the `$output` of the "after" events. With the suggested change, you can now prevent rendering of the undesired template with something along the lines of:
```
Event::on(
    View::class,
    View::EVENT_BEFORE_RENDER_TEMPLATE,
    function(TemplateEvent $event){
    	$event->template = 'something-else';
    }
);
```

If #4570 gets merged in and this causes a conflict, happy to resubmit.

### DIFF
--- a/src/web/View.php
+++ b/src/web/View.php
@@ -1332,7 +1332,7 @@ JS;
      * @param array &$variables The variables that should be available to the template
      * @return bool Whether the template should be rendered
      */
-    public function beforeRenderTemplate(string $template, array &$variables): bool
+    public function beforeRenderTemplate(string &$template, array &$variables): bool
     {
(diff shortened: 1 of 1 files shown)


brandonkelly:
This seems like a bad idea. If you tell Craft to load X template, it should be clear that that’s the exact template that will be loaded.

PR #4861, opened 2019-08-30
Small typo in group filter example

mikewink:
Running the code sample as-is results in an error. I swapped the single quotes around the Y to double quotes and now it works.


brandonkelly:
Thanks!

PR #4879, opened 2019-09-04
Feature/gql def fields event

narration-sd:
This is a feature constructed on basis of some conversation with @andris-sevcenko, which provides the ability of the new Craft-native GraphQL to be amplified with additional or substituted configuration for its handling of Craft fields.

For example, field arguments and resolvers to handle them can now be configured, by returning a straightforward nested array from the Event which defines them per element and field.

This in turn makes it possible for Craft-native GraphQL to provide Craft abilities like `transforms` on image Assets, in a syntax that's widely acceptable, because it doesn't involve GraphQL directives. Custom directives are often proscribed, but the Craft abilities are needed across a wide range of platforms.

Putting these upgrading abilities on an Event it's agreed gives a lot of flexibility, such as Craft providing a standard but modifiable set via a php /config file, the same provided by a dedicated plugin as is being considered, or embedded in a plugin which has its own purposes, while needing to specify upgrading to fully enable them.


andris-sevcenko:
This is not exactly what I was going for and still seems to be a little bit too complex.

I'll see if I can get to this later today.

andris-sevcenko:
Took a closer look at the problem at hand. It would be best solved with an event system similar to what Elements have, where you would listen to a GraphQL Asset type firing an event every time an Asset GraphQL type for a specific volume is generated. That would require some changes in the current architecture. Not major, but still changes enough to warrant a 3.4 release.

I feel like this should come after DotAll conference, and we have more capacity once again. Then I could focus on GraphQL things, including Mutations and make this happen.

narration-sd:
I don't know where you are on this (and  our off-line conversation about it), Andris @andris-sevcenko, but I noticed something extra while testing this correction of most recent merge to keep this pull request current with ongoing Craft cms/develop.

It's that when using the directive form of transform, you get a different target folder for the eventual transformed image - snake-cased, rather than camel from the actual tranform name. .That will cause a mismatch I think between what Craft GraphQL returns, and what a plain Twig script (or CraftQL for anyone wanting to transition) returns.

andris-sevcenko:
Opened a new issue for the transform thing.

narration-sd:
Good to see. As in other note, a reminder it isn't just a 'transform thing'.

Rather, any present and future argument, on any query field, is the range that needs to be able to work that way, for Gatsby, for Gridsome, and no doubt for other Facebook-compliant frameworks, now or to come.

And what range of powers you want to provide - we discussed some of that, but at least to add args and/or change (overall) `resolve`(r) functions for current or added. Taking out args w/suitable `resolve` change is the one thing I didn't already do -- some question in my mind about that one....

Anyway, I think you probably know these things, when remembering, and the several possible arguments for your image Assets probably are a great place to start.

narration-sd:
@andris-sevcenko  Andris, this upgrade appears to successfully handle letting the same server handle both directive-style and element-arguments-style transformation queries. Figured I would leave a more complete approach; perhaps some portion will save you time in framing your apparent other direction. 

There's some assumptive logic you'll want to have a good think about in either case...I've left a note, but it does appear to test ok here on the simple 'transform' part of the needs.

narration-sd:
Not relevant tp PR unless some of this may get used, but the check failing looks like a clock boundary problem in the test harness, rather than anything in the code...

narration-sd:
@gtettelaar  Geil, realized I should probably be notifying you on the apparent timing issue in the test harness showing up here. 

Much appreciation for your intrepid work on getting that which helps out us all, tusen takk!

narration-sd:
@gtettelaar  Geil, this gets complicated, doesn't it. I don't think today's run's fail on pgsql has anything to do with me either. It was also just a dutiful catch-up to present state of cms/develop; nothing changed in the PR code.

The run that had the timing match failure, which I mention just above for your attention, was the preceding test of this pull request. I presume you can find it by some identiy or date/time in the logs, as of course I didn't realize I needed to take that reference before I set off a new test run.

gtettelaar:
@narration-sd I have just fixed the test which failed because of the Time delta (#4996). 

I don't know why the `SearchService` test is failing but I'll ping @angrybrad as he is better at debugging those than me and it looks like it's a [recurring(ish) problem](https://travis-ci.com/craftcms/cms/jobs/237913025).

narration-sd:
@gtettelaar  That sounds very good all around, Giel, appreciated.

And, I will take care not to transpose letters in your name in future...  :)

andris-sevcenko:
Closing this in favor of https://github.com/craftcms/cms/commit/a5dd6f2eef8f1a8a5f81edd13ea123340b5cde69 which accomplishes the same thing and was, in part, inspired by this PR.

PR #4947, opened 2019-09-17
GlobalSet Element Fixture improvements

boboldehampsink:
Adds ability to specify whether to use AR or Element 

@gtettelaar


angrybrad:
@boboldehampsink Thanks... probably should update the docs here, no? https://docs.craftcms.com/v3/testing/testing-craft/fixtures.html#global-set-fixture

gtettelaar:
@boboldehampsink I like this idea! This way we preserve the original functionality of your plugin.

I need to have a look at the implementation as I see 2 problems from my side: 

- `: void` is being used which clashes with Craft's PHP 7 mininum requirement. 
- It looks like the code used in `load()` is copied from your own plugin which would mean it will not work with `craft\test\fixtures\elements\ElementFixture` (mainly because our class does not have a `saveElement()` or `getErrors()` method). In general I'm not a fan of having the GlobalSet class define it's own logic for saving an element as opposed to having that be handled by the main parent class. 

I'll work on this tommorow and then it should be ready. 

@angrybrad I'll also update the docs.

boboldehampsink:
Yeah sorry, didn't test this yet as I just wanted to get the idea out there. Thanks for the help @gtettelaar

gtettelaar:
@boboldehampsink Do you have [allow edits from maintainers](https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork) enabled? I can't seem to push up to your branch.

boboldehampsink:
It is enabled!

boboldehampsink:
@gtettelaar did you get it to work?

gtettelaar:
@boboldehampsink Yea I got it sorted. Git Tower was being annoying on my end. I updated the code, can you check whether this was what you had in mind?

boboldehampsink:
Looks good :shipit:

boboldehampsink:
I tested this and it does what I want. Looks good to me!

angrybrad:
Looks good... merged!  Thanks all.

PR #4959, opened 2019-09-18
Update asset-bundles.md

smockensturm:
You may want to add a warning that registering this bundle - as written - will include ALL the CpAssets. (Unless you get those lines out of there.)


brandonkelly:
Thanks for the typo fix!

Issue #1144, opened 2017-01-31
add additional 'sendActivationEmail' system setting instead of using 'requireEmailVerification' to trigger activation emails


angrybrad:
> Created by: Nate Iler (nate@flipboxdigital.com) on 2015/03/11 14:29:14 +0000
> Votes at time of UserVoice import: 2

-----------------------------------

It is currently impossible to suppress sending activation emails while requiring email verification.  These actions should be evaluated independently as in some instances new registrants should not be notified.  In addition, disabling admins from triggering notifications is also necessary.

brandonkelly:
@nateiler You still feel this is needed?  If so what sort of flow would you hope for?

engram-design:
@brandonkelly Just have a client asking for this. Our use-case is that from a customer/user point of view, they don't want them to have to activate their account if they themselves are the ones signing up. However, they do a lot of phone orders (Commerce site), and they want to be able to create users for phone orders, or in-person.

At this point, they want to be able to create the user on behalf of them, and send the user an activation email. I'm aware that this activation is technically not needed, as the user is already activated, but its a much better AX than creating the user and sending them a password reset email. That's the method at the moment, but its also a little awkward for the registered user (it'd be better to get an email prompting to activate, then reset a password to an account you suddenly have).

The reason we don't have this setting on at all is because we have a password field during checkout so that users can create an account then and there. Its annoying to go through checkout, only to get another email to activate your account, despite literally just buying something (and getting an order confirmation email).

Hope that makes sense? I guess I'm just proposing to take the `{% if requireEmailVerification %}` check out. Or at least have an additional check for `{% if requireEmailVerification or sendActivationEmail %}`.

PR #4981, opened 2019-09-24
Allow users with administrateUsers permissions to choose if verification mail should be sent.

christianruhstaller:
Users with the permission "administrateUsers" should be allowed to choose if a verification mail should be sent when creating a new user.

A typical use case would be that users will be created for a frontend login and have custom fields with documents and co and the creator does not have all information to give access to the user but still wants to create the user and later at some points wants to send the activation mail.

### DIFF
--- a/src/templates/users/_edit.html
+++ b/src/templates/users/_edit.html
@@ -75,7 +75,7 @@
 
                 {{ forms.textField({
                     label: "Email"|t('app'),
-                    instructions: (requireEmailVerification and not currentUser.admin ? 'A verification email will be sent automatically.'|t('app')),
+                    instructions: (requireEmailVerification and not currentUser.admin and not currentUser.can('administrateUsers') ? 'A verification email will be sent automatically.'|t('app')),
                     id: 'email',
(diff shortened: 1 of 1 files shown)


brandonkelly:
Agree, but come to think of it, there’s no reason we can’t give _all_ users that are allowed to register new users, the choice on whether an activation email gets sent out. So I just committed my own change that simply removes the admin checks, rather than replacing them with administrateUsers checks.

engram-design:
@brandonkelly Does this only cover adding new users? Just noticed - and not sure if this is new behaviour - when as an admin, editing a users' email will not change it immediately, instead send a "Verify your email" to the user.

The user setting "Verify email addresses?" is enabled.

Is this new behaviour, a new issue, or unrelated?

brandonkelly:
@engram-design seems unrelated to #1144 to me; can you post a new issue?

brandonkelly:
Nevermind, looks like @lindseydiloreto beat you to it: #5088

engram-design:
Thanks @brandonkelly !

PR #4990, opened 2019-09-25
[BUGFIX] Fix sync when  VolumeRecord has no id

jmverges:


brandonkelly:
Thanks for the PR & bug report! Looks like `_getVolumeRecord()` will _always_ return a record, so I’ve just applied a slightly different fix.

jmverges:
I will try it. Thanks!

PR #4996, opened 2019-09-26
Fix users tests

gtettelaar:
Fixes problem's with a UsersTest -> https://travis-ci.com/craftcms/cms/jobs/236296640 and the UserTest -> https://travis-ci.com/craftcms/cms/jobs/234110077


brandonkelly:
Thank you sir!

PR #5006, opened 2019-09-27
added method getDynamicRangeUrls

hiasl:
I added a new method to the Pagination Object to allow to get dynamic range URLs for paginations like Google has. Please feel free to choose a better name for the method, if you don't like it.


brandonkelly:
Can you explain what exactly that means?

hiasl:
OK, I will try
Google Pagination (and others) provides direct links to 10 (or 11) result pages.

Bildschirmfoto 2019-09-27 um 20 05 22

When you first search (and assuming you get more than 10 pages of results), Google will show its results and then Links for the pages 1 to 10, where 1 is the current page.

Wenn go further down the result pages, e.g. to page 7, Google will show direct links for pages 3 to 12:
Bildschirmfoto 2019-09-27 um 20 08 51

So as you can see in the first and second screenshot, Google is always showing 10 direct links (where of the current page is one). If you're on page 1, 9 other links are shown afterwards. If you are on page 8, 5 are shown before (3-7) and 4 are shown after the current page (9-12).

Now the part, that is most challenging when implementing this with TWIG ,is to actually calculate the amount of links to show before and after the current page. It always takes some time to implement this.

So this is the first thing, my new method does, providing and array with optimized links for this  (the amount of links is the first parameter). I do it a little different than Google though, because my method always shows the same amount of links before and after the current one if possible. Google shows 5 before and 4 afterwards to get an equal number of 10 links.

Some websites in addition always provide a link to the first and the last page, as example it would provide links
1 .... 4,5,6,7,**8**,9,10,11,12 .... 19

Now this is also possible with my method, by setting the second parameter to true.

And the returned array is constructed in the same way the other methods of the Pagination object return the links, with the page number as assoc. array key.

Is it clear what I mean?

brandonkelly:
Thanks for the PR! Just cleaned it up a bit, and merged into the 3.4 branch. I decided to remove the first/last argument, because that would be easy enough to add via Twig, and I think it’s better for this method to be more similar to `getRangeUrls()`.

hiasl:
Your code is very clean, but I think it does not deal with the the case that you're e.g. on page 9 of 10. In this case it will start at 4 but it should start at 1. One of my devs said you must somehow find out if you are in the upper half of the complete set or the lower half and then either start with calculating $end first or $start.

brandonkelly:
Yeah I noticed that after pushing, and fixed it in another commit (d049dad2b16b6e1e09970c11cbf93783ef4423cf).

PR #5007, opened 2019-09-27
Fix bug where label, instructions, and warning text no longer got processed as raw

BenParizek:
The latest change to the `_includes/forms/field` file seems to have broken HTML output in labels and instruction text. Example:
http://share.barrelstrength.co/bf6faf1b3d70

There is a lot going on in this file and I'm not fully convinced I understand the processing order so probably best to check my commit, however this update got things working again for me locally.

The is defined test only works with simple variables, so adding `{{ instructions|raw ?? '' }}`` doesn't work as a shorter solution so I had to use the full `is defined` check: `{{ instructions is defined ? instructions|raw : '' }}`

It appears with several of these variables, the `raw` filter is being used both when setting the variable and when outputting it. Perhaps all those filters can just be moved to when the variables are being set? I didn't make that update as I seemed to get mixed results in my test as to which filters were actually having any effect.


brandonkelly:
Thanks for the PR. I had already fixed it locally though. It was introduced by a24f284a5bb885be03a109bce784ac8079cfae74 which was me attempting to work around a Twig bug that has since been fixed. Twig fix hasn’t been released yet but it’s not a serious problem so I’ll just live with the bug for now.

brandonkelly:
(and 3.3.7 is out now with that fix.)

PR #5026, opened 2019-10-01
Add database replica settings

lesaff:
There's also addition/changes to `config/db.php` and `.env` that is not in this repo.

### DIFF
--- a/src/config/DbConfig.php
+++ b/src/config/DbConfig.php
@@ -107,6 +107,18 @@ class DbConfig extends BaseObject
      * @var string The database username to connect with.
      */
     public $user = 'root';
+    /**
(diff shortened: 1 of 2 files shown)


angrybrad:
Thanks for the PR. Ended up taking it in a bit of a different direction so we could support multiple writers as well and added several new config settings, so will go into Craft 3.4.  There's currently a WIP here for it. https://github.com/craftcms/cms/tree/feature/replication

brandonkelly:
We ended up coming to the conclusion that no change is necessary to support read/write splitting.

We have updated the docs with an example of how to pull it off from `config/app.php` though: https://docs.craftcms.com/v3/config/app.html#database-component

Issue #5049, opened 2019-10-04
Allow custom config to Config service


qrazi:
### Description
Allow adding configs to Config service. This allows for nicer configuration of modules (i.e. not plugins). 

Background: we use modules for customization to Craft CMS that are very specific to a project , often without admin GUI stuff, and thus are not going through Composer managed plugins.

Right now, as far as I can tell, using configurable values specific to a module involves either dumping them in `config/general.php`, or manually create some sort of config structure within the module.

We propose allowing a module to register a `BaseObject`-configuration to the Config-service during init.

Within the module it could look something like:

```php
// ./modules/some-module/config/config.php
return [
    'foo' => 'bar',
];

// ./modules/some-module/SomeModule.php
//....
public function init()
{
    // ...
   
   // Use a defined class or a anonymous class, up to user
   $config = new ConfigClass(@include Craft::getAlias('@ConfigAlias/config.php'));
   $config_service->appendConfig('some-module', $config);   

    // ...
}
//....
```

### Steps to reproduce

1.
2.

### Additional info

- Craft version:
- PHP version:
- Database driver & version:
- Plugins & versions:

brandonkelly:
You have a couple options already:

- Your module actually extends BaseObject itself, so you could make it configurable by adding some public properties to it, and setting their values from `config/app.php`:

  ```php
  'modules' => [
      'my-module-id' => [
          'class' => MyModule::class,
          'foo' => 'bar',
          // ...
       ],
  ],
  ```

- If you want to store your config options in a separate file within `config/`, you can load the contents of that file using [craft\services\Config::getConfigFromFile()](https://docs.craftcms.com/api/v3/craft-services-config.html#method-getconfigfromfile).

  ```php
  $config = Craft::$app->config->getConfigFromFile('my-module');
  Craft::configure($this, $config);
  ```

Either way, you’ll get [multi-environment config](https://docs.craftcms.com/v3/config/environments.html#multi-environment-configs) support.

PR #5050, opened 2019-10-04
Changed: append custom config to Config-service

qrazi:
Related issue: #5049

If accepted / considered, could also expand the docs.


brandonkelly:
I don’t think this is necessary - see my comment at https://github.com/craftcms/cms/issues/5049#issuecomment-539615912.

PR #5058, opened 2019-10-06
Add translation for section types, better Persian translation, fix menu button icon in RTL

vnali:
- update Persian translations
- section types are translatable [in this line](https://github.com/craftcms/cms/blob/4a00146df167dac0beb05f32280f73caaa338d18/src/controllers/SectionsController.php#L107,) but Single and Channel are missing in translation files -fix was only applied to en and fa translation files-
- fix a typo in ElementInterface.php file
- trash and draft icon had bad position in RTL CP -the fix was only applied to _main.scss file-


brandonkelly:
Thanks!

PR #5072, opened 2019-10-09
Add codeception option `projectConfigOptOut`

dhahn:
**This PR offers an option to skip rebuilding the project config before each test is run.**

This change would allow developers to add projectConfigOptOut to their
codeception.yml:

```
modules:
    config:
        \craft\test\Craft:
            configFile: 'tests/_craft/config/test.php'
            entryUrl: 'https://your-site-url.test/index.php'
            projectConfigOptOut: true
            migrations: []
            plugins: []
            cleanup: false
            transaction: false
            dbSetup: { clean: false, setupCraft: false, applyMigrations: false }
```

I've been working through merging my original test setup with the one
y'all provide now because... why wouldn't you? It's been great so far!

The only sticking point is how long the setup takes for each
individual test. I've narrowed the majority of the time spent to project
config being rebuilt between each test. The current setup makes 13 of
our tests take 7 seconds to run as opposed to 1.5 seconds (with opt out)

To further clarify, we're testing a site and custom modules rather than
a plugin. We currently have a setup script that makes a testable db for
us, but allows the developer to run that when it makes sense for them
(e.g. if their local test data gets out of whack), rather than at the
start of every test run.


gtettelaar:
Nice idea! You are right that it can get a bit slow so I also added some minor performance tweaks to reduce filesystem interaction. 

I cleaned up the config setting to make it apart of the `projectConfig` array you [specify as part](https://github.com/gtettelaar/docs/commit/14df6fb543aa8610ecfe08d0191f602878ae2231) of setup in `codeception.yml` and to add a `resetProjectConfig()` method allowing you to reset project config during tests manually.

@angrybrad thoughts?

dhahn:
Yeah this is great! Love the placement of the setting. The projectConfigOptOut definitely felt weird.

angrybrad:
All looks good to me!  We should probably add something in the docs for this here, no? https://github.com/craftcms/docs/blob/v3/testing/framework/config-options.md

gtettelaar:
@angrybrad Yup, just opened craftcms/docs#16 for that.

angrybrad:
Perfect... thanks, all!

PR #5091, opened 2019-10-12
fix for new persian translation

vnali:
@angrybrad , it seems your new commit -https://github.com/craftcms/cms/commit/4476680fc033606d542490ee5e1e1b6220086485 - was based on an old persian translation file -and some commits like https://github.com/craftcms/cms/pull/5058/commits/3a84044a4d12cd23a8b721ae840638e4b1ccaba wasn't applied in your source - and some translation for strings were reverted back to old version, so in this commit i updated them again. although in your new commit, some translations were good and i kept them.
thanks


brandonkelly:
Thanks!

PR #5096, opened 2019-10-14
adding periods to the commands

jasonmccallister:
This is rather nitpicky, but I was working on a plugin that focuses on console commands and noticed some inconsistencies with the use of periods on the command descriptions.


angrybrad:
Nitpicky is what we do.

PR #5100, opened 2019-10-15
Adds before and after events for GraphQL execution

jasonmccallister:
Adding events for before and after the execution of GraphQL execution. 

Spoke with @andris-sevcenko about this on Discord. These events can be used for caching the request from within another plugin.

### DIFF
--- a/src/controllers/GraphqlController.php
+++ b/src/controllers/GraphqlController.php
@@ -9,6 +9,7 @@ namespace craft\controllers;
 
 use Craft;
 use craft\errors\GqlException;
+use craft\events\GraphqlAfterExecuteEvent;
 use craft\events\GraphqlBeforeExecuteEvent;
 use craft\helpers\DateTimeHelper;
 use craft\helpers\StringHelper;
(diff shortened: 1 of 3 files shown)


jasonmccallister:
I'm not sure if this needs a PR for documentation?

andris-sevcenko:
@jasonmccallister looks like regardless of the plugin actions in the `before` event, the query is executed.

Also, it seems like it might be better to move some of GraphQL execution stuff out to the `Gql` service, so that plugins interacting with the service directly can take advantage of the caching benefits.

But the general idea is right along what I was expecting. If you can wait a day or two, I can finish this off for you.

jasonmccallister:
@andris-sevcenko that makes a lot of sense, so are you thinking something like this?

```
Craft::$app->getGql()->execute($args);
```

I'm also assuming moving the events const to the `Gql` service as well? Is there an example of a before event you can send me for reference?

I'm happy to wait, but also happy to make those changes as well.

jasonmccallister:
Thanks @andris-sevcenko!

PR #5103, opened 2019-10-15
Adds docker-compose.yaml for testing

jasonmccallister:
This PR adds a `docker-compose.yaml` and updates the `.env.example.mysql` and `.env.example.pgsql` for easier setup.

I selected the minimum versions of the databases that Craft requires. However, MySQL does not maintain a 5.5 image on Docker Hub so I went with 5.6.


jasonmccallister:
@gtettelaar I think this is failing because travis expects the user root?

gtettelaar:
@jasonmccallister Yup, on setup it copies the `example.env` file for the appropriate DB driver: 
https://github.com/craftcms/cms/blob/7fc874fac910b3ef47cd192c6993a6269a742de3/.travis.yml#L63

Not to worry though. Should probably be as easy as setting editing the build command to set a username and password for the DB rather than depending on Travis's default values. @angrybrad is much better at terminal commands than I'am so I'll leave it to him.

jasonmccallister:
@gtettelaar since these are example envs, I just changed it back to root.

angrybrad:
Since this is for "testing", would this be better off in the `tests` folder instead of root? 

We should probably provide some examples of how to use this on https://docs.craftcms.com/v3/testing/ ?

gtettelaar:
Not sure how this would be used as I don't use Docker myself but I would imagine this should also go into `src/test/internal/example-test-suite` (the directory used for the `tests/setup` command) if we are going down that route right?

jasonmccallister:
Hmm, I was thinking more along the lines of running tests locally. This would remove the need to keep your database installed and maintained locally - as well as testing different versions.

Happy to make a PR to the docs if needed.

angrybrad:
@jasonmccallister Right... if it's meant for testing only, should probably live in the `tests` folder?

jasonmccallister:
@angrybrad moved the file into the test directory. I'm updating the docs and will have a PR soon.

angrybrad:
Could you add it to `src/test/internal/example-test-suite` as well?

jasonmccallister:
@angrybrad done. Although, if this is for testing - do we want to keep the volumes persistent? Maybe we should consider dropping the volumes after a run?

angrybrad:
@jasonmccallister what would the advantage be? Craft/Codeception would already reset the data to a known state between runs.

jasonmccallister:
@angrybrad I'm going to close this out and reopen with a new PR

Issue #5075, opened 2019-10-10
[FR] Preserving Matrix Fields Container Display Value on toggle


svale:
### Description

When expanding a collapsed Matrix block, the blocks fields container is given a style attribute with the value of `opacity:1; display:block` through the `fadeIn`-method of velocity.js.

As opposed to jQuerys `.show()`, velocity.js .`fadeIn()`does not restore the display property to whatever it was initially. Instead it is set to the element's native type. This breaks any custom CSS (like `.fields { display:grid } `) added on top of the default CP-styles.

The display property can however be overridden by passing a custom display value into the `fadeIn()` options object.  

This means a small addition to [MatrixInput.js ](https://github.com/craftcms/cms/blob/2cf33e655bb22866c861a15f1c8f5c79145c6732/src/web/assets/matrix/dist/MatrixInput.js#L545
) allows for more custom styling (directly or through plugins) of Matrix fields. E.g. and specially using CSS grid layouts.

```
var displayValue = this.$fieldsContainer.css('display') || 'block';
this.$fieldsContainer.hide().velocity('fadeIn', {duration: 'fast', display: displayValue});
```

Would this be a possibility and/or a PR be considered?

Thanks!

### Additional info

- Craft version: 3.3.1.2
- Plugins & versions: relevant in the case: doublesecretagency/craft-cpcs

brandonkelly:
We’d take a PR for this, sure.

PR #5108, opened 2019-10-16
Add Matrix block fields container display property to velocity fadeIn options object

svale:
As detailed in FR #5075, this PR adds the existing Matrix block fields container display style value (block, flex, grid, table etc) to the velocity() fade command options object so that it is preserved when a Matrix block is collapsed and expanded.


brandonkelly:
Thanks!

PR #5118, opened 2019-10-18
Fixed user setting for profileTemplates not saving

Mosnar:
The profileTemplates setting currently does not save, as it was left out of the controller action


brandonkelly:
Doh! Thanks so much.

PR #5133, opened 2019-10-21
Prevent stringy from updating past 5.1

Mosnar:
I was doing some local development on Craft and apparently when you symlink a package, `composer install` doesn't respect the composer-lock file in the repo. This resulted in Stringy 5.2.x being installed, which has breaking changes (removal of getLangSpecificCharsArray).

This change prevents Craft from updating to 5.2.0. (Note: The composer.lock file is locked at 5.1.1, which is currently the last 5.1 release of stringy.)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,7 @@
     "craftcms/plugin-installer": "~1.5.3",
     "craftcms/server-check": "~1.1.0",
     "creocoder/yii2-nested-sets": "~0.9.0",
-    "voku/stringy": "^5.1.0",
+    "voku/stringy": "5.1.*",
     "elvanto/litemoji": "^1.3.1",
     "enshrined/svg-sanitize": "~0.9.0",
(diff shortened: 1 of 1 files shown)


brandonkelly:
Those syntaxes do the same thing. `~5.1.0` means it can update to 5.1.∞, but never to 5.2.x. So whatever led to the symlinked Stringy getting loaded, changing it to `5.1.*` wouldn’t have helped.

Mosnar:
@brandonkelly The constraint isn't currently set to `~5.1.0` though, it's set to `^5.1.0`

brandonkelly:
It has used a tilde since 3.3.1.1 (417fbbc13a79aa3ae17375a28187c257fe80d5b9). Maybe you’re just a bit behind?

Mosnar:
D'oh! You're right. I'm looking at the 3.3 branch.

Sorry about that!

PR #5134, opened 2019-10-21
Stream asset download in stead of using local copy

boboldehampsink:


andris-sevcenko:
That certainly makes sense to me.

andris-sevcenko:
Thanks!

boboldehampsink:
You're welcome. You might want to double-check if you can apply it in more places. Wasn't sure but this one.

PR #5136, opened 2019-10-21
Improved asset previews

### LINKED ISSUE #3402: Add ability to extend asset previews
### Description

Currently, the "Preview Asset" button is quite limited. At the time of writing this, it appears to only support some hard coded asset kinds:
```php
[self::KIND_IMAGE, self::KIND_HTML, self::KIND_JAVASCRIPT, self::KIND_JSON]
```

Adding some events to register custom asset preview handlers would go a long way to opening up doors for developers. I started investigating this after a client requested the ability to preview PDFs in the CMS. I can easily visualize other use-cases with things like 3D objects, video files, CAD drawings, Unity files, etc.

Mosnar:
This was more of a fun experiment that I was working on this weekend, but I thought I'd submit a draft PR since it's functional and clean and is _nearly_ production ready if desired.

I've implemented the following changes:
- Moved Asset previewability out of the Asset Element and into the Assets service, deprecating `getSupportsPreview`
- Moved preview generation logic out of the `AssetsController`
- Added concept of an `AssetPreview` via an abstract class, trait, and interface, which is responsible for returning modal HTML given an Asset
- Removed inline HTML for previews and replaced with Twig templates and corresponding AssetPreview concrete classes
- Added event for overriding and supplying AssetPreviews via `Assets::EVENT_GET_ASSET_PREVIEW`
- Allow AssetPreviews to register their own head and footer Javascript, allowing us to remove preview-specific logic from `PreviewFileModal.js` and into the appropriate Twig template.

image

I won't be offended if this isn't the direction you want to go, but hopefully it at least gives you ideas, as I think this is code that needs to be refactored at some point.

(Closes #3402)

### DIFF
--- a/src/assetpreviews/HtmlPreview.php
+++ b/src/assetpreviews/HtmlPreview.php
@@ -49,7 +49,8 @@ class HtmlPreview extends AssetPreview
     /**
      * @inheritDoc
      */
-    public function getModalHtml(): string {
+    public function getModalHtml(): string
+    {
         return $this->template;
     }
 
--- a/src/assetpreviews/HtmlPreview.php
+++ b/src/assetpreviews/HtmlPreview.php
@@ -1,0 +1,60 @@
+<?php
+/**
+ * @link https://craftcms.com/
+ * @copyright Copyright (c) Pixel & Tonic, Inc.
+ * @license https://craftcms.github.io/license/
+ * @since 3.4.0
+ */
+
+namespace craft\assetpreviews;
+
+use Craft;
+use craft\base\AssetPreview;
+use craft\elements\Asset;
+
+/**
+ * Provides functionality to preview text files as HTML
+ *
+ * @author Pixel & Tonic, Inc. <support@pixelandtonic.com>
+ * @since 3.3.0
+ */
+class HtmlPreview extends AssetPreview
+{
+
+    private $template = '';
+    private $foot = '';
+
+    // Public Methods
+    // =========================================================================
+
+    public function init()
+    {
+        $localCopy = $this->asset->getCopyOfFile();
+        $content = htmlspecialchars(file_get_contents($localCopy));
+        unlink($localCopy);
+        $language = $this->asset->kind === Asset::KIND_HTML ? 'markup' : $this->asset->kind;
+        $view = Craft::$app->getView();
+        $this->template = $view->renderTemplate('assets/_previews/html', [
+            'asset' => $this->asset,
+            'language' => $language,
+            'content' => $content
+        ]);
+
+        $this->foot = $view->getBodyHtml();
+
+        parent::init();
+    }
+
+
+    /**
+     * @inheritDoc
+     */
+    public function getModalHtml(): string {
+        return $this->template;
+    }
+
+    public function getFootHtml()
+    {
+        return $this->foot;
+    }
+}
--- a/src/assetpreviews/HtmlPreview.php
+++ b/src/assetpreviews/HtmlPreview.php
@@ -16,7 +16,7 @@ use craft\elements\Asset;
  * Provides functionality to preview text files as HTML
  *
  * @author Pixel & Tonic, Inc. <support@pixelandtonic.com>
- * @since 3.3.0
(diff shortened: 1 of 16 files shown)


src/assetpreviews/HtmlPreview.php
```diff
@@ -0,0 +1,60 @@
...
+        $this->foot = $view->getBodyHtml();
+
+        parent::init();
+    }
+
+
+    /**
+     * @inheritDoc
+     */
+    public function getModalHtml(): string {
```
andris-sevcenko: Curly brace on the wrong line.

src/assetpreviews/ImagePreview.php
```diff
@@ -0,0 +1,45 @@
...
+class ImagePreview extends AssetPreview
+{
+
+    // Public Methods
+    // =========================================================================
+
+    /**
+     * @inheritDoc
+     */
+    public function getModalHtml(): string {
```
andris-sevcenko: Curly brace on the wrong line.

src/assetpreviews/ImagePreview.php
```diff
@@ -0,0 +1,45 @@
...
+
+        $volume = $this->asset->getVolume();
+        if ($volume->hasUrls) {
```
andris-sevcenko: No newline before `if`

src/assetpreviews/NoPreview.php
```diff
@@ -0,0 +1,34 @@
...
+{
+
+    // Public Methods
+    // =========================================================================
+
+
+    /**
+     * @inheritDoc
+     */
+    public function getModalHtml(): string {
```
andris-sevcenko: Curly brace on the wrong line.

src/assetpreviews/PdfPreview.php
```diff
@@ -0,0 +1,44 @@
...
+{
+
+    // Public Methods
+    // =========================================================================
+
+
+    /**
+     * @inheritDoc
+     */
+    public function getModalHtml(): string {
```
andris-sevcenko: Curly brace on the wrong line.

src/assetpreviews/PdfPreview.php
```diff
@@ -0,0 +1,44 @@
...
+
+        $volume = $this->asset->getVolume();
+        if ($volume->hasUrls) {
```
andris-sevcenko: No newline before `if`

andris-sevcenko:
This is so much better than the "I wonder if we can make this happen two days before release" preview we had going on :)

Mosnar:
@andris-sevcenko Thanks for the feedback, I'm glad you like it! I've made the code style updates you requested as a few other doc tweaks.

A few thoughts I had about this approach that might be worth considering:
- In an effort to make this as light-weight as possible, I elected to avoid a "register components" style event in favor of a "before preview" event. The downside is we can't easily get a list of all preview handlers. This could be an issue if you ever wanted to make some sort of "preview handler" selection setting or, in my case, I was hoping to gray out the "Preview Asset" element action if there wasn't a preview handler for it. There are of course work-arounds for this, such as including the preview handler class on the element in the index as a data-attribute, but then you're firing a lot of events every time you load the element index.
- I'm not super happy about of the Javascript part in the preview-handler templates, especially since I had to access some private methods on the modal class to trigger a resize. Probably a good idea to have someone with more Garnish experience look at this.
- The PDF preview is super rudimentary, and might have some weird behavior if a PDF has a public URL that has a `Content-Disposition: Attachment` header, though I haven't tried it yet.

andris-sevcenko:
There are drawbacks to most things. This is just such an improvement over the existing thing, that we can think about them later. To address them, though:

* Currently Craft also allows you to preview unsupported files, only to tell you "Preview not available"
* Again, this is what the current preview did, too. This will become less of an eyesore, as Craft moves away from Garnish in later versions.
* Once we have people complaining about it with specific steps to reproduce, we can think about dealing with that :)

andris-sevcenko:
Think that changelog covers it. Thank you so much!

PR #5137, opened 2019-10-21
[testing] Fixes a functional test re-bootstrapping bug

gtettelaar:
Plugins weren't being loaded with their specific configuration file on request reset causing obvious problems.


angrybrad:
Thanks!

PR #5138, opened 2019-10-21
fix: return 200 instead of 302 if ajax request

oddnavy:
### Description

In our headless setup we POST to the actionSetPassword from our frontend app. This is working as intended except that on success, Craft returns a 302 response as per:

https://github.com/craftcms/cms/blob/edbcd4a67755ef367b86be2217647371110c1327/src/controllers/UsersController.php#L471

It could be better to check if the request accepts JSON and return a 200 with success = true in these instances


brandonkelly:
Thanks!

Issue #4642, opened 2019-07-23
Draft "Update Entry" opaque behavior in multi site setup (Could lead to unexpected content changes) 


christianruhstaller:
### Description
If you have a multisite Craft installation and you want to work with drafts the behavior when updating an entry from a draft is not what a normal user would expect and could lead to changed content on the entry on a different site.

If you create a draft for an entry this draft will be created for all the sites (If configured) and this is expected. Now you want to use this draft only for the english site. This draft is scheduled to be published in 3 days. In the meantime you fix an error  on the german site of the same entry (not via the draft) und save the entry. 

After 3 days you will publish the draft from the english site. The english sites draft is published and works perfectly. But what you do not know is that your fix on the german site was overwritten by the german sites draft entry which was not updated by the fix you made on the german entry which is in my opinion a valid and practical workflow.

### Steps to reproduce

1. Create an entry on a multisite instance (english + german)
2. Create a draft on the english site (Will create also a draft for the german site)
3. Edit entry for the german site
4. Publish draft on english site
 
#### Observed behavior
After you publish the draft your changes on the german sites are overwritten by the state of the german draft

#### Expected behavior
You only want to publish the draft for the english site and not the german site. 
Actually it could be quite confusing that a draft will get automatically created for another site.

A typical use case could be a promo action only for the us site and not the german site. 
I think this "feature" needs some refinement. 

### Additional info

- Craft version: 3.2.5.1

rhoffmann:
I would just like to add more priority to this. 
Currently we have an issue regarding multiple languages with drafts. When an editor creates a draft in one language (which automatically happens when you open the preview), and then applies that draft, all the other Languages/Sites are *overwritten* with the content of that language. Is that an issue that is specific to our setup in the structure? That is really a major issue for us (our customers) and i cannot really see how this can be called working as expected.

ctmaloney:
There should be a config option to go between the current functionality (versions and drafts and propagate across sites) and the functionality that sparked the change here: https://github.com/craftcms/cms/issues/2669

ctmaloney:
I'll also add that switching section settings back to "Only save entries to the site they were created in" shouldn't automatically delete all entries in the other sites as that creates a whole different set of issues.

mdominguez:
>Unfortunately, the way "Drafts" has been designed for Craft 3.2 breaks the basic translation workflow model for any content managers who want to update and manage their Target Entries and Sites on an individual basis.

[...]

Issue #4149, opened 2019-04-16
[FR] Ability to see dirty attributes programmatically before element save


Mosnar:
### Description
Currently, it doesn't look like there's a way to see what attributes are updated on elements without doing an additional query before an element is saved. Ideally, there would be a layer of abstraction on elements to help keep track of dirty fields and basic attributes, such as slugs and content fields. This functionality could mirror Yii's ActiveRecord:

`public array getDirtyAttributes ( $names = null )`
`public mixed getOldAttribute ( $name )`

Due to the complexity of elements and their usage of sister tables, I imagine this may need to be an "opt-in" based functionality in certain cases; particularly because certain elements query their sister tables directly, bypassing ActiveRecord.

The benefits of this would be pretty substantial:
- Rebuild sitemaps/create 301 redirects only if slug or title changed
- Programmatically clear coldcaches or cacheflags on heavy templates only when a specific attribute is updated
- Paves the way for keeping track of element histories efficiently

### Additional info

- Craft version: 3.1.22

brandonkelly:
Added for Craft 3.4 :)

Mosnar:
@brandonkelly Not exactly related to this issue, but I noticed the delta updates don't currently work on the draft editor for live preview. Is that something that's more trouble than it's worth or is it still on your radar? I imagine it could mean a pretty big bump in live-preview speed, especially for larger entries.

brandonkelly:
Definitely part of the plan :) It will be a prerequisite for #4642 which is next up.

brandonkelly:
@Mosnar Drafts support delta updates now (2e1c70a6d16b38c7f1ae0b84548a7c866646e4c3).

PR #5146, opened 2019-10-23
Delta updates

### LINKED ISSUE #4064: Saving entries in a multi-site setup takes too long
### Our Setup
We have a multi-site setup with 24 sites. Each site represents a country specific language. We use a variety of translation methods for each of our entries' fields, often utilizing site groups as common languages (17 site groups).

### Issue
As we've added more sites (more to come), we've noticed a drastic slowdown when saving entries and when re-saving entries tasks are triggered.

Re-saving an entry in a `channel` section takes almost 6 seconds.

In comparison, a `single` entry that is only enabled for our US English site takes about 1/2 a second to save.

This makes it difficult to deploy field changes that trigger re-saving entries tasks, as they take far too long. This also makes the CP feel sluggish for content authors.

### Help!

We're afraid that this will only get worse as more sites are added.

Is this expected behavior for multi-sites in Craft?
Is there something we could do to improve performance?

### Additional info

- Craft version: 3.1.20.1
- PHP version: 7.1.25

brandonkelly:
Will resolve #4064, #4149, and pave the way for a proper fix for #4642.

- [x] Elements should keep track of dirty field values.
- [x] Only update content for dirty fields
- [x] Track deltas for Edit Entry forms
- [x] Track deltas for Edit User forms
- [x] Track deltas for Edit Category forms
- [x] Track deltas for Element Editor HUDs
- [x] Matrix field support
  - [x] Only process field if dirty
  - [x] Merge delta blocks with existing blocks
  - [x] Only save dirty blocks
  - [x] Update sort order separately
- [x] Relational field support
- [x] Only update search indexes for dirty fields


Mosnar:
I just wanted to say that this is awesome and it's working great in my tests!

Gonna casually drop some ideas...
- Could start tracking fields accessed on elements in `templatecacheelements` and invalidating template caches more precisely based on the fields used in them. I couldn't help myself and actually developed a rough implementation of this and it was magical, though there are some table growth implications to consider.
- Could automatically set revision notes to include a list of updated fields if left unspecified
- Could have a "hard-save" keyboard short-cut, such as <kbd>CMD</kbd><kbd>SHIFT</kbd><kbd>S</kbd> that would save all fields to appease the inevitable complaints of not re-saving all fields.
- Could start updating the search index on a per-field-basis as a further optimization

brandonkelly:
@Mosnar Great ideas!

> Could start tracking fields accessed on elements in `templatecacheelements` and invalidating template caches more precisely based on the fields used in them. I couldn't help myself and actually developed a rough implementation of this and it was magical, though there are some table growth implications to consider.

We’d also have to start tracking other element properties for this to work effectively. Otherwise let’s say you have a cached template that outputs an entry’s author, and then edit the entry to change the author (without modifying custom field values), the cache wouldn’t get cleared. We could track whether _any_ element properties besides their custom fields were accessed, but at that point there’s not much to gain here.

That said, we’re probably going to swap out the entire template caching system with something else in Craft 4, so I’ll keep this in mind for that.

> Could automatically set revision notes to include a list of updated fields if left unspecified

The plan to fix #4642 is going to involve keeping track of custom field changes, and we’ll likely find a way to show what’s changed within the UI, so this should be unnecessary. 

> Could have a "hard-save" keyboard short-cut, such as <kbd>CMD</kbd><kbd>SHIFT</kbd><kbd>S</kbd> that would save all fields to appease the inevitable complaints of not re-saving all fields.

Yeah it will probably be needed from time to time. I think I’ll make it possible from `resave/*` commands, and see if that is enough.

> Could start updating the search index on a per-field-basis as a further optimization

Yeah we should definitely do that.

Issue #2566, opened 2018-03-13
@webroot doesn't return '/web' on console


ostark:
Simply setting the @webroot alias to `dirname($this->getScriptFile()))` doesn't make much sense as the cli is not executed in `/web`. But the value of the alias should be consistent, otherwise we can not rely on that.

As guessing or detecting the actual webroot from the console is not easy, a configuration option might be useful. This way we can overwrite it, if we don't use the default `web` folder.

### Additional info

- Craft version: 3.0.0-RC15

TUSAB-web:
Another common name for a webroot would be wwwroot/ if you wanted to add it as something craft checks for.

PR #5179, opened 2019-10-28
Use CRAFT_BASE_PATH as dir for @webroot creation

leevigraham:
We moved our `craft` executable to `bin/craft` which meant the `@webroot` alias could not be automatically determined when running a console command becuase `$dir` was the `bin` directory.

This PR replaces `$this->getScriptFile()` with the `CRAFT_BASE_PATH` constant which in our case is `define('CRAFT_BASE_PATH', __DIR__.'/..');`

### DIFF
--- a/src/console/Request.php
+++ b/src/console/Request.php
@@ -35,7 +35,7 @@ class Request extends \yii\console\Request
         // Set the @webroot and @web aliases, in case they are needed
         if (Craft::getRootAlias('@webroot') === false) {
             // see if it's any of the usual suspects
-            $dir = dirname($this->getScriptFile());
+            $dir = realpath(CRAFT_BASE_PATH);
             foreach (['web', 'public', 'public_html', 'html'] as $folder) {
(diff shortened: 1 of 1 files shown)


leevigraham:
Related:

* https://github.com/craftcms/cms/issues/2566
* https://github.com/putyourlightson/craft-blitz/issues/22

brandonkelly:
This would break the `@webroot` alias for installs where the web root isn’t in the `CRAFT_BASE_PATH` directory, such as:

```
my-project/
├── craft/   <-- base path
├── cli      <-- executable
└── web/     <-- web root
```

Probably an edge case, but so is putting the executable in a `lib/` folder, so I don’t think the change would be worth the potential breaks.

You can work around it by defining the `@webroot` alias yourself from `config/general.php`.

```php
'aliases' => [
    '@webroot' => dirname(__DIR__) . '/web',
],
```

leevigraham:
>This would break the @webroot alias for installs where the web root isn’t in the CRAFT_BASE_PATH directory, such as:

Righto…  I see how that would be an issue.

PR #5196, opened 2019-11-01
Plugin Repository URL in new tab

JayBox325:
Updating the Repository URL for a plugin page in the plugin store to open in a new tab so I don't lose my project tab.


benjamindavid:
Thank you @JayBox325!

PR #5198, opened 2019-11-01
Adding new checks to old migrations scripts

tmulry:
I was running into some existing table errors during core craft upgrades.  The following tables were already included
in the Install.php script and thus were causing an error during the DB Migration Process:

  Table::DRAFTS
  Table::REVISIONS
  Table::GQLSCHEMAS
  Table::SEQUENCES
  {{%entryversionerrors}}
  {{%entrydrafterrors}}

Steps to reproduce:
You will need to trigger one of the affected migrations. this entails rolling back to a snapshot from a previous craft version or removing the respective entry from the craft migration table.
With your composer version pointed to the latest craft release, Remove the entire vendor directory and run a fresh composer install.
Go to the /admin install page and click 'Finish Up'.
You see migration run for awhile before breaking on the error "craft_xxxx" table already exists.

Why this should work:
It is a standard pattern in the migration scripts to run `dropTableIfExists` on a table before running `createTable`.  This will allow the migration process to continue for tables that are created as part of Install.php.

### DIFF
--- a/src/migrations/m181112_203955_sequences_table.php
+++ b/src/migrations/m181112_203955_sequences_table.php
@@ -15,6 +15,7 @@ class m181112_203955_sequences_table extends Migration
      */
     public function safeUp()
     {
+        $this->dropTableIfExists(Table::SEQUENCES);
         $this->createTable(Table::SEQUENCES, [
             'name' => $this->string()->notNull(),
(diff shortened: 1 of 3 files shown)


brandonkelly:
Good idea, thanks for the PR!

PR #5229, opened 2019-11-12
Updates return type of `$serializedFieldValues`

brianjhanson:
While working on a client site I noticed that `serializedFieldValues` was throwing a warning because I was treating it like an array but PHPStorm thought it could be a string. This PR just updates return type  to be `array|null` instead of `string|null`. 

I don't believe `serializedFieldValues` is ever a string (especially since the description says it's an array), but it's possible I'm missing something.


brandonkelly:
Thanks for pointing that out! It can actually *only* return an array (not `null`), so just pushed my own commit to fix it.

PR #5272, opened 2019-11-27
Fix bug where asset preview modals can be left orphaned

Mosnar:
Currently, if you preview an asset in Craft via the Element Action, the modal is initialized with the incorrect parameter for the ElementSelect instance. This results in two issues:
- Error thrown in CP when the modal is closed:
`Craft.js:17921 Uncaught TypeError: this.elementSelect.focusItem is not a function`
- The modal is never destroyed on close

The asset preview works as expected when previewing via the keyboard shortcut since the instance is created correctly in the AssetIndex (https://github.com/craftcms/cms/blob/develop/src/web/assets/cp/src/js/AssetIndex.js#L1039)


brandonkelly:
Thanks!

PR #5282, opened 2019-11-28
Fixes bug with token when params are empty

bencroker:
If a token is created without any params, for example:
```
Craft::$app->getTokens()->createToken('my/token/route');
```

and the token is later used, an `Unsupported operand types` error is generated because `$params` is `null`:
```
return [$route, $params + $this->getQueryParams()];
```

This PR ensures that `$params` is an array before applying the `+` operator.

### DIFF
--- a/src/web/Request.php
+++ b/src/web/Request.php
@@ -1017,6 +1017,9 @@ class Request extends \yii\web\Request
 
         list($route, $params) = $result;
 
+        // Ensure params is an array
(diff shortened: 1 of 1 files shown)


brandonkelly:
That change wouldn’t have completely fixed the issue, since if you only pass a string to `createToken()`, `$result` would have ended up being that string, not an array, in the first place.

Just pushed a full fix for this.

bencroker:
Perfect, thanks!

PR #5315, opened 2019-12-06
Respect slugWordSeparator for section URIs if configured

chrisrowe:
I guess `slugWordSeparator` should be `uriWordSeperator` too?


brandonkelly:
Good idea, thanks!

PR #5327, opened 2019-12-11
Improve _ensureSingleEntry exception message

nstCactus:
Make debugging easier by including the name of the section having validation errors in the exception message


brandonkelly:
Good call, thanks!

PR #5328, opened 2019-12-11
Fixed typo in variable (3.4)

janhenckens:
I tried renaming a site group on a 3.4 beta install and noticed that it crashed on this.


brandonkelly:
Thanks!

PR #5342, opened 2019-12-16
Remove import _self as it no longer works on 2.8.0

mblode:
`{% from _self import uploadLocationInput %}` no longer works in Craft 2.8.0 because of the changes to Twig:
https://symfony.com/blog/simpler-macros-in-twig-templates

This is one fix but there would be issues throughout the codebase.

### DIFF
--- a/src/templates/_components/fieldtypes/Assets/settings.html
+++ b/src/templates/_components/fieldtypes/Assets/settings.html
@@ -27,9 +27,6 @@
 	</table>
 {% endmacro %}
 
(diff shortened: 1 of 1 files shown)


brandonkelly:
Thanks! Importing from `_self` is still supported. Verified that all other places we are doing that are still working fine. So I suspect this is a Twig bug.

brandonkelly:
We just released Craft 2.8.0.1 with this.

PR #5367, opened 2019-12-20
Allow anonymous for session-info

### LINKED ISSUE #5355: Make it easier for headless apps to retrieve user/session/CSRF cookies
Suppose I have a headless Craft site, and I'm trying to render a simple widget to display the current user's username if logged-in, or a login link if logged out.

Ideally we wouldn't have to make an ajax request on every page to do this, as there is a username/identity cookie that is set on login.

However, these are encoded (with a key you wouldn't want exposed to client-side code), and http-only, so they can't be read with JS.

So, you're left with making an additional ajax request every render…only there is no "get current user" controller with anonymous access to get the information. Without writing you'r own, `UsersController::actionGetRemainingSessionTime` is as close as you can get, which can tell you if the user is logged in, but won't give you their username.

So…you can write your own anonymous controller that returns the username, fine.

You're still left with the trouble of CSRF tokens. Again, you don't want to have to keep retrieving these with AJAX requests, but you can't access the `CRAFT_CSRF_TOKEN` vis JS.

Any ideas how to make this less of a headache?

It seems like maybe a config option like `$encodeUserCookies = false` might makes sense?

This is could also be `true` by default when `$headlessMode = true`.

timkelty:
Fixes #5355

### DIFF
--- a/src/controllers/UsersController.php
+++ b/src/controllers/UsersController.php
@@ -98,6 +98,7 @@ class UsersController extends Controller
         'save-user' => self::ALLOW_ANONYMOUS_LIVE,
         'send-activation-email' => self::ALLOW_ANONYMOUS_LIVE | self::ALLOW_ANONYMOUS_OFFLINE,
         'send-password-reset-email' => self::ALLOW_ANONYMOUS_LIVE | self::ALLOW_ANONYMOUS_OFFLINE,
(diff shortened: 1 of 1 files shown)


brandonkelly:
sorry missed this, and already added via 53d6a614a015a0804b178516b804ea063cf8ce84

Issue #5366, opened 2019-12-20
StringHelper::containsMb4 runs out of memory


gglnx:
### Description

We got a `project.yaml` which transformed to JSON has a length of 2192129 (~ 2.2 MB). Our `memory_limit` is 256 MB. Craft CMS checks before saving the new project configuration into the MySQL database with `StringHelper::containsMb4`. This needs around 320 MB, so the process fails with a fatal error (`Fatal error: Allowed memory size of XXXX bytes exhausted`).

`containsMb4` uses `str_split` and `array_map` (to apply `ord`). Means the string will be converted into an array saved in memory. Here a profile with the text of the Bible (4258801 chars, ~ 4.1 MB):

```
Start: 400.28 kb
After File Read: 4.45 mb
After containsMb4 Start: 4.46 mb
After str_split: 390.43 mb
After array_map: 646.43 mb
After max: 646.43 mb
bool(false)
This process used 406 ms for its computations
It spent 181 ms in system calls
After containsMb4: 4.46 mb
Peak: 646.43 mb
```

I wrote a replacement using `for`:

```php
    public static function containsMb4(string $str): bool
    {
		$length = strlen($str);

		for ($i = 0; $i < $length; $i++) {
	        if (ord($str[$i]) >= 240) {
	            return true;
	        }
	    }

        return false;
    }
```

This uses less memory and is quicker:

```
Start: 400.3 kb
After File Read: 4.45 mb
After containsMb4Fixed Start: 4.46 mb
After strlen: 4.46 mb
After for: 4.46 mb
bool(false)
This process used 243 ms for its computations
It spent 4 ms in system calls
After containsMb4Fixed: 4.46 mb
Peak: 4.47 mb
```

### Steps to reproduce

1. Download the bible from: https://bereanbible.com/bsb.txt
2. Run the scripts: https://gist.github.com/gglnx/b95b237b3790cfe2ed5b90afa4540456

### Additional info

- Craft version: Craft Pro 3.3.18.3
- PHP version: 7.3.13
- Database driver & version: MySQL 5.5.5

brandonkelly:
Can you submit your changes as a pull request?

gglnx:
Done :)

brandonkelly:
3.3.19 is released now with your change!

PR #5370, opened 2019-12-21
Refactor StringHelper::containsMb4

gglnx:
Replace str_split/array_map with more efficient for-loop.

See #5366


brandonkelly:
Thanks!

PR #5403, opened 2020-01-07
Allow categories to be enabled / disabled per site

stevenvandemoortele:


brandonkelly:
Thanks for the PR! But it’s actually a bit more involved than this, and categories are sortof feature-locked right now, as the concept will be going away in Craft 4, in favor of more feature-robust entries that gain all the category features.

PR #5419, opened 2020-01-10
Skip project.yaml matrix logic on deleted block types when no longer a matrix

markhuot:
Inside the `handleChangedBlockType` you account for when a field is changed from a matrix field to another type of field. You currently do this by ignoring the matrix-based updates on fields that are no longer matrix fields. E.g., if a matrix field of "Authors" changes to a Users field then there's no need to update the block types of that field since Users fields don't have block types.

Existing code is here,

https://github.com/craftcms/cms/commit/921f128ead2fcf7fc202b34bafd4b65e0d997ba4?w=1#diff-fb553bd95b8117c0eadba13cfa01e15a

This PR adds the same check to `handleDeletedBlockType`. This comes in to play in the following scenario,

1. A matrix field is created called "Authors"
2. The matrix field has a block type called "Author" that contains several fields such as name, title, bio, etc…
2. The matrix field is retyped to a Users field
3. (currently there seems to be a bug where the `project.yaml` retypes the field correctly but retains the Author block type)
4. Manually remove the Author block type from `project.yaml`
5. Run `project-config/sync` and it'll yell at you because of the orphaned block type

This PR fixes 5 and allows `project-config/sync` to finish. There may be deeper work to be done with regard to orphaned block types in `project.yaml`.

Note: view this PR with `?w=1` in GitHub. The change is minimal but I re-indented inside the conditional so it's showing bigger than it really should be.

### DIFF
--- a/src/services/Matrix.php
+++ b/src/services/Matrix.php
@@ -482,35 +482,39 @@ class Matrix extends Component
 
             /** @var MatrixField $matrixField */
             $matrixField = $fieldsService->getFieldById($blockType->fieldId);
-            $contentService->contentTable = $matrixField->contentTable;
 
-            // Set the new fieldColumnPrefix
-            $originalFieldColumnPrefix = Craft::$app->getContent()->fieldColumnPrefix;
-            Craft::$app->getContent()->fieldColumnPrefix = 'field_' . $blockType->handle . '_';
+            // Ignore it, if the parent field is not a Matrix field.
+            if ($matrixField instanceof MatrixField) {
+                $contentService->contentTable = $matrixField->contentTable;
 
-            // Now delete the block type fields
-            foreach ($blockType->getFields() as $field) {
-                Craft::$app->getFields()->deleteField($field);
+                // Set the new fieldColumnPrefix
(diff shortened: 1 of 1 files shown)


andris-sevcenko:
> `?w=1` in GitHub

WHAT SORCERY IS THIS? OMG!

As far as (4) goes, yeah. We're still to figure out what to do with orphaned block types and whether to delete them or not.

andris-sevcenko:
Thank you!

markhuot:
> WHAT SORCERY IS THIS? OMG!

Screen Shot 2020-01-10 at 10 23 05 AM

Same as clicking this checkbox but I always forget that checkbox exists and just instinctively add `?w=1` to the end of all my PR URLs.

> As far as (4) goes, yeah. We're still to figure out what to do with orphaned block types and whether to delete them or not.

Yea, either delete the data and make half the world mad or retain the data and make the other half the world mad… I don't mind either way, as long as `project-config/sync` doesn't get hung up on it.

Thanks @andris-sevcenko!

proimage:
> Yea, either delete the data and make half the world mad or retain the data and make the other half the world mad… 

Why not present the data to the admin, explain the situation to them, and allow them to decide to delete the data? image

Issue #5422, opened 2020-01-10
Documentation for craft\web\Controller $allowAnonymous


abryrath:
### Description
craft\web\Controller no longer accepts `true` as a valid value for protected attribute $allowAnonymous. Inline documentation still shows `true` and `bool` as possible values.

I am submitting a PR to update the documentation.

### Steps to reproduce

1. N/A

### Additional info

- Craft version: 3.3.19
- PHP version: N/A
- Database driver & version: N/A
- Plugins & versions: N/A

brandonkelly:
`true`/`false` are still supported:

https://github.com/craftcms/cms/blob/49e143d4f866ce477b4a8b4c97f61a8d23fb3dc7/src/web/Controller.php#L75-L77

If it’s not working for you, check if your controller is overriding the `init()` method. If so, make sure that you are calling `parent::init()` within it.

PR #5423, opened 2020-01-10
`true` is no longer an accepted value for

abryrath:
#5422


brandonkelly:
See https://github.com/craftcms/cms/issues/5422#issuecomment-573125706.

PR #5445, opened 2020-01-15
Improved Sendmail mail adapter

bertoost:
Since the command behind the sendmail adapter is coming from Swiftmailer.
It should respect the php.ini sendmail_path setting, but it doesn't. 
And besides that, I think it would be nice to even overwrite that via the general config.

Since it's not possible for a plugin to hook in on this (plugin will be loaded to late)

Let me know your thoughts :-)

### DIFF
--- a/src/config/GeneralConfig.php
+++ b/src/config/GeneralConfig.php
@@ -932,6 +932,12 @@ class GeneralConfig extends BaseObject
     public $verifyEmailSuccessPath = '';
 
     /**
+     * @var string The sendmail command, used as alternative for the php.ini setting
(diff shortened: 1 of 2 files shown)


brandonkelly:
Good idea! But I think it makes more sense for this to be a setting on the Sendmail transport adapter, rather than a general config setting. I’ve just added that for the next 3.4 release. It can even be set to an environment variable, as it’s likely to change from environment to environment.

The new Sendmail Command setting

bertoost:
Nice @brandonkelly .. That's even better!

PR #5454, opened 2020-01-17
Add `tableId` attribute to AdminTable component

engram-design:
This is mostly to make admin tables compatible with [Cloner](https://github.com/verbb/cloner).

For example, take the section index HTML:

```
<div id="sections-vue-admin-table"></div>
```

This is completely replaced when the component is mounted to this element. I actually didn't think this replaced all attributes on the HTML element, but it does, resulting in:

```
<div class="vue-admin-table">
    <div class="toolbar" style="display: none;">
    <div class="flex">
    ...
```

This adds back the original container ID, so you'll get:
```
<div id="sections-vue-admin-table" class="vue-admin-table">
    <div class="toolbar" style="display: none;">
    <div class="flex">
    ...
```

From this, I can target the table and go from there!


brandonkelly:
Thanks!

PR #5490, opened 2020-01-23
Update or insert searchindex, instead of only inserting, which can fail.

ryanpcmcquen:
The current `searchindex` fails if there is a duplicate entry, it should be an [`upsert()`](https://www.yiiframework.com/doc/api/2.0/yii-db-querybuilder#upsert()-detail) so that the current record is updated instead of just failing.
```
SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '349095-slug-0-1' for key 'PRIMARY'
The SQL being executed was: INSERT INTO `searchindex` (`elementId`, `attribute`, `fieldId`, `siteId`, `keywords`) VALUES (349095, 'slug', 0, 1, ' coming soon pop icons bill nye ')
```

### DIFF
--- a/src/services/Search.php
+++ b/src/services/Search.php
@@ -396,7 +396,7 @@ SQL;
 
         // Insert/update the row in searchindex
         $db->createCommand()
-            ->insert(Table::SEARCHINDEX, $columns, false)
+            ->upsert(Table::SEARCHINDEX, $columns, false)
             ->execute();
     }
 


src/services/Search.php
```diff
@@ -396,7 +396,7 @@ private function _indexElementKeywords(int $elementId, string $attribute, string
 
         // Insert/update the row in searchindex
         $db->createCommand()
-            ->insert(Table::SEARCHINDEX, $columns, false)
+            ->upsert(Table::SEARCHINDEX, $columns, false)
```
ryanpcmcquen: @brandonkelly, we may want to pass `true` to `$updateColumns` instead of `false`, thoughts?
```
The column data (name => value) to be updated if they already exist. If true is passed, the column data will be updated to match the insert column data. If false is passed, no update will be performed if the column data already exists.
```

The newer data triggered by the save is likely more desirable than the older data.

brandonkelly:
The `Insert/update the row in searchindex` comment is a little misleading there, as by this point, any existing keyword indexes for the given element/attribute will have already been deleted:

https://github.com/craftcms/cms/blob/4d11acc31e4820a56fcc4f9f6b15e6936fe0f174/src/services/Search.php#L119-L125

Did you actually run into a SQL error due to that `insert()` call? If so, how did you get there?

ryanpcmcquen:
Yes, we get a ton of these errors all the time. It happens merely by resaving elements where the primary key has not changed.

brandonkelly:
Can you share the stack trace that is added to the logs (`storage/logs/web.log`) alongside this error?

brandonkelly:
Also, the Search service has been refactored a bit in Craft 3.4. Any chance you can try updating to that and see if it helps?

To update, change your `craftcms/cms` requirement in composer.json to:

```json
"require": {
  "craftcms/cms": "^3.4.0-RC3,
  "...": "..."
}
```

Then run `composer update`.

ryanpcmcquen:
@brandonkelly, I don't see anything useful in `web.log`, but `queue.log` has plenty of stuff, here are some examples:

```
./queue.log:2020-01-23 14:20:55 [-][140926][-][error][craft\queue\QueueLogBehavior::afterError]  [16707] Updating search indexes (attempt: 1) - Error (time: 0.014s): SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '178180-slug-0-1' for key 'PRIMARY'
./queue.log:2020-01-23 14:20:55 [-][140926][-][error][craft\queue\QueueLogBehavior::afterError]  [16729] Updating search indexes (attempt: 1) - Error (time: 0.028s): SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '178195-title-0-1' for key 'PRIMARY'
./queue.log:2020-01-23 14:20:56 [-][140926][-][error][craft\queue\QueueLogBehavior::afterError]  [16748] Updating search indexes (attempt: 1) - Error (time: 0.009s): SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '94606-slug-0-1' for key 'PRIMARY'
./queue.log:2020-01-23 14:20:56 [-][140926][-][error][craft\queue\QueueLogBehavior::afterError]  [16767] Updating search indexes (attempt: 1) - Error (time: 0.020s): SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '148629-title-0-1' for key 'PRIMARY'
./queue.log:2020-01-23 14:20:57 [-][140926][-][error][craft\queue\QueueLogBehavior::afterError]  [16814] Updating search indexes (attempt: 1) - Error (time: 0.033s): SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '148641-field-570-1' for key 'PRIMARY'
./queue.log:2020-01-23 14:20:57 [-][140926][-][error][craft\queue\QueueLogBehavior::afterError]  [16823] Updating search indexes (attempt: 1) - Error (time: 0.037s): SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '148644-field-570-1' for key 'PRIMARY'
```

brandonkelly:
What version of Craft have you bee running?

ryanpcmcquen:
3.3.20.1.

ryanpcmcquen:
>The Insert/update the row in searchindex comment is a little misleading there, as by this point, any existing keyword indexes for the given element/attribute will have already been deleted:

I think the cleanup may not be happening fast enough on our system, perhaps it is just a scale/race condition issue? We do have a _lot_ of entries.

Still, this fix seems reasonable and would alleviate the issue. Doesn't really seem necessary to insert a record that may already exist. Is there a downside to changing it?

brandonkelly:
> perhaps it is just a scale/race condition issue? We do have a _lot_ of entries.

Ah, perhaps. The search service does try to get a lock on the element for the duration of its work (deleting old values, inserting new ones).

https://github.com/craftcms/cms/blob/49e143d4f866ce477b4a8b4c97f61a8d23fb3dc7/src/services/Search.php#L109-L117

If you have a multi-environment setup and environments aren’t sharing the same `storage/runtime` directory, then it’s possible the locks just aren’t effective.

You can swap out the `mutex` component another mutex driver, for example [yii\mutex\MysqlMutex](https://www.yiiframework.com/doc/api/2.0/yii-mutex-mysqlmutex) or [yii\mutex\PgsqlMutex](https://www.yiiframework.com/doc/api/2.0/yii-mutex-pgsqlmutex), depending on which DB you’re using, or [yii\redis\Mutex](https://www.yiiframework.com/extension/yiisoft/yii2-redis/doc/api/2.0/yii-redis-mutex) if you are [using Redis](https://docs.craftcms.com/v3/config/app.html#redis-example).

In `config/app.php`, add this:

```php
return [
    'components' => [
        'mutex' => 'yii\mutex\MysqlMutex',
    ],
];
```

> Still, this fix seems reasonable and would alleviate the issue. Doesn't really seem necessary to insert a record that may already exist. Is there a downside to changing it?

`upsert()` can be very slow on servers with a lot of databases, so we try to avoid it whenever possible, unless it’s something that doesn’t happen very often. If it weren’t for that, I’d be more inclined to accept the PR right away. But first I’m hoping we can figure out why the current logic isn’t working as expected. Hopeful it may just be because of the mutex issue.

ryanpcmcquen:
> `upsert()` can be very slow on servers with a lot of databases, so we try to avoid it whenever possible, unless it’s something that doesn’t happen very often. If it weren’t for that, I’d be more inclined to accept the PR right away. But first I’m hoping we can figure out why the current logic isn’t working as expected. Hopeful it may just be because of the mutex issue.

Wouldn't `$updateColumns` being set to `false` keep it from actually doing the update? Seems the speed would be the same then but the `insert` would fail gracefully.

brandonkelly:
The performance is just in preparing for the query – Yii does a lot of internal schema analysis before executing it, which can take a while.

ryanpcmcquen:
Ah, we are testing the Mutex change now to see if it helps.

ryanpcmcquen:
@brandonkelly, unfortunately, we are still getting locks even using Redis as our Mutex handler.

src/services/Search.php
ryanpcmcquen: Thinking about it more I like sticking with false, I really just want it to fail gracefully, as the user experience for jobs that continually fail is quite bad (it slows Craft's dashboard down for the end user).

brandonkelly:
Sorry for the delay; last week was a little busy with the 3.4 launch.

Looking through the code, the only way I can see this issue happening (besides a bad mutex), is if something is calling the now-deprecated `indexElementFields()` method:

https://github.com/craftcms/cms/blob/ce710a2d9ac046efff78827f8d95d00b265842df/src/services/Search.php#L171-L181

Do you have any custom module/plugin code that may be calling that? Or maybe a 3rd party plugin?

I’ve just made a chance for the next 3.4 release so that the full stack trace will get logged when an exception is thrown from a queue job. If you can update to that, we should be able to get more insight into how the error is occurring. To do that, change your `craftcms/cms` requirement in `composer.json` to:

```json
"require": {
  "craftcms/cms": "dev-develop#15ab59b487b0e72a4bae7d51d92340e612e88ded as 3.4.2",
  "...": "..."
}
```

Then run `composer update`.

ryanpcmcquen:
I just did a search of our codebase, including the `vendor` directory, nothing we have is calling that. The only calls are inside of Craft's own `Search.php`.

Unfortunately we cannot easily upgrade to 3.4 as we were using the standard property assignment in a lot of places (not `setFieldValue`), so it will be a bit until we can try `3.4.x`.

brandonkelly:
Sorry, I thought you were already on 3.4 for some reason. What version are you currently running?

ryanpcmcquen:
3.3.20.1

brandonkelly:
Right, ok. Well, would it be possible for you to make the change in 15ab59b487b0e72a4bae7d51d92340e612e88ded on your install, and post the full stack trace the next time this error occurs?

brandonkelly:
> Thinking about it more I like sticking with `false`, I really just want it to fail gracefully, as the user experience for jobs that continually fail is quite bad (it slows Craft's dashboard down for the end user).

A big benefit of moving to 3.4 will be that lots of queued/failed jobs won’t impact the control panel performance anymore, by the way.

ryanpcmcquen:
Somewhat related, but the same change would also be nice on forced project config applications:
```
b420207cfe72:/var/www/html$ ./craft project-config/apply --force
Applying changes from your project config files ... error: SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '238-854' for key 'fieldlayoutfields_layoutId_fieldId_unq_idx'
The SQL being executed was: INSERT INTO `fieldlayoutfields` (`layoutId`, `tabId`, `fieldId`, `required`, `sortOrder`, `dateCreated`, `dateUpdated`, `uid`) VALUES (238, 9775, 854, 0, 0, '2020-11-25 22:14:44', '2020-11-25 22:14:44', 'fab36e6f-573b-4241-9846-9cff8aa184fd')
```

brandonkelly:
Going to close this because `upsert()`  has a performance cost, so it’s not a good miracle cure for every possible integrity constraint violation. If you can figure out steps to reliably reproduce these errors on a project, send the Composer files, DB backup, etc., files into support@craftcms.com and we can work out a proper fix.

PR #5509, opened 2020-01-28
S3AssetSourceType: Non-AWS Compatibility

antom:
- Add overridable default configuration for various S3 Asset Source attributes to allow compatibility with other S3-Compatible services (eg. DigitalOcean Spaces).
- Update/amend S3AssetSourceType to utilise new configuration.


antom:
I've recently had to migrate a number of legacy Craft 2 websites over from AWS to DigitalOcean, which included migrating the existing S3 buckets over to DigitalOcean Spaces.

As there's no way to create custom Asset Source Types via plugins, I've patched the code as per this pull request to allow some basic configuration to be set to help add support for S3-compatible services through the Craft configuration.

An example of overriding the defaults with the patch for compatibility with DigitalOcean Spaces is as follows:

```
// Save this in /config/s3assetsource.php
return array(
    'assetSourceName' => 'DigitalOcean Spaces',
    'host' => 'digitaloceanspaces.com',
    'region' => 'fra1',
    'endpoint' => '{region}.{host}',
    'endpointByLocation' => '{location}.cdn.{host}',
    'predefinedEndpoints' => array(),
    'urlPrefix' => 'https://{bucket}.{endpointByLocation}/',
    'putObjectForceContentLength' => 1,
);
```

Figured that this may come in use for anyone else in a similar position & updating to Craft 3 isn't an immediate option.

brandonkelly:
Hey @antom, thanks for the PR! However we are [dropping](https://craftcms.com/guides/supported-versions) active support for Craft 2 in two days, so it’s not a good idea for us to be making changes at this point.

antom:
As this hasn't been officially merged, I've made a [git patch](https://gist.github.com/antom/f9c64370b2393cb18e64f3b4d243ce4c) available for anyone that may need it :)

angrybrad:
Thanks, @antom!

narration-sd:
@antom - that link doesn't appear to be good; findable though, and this link appears to be good: https://gist.github.com/antom/8705076ffb1a662aa1138b0e09043d29

(just wanted to have a look at it, not using, but nice contribution!)

PR #5560, opened 2020-01-31
Adds details hook to assets edit template

markhuot:
Mirrors the functionality plugins have on entry edit templates.

Specifically, looking to extend the Sequential Edit plugin to assets now that they have laudable URLs in the CP just like entries!


brandonkelly:
Good call, thanks! Went ahead and added `cp.assets.edit`, `cp.assets.edit.settings`, and `cp.assets.edit.meta` hooks as well.

brandonkelly:
Craft 3.4.2 is out now with those new hooks.

Issue #5564, opened 2020-02-01
FR: Add video preview in CP


urbantrout:

urbantrout:
Solved with v3.4.3

PR #5565, opened 2020-02-01
Add video preview for assets

urbantrout:
This adds a simple preview for videos via HTML5 video player.

Related to #5564

### DIFF
--- a/src/assetpreviews/Video.php
+++ b/src/assetpreviews/Video.php
@@ -1,0 +1,29 @@
+<?php
+namespace craft\assetpreviews;
+
+use Craft;
+use craft\base\AssetPreviewHandler;
+
+class Video extends AssetPreviewHandler
+{
+    /**
+     * @inheritdoc
+     */
+    public function getPreviewHtml(): string
+    {
+        /** @var Volume $volume */
+        $volume = $this->asset->getVolume();
+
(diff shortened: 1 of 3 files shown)


src/assetpreviews/Video.php
```diff
@@ -0,0 +1,41 @@
...
+    public function getPreviewHtml(): string
+    {
+        /** @var Volume $volume */
+        $volume = $this->asset->getVolume();
+
+        if ($volume->hasUrls) {
+            $url = $this->asset->getUrl();
+        } else {
+            $source = $this->asset->getTransformSource();
+            $url = Craft::$app->getAssetManager()->getPublishedUrl($source, true);
```
brandonkelly: I don’t think we should be downloading video files just for previewing. Can you just throw a \yii\base\NotSupportedException` here instead? (I’ve updated the controller to be prepared for that possibility – 0319e770e6fa78f5140869a7aac955bcbd664398.)

urbantrout: That's a good idea!

urbantrout:
I added `* @since 3.4.x` since I didn't know whether this would be merged nor when it would be merged,

brandonkelly:
Merged in for 3.4.3 :)

PR #5715, opened 2020-02-25
Add `ElementIndexAvailableTableAttributesEvent` event

engram-design:
I know that event name is a bit of a mouthful, but wanted to prefix the event name so it relates to an element index event.

This adds an event to give plugins a chance to modify the available table attributes of any element. In my use-case, this is for [Events](https://verbb.io/craft-plugins/events), where one element type, a Purchased Ticket actually uses the custom fields of another element, the Ticket, so that they're actually linked. If you remove a field from Ticket element's field layout, it goes from the Purchased Ticket.

As-is, this function to generate the available table attributes assumes you want to get element attributes and custom fields from the same element. This is most commonly the case, but adding this function allows us to modify this behaviour.

If I could do this in the Purchased Ticket element, I would, but table functions are static, and are more for attributes, than this later step where it amalgamates attributes and custom fields.

### DIFF
--- a/src/events/ElementIndexAvailableTableAttributesEvent.php
+++ b/src/events/ElementIndexAvailableTableAttributesEvent.php
@@ -1,0 +1,34 @@
+<?php
+/**
+ * @link https://craftcms.com/
+ * @copyright Copyright (c) Pixel & Tonic, Inc.
+ * @license https://craftcms.github.io/license/
+ */
+
+namespace craft\events;
+
+use yii\base\Event;
+
+/**
+ * ElementIndexAvailableTableAttributesEvent event class.
+ *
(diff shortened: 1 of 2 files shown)


brandonkelly:
I think for now a better way to pull this off will be to just include the event custom fields in your ticket `defineTableAttributes()` method.

```php
protected static function defineTableAttributes(): array
{
    $attributes = [];
    // ...

    // Include event custom fields
    foreach (Craft::$app->elementIndexes->getAvailableTableFields(Event::class) as $field) {
        /** @var \craft\base\Field $field */
        $attributes['field:' . $field->id] = ['label' => Craft::t('site', $field->name)];
    }

    return $attributes;
}
```

engram-design:
Yeah, way to over-complicate things @engram-design thanks @brandonkelly

Issue #5656, opened 2020-02-14
Many-columned tables don't fit on Entry edit pages


rbaud:
### Description

Tables with many columns overflow, and flow under the sidebar, on the Entry edit page. The page and table/wrapper aren't horizontally-scrollable and with the overlap of the sidebar, it makes large tables unusable.

![image](https://user-images.githubusercontent.com/2589656/74552096-18151400-4f1a-11ea-90d0-d1d308b4bc19.png)

### Steps to reproduce

1. Add a table field with many columns
2. View it on an Entry edit page

### Additional info

- Craft version: 3.4.5
- PHP version: 7.2.24
- Database driver & version: MySQL 5.7.29
- Plugins & versions: None

rbaud:
Please ignore the CP Field Inspect cogs here, I took the picture above before testing on a fresh, plugin-free install.

jonleverrier:
@rbaud I think this has been fixed for the next release. See https://github.com/craftcms/cms/issues/5611#issuecomment-583840394

brandonkelly:
Looks like this is a slightly different but related bug. Will look into it…

brandonkelly:
Fixed for the next release.

brandonkelly:
Craft 3.4.6 is out now with this fix.

PR #5746, opened 2020-03-03
Added `white-space: nowrap` to thin table cells

bencroker:
When `.thin` table cells contain text with spaces, the text is wrapped and looks quite ugly. This PR fixes that with `white-space: nowrap`.

Before:
Screenshot 2020-03-03 at 08 39 19

After:
Screenshot 2020-03-03 at 08 39 49

### DIFF
--- a/src/web/assets/cp/dist/css/_main.scss
+++ b/src/web/assets/cp/dist/css/_main.scss
@@ -1525,6 +1525,7 @@ table {
   th.thin,
   td.thin {
     width: 0.01% !important;
+    white-space: nowrap;
   }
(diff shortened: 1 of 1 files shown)


brandonkelly:
Thanks, looks like this was introduced in 141a2d9935c620e8fe7fd093081ee155c0e71a6d to fix #5656, but I think your idea of limiting it to `.thin` cells is a good compromise.

brandonkelly:
3.4.10 is out now with this change.

bencroker:
Thanks!

Issue #5810, opened 2020-03-19
Add a new craft cli command which checks if craft is installed.


urbantrout:
This would make things easier for automated setups and custom scripts.

For example: Check if craft installed and only then run `./craft migrate/all`

urbantrout:
The PR allows us to create composer scripts like this one:

`"@php craft setup/is-installed && @php craft migrate/all || return 0"`

khalwat:
This would be really useful for CI and for Docker-ized setups

brandonkelly:
Good idea! Just added an `install/check` command for Craft 3.5.

PR #5811, opened 2020-03-19
add setup/is-installed command to craft cli

urbantrout:
Related to #5810


khalwat:
This would be really useful for CI and for Docker-ized setups

brandonkelly:
Thanks for the PR! `install/check` seemed like a better place for this than `setup/is-installed`, so will go ahead and close this.

PR #5814, opened 2020-03-19
Unset ID if null to avoid Postgres throwing an error.

bencroker:
This PR unsets the ID of all ActiveRecord classes if null to avoid Postgres throwing an error.

This will eliminate the need to do this manually in various places, including plugins:
https://github.com/craftcms/cms/blob/3.4.10.1/src/services/AssetIndexer.php#L415-L416


brandonkelly:
Thanks for the suggestion! Just implemented a similar change for Craft 3.5, but made it a bit more dynamic by looping over `static::primaryKey()` rather than assuming it's going to be `id`.

bencroker:
Perfect, thanks!

PR #5861, opened 2020-03-30
Make `siteId` more flexible in `getElementById()`

bencroker:
This PR makes the `siteId` parameter in the `getElementById` method of the `Elements` service more flexible. Rather than being limited to `int|null`, it is expanded it to `int|int[]|string|null` as in `ElementQuery::$siteId`:

```php
    /**
     * @var int|int[]|string|null The site ID(s) that the elements should be returned in, or `'*'` if elements
     * should be returned in all supported sites.
     * @used-by site()
     * @used-by siteId()
     */
    public $siteId;
```


brandonkelly:
Good call, thanks!

brandonkelly:
Just released Craft 3.4.12 with this change.

bencroker:
Excellent, thanks!

Issue #3631, opened 2019-01-15
Problems with multi-thread php for dotenv


smcyr:
### Description

When you have ajax calls or reload/load multiples pages rapidly in a multi-thread environment (for example, Windows Apache with mod_php), some values from .env don't load with getenv(). See these Github issues:
https://github.com/vlucas/phpdotenv/issues/203
https://github.com/vlucas/phpdotenv/issues/219
https://github.com/vlucas/phpdotenv/issues/248
https://github.com/vlucas/phpdotenv/issues/266
https://github.com/laravel/framework/issues/24510

To circumvent this issue, Laravel cache the values from dotenv in a compiled config file:
https://laravel.com/docs/5.7/configuration#configuration-caching

Craft CMS should use a similar solution to prevent these problems on many environments.

### Steps to reproduce

1. Make multiple ajax calls or load multiple pages in multiple tabs in a Craft CMS website.

### Additional info

- Craft version: 3.0.37
- Windows 10, Apache 2.4.37 with mod_php
- PHP version: PHP 7.2.14 thread safe x64

smcyr:
Yes, but in my case, the problem is in our work environment (it works fine in production). We all use Windows and I have Apache with multiples sites using dotenv. For now, I configured Apache to use mod_fcgid (cgi implementation) instead of mod_php to make it work in local. I had the problem for weeks but just figured it out,

See this comment from the creator of phpdotenv:
https://github.com/vlucas/phpdotenv/issues/219#issuecomment-450900648
> Hmm, I don't think this is necessary. I'd strongly recommend people use this package to one-time use this library to populate a compiled config file, exactly how Laravel does it. I'd consider that to be the correct way to use this library in a multi-threaded environment (that is, by not using it in the multi-threaded environment).

brandonkelly:
Good to know! I didn’t mean to give the impression we’re not interested; just offering a workaround for the meantime.

smcyr:
Thank you for answering so fast and your amazing work on Craft!

GrahamCampbell:
I think it's likely the following example will work in a multithreaded environment. This avoids using the adapter that would have called `putenv` and `getenv`, which are not threadsafe.

```php
<?php

use Dotenv\Environment\Adapter\EnvConstAdapter;
use Dotenv\Environment\Adapter\ServerConstAdapter;
use Dotenv\Environment\DotenvFactory;
use Dotenv\Dotenv;

[...]

PR #5893, opened 2020-04-06
Dotenv stability on multithread envs

TimvanDijkhuizen:
The getenv function is very unstable on a multithread env. This commit replaces all calls to getenv() with Craft::getEnv(), which grabs the variable from $_SERVER instead.

The behavior of Craft::getEnv() is pretty much the same as getenv(). The only difference is that varname is required for this function.


TimvanDijkhuizen:
This PR fixes https://github.com/craftcms/cms/issues/3631

TimvanDijkhuizen:
Before this change my install would break every few requests because of missing variables. After this change it loads without any problems.

TimvanDijkhuizen:
Update: Even though this commit seems to reduce how often this bug occurs it does not seem to have fixed it.

TimvanDijkhuizen:
> If you try changing how you load dotenv to this in a multithreaded environment [#3631 (comment)](https://github.com/craftcms/cms/issues/3631#issuecomment-457849474) does it work for you?

I've implemented the code suggested by that user which broke the getenv() function. After replacing all getenv() calls in my config and bootstrap files with $_SERVER[] everything seems to be working really well. All of my requests 1800+ test requests survived.

It looks like this PR combined with replacing the dotenv load code fixes this issue. All getenv() calls before the Craft bootstrap is loaded would have to be replaced by $_SYSTEM which is not very clean. So that is something to think about.

angrybrad:
> which broke the getenv() function

Can you elaborate on that?  What sort of error(s) were you getting?

TimvanDijkhuizen:
> > which broke the getenv() function
> 
> Can you elaborate on that? What sort of error(s) were you getting?

There weren't any errors but none of my variables were returning their value. I placed the following piece of code right after the code used to load dotenv:

```php
var_dump(getenv('SECURITY_KEY'));
exit();
```

Which returned the boolean value false. According to the PHP docs that means the variable doesn't exist.

angrybrad:
Got it... looks like by overriding the default adaptors (https://github.com/vlucas/phpdotenv/blob/3.6/src/Environment/DotenvFactory.php#L34), the putenvadaptor isn't loaded, so there's nothing for getenv to load.

brandonkelly:
New Craft projects aren’t even setting environment variables to the `$_SERVER` array anymore, as of last week (see craftcms/craft#55).

TimvanDijkhuizen:
@brandonkelly Would it be safe to use $_ENV instead? That constants doesn't seem to be getting logged.

I'm still trying to work something out as my Craft installations are pretty much unusable on Windows with WAMP. My env variables fail to load 1 out of 5 times and they never load for job and CLI requests. The project I'm working on uses the queue for important API requests and I am unable to work on those now.

Is there anything we can do about this?

brandonkelly:
That change would have affected both `$_ENV` and `$_SERVER`. I ended up reverting craftcms/craft#55 though, so we can revisit this. `$_ENV` does seem like a better place to check than `$_SERVER`.

I’d suggest you make these changes though:

- Rename `getEnv()` to `env()` and move it to `craft\helpers\App`.
- Change it to simply return:
    ```php
    return $_ENV[$name] ?? getenv($name);
    ```
    so `getenv()` is still used as a fallback in case environment variables didn’t make it into `$_ENV`.

brandonkelly:
Actually I’ll go ahead and push up those changes…

brandonkelly:
Craft 3.4.18 is out now with this change.

PR #5901, opened 2020-04-07
Fix `securityKey` description.

mattstein:
Per [recent discussion and clarification](https://discordapp.com/channels/456442477667418113/456442477667418115/697164185494356079), it’s safe to have the key differ across environments because it only impacts cookies and hashed form values.


mattstein:
...and I’m wondering if changing it from one value to another in the *same* environment will matter per the previous warning. Probably not?

brandonkelly:
It’s safe _for now_ but it’s still recommended that each environment has the same security key, and could be more important down the road than it is today.

PR #5925, opened 2020-04-15
Bump composer deps

leevigraham:
v 1.10.x allows Symonfy 5.0 components.


brandonkelly:
Whoops forgot to follow up here, but yes this is now resolved for Craft 3.5 :)

PR #5938, opened 2020-04-17
Use PHP's ErrorException instead of Yii's

boboldehampsink:
So we also catch native `unlink` exceptions

Should fix errors like these: https://sentry.io/share/issue/b76efc6f00934b6392d6344004fcfc96/


brandonkelly:
Thanks!

PR #5991, opened 2020-04-29
Prevent “node_modules” scan for template suggestions

billmn:
Currently the "template suggestion" feature scans the `node_modules` folder if exists inside the template.

This slows down considerably the loading of the page (e.g. Edit of a Section) inside the control panel.

With this PR we can prevent the scanning of unwanted folders (like `node_modules`) speeding up the loading of the pages that use this feature.

I was thinking of moving the `$excludes` variable to a configuration file but I'm not sure which one is correct. Thoughts?

### DIFF
--- a/src/web/twig/variables/Cp.php
+++ b/src/web/twig/variables/Cp.php
@@ -459,9 +459,13 @@ class Cp extends Component
         $files = [];
         $pathLengths = [];
 
+        $excludes = ['node_modules/'];
+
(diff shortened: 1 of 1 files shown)


brandonkelly:
Same idea was PR’d earlier by #4122. Just merged that one in, as it’s a bit more efficient.

brandonkelly:
Craft 3.4.18 is out now with #4122 merged in.

PR #6046, opened 2020-05-11
Disable caching of compiled templates when `devMode` is enabled

bencroker:
This PR disables caching of compiled templates when `devMode` is enabled, for a better local development experience. https://twig.symfony.com/doc/3.x/api.html#environment-options


brandonkelly:
This would drastically slow down pages as Twig would need to start recompiling all loaded templates for each request. Seeing a typical Edit Entry page go from ~600ms on average up to about ~900 when I make this change.

bencroker:
Fair enough.

PR #6089, opened 2020-05-15
Use systemName over siteName in emails

timkelty:
`systemName` might make more sense than `siteName` here, as users aren't site-specific.


brandonkelly:
Good call!

PR #6090, opened 2020-05-16
Move login delay to failure handler

johnnynotsolucky:
This PR moves the random delay applied to all login attempts to the `_handleLoginFailure` method. 

As brute-force mitigation, I think it should only really necessary to add a delay to failed attempts and any successful login shouldn't be met with a potential delay.


brandonkelly:
Good call, thanks!

brandonkelly:
Craft 3.4.20 is out now with this change.

PR #6096, opened 2020-05-18
make first draft name translatable

kringkaste:
The "First draft" name was not translatable, so we fixed that and added the German translation :)


brandonkelly:
Thanks!

brandonkelly:
Craft 3.4.20 is out now with this change.

PR #6101, opened 2020-05-19
Fix allowAnonymous checking for CP offline

### LINKED ISSUE #6100: Anonymous/Offline access isn't working
### Description
In cms/src/web/Controller.php, there is broken syntax which breaks anonymous access when the site status is offline

### Steps to reproduce

1. Set `$allowAnonymous` to be true for something
2. Set the site status offline
3. Try to access it

### Additional info

- Craft version: newest
- PHP version: 7.x
- Database driver & version: n/a
- Plugins & versions: n/a

paullegrand:
Only one `&` breaks using `$allowAnonymous` when the site status is toggled off, because the comparison fails.

Fixes #6100

### DIFF
--- a/src/web/Controller.php
+++ b/src/web/Controller.php
@@ -145,7 +145,7 @@ abstract class Controller extends \yii\web\Controller
             $allowAnonymous = $this->allowAnonymous[$action->id] ?? self::ALLOW_ANONYMOUS_NEVER;
         }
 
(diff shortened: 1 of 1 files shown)


brandonkelly:
Thanks for the PR but this is working as intended – see https://github.com/craftcms/cms/issues/6100#issuecomment-630565903.

PR #5929, opened 2020-04-16
Fixtures: generate element query with field:handle

boboldehampsink:

PR #6103, opened 2020-05-19
Fixtures: Use relations as primary key

boboldehampsink:
A relation field is always passed as an array (of id's or elements), never a scalar.

Had this in #5929 but reverted it, thinking it wasn't needed anymore. I was wrong.


angrybrad:
> - Fixed a bug in test fixtures where primary keys were not being detected on relational fields.

Sound about right, @boboldehampsink?

boboldehampsink:
That's correct @angrybrad !

angrybrad:
Thanks!

brandonkelly:
Craft 3.4.21 is out now with this change.

PR #6149, opened 2020-05-27
Throw UnauthorizedHttpException for guest users

johnnynotsolucky:
When a guest tries to access an action which doesn't allow anonymous access, and doesn't `requireLogin()` we get a `ServiceUnavailableException`.

To reproduce we used:

```php
class FooController extends \craft\web\Controller
{
    public function actionBar()
    {
        return $this->asJson(['foo' => 'bar']);
    }
}
```

The CraftCMS docs say:

> You can easily control whether the controller should allow anonymous access by overriding $allowAnonymous . (An active user session is required by default).

This PR aligns the error with the requirements in the docs.

### DIFF
--- a/src/web/Controller.php
+++ b/src/web/Controller.php
@@ -18,6 +18,7 @@ use yii\base\InvalidConfigException;
 use yii\base\UserException;
 use yii\web\BadRequestHttpException;
 use yii\web\ForbiddenHttpException;
+use yii\web\UnauthorizedHttpException;
 use yii\web\HttpException;
 use yii\web\JsonResponseFormatter;
 use yii\web\Response as YiiResponse;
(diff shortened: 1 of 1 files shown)


sjcallender:
If accepted, please consider adding this as a patch to version 3.4. 🙏

brandonkelly:
`UnauthorizedHttpException` gives you a 401 Unauthorized response, which is more about authentication than authorization, per https://stackoverflow.com/a/50143519/1688568. So switched that to a `ForbiddenHttpException`, which gives you a 403 Forbidden response. (And it will still throw a `ServiceUnavailableHttpException` (503) in the event that the system is offline.

brandonkelly:
Craft 3.4.21 is out now with that change.

PR #6166, opened 2020-05-30
Remove warning log when removing local image source

johnnynotsolucky:
This PR removes the warning log which is logged when attempting to unlink an image source which doesn't exist prior to retrieving a local image source.

I can replicate this by loading the assets in CP; waiting for the the thumbnails to load and searching my logs for the warning.


brandonkelly:
Good call, thanks!

brandonkelly:
Craft 3.4.23 is out now with this change.

PR #6169, opened 2020-06-01
Added wrapper to make button containers consistent

leevigraham:
This PR adds a button container with an ID to make it consistent with the `#save-btn-container` and provide a hook for plugins like https://github.com/spicywebau/craft-fieldlabels 

In fact this change may be required to fix https://github.com/spicywebau/craft-fieldlabels/issues/51


brandonkelly:
Thanks!

leevigraham:
Noice :)

PR #6185, opened 2020-06-05
Override `evaluateDynamicContent()` to prevent unintended code eval

michaelrog:
This short-circuits some potential security vulnerabilities at the source, by overriding `View::evaluateDynamicContent()` as a no-op, unless the environment explicitly permits dynamic evaluation via the `enableDynamicContentEvaluation` config setting.

(Prevents _unintended_ execution of untrusted code, but still keeps the API intact for the [brave souls](https://gist.github.com/elfacht/bc2c8a997c1801c45d29e4e374cd550a) who are utilizing that method already.)


khalwat:
Nicely done, sir.

brandonkelly:
Thanks for the PR! We decided this isn’t worth a config setting, so made it a public property on `craft\web\View` instead. So if you _really_ need to be able to call `evaluateDynamicContent()`, you can add this to your config/app.php file:

```php
return [
  'components' => [
    'view' => [
      // todo: delete this line asap!
      'allowEval' => true,
    ],
  ],
],
```

PR #6202, opened 2020-06-10
plugin-cards are now links and can be opened in tabs

leevigraham:


brandonkelly:
Thanks for the PR! We’ve resolved this with e623d2dd83a62d8e3120af96447d01e630af7969.

PR #6263, opened 2020-06-25
Register TestMailer on init instead of after init

boboldehampsink:
This registers Craft's TestMailer on init instead of after init (which can cause trouble in some cases).

### DIFF
--- a/src/test/CraftConnector.php
+++ b/src/test/CraftConnector.php
@@ -61,17 +61,18 @@ class CraftConnector extends Yii2
     }
 
     /**
-     *
+     * @inheritdoc
      */
-    public function startApp()
+    protected function mockMailer(array $config)
     {
-        parent::startApp();
-
-        Craft::$app->set('mailer', [
+        $config = parent::mockMailer($config);
+        $config['components']['mailer'] = array_merge($config['components']['mailer'], [
             'class' => TestMailer::class, 'callback' => function(MessageInterface $message) {
                 $this->emails[] = $message;
             }
         ]);
+
+        return $config;
     }
 
     /**


angrybrad:
> which can cause trouble in some cases

What are the specific cases?

The method signature and name change from startApp to mockMailer is a very breaking change, no?  startApp comes from the Yii2 codeception connector parent class.

boboldehampsink:
Let me try to explain:

Right now, parent::startApp starts Craft, and after initializing Craft, the child startApp overrides the email component. But, if you call getMailer() from a plugin's init() method, it fails with the following error:

```

Test  tests/functional/DepartmentCest.php:testDepartment
--
  |  
  | [yii\base\UnknownMethodException] Calling unknown method: Codeception\Lib\Connector\Yii2\TestMailer::getTransport()
  |  
  | #1  /app/user/vendor/yiisoft/yii2/base/Component.php:300
  | #2  /app/user/vendor/bertoost/craft-mailjet/src/adapters/MailjetAdapter.php:108
  | #3  /app/user/vendor/bertoost/craft-mailjet/src/traits/PluginEventsTrait.php:39
  | #4  /app/user/vendor/bertoost/craft-mailjet/src/Plugin.php:35
  | #5  /app/user/vendor/yiisoft/yii2/base/BaseObject.php:109
  | #6  /app/user/vendor/yiisoft/yii2/base/Module.php:158
  | #7  /app/user/vendor/craftcms/cms/src/base/Plugin.php:115
  | #8  craft\base\Plugin->__construct
  | #9  /app/user/vendor/yiisoft/yii2/di/Container.php:400
  | #10 /app/user/vendor/yiisoft/yii2/di/Container.php:159


```

As you can see, at that point it still uses Yii2's TestMailer instead of Craft's TestMailer. By moving the logic to mockMailer it gets registered when initializing the app straight away.

> The method signature and name change from startApp to mockMailer is a very breaking change, no? startApp comes from the Yii2 codeception connector parent class.

It's not breaking. mockMailer also comes from the Yii2 codeception connector parent class.

angrybrad:
Got it... thanks!  And here's the changelog for it: https://github.com/craftcms/cms/commit/643d871155d6a2d5c47d62d33d22a268c25872e3

brandonkelly:
Craft 3.4.26 is out now with this change.

PR #6330, opened 2020-07-06
Add docker-compose.yaml for tests

jasonmccallister:
Added a `docker-compose.yaml` in `tests` that contains databases for quick setup.

### DIFF
--- a/tests/.env.example.mysql
+++ b/tests/.env.example.mysql
@@ -1,8 +1,8 @@
 # Set in accordance to your environment
 
-DB_DSN="mysql:host=localhost;port=3306;dbname=craft-test"
-DB_USER="root"
-DB_PASSWORD=""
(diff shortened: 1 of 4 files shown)


tests/.env.example.mysql
```diff
@@ -1,6 +1,8 @@
 # Set in accordance to your environment
 
-DB_DSN="mysql:host=127.0.0.1;port=33060;dbname=craft-test"
+DB_DSN="mysql:host=localhost;port=3306;dbname=craft-test"
+# if you are using the docker-compose.yaml file use these
+# DB_DSN="mysql:host=127.0.0.1;port=33060;dbname=craft-test"
 DB_USER="craft"
```
angrybrad, line 6: Is the docker image using the same creds? If not, we can add them commented out as well. Probably just group all of the docker specific lines commented out at the bottom together.

Issue #6309, opened 2020-07-02
[Testing] Cannot successfully extend AssetFixture


davist11:
### Description

I'm working on adding an asset fixture to my test suite, and having a number of issues. Here is the asset fixture:

```php
<?php

namespace client\tests\fixtures;

use Craft;
use craft\test\fixtures\elements\AssetFixture;

class AssetsFixture extends AssetFixture
{
    public $dataFile = __DIR__ . '/data/assets.php';
}
```

Here is the data:

```php
<?php

return [
    [
        'tempFilePath' => CRAFT_STORAGE_PATH . '/runtime/temp/sample.png',
        'filename' => 'sample.png',
        'volumeId' => $this->volumeIds['parts'],
        'folderId' => 1,
    ],
];
```

The asset exists:

![image](https://user-images.githubusercontent.com/199985/86364046-ada10e00-bc45-11ea-97d3-7a0110ca52e2.png)

Here's where I'm loading the fixture:

```php
<?php

namespace client\tests;

use Codeception\Test\Unit;

use UnitTester;
use Craft;
use client\tests\fixtures\AssetsFixture;

class WebTest extends Unit
{
    /**
     * @var UnitTester
     */
    protected $tester;

    public function _fixtures(): array
    {
        return [
            'assets' => [
                'class' => AssetsFixture::class,
            ],
        ];
    }
}
```

When I run that, I get the following error:

```
[yii\base\ErrorException] copy(/var/www/html/vendor/craftcms/cms/tests/_craft/assets/sample.jpg): failed to open stream: No such file or directory  
```

When I dig into the AssetFixture class I'm extending, this makes sense because of how these properties are being set:

https://github.com/craftcms/cms/blob/74684923842c6667690e8a85f543f5e5a976adf8/src/test/fixtures/elements/AssetFixture.php#L76-L77

Am I doing something wrong? I ended up overriding the `init()` method in my fixture and got some progress:

```php
public function init()
{
    // parent::init();

    foreach (Craft::$app->getVolumes()->getAllVolumes() as $volume) {
        $this->volumeIds[$volume->handle] = $volume->id;
        $this->folderIds[$volume->handle] = VolumeFolder::findOne([
            'parentId' => null,
            'name' => $volume->name,
            'volumeId' => $volume->id,
        ])->id;
    }

    $this->sourceAssetPath = dirname(__FILE__, 3) . '/tests/_craft/assets/';
[...]

PR #6331, opened 2020-07-07
potentially fix asset fixtures for non-craft tests

angrybrad:
Potentially fixes https://github.com/craftcms/cms/issues/6309

### DIFF
--- a/CHANGELOG-v3.md
+++ b/CHANGELOG-v3.md
@@ -1,5 +1,10 @@
 # Release Notes for Craft CMS 3.x
 
+## Unreleased
+
+### Fixed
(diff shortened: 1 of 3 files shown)


boboldehampsink:
This has broken the use of the `craft fixture/load` command for me. Giving the following error:

```
PHP Warning 'yii\base\ErrorException' with message 'Use of undefined constant CRAFT_TESTS_PATH - assumed 'CRAFT_TESTS_PATH' (this will throw an Error in a future version of PHP)'
```

This is because the `CRAFT_TESTS_PATH` constant is only defined in `tests/_bootstrap.php`, while `craft fixture/load` doesn't use that file at all.

angrybrad:
@boboldehampsink this should take care of both issues https://github.com/craftcms/cms/commit/2bbe993bbad4ab7472aa68086699a530307deac9

brandonkelly:
Craft 3.4.29 is out now with this change.

PR #6353, opened 2020-07-09
add support for inlineOnly to markdown GraphQL directive

oddnavy:
### Description
Add inlineOnly argument to match twig filter implementation:

https://github.com/craftcms/cms/blob/cd07fdb05d12b4594fc8c2ce836b9fa083836ae6/src/web/twig/Extension.php#L836


andris-sevcenko:
Thank you!

brandonkelly:
Craft 3.4.28 is out now with this change.

PR #6361, opened 2020-07-12
Adds static translations for Heading and Tip field layout elements

mmikkel:
### Description

Adds missing static translation (using the `'site'` translation category) for Heading and Tip field layout elements' values in the new field layout designer.


brandonkelly:
Doh, thanks!

PR #6367, opened 2020-07-13
Eager-loading fails in queries spanning multiple sites

### LINKED ISSUE #6366: Eager-loading fails in queries spanning multiple sites
### Description

My entries have an asset field named `teaserImage`. When I query with multiple site IDs (`.siteId([1, 2])`), the eager-loading fails and `teaserImage` is null.

It's looks like that the query build by `BaseRelationField::getEagerLoadingMap` is broken. Instead of 'AND' between the IDs of one site there should be an 'OR'.

### Steps to reproduce

1. Create a BaseRelationField field like an Asset field, attach it to an entry type
2. Create two posts in two sites with with an attached asset
3. Query with eager-loading for both sites, the assets field will be empty

### Additional info

- Craft version: 3.4.24
- PHP version: 7.4.6
- Plugins & versions: None

gglnx:
Fixes #6366

### DIFF
--- a/src/fields/BaseRelationField.php
+++ b/src/fields/BaseRelationField.php
@@ -647,8 +647,7 @@ JS;
         }
 
         $condition = [
-            'and',
-            ['fieldId' => $this->id],
+            'or',
         ];
 
         foreach ($sourceElementIdsBySiteId as $siteId => $elementIds) {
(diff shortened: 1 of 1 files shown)


brandonkelly:
Thanks for fixing that!

PR #6386, opened 2020-07-15
Fixes #6351

### LINKED ISSUE #6351: Cannot replace SVGs in Assets
I just get the following error.

```
Failed to load the SVG string.
```

#### Versions

- Craft CMS 3.5.0-beta.3
- PHP 7.4.4
- Imagick 3.4.4 (ImageMagick 7.0.10-1)
- Imagine 1.2.3-dev
- Amazon S3 1.2.8

joshuabaker:
Relates to #6351.

This change seems rather brute force to me, I’ll admit, however, I’m not sure how else you might go about this.

`image/svg` is not a valid mime type ([IANA](https://www.iana.org/assignments/media-types/media-types.xhtml)). Browsers cannot display it and Craft cannot read/replace assets with this mime type assigned.

As content managers cannot be expected to ensure all SVGs have the XML declaration and because most software (Adobe Illustrator, SVGOMG, et al.) now omit this when preparing for web, we should fix this in Craft.

### DIFF
--- a/src/helpers/FileHelper.php
+++ b/src/helpers/FileHelper.php
@@ -261,6 +261,11 @@ class FileHelper extends \yii\helpers\FileHelper
         if ($checkExtension && ($mimeType === null || !static::canTrustMimeType($mimeType))) {
             return static::getMimeTypeByExtension($file, $magicFile) ?? $mimeType;
         }
+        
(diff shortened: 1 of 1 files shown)


brandonkelly:
Thanks!

Issue #6097, opened 2020-05-18
[FR] Use Craft::createObject instead of "new" Object


Anubarak:
### Description
This is not really an issue for one single spot, but a general request for "coding style". I notice many times you create classes with the `new` keyword which makes it hard to fully use the potential of Yii2. Things like Exceptions and Queries don't need to be created via DI but especially components in your [ComponentHelper](https://github.com/craftcms/cms/blob/develop/src/helpers/Component.php#L110) could be created in that way to achieve Dependency Injection. 

I use Yii2 outside of the Craft world too and since the upcoming release (Yii3) is going to (kinda) drop the Service Locator my modules for Yii3 are going to drop all of the `Yii::$app->` lines as well. In order to prepare for that move tt would be awesome if Craft would let me do that easier by also using the `createObject` method to  allow me to pass dependencies.

Of course if you are going to not use Yii3 in the future and use another Framework due to the heavy changes or if you are going to stay with Yii2 for the next years it won't make any differences for you,

PS I'm talking about [this](https://github.com/yiisoft/docs/blob/master/guide/en/intro/upgrade-from-v2.md#use-di-instead-of-service-locator)

brandonkelly:
We were just talking about how we need to start refactoring code in preparation for that, whether we go with Yii 3 or Laravel, etc., it seems DI is the way to go these days.

Anubarak:
Yeah I really don't want to make that decision. Of course you could bring back the Service Locator in Yii3 (only a one liner) but to choose between the ongoing trend/"best practice" and easy to use code can be quite difficult.
I guess your internal discussion is going to be very interesting. 

In fact Craft can be moved to Yii3 even with all those upcoming breaking changes without that much effort to restructure logic. I searched all of your components and you avoided dependency cycles quite well but for example the Commerce package won't be that easy to refactor without a service locator because some classes would look like

```php
class Customers{
    public function __construct(Carts $carts){}
}
//... 
class Carts{
    public function __construct(Customers $customers){}
}
// ...
$customers = Commerce::getInstance()->getCarts();
// fatal error, out of memory, Carts tries to create a Customers instance, 
// Customers tries to create a Carts instance and so on
```

[...]

PR #6391, opened 2020-07-16
Create objects via Craft::createObject instead of new

Anubarak:
### Description
Already mentioned https://github.com/craftcms/cms/issues/6097#issuecomment-658590171 using `Craft::createObject` instead of `new` will allow us to include custom HTML into the overlay really easily so we can include settings and options for the site or where it's required.

### Break
This **may** contain breaking changes in case someone added random dependencies to the DI container. For example if someone did
```PHP
Craft::$container->set('craft\\fieldlayoutelements\\CustomField' => 'Foobar');
```
Before this pull request the code above did nothing, but now it will likely throw an error.
But to be honest I highly doubt someone without experience in Yii2 has ever used it and someone experienced know that including some random code that does nothing might not be a good idea.

### What the change allow us to do

We can set a different class as CustomField and include custom properties to it. Just as an example:
```PHP
\Craft::$container->set(CustomField::class, ['class' => \modules\fieldlayoutelements\CustomField::class]);

// ...
namespace modules\fieldlayoutelements;

use Craft;

class CustomField extends \craft\fieldlayoutelements\CustomField
{
    public $sortable = true;

    public function settingsHtml()
    {
        $settingsHtml = parent::settingsHtml();
        $settingsHtml .=  $view->renderTemplateMacro('_includes/forms', 'textField', [
            [
                'label' => Craft::t('app', 'Is this field sortable via frontend?'),
                'id' => 'sortable',
                'name' => 'sortable',
                'value' => $this->sortable,
                'placeholder' => 'sortable',
            ],
        ]) ;
        return $settingsHtml;
    }
}
```

### issues
Currently rendering a matrix field will throw an 
> Calling unknown method: craft\fieldlayoutelements\CustomField::hasErrors()

error but that was the case before as well. The function is missing in your base class, besides that everything works


brandonkelly:
Thanks! All merged in. I switched back to defining class names as `class` rather than `__class` to be consistent with the rest of the codebase, and also because `__class` support isn’t actually documented in [BaseYii::createObject()](https://www.yiiframework.com/doc/api/2.0/yii-baseyii#createObject()-detail), so don’t really trust it.

Anubarak:
Thank you so much. Yeah `__class` is rather new and came because of yii3. They started it during development for yii3 and integrated it into yii2 but didn't update the docs.
Yii3 DI will be more and more ported to yii2. With the latest version dependencies will be automatically injected into controller actions as well. It's not mentioned tho. 
Whatever. Thanks again for your time. This will allow us to define our frontend forms for sections via cp Fieldlayout programmatically. Really awesome

PR #6397, opened 2020-07-16
Store public schema’s token settings in the project config.

### LINKED ISSUE #6078: GraphQL public schema changes
### Description
I'm having a problem using project.yaml and setting up GraphQL Public Schema. I have `allowAdminChanges` false in staging and production environments.

We needed to make some changes to our public schema — somehow we ended up with multiple public schema entries in our `project.yaml`. We manually deleted the extra ones. This probably would have worked fine if we'd managed to keep the same public schema ID that was originally created.

In development, the Public Schema works as it should (presumably because it was created in development environment) — when making GraphQL queries it doesn't require the Authorization header. In staging and production, after synching with the project.yaml, Craft won't allow me to make queries to CraftQL, they all get rejected with Status 400 `{"error":"Missing Authorization header"}`.

I've verified that in the staging and production databases the `gqlschemas` and `gqltokens` are created.

What I noticed was that the `gqltokens` record for the `__PUBLIC__` accessToken has an empty schemaId field after the project.yaml sync. It appears this field isn't getting populated. Manually setting it to the created `gqlschemas` record ID is a workaround but sorts the problem.

### Steps to reproduce

1. Create a development site + database and enable GraphQL public schema.
2. Set `allowAdminChanges` false and ensure project.yaml is enabled.
3. Launch the site in staging environment with a separate database, sync with project.yaml
4. Delete and recreate the public schema in development, save new project.yaml with updated public schema. Manually delete old public schema entry from project.yaml.
5. Try make a request to GraphQL public schema on staging site

### Additional info

- Craft version: Craft Pro 3.4.19.1
- PHP version: 7.4.5
- Database driver & version: MySQL 5.7.19
- Plugins & versions: none required for this issue

andris-sevcenko:
This makes it possible to tweak the public access settings on sites with `allowAdminChanges` setting set to `true`.
The migration will also create a disabled public schema token on sites where it is missing.

Resolves #6078


brandonkelly:
Boom!

Issue #6240, opened 2020-06-19
Frontend entry form - name of variable containing submission data


piotrpog:
When we submit frontend entry form with errors, variable is injected in template that contains our submission data and errors information. 

Problem is, this variable is called `entry`, which conflicts with the existing `entry` variable (the page that contains entry form might have has its own `entry`).

I know that changing variable name would be breaking change, so why now allow changing its name using config setting?

timkelty:
@piotrpog I swear I made this FR a couple years ago, but now can't find it…so I'm glad you posted!

brandonkelly:
As of Craft 3.5 it will be possible to specify the variable names that the asset/category/entry/user should be passed back to the template as if they contain validation errors, like so:

```
{# asset forms #}
{{ hiddenInput('assetVariable', 'badAsset'|hash) }}

{# category forms #}
{{ hiddenInput('categoryVariable', 'badCategory'|hash) }}

{# entry forms #}
{{ hiddenInput('entryVariable', 'badEntry'|hash) }}

{# user forms #}
{{ hiddenInput('userVariable', 'badUser'|hash) }}
```

Thanks to @timkelty for the PR!

PR #6401, opened 2020-07-16
Configurable route variables from request

timkelty:
Listen in the request for parms to configure route variables to avoid template conflicts.

I did it for every element you can potentially save from the frontend, so `entryVariable`, `userVariable`, `categoryVariable`, `assetVariable`.

Fixes https://github.com/craftcms/cms/issues/6240

/cc @piotrpog


brandonkelly:
Can you change the `getBodyParam()` calls to `getValidatedBodyParam()`? Otherwise this opens the door for people to tamper with the forms and start changing variables in a way the template wasn’t expecting.

timkelty:
Good call, done.

brandonkelly:
Brilliant, thanks!

Issue #6478, opened 2020-07-28
[3.5.0-RC3] One field is instantiated multiple times


sebastian-lenz:
### Description

Up until now, a field (meaning an instance of `craft\base\Field`) has only been created exactly one time per configured field in the control panel. In the current release candidate one configured field can be instantiated multiple times, meaning the field class constructor will be invoked multiple times with the same field record.

### Details

In previous versions this happened in `craft\services\Fields::getAllFields` in [Fields.php#L572](https://github.com/craftcms/cms/blob/develop/src/services/Fields.php#L572). All other field utility methods basically just filtered this list of fields once loaded (e.g. `craft\services\Fields::getFieldByUid` in [Fields.php#L631](https://github.com/craftcms/cms/blob/develop/src/services/Fields.php#L631))

In Craft 3.5-RC3 there is another call to the field constructor version in `craft\services\Fields::getFieldsByLayoutId` in [Fields.php#L572](https://github.com/craftcms/cms/blob/develop/src/services/Fields.php#L1113).

I am not sure whether this is an intended behavior introduced cause users now can e.g. edit the label of fields. As a field might be pretty complex (e.g. matrix fields) it seems rather unnecessary to recreate them entirely just to change the field label. Also fields might rely on the fact that they are only created once (mine did), this might lead to inconsistencies as the field instance that loads a field from the database might not be the same as the one that displays the field.

### Steps to reproduce

1. Assuming we have a fresh and clean Craft installation, create a new field in the control panel. Attach it to a new entry type.
2. Place a debugger breakpoint within the constructor function of that field.
3. Visit the entry edit page in the control panel. The debugger will halt two times as the field is created twice. One time within `craft\services\Fields::getAllFields` and one time within `craft\services\Fields::getFieldsByLayoutId`.

### Additional info

- Craft version: 3.5.0-RC3
- PHP version: 7.4.7
- Database driver & version: MySQL 5.7.29

brandonkelly:
That actually has been the case since 3.0.0.

https://github.com/craftcms/cms/blob/0ec7014ffd69109642a4cb72a7a7f4bf5d6b5f04/src/services/Fields.php#L1027-L1049

sebastian-lenz:
Hi @brandonkelly, thanks for having a look! You are totally right, my conclusions are wrong and the double field creation existed like this in prior versions of Craft. So let us scrap my explanation. Yet the underlying observation of the changed behavior in Craft 3.5 is still true.


### Observation

[...]

PR #6486, opened 2020-07-29
Reuse loaded fields

sebastian-lenz:
Use already loaded field definitions, deprecate field layout attributes on fields, see #6478


brandonkelly:
Oh my gosh, I completely missed this PR when I was looking into your issue, and ended up implementing the exact same thing (87436c5d634a7acc8e2bf16e0043d52c37e5afd9). Great minds think alike!

PR #6487, opened 2020-07-29
Allow basic auth by username or password

timkelty:


brandonkelly:
Thanks!

brandonkelly:
Craft 3.5 RC5 is out now with that change.

PR #6498, opened 2020-07-30
GQL: Add asset uploader

timkelty:
Adds uploader to GQL asset arguments, interfaces, and mutations.

Note: based on how other things are structured, the `uploader` argument should probably be `uploaderId`? I'll leave that to you, since I didn't want to add the method to `AssetQuery` in this PR.

### DIFF
--- a/src/gql/arguments/elements/Asset.php
+++ b/src/gql/arguments/elements/Asset.php
@@ -81,6 +81,11 @@ class Asset extends ElementArguments
                 'name' => 'withTransforms',
                 'type' => Type::listOf(Type::string()),
                 'description' => 'A list of transform handles to preload.'
+            ],
+            'uploader' => [
+                'name' => 'uploader',
+                'type' => QueryArgument::getType(),
+                'description' => 'Narrows the query results based on the user the assets were uploaded by, per the user’s ID.'
             ]
         ]);
     }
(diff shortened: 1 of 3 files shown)


andris-sevcenko:
Looking good, will add changelog and merge shortly!

timkelty:
👍

@andris-sevcenko You think an `AssetQuery::uploaderId` getter should be added so the argument could be `uploaderId`? Only reason being it would match the other args like `folderId`, `volumeId`, etc.

andris-sevcenko:
>  You think an AssetQuery::uploaderId getter should be added so the argument could be uploaderId

Yeah, maybe. I would ditch the `uploader` argument in that case, though. What do you think?

timkelty:
Right – nix uploader arg and just have uploaderId was might thought

brandonkelly:
Craft 3.5 RC6 is out now with this PR.

andris-sevcenko:
@timkelty getting too close to release to change this property. Also, there's no `uploaderId` method on the `AssetQuery`, so it would require some mapping between arguments -> methods. Not hard to do, but unneeded complexity. Let's go with `uploader` for now.

timkelty:
> Also, there's no uploaderId method on the AssetQuery

Yup, that's why I held off initially.

PR #6502, opened 2020-07-31
Fixed a wrong translation

Saboteur777:


brandonkelly:
Thanks for that! We have actually removed that message in 3.5 (b10423fdd8f8780ccb0d1a805f7c9eb5e7535f8f), which is likely going to be the next stable release.

FFR, we manage our translations on [Crowdin](https://crowdin.com/project/craft-cms), which you should be able to contribute to directly, and provides a much nicer translation workflow.

Saboteur777:
Thanks Brandon, I will submit translations there in the future.

PR #6505, opened 2020-07-31
fix lightswitch id when id is available

vnali:
it seems currently random() is added to end of lightswitch id even if id is available. this PR prevents this.


brandonkelly:
Thanks for the PR!

vnali:
thanks for merging this @brandonkelly 
actually i submitted this PR to have control over id of lightswitch's container.
as you mentioned on release note,  for and ID attributes are same now. 
but if i click on label of lightswitch, the state of lightswitch doesn't change. is it expected?

brandonkelly:
It was supposed to set focus on the lightswitch, but just realized that’s not happening, so fixed it for the next 3.5 release. (dabed800c2cc1553556a7b979257766152863d41)

brandonkelly:
Craft 3.5 RC6 is out now with those changes.

PR #6523, opened 2020-08-03
GQL Asset interface: Include conditional fields

timkelty:
Missed this in https://github.com/craftcms/cms/pull/6498


andris-sevcenko:
Appreciate it, @timkelty!

PR #6525, opened 2020-08-03
Remove |e filter to render permissions properly

timkelty:
Not sure why the label here was being escaped initially, but it screwed up the output when the label has an unescaped character (e.g. **&**):

Screen Shot 2020-08-03 at 1 13 31 PM

### DIFF
--- a/src/templates/_includes/permissions.html
+++ b/src/templates/_includes/permissions.html
@@ -31,7 +31,7 @@
 
             <li>
                 {{ checkbox({
-                    label: props.label|e,
+                    label: props.label,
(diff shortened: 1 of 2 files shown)


brandonkelly:
Thanks!

Looks like at the time that `|e` was added, checkbox labels were output without encoding:

https://github.com/craftcms/cms/blob/878b18f1036e1612f10049dc92e289edac862f9c/src/templates/_includes/forms/checkbox.html#L27

Whereas that is no longer the case in Craft 3.5, per 73d448e8c5ff7418905c04c8b83b1a0bd5ab5764.

PR #6539, opened 2020-08-04
Display author for "revisionCreator/Last Edited By" if no revisions yet

timkelty:
It seems like adding the "Last Edited By" column to the element table attributes should default to the author if there aren't yet any revisions.

This makes it analogous to how "Date Updated" works.

Or, if you wanted to get even more granular, there could be a new "Last Updated By" column with this value.

### DIFF
--- a/src/elements/Entry.php
+++ b/src/elements/Entry.php
@@ -1242,6 +1242,9 @@ class Entry extends Element
                     ($creator = $behavior->getCreator()) === null
                 ) {
                     return '';
+                    if (!$creator = $this->getAuthor()) {
+                        return '';
+                    }
                 }
                 return Craft::$app->getView()->renderTemplate('_elements/element', ['element' => $creator]);
         }


brandonkelly:
Just because someone is the author doesn’t mean they were the last person to edit the entry. I’d rather just display a blank value if we don’t know for sure.

timkelty:
> Just because someone is the author doesn’t mean they were the last person to edit the entry

Right, but I'm only falling back to the author if there is no revision creator…At any rate, here's an example of how to do this if someone else ends up here:

```
        Event::on(
            Entry::class,
            Element::EVENT_REGISTER_TABLE_ATTRIBUTES,
            function (RegisterElementTableAttributesEvent $event) {
                $event->tableAttributes['lastUpdatedBy'] = ['label' => 'Last Updated By'];
            }
        );

        Event::on(
            Entry::class,
            Element::EVENT_SET_TABLE_ATTRIBUTE_HTML,
            function (SetElementTableAttributeHtmlEvent $event) {
                if ($event->attribute !== 'lastUpdatedBy') {
                    return;
                }

                $entry = $event->sender;
                $revision = $entry->getCurrentRevision();
                $behavior = $revision ? $revision->getBehavior('revision') : null;
                $creator = $behavior ? $behavior->getCreator() : null;
                $lastUpdatedBy = $creator ?? $entry->getAuthor();

                $event->html = Craft::$app->getView()->renderTemplate('_elements/element', ['element' => $lastUpdatedBy]);
            }
        );
```

PR #6558, opened 2020-08-05
Updated listview_sort to mirror proportions of other UI elements

kerns:
Replacing off-proportion sort icon used in column sorting in the entry list view. Not sure yet where the source UI files live (if anywhere) and what format you keep them in but can provide as .psd, sketch, figma or other.
listview_sort


brandonkelly:
Thanks!

brandonkelly:
Actually, something is up with that PR (as well as #6559) – they just delete the images, rather than replace them. So have reverted this for now.

kerns:
> Actually, something is up with that PR (as well as #6559) – they just delete the images, rather than replace them. So have reverted this for now.

Sorry about that. Learning by doing. Hope it works with [this one](https://github.com/craftcms/cms/pull/6563).

PR #6572, opened 2020-08-06
add sitegroup handle & more defined template paths

ThomasDeMarez:
### Description
Simple implementation of Site specific templates inside of a specific SiteGroup.
Uses auto generated SiteGroup handles based on the group name.

```
.
└── templates/
    ├── SiteGroupHandle/
    │   ├── SiteHandle/
    │   │   └── index.twig ----> specific template for site with this handle in this group
    │   └── index.twig     ----> default template for all sites in this group
    └── index.twig.        ----> fallback for all other sites
```

### Related issues
#3101

### DIFF
--- a/src/controllers/SitesController.php
+++ b/src/controllers/SitesController.php
@@ -14,6 +14,7 @@ use craft\models\Site;
 use craft\models\SiteGroup;
 use craft\web\assets\sites\SitesAsset;
 use craft\web\Controller;
+use yii\helpers\Inflector;
 use yii\web\BadRequestHttpException;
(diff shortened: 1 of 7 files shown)


felds:
@brandonkelly Any news on this one? I was looking for this feature and this looks like a good implementation. Also, in the case this won't be accepted, do you have any pointers to how implement it as a plugin or userland code?

brandonkelly:
We’ve decided against pulling this in. Here’s one way you could pull it off using a custom template root: https://github.com/craftcms/cms/discussions/8564#discussioncomment-5851314

PR #6582, opened 2020-08-07
Update field layout designer to fit with LTR and RTL a little better

Harry-Harrison:
### Description

This is entirely an opinion thing, sorry if it's not really suitable. It may make more sense as a plugin?

I noticed while using the new field layout designer that the general flow of adding fields didn't feel quite right. Flipping the location of the field-pool so it's on the left of the workspace I think fits with the natural language approach of things. It also feels a lot easier to use if you're on a wide screen.

This uses the same mixins, so it's correctly on the right side on right-to-left locales.


brandonkelly:
The current layout is pretty conventional. Putting the sidebar on the left doesn’t make sense hierarchically because it contains a library of things that can be added to the layout; they are not in any way a parent of the workspace.

PR #6600, opened 2020-08-10
Give EVENT_BEFORE_CREATE_REVISION a chance to set creatorId, source

timkelty:
Using `Revisions::EVENT_BEFORE_CREATE_REVISION`, you can set `revisionNotes`, but not `creatorId` or `source`.

Here's my use-case:

My client wants to be able to track the last person (of their team) that has edited an entry, so I added the "Last Edited By" column.

However, this section is also hooked up to a frontend submission form. When submitted via GuestEntries, it works as they want – "last edited by" is empty until one of them edits in the CP. However, if a user is logged in and submits via the normal save-entries controller, the revision automatically gets mapped to them.

This all makes sense, but in this case, if I were able to change the `creatorId` form the `EVENT_BEFORE_CREATE_REVISION`, I could just set it to `null` in the cases I want and be done.

Instead, I ended up having to add a new users field (`lastCPEditor`), that I update via `BEFORE_SAVE_ENTRY`, and add that column to my element indices.

### DIFF
--- a/src/services/Revisions.php
+++ b/src/services/Revisions.php
@@ -126,6 +126,8 @@ class Revisions extends Component
         ]);
         $this->trigger(self::EVENT_BEFORE_CREATE_REVISION, $event);
         $notes = $event->revisionNotes;
+        $creatorId = $event->creatorId;
+        $source = $event->source;
 
         $elementsService = Craft::$app->getElements();
 


brandonkelly:
Gracias!

brandonkelly:
Craft 3.5.3 is out now with this change.

PR #6646, opened 2020-08-16
Remove checkerboard from small thumbs when img present

johnfmorton:
In the user listing, small thumbnail images showed a checkerboard pattern in their transparent area instead of the solid background color.

### Description

Here is a screenshot of the proposed change. https://share.jmx2.com/cgv9zQ

I could fix it in my CSS directly, but since you're using SCSS, I've attempted to track down the spot in the SCSS that created the CSS line as shown here: https://share.jmx2.com/W0nqES

### DIFF
--- a/src/web/assets/cp/src/css/_main.scss
+++ b/src/web/assets/cp/src/css/_main.scss
@@ -2046,7 +2046,7 @@ $elementInnerSpacing: 5px;
     display: inline-block;
   }
 
-  &.small,
+  &.small:not(.hasthumb),
   &.large:not(.hasthumb) {
     display: inline-block;
     padding: $baseElementSidePadding;


brandonkelly:
Just made this change, but only for user photos, and excluding thumbnail view. I’ve also made small user photos circular again, without affecting assets or any other element types that have thumbnails.

carlcs:
Does this also remove checkerboard in thumbs for non-image files?
image

mmikkel:
@carlcs Probably not based on Brandon's comment above, but those have been bothering me, too :)

brandonkelly:
@carlcs Good call! That’s a bit more complicated to do purely from CSS, so went ahead and added new `getHasCheckeredThumb()` and `getHasRoundedThumb()` element methods, so element types have control over each of those behaviors now.

brandonkelly:
Craft 3.5.5 is out now with those changes.

PR #6655, opened 2020-08-17
Eager loads Author in Recent Entries widget

mildlygeeky:
Currently the Recent Entries widget does two queries on entry.author for every entry in Recent Entries - this gets it to properly eager-load the author and keeps queries constant.

### DIFF
--- a/src/widgets/RecentEntries.php
+++ b/src/widgets/RecentEntries.php
@@ -180,6 +180,7 @@ class RecentEntries extends Widget
         $query->sectionId($targetSectionId);
         $query->editable(true);
         $query->limit($this->limit ?: 100);
+        $query->with(['author']);
         $query->orderBy('elements.dateCreated desc');
 
         return $query->all();


mildlygeeky:
@brandonkelly - could this get merged in? Really small change, but can really improve the first-load experience for people logging into the Craft CP, as Recent Entries is a default widget - putting this into a typical Dashboard with a single 10 entry Recent Entries widget will reduce the number of queries by 18. If someone is using multiple Recent Entries widgets or showing more than 10, than the effect will be even more drastic.

mildlygeeky:
@brandonkelly Calling this out - I think this is the only spot in the CP where there is some inefficient query joins

brandonkelly:
Sorry dude! Not sure why this slipped through the cracks. Merged for the next release.

mildlygeeky:
No worries! Not a groundbreaking PR, but I can't help but think of all those lost person-milliseconds lost on every Dashboard load. ;)

brandonkelly:
We just released Craft 3.6.12 with this change.

PR #6692, opened 2020-08-21
Improve downloading assets from stream

boboldehampsink:
Sometimes downloading an asset results in an error like `yii\base\ErrorException: fseek(): stream does not support seeking in /app/vendor/yiisoft/yii2/web/Response.php:586`

We know the Asset size and mimeType so we can just feed them as options to the `sendStreamAsFile` method.
By not having to guess the filesize and mimetype we can improve the stream stability.


brandonkelly:
Craft 3.5.6 is out now with this PR. Thanks again!

PR #6730, opened 2020-08-26
Fix issue with Globals redirect to first set where the array of sets might not be zero indexed

tjdraper:
# Description

If a user that does not have access to the alphabetically first global set clicks on "Globals" in the CMS nav, an exception will be thrown with the message `Undefined offset: 0`.

Screen Shot 2020-08-26 at 12 16 34 PM

Screen Shot 2020-08-26 at 12 17 59 PM

Digging into this, I found that the user group did not have access to the first global set alphabetically.

Screen Shot 2020-08-26 at 12 18 33 PM

A dump of `$editableSets` revealed that the array index in this case did not start at `0`.

Screen Shot 2020-08-26 at 12 19 32 PM

Screen Shot 2020-08-26 at 12 19 39 PM

Wrapping `$editableSets` with an `array_values` function produces the editable sets array with a 0 index, allowing the user access to the Global Sets they have access to.

Screen Shot 2020-08-26 at 12 20 25 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@
 ### Changed
 - The GraphQL API now supports `variables` and `operationName` query string parameters. ([#6728](https://github.com/craftcms/cms/issues/6728))
 
+### Fixed
+- Fixed a bug where a user trying to access globals that does not have access to the alphabetical first global set would throw an `Undefined offset: 0` exception
+
 ## 3.5.7 - 2020-08-26
 
 ### Changed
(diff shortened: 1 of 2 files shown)


brandonkelly:
Thanks for reporting that! Just fixed the real culprit here, which is that `craft\services\Globals::getEditableSets()` should be returning the array with the correct indexes.

tjdraper:
Awesome!

CreateSean:
is this fix in 3.5.7 because I updated this morning and am still seeing this.

As a work around I renamed the global set so that it's not first.

andris-sevcenko:
That fix has not yet been released, no. (@CreateSean)

brandonkelly:
Craft 3.5.8 is out now with this fix.

PR #6731, opened 2020-08-26
Add `configure` Twig function

AugustMiller:
This allows you to do something like this:

```twig
{% set myCustomizedResults = configure(entry.myMatrixField, {
  type: ['blockTypeHandleA', 'blockTypeHandleB'],
  with: [['myAssetField']]
}).all() %}
```

In addition, it basically makes it possible to set properties on whatever objects you come across—or your own…

```twig
{% set baseCriteria = {
  section: ['posts']
} %}

{% if currentUser.isAdmin %}
  {% do configure(baseCriteria, { onlyVips: true }) %}
{% endif %}

{% entries = craft.entries(myCriteria).all() %}
```

…or even worse…

```twig
{% do configure(entry, { title: 'New Title!' }) %}
```

I've got another PR coming for docs, should y'all like the idea!

### DIFF
--- a/src/web/twig/Extension.php
+++ b/src/web/twig/Extension.php
@@ -955,6 +955,7 @@ class Extension extends AbstractExtension implements GlobalsInterface
             new TwigFunction('className', 'get_class'),
             new TwigFunction('clone', [$this, 'cloneFunction']),
             new TwigFunction('combine', 'array_combine'),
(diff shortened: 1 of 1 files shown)


brandonkelly:
Makes sense since we already have `create()`. Thanks for the PR!

brandonkelly:
Craft 3.5.8 is out now with the new `configure()` Twig function.

PR #6769, opened 2020-09-02
Set ApplicationTrait::$env in test setup

boboldehampsink:
This makes Craft::$app->env available in tests


brandonkelly:
Good call, thanks.

PR #6791, opened 2020-09-04
Update example test suite

qrazi:
### Description
Add missing constant to example test suite, bringing it in line with official docs


angrybrad:
Thanks, @qrazi!

PR #6805, opened 2020-09-08
Update translations for swiss german (de-CH)

christianruhstaller:
### Description
This pull requests updates some missing translations and fixes wrong quotation marks.
Swiss german uses `«…»` compared to high german which uses `„…“`.


brandonkelly:
Thanks! FFR we manage translations over at [CrowdIn](https://crowdin.com/project/craft-cms). You might find it easier to contribute to translations from there rather than a PR.

PR #6814, opened 2020-09-10
Fixed missing white space in install agreement.

wrux:
### Description

Fixed a missing space below the "got it" button in the license agreement.

I've attached screenshots:

**Before**
Screenshot 2020-09-10 at 23 26 24

**After**
Screenshot 2020-09-10 at 23 26 41


brandonkelly:
Was able to reproduce that in Firefox, however applying your change added double the padding in Chrome, Safari, etc. Came up with a different fix that works the same across all browsers.

Issue #2872, opened 2018-05-11
Request: Ctrl + S on Field pages


mark-chief:
Since Ctrl + S 'ing is so damn handy it would be good to have this functionality when adding or editing a field.

I am quite often making small tweaks then clicking save to then be sent back to the fields overview page. Especially irritating when editing a matrix field block as you have to click on the field, and then drill down through the blocks to find the field you were editing.

Whilst I am here, "Save and add another field" would also be super handy.

We cool though, there's worst things in the world that not having this.

j-greig:
Control + S would be nice to have on _any_ page with a Save button IMO :)

brandonkelly:
@j-greig that should already be the case ;) This request is more about post-shortcut behavior (you should probably be kept on the same page rather than redirected to the parent page).

brandonkelly:
This has been added for the next release 🎉

![The Save button on an Edit Field page, with its new expanded menu, revealing a “Save and continue editing” option](https://user-images.githubusercontent.com/47792/92963423-4498fc00-f427-11ea-8d3c-86baab4dd1f5.png)

brandonkelly:
Craft 3.5.10 is out now with this.

carlcs:
This would be nice to have on the Entry Type settings page as well!

brandonkelly:
Added for the next release, as well as for Edit Section pages.

brandonkelly:
Craft 3.5.17 is out now with the change to Section / Entry Type pages.

PR #6850, opened 2020-09-17
Ctrl/Cmd + Shift + S: save field settings and go back

lhs-dev:
Add a new shortcut for the old behavior of the field settings "Save" button: Ctrl/Cmd + Shift + S will save the field settings and take you back to the list of fields.

### Description
#2872 Requested a new behavior for the Ctrl/Cmd + S on edit field settings page. This PR adds a shortcut for the previous behavior: save and go back to the list of fields.

### DIFF
--- a/src/templates/settings/fields/_edit.html
+++ b/src/templates/settings/fields/_edit.html
@@ -7,6 +7,11 @@
         shortcut: true,
         retainScroll: true,
     },
+    {
+        label: 'Save'|t('app'),
+        shortcut: true,
+        shift: true,
+    },
 ] %}
 
 {% import "_includes/forms" as forms %}


brandonkelly:
This breaks existing conventions. Currently the only place we support a <kbd>Command</kbd>+<kbd>Shift</kbd>+<kbd>S</kbd> shortcut is the “Save and add another” action on Edit Entry pages, and I would expect the same shortcut on Edit Field pages would have the same behavior (save the field and bring you to a new field creation page).

Plus, you can trigger a normal form submit by focusing on anything besides multi-line text inputs / editable tables, and clicking the <kbd>return</kbd> key.

Issue #6849, opened 2020-09-16
GraphQL returns an error with site "*"


smcyr:
### Description
With a long query with (`site: "*"`), GraphQL returns this error:
`"Invalid site param: [*, *, *, *, *, *, *, *, *, *, *, *]"`

For some reason, CraftCMS do some eager loading for multiSiteContent but results in a site array with multiple * instead of one.

You already have my composer.json/lock and database from this issue: https://github.com/craftcms/cms/issues/6844

Here is the query:

```graphql
query FetchPage($uri: [String]) {
  pageEntry: entry(site: "*", uri: $uri) {
    ...PageFragment
    __typename
  }
}

fragment PageFragment on EntryInterface {
  id
  title
  level
  uri
  url
  postDate
  expiryDate
  dateUpdated
  ... on blogPost_blogPost_Entry {
    alternateUri
    siteId: currentSiteId
    seoDescription
    seoTitle
    seoRobotFollow
    seoRobotIndex
    seoAlternateUrl
    seoOpenGraphImage {
      ...ImageFragment
      __typename
    }
    blocks {
      ...BlocksFragment
      __typename
    }
    ...BlogPostFragment
    __typename
  }
  ... on blogPostList_blogPostList_Entry {
    alternateUri
    siteId: currentSiteId
    seoDescription
    seoTitle
    seoRobotFollow
    seoRobotIndex
    seoAlternateUrl
    seoOpenGraphImage {
      ...ImageFragment
      __typename
    }
    __typename
  }
  ... on job_job_Entry {
    alternateUri
    siteId: currentSiteId
    seoDescription
    seoTitle
    seoRobotFollow
    seoRobotIndex
    seoAlternateUrl
    seoOpenGraphImage {
      ...ImageFragment
      __typename
    }
    blocks {
      ...BlocksFragment
      __typename
    }
    ...JobFragment
    __typename
  }
  ... on jobsList_jobsList_Entry {
    alternateUri
    siteId: currentSiteId
    seoDescription
    seoTitle
    seoRobotFollow
    seoRobotIndex
    seoAlternateUrl
    seoOpenGraphImage {
      ...ImageFragment
      __typename
    }
    blocks {
      ...BlocksFragment
      __typename
    }
    ...JobsListFragment
    __typename
  }
  ... on pressRelease_pressRelease_Entry {
    alternateUri
    siteId: currentSiteId
    seoDescription
    seoTitle
    seoRobotFollow
    seoRobotIndex
[...]

PR #6865, opened 2020-09-18
Fix for GraphQL where it returns an error with site: "*"

smcyr:
### Description

Replace `array_merge_recursive` by `ArrayHelper::merge` for eager loading rules in Graphql queries. Instead of merging string values (like *) in an array, it just replace it. For numeric values, it's the same behavior as before.

### Related issues

https://github.com/craftcms/cms/issues/6849


andris-sevcenko:
Switching to `ArrayHelper::merge()` will have negative consequences here - while it wouldn't merge `*` together in an array, it would also not merge different strings together in an array, either, which would break eager loading in some cases.

Opted for a cleanup function, instead. Thanks for figuring out the root cause, though! Appreciate it!

PR #6889, opened 2020-09-22
Tidy settings and add group tags for docs

mattstein:
- Adds `@group` tags to general config.
	- Example output: [markdown](https://gist.github.com/mattstein/fbc2188438359dc94f018146fff4e00a) + [weird screenshot](https://user-images.githubusercontent.com/2488775/93951614-70ac5b00-fd0c-11ea-8234-0abe5202d5fd.png)
- Improves consistency of line lengths (<150 characters) and breaks within general and db config docblocks.
- Adds line breaks to general and db config so properties and their docblocks are grouped visually.
- Includes dazzlingly pedantic nitpicks and minor corrections.

## Task List

- [x] Add `@group` tags to all general config properties.
- [x] Clean up formatting+wording for general and db configs.
- [x] Post example markdown with groupings from revised doc generator.


mattstein:
@brandonkelly This is ready to go unless the _Image Handling_ group should be folded into _Assets_. (Opted not to since not all Assets are images.)

PR #6931, opened 2020-09-30
Skip eager-loading this segment if there is no filtered elements

smcyr:
### Description
If the filtered elements is empty, it causes an exception.

### Related issues
Issue described here:
https://github.com/craftcms/cms/issues/6849


brandonkelly:
Thanks for the PR! Only necessary to make that check if we actually needed to filter the elements, so applied a slightly different fix.

PR #6936, opened 2020-10-01
Adding a progress count to the output of the resave console commands.

JamesNuttall:
Pulling in the query count to the `beforeCallback` and using the `BatchElementActionEvent` position property to display a progress count in the console when resaving elements. 

Example:
```
- Resaving (32/1341) Graphic Design (283204) ... done
```


brandonkelly:
Thanks for the PR!

brandonkelly:
Craft 3.5.13 is out now with that change ✨

PR #6942, opened 2020-10-02
Add hover style to improve contrast of Craft version info

marzepani:
### Description

I have a hard time reading Craft's version info in the control panel. This PR adds a hover style to increase contrast. I added a delay to the transition to prevent flickering when moving the pointer over.

default state (like now)
Screenshot-2020-10-02-at-13 01 37

hover state
Screenshot-2020-10-02-at-13 01 42


brandonkelly:
Thanks for the PR! Coincidentally we had just started talking about how to improve that last week. We decided it would be better to move it out of the sidebar, where we can give it a normal contrast ratio without being quite so glaring. That’s done now for the next release.

Issue #6945, opened 2020-10-02
Using the renderTemplate() in the console given error


mostafasoufi:
I wonder why the render template on the console given me the YII error

```
Calling unknown method: yii\web\UrlManager::getMatchedElement()
```

And here is the method in the controller.
```
public function sendEmailToCustomer(Order $order)
{
    Craft::$app->getView()->setTemplatesPath(CRAFT_BASE_PATH . '/templates');
    $html = Craft::$app->getView()->renderTemplate('_email/order/customer-invoice', [
        'order'           => $order,
    ], View::TEMPLATE_MODE_SITE);

    $message = new Message();
    $message->setSubject("Invoice for order #{$order->getId()} is ready!")
        ->setHtmlBody($html)
        ->send();
}
```

### Additional info

- Craft version: 3.5.12
- PHP version: 7.2.19
- Plugins & versions: Commerce 3.2.7

brandonkelly:
Thanks for reporting that. Craft 3.5.12.1 is out now with the fix for this.

brandonkelly:
Craft 3.5.13 is out now with that fix.

PR #6952, opened 2020-10-05
Avoide to call _getMatchedElementRoute when UrlManager uses in console

mostafasoufi:
### Description

When you're using the `Craft::$app->getView()->renderTemplate()` the below error will be occurred.

> Exception 'TypeError' with message 'Argument 1 passed to craft\web\UrlManager::_getMatchedElementRoute() must be an instance of craft\web\Request, instance of craft\console\Request given, called in www\nova\vendor\craftcms\cms\src\web\UrlManager.php on line 229'

So the PR will avoid to call the `_getMatchedElementRoute()`

### Related issues
https://github.com/craftcms/cms/issues/6945

### DIFF
--- a/src/web/UrlManager.php
+++ b/src/web/UrlManager.php
@@ -222,6 +222,10 @@ class UrlManager extends \yii\web\UrlManager
             return $this->_matchedElement;
         }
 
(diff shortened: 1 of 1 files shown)


brandonkelly:
Thanks!

Issue #6900, opened 2020-09-24
UrlHelper ignores/strips valueless parameters?


evanwarner:
I filed the following bug with Solspace and they responded that it's a Craft issue.

> It seems that an overridden [`returnUrl `](https://docs.solspace.com/craft/freeform/v1/template-functions/freeform.form.html#param-returnurl) property ignores/strips valueless query parameters. If I set the value (via a hidden field) to `{{ 'my/path/?thanks' | hash }}`, it shows correctly in the field, but the form only returns to `my/path/`. I have to give the parameter a value (e.g. `{{ 'my/path/?thanks=1' | hash }}`) in order for it to return to `my/path/?thanks=1`.
> 
> I can use a valueless parameter just fine if it is set within Craft on the form's **Return URL** field.

Their developer responded that it's an issue with `UrlHelper::url()` removing query params that have no value. A quick look at their source code only revealed use of `UrlHelper::cpUrl()`, but I wanted to lob this to you and see if the issue was an obvious fix in Craft.

### Additional info

- Craft version: 3.5.10
- PHP version: 7.4.2
- Database driver & version: MySQL 5.7.26
- Plugins & versions: Freeform 3.8.1

brandonkelly:
Thanks for reporting that! Just fixed for the next release.

brandonkelly:
Craft 3.5.12 is out now with that fix.

bencroker:
The fix in https://github.com/craftcms/cms/commit/7e2642cffef78fcd4f25a14daa797ce59c724da3 introduces a change that is breaking things in some plugins (https://github.com/putyourlightson/craft-campaign/pull/181). Previously when the value was a blank string, `?token=` was output instead of what is output since 3.5.12 which is`?token`.

Submitting a PR with a test now to fix this regression.

bencroker:
FYI I applied a fix to the Campaign plugin in https://github.com/putyourlightson/craft-campaign/commit/b2c6e467ff37a5508ec71e14634cf6aae16cecfc as this introduced a rather nasty bug, so it is no longer affected by the issue.

brandonkelly:
@bencroker Sorry about the unintended break. PHP will treat both `?lid=` and `?lid` the same though (as `$_GET['lid'] = ''`), so I’m confused on how the change would have broken anything?

bencroker:
[...]

PR #6965, opened 2020-10-07
Fixes change in how blank parameter values are applied in URL

bencroker:
### Description

Fixes https://github.com/craftcms/cms/issues/6900#issuecomment-705166068


brandonkelly:
This was an intentional change. To PHP, `?foo=` and `?foo` are treated the same (`$_GET['foo'] = ''`).

Issue #6028, opened 2020-05-05
Lightswitch fields are not properly searchable.


msimpson:
### Description

The `keywords` for a lightswitch field within the `searchindex` table are erroneous. While `true` values are stored correctly as `1`, `false` values result in an empty string (expected `0`). Thus making it impossible to search for the `false` state due to the `LIKE` operator.

### Additional info

- Craft version: Craft Pro 3.4.18
- PHP version: 7.4.3
- Database driver & version: MySQL 5.5.5

brandonkelly:
Instead of using the `search` parameter, you should use your Lightswitch fields’ dedicated query parameters to filter results by their values.  See https://docs.craftcms.com/v3/lightswitch-fields.html#templating

brandonkelly:
Instead of using the `search` parameter, you should use your Lightswitch fields’ dedicated query parameters to filter results by their values.  See https://docs.craftcms.com/v3/lightswitch-fields.html#querying-elements-with-lightswitch-fields for details.

msimpson:
Okay. I'll just avoid the issue, then ...

brandonkelly:
To clarify, `search` is not meant to be a precise tool, so little idiosyncrasies like this are normal and out of our control for the most part.

msimpson:
I understand that `search` isn't meant to be a precise tool.

But, if `craft\fields\Lightswitch` implemented its own `getSearchKeywords` method which returned numeric values, directly, instead of relying on the version inhereted from `craft\base\Field` (where `StringHelper::toString` causes this discrepancy), this wouldn't be a problem.

So this seems like an easily resolvable oversight and not some idiosyncrasy outside your control.

brandonkelly:
Honestly the most awkward thing about this to me is that it’s even returning `1` for enabled values. That’s not meant to be a search keyword. If I were to add that method to Lightswitch fields, it would be so that an empty string is always returned. You’re going to be much better off using the field’s element query param, which is significantly faster than going through `search`.

msimpson:
I've since refactored my code to use query parameters, instead. So while this is a non-issue for my project, I was still curious about the implications on search overall. And to your point, it doesn't seem the lightswitch field should be searchable at all given the current implementation. Search would need to be refactored to require field name specification on such field types to even provide relevant results.

brandonkelly:
[...]

PR #7054, opened 2020-10-23
Adds the ability for plugins/modules to intercept an entries keywords before saving

OwenMelbz:
### Description
After discussing with Oli over email, it was decided that the "cleanest" way for users to be able to manipulate keyword data would be to introduce a new event that works like others such as `RegisterComponentTypesEvent::class`

**This PR is really a proof-of-concept, and not completely polished yet, e.g. hense no code comments etc - It is here as a bit of an RFC to discuss how to move forward**

The naming and location of the events might need changing etc, e.g. Maybe the event should be `SearchKeywordsEvent` for example rather than `ElementKeywordsEvent` etc.

In a nut shell, this PR adds the ability via an Event for users to intercept the keywords that will be saved for `field`.

This gives them the opportunity to overwrite or manipulate them in anyway, such as appending author names, or other information into the keywords via an automated and seamless method.

Example usage:

```php
Event::on(Search::class, Search::EVENT_ELEMENT_KEYWORDS, function (ElementKeywordsEvent $event) {
    $event->keywords = 'overwrite the default keywords with this';
    return $event;
});
```

### DIFF
--- a/src/events/ElementKeywordsEvent.php
+++ b/src/events/ElementKeywordsEvent.php
@@ -1,0 +1,30 @@
+<?php
+/**
+ * @link https://craftcms.com/
+ * @copyright Copyright (c) Pixel & Tonic, Inc.
+ * @license https://craftcms.github.io/license/
+ */
+
+namespace craft\events;
+
+use craft\base\ElementInterface;
+use yii\base\Event;
+
+/**
+ * Element keywords event class.
+ *
+ * @author Pixel & Tonic, Inc. <support@pixelandtonic.com>
+ * @since 3.0.0
+ */
+class ElementKeywordsEvent extends Event
+{
+    /**
+     * @var string List of keywords to be saved to the index.
+     */
+    public $keywords;
+
+    public $element;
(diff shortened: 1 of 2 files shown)


OwenMelbz (commented):
ready for feedback

brandonkelly:
This is already possible as of 3.5.0 – see https://github.com/craftcms/cms/issues/6028#issuecomment-625568337.

OwenMelbz:
Hi,

Sorry that’s not what this PR is for.

That comment you’ve referenced is relating to a specific field on your entry type.

Eg that effectively allows you to convert one fields data into another format.

However it requires you to have a field in the first place.

This PR allows you to physically manipulate the keywords before they enter.

This doesn’t require you to have a field - especially one that you would never use.

E.g. maybe you want to include author information within the keywords, or maybe you need to lookup some remote data on an API to insert it as well.

——-

To re-iterate.

This allows you to manipulate the keywords being saved.

It has nothing to do with fields on the entry.

Issue #6833, opened 2020-09-15
aria-label attributes should be used in addition to title


brandonkelly:
Per [this Dev.Opera article](https://dev.opera.com/articles/ux-accessibility-aria-label/#accessible-name-calculation), `aria-label` is preferred over `title` for screen reader support.

There’s a ton of elements in the CP that specify a `title` but are missing an `aria-label` attribute.

missmatsuko:
Related: noticed that many of the pages have H1 with an unnecessary title attribute containing the same text.

brandonkelly:
@missmatsuko That is generally intentional, done in places where the text could be truncated, so there’s still a way to see the full value by hovering your cursor over it. Is there a practical downside to doing it?

missmatsuko:
Is there a reason not to show the whole title in the heading, and have it wrap instead of truncating? 

As it is now, the full title wouldn't be available for keyboard users who don't use a screen reader since the full title is only visible on mouse hover.

The title attribute often isn't exposed to screen readers, and in other cases it is read in addition to the original text content.

brandonkelly:
> Is there a reason not to show the whole title in the heading, and have it wrap instead of truncating?

Context: #4548

Happy to consider taking the styles in a different direction though! Maybe @benjamindavid has some ideas…

PR #7069, opened 2020-10-29
Add aria-label where title is used

missmatsuko:
### Description
Add aria-label where title is used

### Related issues
#6833


brandonkelly:
Nice!!

PR #7072, opened 2020-10-29
Use $event->query in `executeQuery`

jamesedmonston:
### Description

Currently, we're unable to modify the query passed into `GraphQL::executeQuery`.

Although the `EVENT_BEFORE_EXECUTE_GQL_QUERY` trigger is being called, the variable passed into the function initially is being used to create the cache keys and run the query, rather than the modified query from `$event`.

### DIFF
--- a/src/services/Gql.php
+++ b/src/services/Gql.php
@@ -455,7 +455,7 @@ class Gql extends Component
         if ($event->result === null) {
             $cacheKey = $this->_getCacheKey(
                 $schema,
-                $query,
+                $event->$query,
                 $event->rootValue,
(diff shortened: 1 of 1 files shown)


brandonkelly:
It’s an object so it is passed by reference, so changes made on the query will persist without this change.

andris-sevcenko:
Thanks for doing this and sorry for the delayed response on this.

I ended up adding this to Craft 3.6, on the off-chance that something _might_ have been modifying that query on the event and relying on the fact that it is not changed.

3.6 should be out pretty soon!

PR #7081, opened 2020-10-31
Fixing typing for craft\records\Site::baseUrl

dsmrt:
### Description
Should be string, not boolean. Just a typing issue.


brandonkelly:
Thanks!

PR #7090, opened 2020-11-03
Fix test class namespaces

bencroker:
Changed test class namespaces to be PSR-4 compliant following composer warnings.


brandonkelly:
Thanks!

PR #7099, opened 2020-11-06
Set notice error when the email is already verified

mostafasoufi:
### Description
I don't know why this notice was not been set on the user controller, anyway, this PR will do that.


brandonkelly:
Not quite that simple, as they could also be trying to reset their password.

PR #7101, opened 2020-11-06
Zip backups by default

bencroker:
This PR zips backups by default in the console controller action. This makes sense to save space on the server as the command will generally be run through scheduled cron jobs. It is also in-line with how backups are zipped in the utilities controller action:
https://github.com/craftcms/cms/blob/develop/src/controllers/UtilitiesController.php#L352-L353

### DIFF
--- a/src/console/controllers/BackupController.php
+++ b/src/console/controllers/BackupController.php
@@ -30,7 +30,7 @@ class BackupController extends Controller
      * @var bool Whether the backup should be saved as a zip file.
      * @since 3.5.0
      */
-    public $zip = false;
+    public $zip = true;
 
     /**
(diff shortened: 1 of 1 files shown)


brandonkelly:
In general boolean console command options are better to be `false` by default, so you can override just by typing their name (e.g. `--zip`). With this PR, you would have to start specifying the value, which is awkward for booleans (`--zip=0`).

If we were going to do this, it would be better if the option was renamed to `$noZip`, so it could be passed as `--no-zip` when a `.sql` file is desired.

bencroker:
That's a good point, and the `--zip` option is anyway documented when running `./craft help db/backup`.

Screenshot 2020-11-09 at 20 31 57

PR #7131, opened 2020-11-13
Fixed redirection when field layout is submitted

nicosomb:
### Description

When we submit the Field layout form, we are redirected to the main settings page. It's not user friendly when you want to do many changes. 

### Related issues

N/A

### DIFF
--- a/src/templates/settings/users/fields.html
+++ b/src/templates/settings/users/fields.html
@@ -8,7 +8,7 @@
 {% block content %}
     <form method="post" accept-charset="UTF-8" data-saveshortcut data-confirm-unload>
         {{ actionInput('users/save-field-layout') }}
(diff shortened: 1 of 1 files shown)


brandonkelly:
Thanks for the PR! The `redirect` input isn’t necessary if the URL should stay the same. Just removed it from that page as well as Settings → Users → Settings.

nicosomb:
Thank you @brandonkelly. So, do you want that I update the PR?

brandonkelly:
Sorry to be clear, I’ve already made the change for the next release (0e2e263d0cd17e1954b85f768f1127e0880cc570).

brandonkelly:
Craft 3.5.16 is out now with that change.

Issue #7110, opened 2020-11-09
Ability to access reverse relations within entry query in GQL


andrewfairlie:
It would be really useful if when querying entries in GraphQL you could get _reverse_ related elements to that entry, which would avoid you having to do multiple GraphQL calls of "get the entry ID, and then run another GraphQL query with `relatedTo()` because this can quickly add up and become clunky. 

## Original comment from other thread

Just a thought, and apologies if this was covered in other thoughts above and I misunderstood, would a good syntax for this be something like `relatedEntries`, `relatedAssets` etc. which behave like a normal `entries` / `asset` loop but with the `relatedTo` pre-populated with the current entry. So you could do things like this very weird, but hopefully useful example...?

```
entry(section:"offices",slug:"london"){
   title,
   id,
   ... on offices_offices_entry {
      peopleInThisOffice: relatedEntries(section:"people"){
        title
      }
      facilitiesInThisOffice: relatedEntries(section:"facilities"){
        title
      }
      imagesTaggedInThisOffice: relatedAssets(source:"images"){
        title
      }
   }
}
```

_Originally posted by @andrewfairlie in https://github.com/craftcms/cms/issues/5143#issuecomment-721702205_

andris-sevcenko:
This has been added for Craft 3.6

andris-sevcenko:
Well, in its own branch for now.

brandonkelly:
Merged now. Note that it’s `relatedToEntries`, etc., not `relatedEntries`.

brandonkelly:
Craft 3.6 RC2 is out now with these new GraphQL arguments.

brandonkelly:
Craft 3.6 has now been officially released ✨

membla:
Just had a conversation on Discord with @andris-sevcenko on this – the actual issue at hand here is not resolved and even if it may not be feasible to do it in the GraphQL API without changes to the Element Query API or Element itself I think the issue should be reopened because the use case is very valid (and well explained in the original post above).

brandonkelly:
Whoops, you’re right @membla. Guess we were on a different brainwave. We’re considering how to best address this.

Issue #6911, opened 2020-09-28
[FR] Allow filtering of elements by other criteria in GraphQL queries


joshuabaker:
### Summary

We currently cannot filter elements by other criteria in GraphQL queries like we can in Twig.

### Example use case

A common use case is to return **all categories that are in use** (i.e. related to any entry). 

In Twig, we’d approach this using a subquery, like the following.

```twig
{% set allEntriesQuery = craft.entries.limit(null) %}

{% set usedCategoriesQuery = craft.categories.relatedTo(allEntriesQuery) %}
```

This is performant because we’re not actually getting the entries in the `allEntriesQuery` here, we’re just limiting the `usedCategoriesQuery` by the subquery. It’s also not limited, should we need to get all elements that are related to a set of elements, that are related to a set of elements, and so on.

In GraphQL, there’s currently no way of achieving this without actually getting _all_ of the entries first (potentially thousands) and then sending a subsequent request to get the categories (constrained by the `id` parameter).

### Proposal

I’ve spent some time thinking about how this could work in the context of Craft without having to implement reverse relations. 

If we allow full access to the criteria as an object (see below) and leverage [enumeration types](https://graphql.org/learn/schema/#enumeration-types) to define the different element types (`Category` and `Entry` below) it’d be possible to map to the existing element queries like Twig does, I believe.

```gql
{
  elements(
    criteria: {
      elementType: Category, 
      relatedTo: {
        criteria: {
          elementType: Entry, 
          limit: null
        }
      }
    }
  ) {
    ...categoryFragment
  }
}
```

This may not be the perfect schema, but I’d love to get your thoughts on how we could approach this.

brandonkelly:
> This is performant because we’re not actually getting the entries in the `allEntriesQuery` here, we’re just limiting the `usedCategoriesQuery` by the subquery.

To clarify, when you pass an element query to `relatedTo`, it will execute the query’s `ids()` method internally – not include it as a subquery of the main query. So both of these lines are exactly the same, really:

```twig
{% set usedCategoriesQuery = craft.categories.relatedTo(allEntriesQuery) %}
{% set usedCategoriesQuery = craft.categories.relatedTo(allEntriesQuery.ids()) %}
```

So the only actual performance gain there could possibly be here would come from going from two HTTP requests down to one.

brandonkelly:
We’ve made this possible for Craft 3.6. Going forward you will be able to do this:

```graphql
query {
[...]

Issue #5143, opened 2019-10-22
Pass in slugs to GraphQL relatedTo argument


dangayle:
We need to be able to pass slugs as arguments in GraphQL requests made using the relatedTo query. 

This would be essential for things like having a url `/blog/cats` that makes a request for a list of blogposts related to the cats category. In this case, you won't know the id, so you would have to make two requests to get the list of categories first just to get the id.

dangayle:
Initially discussed here: https://github.com/craftcms/cms/issues/5071

andris-sevcenko:
GraphQL basically wraps around Element queries, so this would first have to be implemented in element queries. If this were a twig page, it would still require two separate element queries, right? Of course, there's the difference that it would happen on a single HTTP request, but the point still stands.

So, there are three ways of going forward:

1) do two requests (not optimal, but currently the only way out of the box)
2) implement matching elements by slugs in element queries (feels overkill-y to me, but it's just an opinion)
3) facilitate using the results of a GraphQL query in another GraphQL query in a single request.

My personal favorite is (3). Even though it's not trivial, this seems like the option that offers the most flexibility. But I can't promise it will happen for 3.4.

narration-sd:
@andris-sevcenko  This of course attracted my divergent thoughts tendencies, probably as it models a particular aspect not yet fully in mind yet about GraphQL. 

A couple of things found might be useful to your thinking?

- https://stackoverflow.com/a/47241014 (plus the further answer)
- https://stackoverflow.com/a/54203320/2113528

back to what I'm supposed to be doing...

denisyilmaz:
Hi @andris-sevcenko, i checked the Changelog in the 3.5 branch to see if this feature will be added there but could not find anything related to this issue here.

Is this feature still planed to be part of 3.5?

andris-sevcenko:
@denisyilmaz probably not. 3.5 (in terms of GraphQL) will be mainly a mutation-themed release.

This lack of urgency is because there _is_ a way of doing this out of the box by running two queries instead of one (one to fetch ids by slugs, the second one to fetch the relations by ids).

In the long term, I still want to implement the option (3) from the list above as that would add the flexibility of rolling two requests into one.

denisyilmaz:
@andris-sevcenko thanks for the quick reply. I got this to work in my project the way you described with the two queries. Its working but introduces a delay for page load around 50-100ms depending on server conditions because we have now two roundtrips. 

[...]

PR #7202, opened 2020-11-30
Feature/graphql relation arguments

andris-sevcenko:
### Description

This branch adds advanced relation input types to GraphQL API.

- Added `relatedAssets`, `relatedCategories`, `relatedEntries`, `relatedTags`, `relatedUsers` InputTypes.
- Possible Input Type argument values are the same as the matching GraphQL element queries have.
- It's possible to nest multiple relation arguments.

### Related issues

#7110 
#5143 
#6911


toddpadwick:
Hi @brandonkelly  @andris-sevcenko 
is there any documentation to how to use this new feature? I tried searching `relatedCategories` in the CraftCMS graphql documentation section but nothing came up.
I dont know what the query augument type should be for a category? am i passing in the whole category object? or just a slug? or an ID?

and if no categories are passed in, will it return all or none?
here is my code so far:
```
query($limit:Int, $relatedCategories[????????what goes in here???????????) {
  articles:entries(section: "articles", limit: $limit, relatedToCategories ) {
    postDate
    ... on articles_articles_Entry {
      title
      slug
      body
      description
      category {
        title
        slug
        url
      }
      author {
        name
        id
      }
    }
  }
}
```

andris-sevcenko:
> I tried searching relatedCategories in the CraftCMS graphql documentation section but nothing came up.

Oops, I'll look into that!

> I dont know what the query augument type should be for a category? am i passing in the whole category object? or just a slug? or an ID?

You pass in the same arguments as you would for a Category query, for example:
`relatedToCategories: [{slug: "mySlug"}]`
This would require a category to be related with the slug "mySlug". 

You can add more of the inputs there, but all of them must be satisfied, so:
`relatedToCategories: [{slug: "mySlug"}, {slug: "otherSlug"}]`
would require for two categories be related with the slugs above.


> and if no categories are passed in, will it return all or none
The argument acts as a filter. If the filter is empty, then the filter is not applied and all the entries are returned.

toddpadwick:
excellent. thanks, @andris-sevcenko . THis is very helpful.

PR #7207, opened 2020-11-30
Update TestSetup.php

brianjhanson:
Add `appicons` alias

### Description
While running some functional tests that involve the CP I got an error related to the `@appicons` alias. 

```                                                                                                                                         
  [Twig\Error\RuntimeError] An exception has been thrown during the rendering of a template ("Invalid path alias: @appicons/c-outline.svg").  
```

It looks like this alias just wasn't ported over to `TestSetup.php` when it was added in 8335098e548ca2fb4e90cea1cc0da189f676a141

### DIFF
--- a/src/test/TestSetup.php
+++ b/src/test/TestSetup.php
@@ -344,6 +344,7 @@ class TestSetup
         Craft::setAlias('@vendor', $vendorPath);
         Craft::setAlias('@lib', $libPath);
         Craft::setAlias('@craft', $srcPath);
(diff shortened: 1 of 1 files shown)


brianjhanson:
In case anyone else runs into this. Quick workaround is to add an `appicons` alias to your `tests/_craft/config/general.php` file.
```php
'aliases' => [
  'appicons' =>  './'
]
```

angrybrad:
Thanks!

brandonkelly:
Craft 3.5.17 is out now with a fix for this.

PR #7247, opened 2020-12-10
Use svg() for inline svgs

missmatsuko:
### Description

Sanitizes SVG so there are no duplicate IDs.

For valid HTML and [minor a11y improvement](https://dequeuniversity.com/rules/axe/4.0/duplicate-id).


brandonkelly:
Decided to just namespace each of the system’s SVG icon files instead, since these are trusted and we don’t need to actually sanitize them.

brandonkelly:
Reverted as I overlooked the possibility of including the same SVG multiple times on the same page.

Issue #7268, opened 2020-12-15
Login page accessibility


missmatsuko:
### Description

<img width="477" alt="Screenshot of the log-in form." src="https://user-images.githubusercontent.com/17884198/102152934-ee26c980-3e2a-11eb-8cb1-1f7859dd6c54.png">

Some ares for accessibility improvements on the CP login page (`/admin/login`):
- Text in notice overlay when cookies are disabled has insufficient contrast (should be fixed with https://github.com/craftcms/sass/pull/1)
- If it's convenient, pull out `h1` out of the `form` element. Elements in a form that aren't form inputs/buttons/labels can get missed in screen readers' form modes, but the page title would be announced when the page is first loaded anyways.
- When a site logo is uploaded, there should still be an `h1` element containing the site name. It should be the site logo image with the site name as its alt value nested in an `h1`.
- It would be more clear if the login page form fields had persistent visible labels. Since it's a log-in form, users generally likely expect to enter a username/email and password. Maybe can be something to redesign for 4.0 if not sooner. **Moved to: #7296**
- Required form fields are not announced to AT
- Password visibility toggle is not keyboard accessible 
- The "Forgot your password" link doesn't have an `href` value so it isn't focusable/usable with the keyboard. Change to a valid `a` tag or `button` element. I would go with `button` since the URL isn't changing. **Related: #7287**
- There should be some way to get back to the log-in form view after going to the reset password form view.
- The text should be updated for the forgot password form's submit button. It currently still says "Login". Should be changed to "Reset password" or similar.
- The "Craft CMS" image link has [insufficient contrast of 1.98:1](https://webaim.org/resources/contrastchecker/?fcolor=A2AAB4&bcolor=E4EDF6). **Related: #7289**
- [Don't disable the submit button](https://axesslab.com/disabled-buttons-suck). Instead, let the user submit the form and then display form errors if necessary.
- Use ARIA live regions to announce form errors and updates. **Related: #7288**

### Steps to reproduce
1. Go to `/admin/login`

### Additional info
- Craft version:  3.5.16

missmatsuko:
Closes #7288
Closes #7288

PR #7276, opened 2020-12-16
Login page accessibility improvements

missmatsuko:
### Description
- Addresses several items in #7268:
  - [x] Text in notice overlay when cookies are disabled has insufficient contrast (should be fixed with https://github.com/craftcms/sass/pull/1)
  - [x] If it's convenient, pull out `h1` out of the `form` element. Elements in a form that aren't form inputs/buttons/labels can get missed in screen readers' form modes, but the page title would be announced when the page is first loaded anyways.
  - [x] When a site logo is uploaded, there should still be an `h1` element containing the site name. It should be the site logo image with the site name as its alt value nested in an `h1`.
  - [ ] It would be more clear if the login page form fields had persistent visible labels. Since it's a log-in form, users generally likely expect to enter a username/email and password. Maybe can be something to redesign for 4.0 if not sooner. **Moved to: #7296**
  - [x] Required form fields are not announced to AT
  - [x] Password visibility toggle is not keyboard accessible
  - [x] The "Forgot your password" link doesn't have an `href` value so it isn't focusable/usable with the keyboard. Change to a valid `a` tag or `button` element. I would go with `button` since the URL isn't changing.
  - [x] There should be some way to get back to the log-in form view after going to the reset password form view.
  - [x] The text should be updated for the forgot password form's submit button. It currently still says "Login". Should be changed to "Reset password" or similar.
  - [x] The "Craft CMS" image link has [insufficient contrast of 1.98:1](https://webaim.org/resources/contrastchecker/?fcolor=A2AAB4&bcolor=E4EDF6).
  - [x] [Don't disable the submit button](https://axesslab.com/disabled-buttons-suck). Instead, let the user submit the form and then display form errors if necessary.
  - [x] Use ARIA live regions to announce form errors and updates (e.g. switching between the login and forgot password view).
- Cleans up login page code


missmatsuko:
Closes #7288
Closes #7288

brandonkelly:
Merged! Thanks so much @missmatsuko!

Issue #7280, opened 2020-12-16
The Garbage Collector delete all soft-deletes when run with --interactive=0 on commandline


caalholm:
### Description
When you run `./craft gc` on the commandline you get asked if you want to delete all thrashed or soft-deleted items, the default answer is [no]. If you suppress this question by passing `--interactive=0`, the script will delete all soft-deletes, even though they are not eligible according to softDeleteDuration!

The problem starts here, `$gc->deleteAllTrashed = ($this->deleteAllTrashed || $this->confirm('Delete all trashed items?'));`
```php
//src/console/controllers/GcController.php
   ...
    public function actionRun(): int
    {
        $gc = Craft::$app->getGc();
        $deleteAllTrashed = $gc->deleteAllTrashed;
        $gc->deleteAllTrashed = ($this->deleteAllTrashed || $this->confirm('Delete all trashed items?'));
        $this->stdout('Running garbage collection ... ', Console::FG_YELLOW);
        $gc->run(true);
        $this->stdout('done' . PHP_EOL, Console::FG_GREEN);
        $gc->deleteAllTrashed = $deleteAllTrashed;
        return ExitCode::OK;
    }
 ```
Since `$this->deleteAllTrashed` is false it tries to confirm it in  Yii\console\Controller
```php
//yii2/console/Controller.php
...
    public function confirm($message, $default = false)
    {
        if ($this->interactive) {
            return Console::confirm($message, $default);
        }

        return true;
    }
```
But since `$this->interactive` is false, the method return `true`, and all of a sudden, `$gc->deleteAllTrashed` is true.. 

### Steps to reproduce

1. Soft-delete an entry
2. run `./craft gc --interactive=0` or even `./craft gc --interactive=0 --delete-all-trashed=0`
3. The soft-deleted entry is hard-deleted, even if softDeleteDuration is set to default 2592000 seconds (30 days)

### Additional info

- Craft version: 3.5.16
- PHP version: 7.4

brandonkelly:
Thanks for reporting that! Just fixed for the next release.

brandonkelly:
Craft 3.5.17.1 is out now with that fix.

PR #7283, opened 2020-12-16
Check if command is run in interactive mode before asking for input

caalholm:
### Description
If the Garbage Collector is not run in interactive mode, non eligible sofe-deletes wil not be deleted

### Related issues
#7280


brandonkelly:
Doh, missed that you PR’d this. Thanks anyway!

PR #7302, opened 2020-12-20
Hide or override CSS icon content text for AT

missmatsuko:
### Description
The CSS `content` value used for icon fonts may be announced to screen readers.

I went through several instances of this and added aria-hidden where the icon was duplicating nearby text, and an overriding aria-label where it was standalone.

Also, I think the Position Select field type was deprecated, so not sure if its code is still needed in the repo.


brandonkelly:
Thanks!

PR #7315, opened 2020-12-22
Improve lightswitch field accessibility

### LINKED ISSUE #7313: Lightswitch field accessibility
### Description
Screenshot of 2 lightswitch fields.

- The on/off text should not be label elements. The label above the lightswitch field should be its only label element, or else it's unclear which label will be announced for the input.
- Add `aria-describedby` to describe how the checked status is related to the on/off labels. 
- The focus style is too subtle, but this should be a global fix.
- [the switch role doesn't support indeterminate/mixed values](https://www.w3.org/TR/wai-aria-1.1/#switch), so change to a checkbox which does
- The indeterminate/mixed style looks very similar to the off style:
  Screenshot of lightswitch field in an indeterminate/mixed state.

- Tangentially, the labels for the on/off fields on the field settings side seem like they shouldn't be uppercase:
  Screenshot of lightswitch field settings.

### Steps to reproduce
1. Look at a lightswitch field. E.g. "System Status" at `admin/settings/general`.

### Additional info
- Craft version: 3.5.17.1

missmatsuko:
### Description
Closes #7313


brandonkelly:
@missmatsuko Can you update `Craft.UI.createLightswitch()` as well?

https://github.com/craftcms/cms/blob/c4b999b325736968a05d3bcc1a5a2444dfbe9323/src/web/assets/cp/src/js/UI.js#L377-L433

(Used by fields’ “Required” settings within field layout designers.)

missmatsuko:
@brandonkelly Okay now that should be done!

PR #7316, opened 2020-12-22
Stop screen reader from announcing random periods

missmatsuko:
### Description
Change CSS `content: '.'` from periods to empty so the screenreader doesn't announce random periods.


brandonkelly:
@missmatsuko Did you verify that these pseudo elements are still visible as intended?

missmatsuko:
@brandonkelly Yep, I've gone through them and they look good.

brandonkelly (approved):

Issue #7303, opened 2020-12-20
Don't allow placeholders for field settings


missmatsuko:
### Description
Placeholders have a number of accessibility and usability issues. 

They're often misused in place of labels, instructions, and formatting hints.

I think it's very rare for them to be useful, and having this option available may encourage people to use them instead of the instructions field.

Alternative features that could be used instead of placeholder:
- Front-end validation (accept formatting rules, maybe with RegEx)
- Formatting on blur, as with [Politespace](https://www.filamentgroup.com/lab/politespace.html)
- [Formatting preview](https://adamsilver.io/articles/form-design-multiple-inputs-versus-one-input/#2.-to-help-users-check-their-answers)

Another accessibility boost with these methods is that we can automatically include formatting instructions.

Currently, for example, it's not clear that the URL fields only accept full URLs with protocols unless the developer has explicitly included this information themselves.

We may also be able to combine several plain text-like field types into 1 field type.

**Currently affected field types:**
- Email
- Plain Text
- URL
- Dropdown (with a valueless option containing text content, rather than a placeholder attribute)

**References:**
- [Smashing Magazine: Don’t Use The Placeholder Attribute](https://www.smashingmagazine.com/2018/06/placeholder-attribute/)
- [Nielsen Norman Group: Placeholders in Form Fields Are Harmful](https://www.nngroup.com/articles/form-design-placeholders/)
- [Deque: The Anatomy of Accessible Forms: The Problem with Placeholders](https://www.deque.com/blog/accessible-forms-the-problem-with-placeholders/)

Issue #5497, opened 2020-01-24
Mailto links should be allowed in URL fields


davist11:
### Description
`mailto:test@test.com` is a valid URL for `<input type="url">`, but they fail the validation in `craft\validators\UrlValidator`.

### Steps to reproduce

1. Add a URL field to a section
2. Enter `mailto:test@test.com` in the field and save.
3. See the error message `[fieldname] is not a valid URL.`

### Additional info

- Craft version: Craft Pro 3.3.20.1
- PHP version: 7.1.32
- Database driver & version: MySQL 5.5.5
- Plugins & versions:

<table class="data fullwidth fixed-layout" dir="ltr">
            <tbody>
                                    <tr>
                        <th class="light">Asset Rev</th>
                        <td>6.0.2</td>
                    </tr>
                                    <tr>
                        <th class="light">Azure Blob Remote Volume</th>
                        <td>1.0.2</td>
                    </tr>
                                    <tr>
                        <th class="light">Calendar</th>
                        <td>2.0.25</td>
                    </tr>
                                    <tr>
                        <th class="light">Navigation</th>
                        <td>1.2.7</td>
                    </tr>
                                    <tr>
                        <th class="light">Redactor</th>
                        <td>2.4.0</td>
                    </tr>
                                    <tr>
                        <th class="light">Redirect plugin for Craft 3</th>
                        <td>1.0.23</td>
                    </tr>
                                    <tr>
                        <th class="light">Super Table</th>
                        <td>2.3.3</td>
                    </tr>
                                    <tr>
                        <th class="light">Video Embed</th>
                        <td>1.2.1</td>
                    </tr>
                                    <tr>
                        <th class="light">VZ Address</th>
                        <td>2.0.2</td>
                    </tr>
                                    <tr>
                        <th class="light">Width Fieldtype</th>
                        <td>1.0.6</td>
                    </tr>
                            </tbody>
        </table>

am0:
Just giving this a bump as I had to switch a field from a URL field to a Plain Text field today so that users could use mailto: links.

nathansnelgrove:
Giving this a bump as well; I find I _frequently_ need to use Plain Text fields instead of url fields because of this issue.

brandonkelly:
This is resolved for the next 3.6 release ✨ 

URL fields now have an “Allowed URL Types” setting, with “Web page”, “Telephone”, and “Email” options:

[...]

PR #7374, opened 2021-01-11
Allowed URL Types setting

brandonkelly:
### Description

This PR adds a new “Allowed URL Types” setting to URL fields, with “Web page”, “Telephone”, and “Email” options:

The new “Allowed URL Types” setting

When 2+ options are checked, a dropdown will be displayed to the left of the URL input, which lists the options you’ve selected.

A URL field with the “Email” input type selected

Phone numbers will get a `tel:` prefix added; emails will get a `mailto:` prefix. But those will be hidden from the actual UI.

I also ended up removing the Placeholder setting, because we would have needed to add URL type-specific placeholder settings, which is the wrong direction we should be going (#7303).

### Related issues

#5497


src/fields/Url.php
```diff
@@ -95,29 +138,167 @@ public function getSettingsHtml()
...
+                    case self::TYPE_TEL:
+                        $value = str_replace(' ', '-', $value);
+                        $value = StringHelper::ensureLeft($value, 'tel:');
+                        break;
+                    case self::TYPE_EMAIL:
+                        $value = StringHelper::ensureLeft($value, 'mailto:');
+                        break;
+                    case self::TYPE_URL:
+                        if (!UrlHelper::isFullUrl($value)) {
+                            $value = StringHelper::ensureLeft($value, 'http://');
```
missmatsuko, line 161: Not sure, but would this be better off using `https://` or `//` instead of `http://`?
https://www.paulirish.com/2010/the-protocol-relative-url/

missmatsuko (commented):
- Left a note on `src/fields/Url.php` line 161 re: the URL protocol, not sure about it though.
- Looks like the `legend` element is getting a `for` attribute with the input's ID, but this is not necessary or valid HTML.
- I think it might be better to keep the text input's value even after changing the URL type, since someone may type the correct URL and then realize they need to switch the URL type after.
- I think it's a bit inconsistent that the saved data shows a full URI for website URLS but not email or telephone... but maybe most content authors wouldn't mind. Seems like it saves either way it's entered so that's good.

src/fields/Url.php
brandonkelly: Most sites that support SSL will automatically redirect to it, whereas there are plenty of sites that _only_ support `http` still, so I figured `http` is a better default.

brandonkelly:
> I think it's a bit inconsistent that the saved data shows a full URI for website URLS but not email or telephone... but maybe most content authors wouldn't mind. Seems like it saves either way it's entered so that's good.

I went a little back and forth on this, but figured the best AX would be to use `type="tel"` + `inputmode="tel"` (or `="email"`), which wouldn’t make sense if we were going to show the `tel:`/`mailto:` value prefixes.

brandonkelly:
> Looks like the `legend` element is getting a `for` attribute with the input's ID, but this is not necessary or valid HTML.

Fixed on the `3.6` branch via aeb15640e441125fd14995cc952f29ff80396728.

> I think it might be better to keep the text input's value even after changing the URL type, since someone may type the correct URL and then realize they need to switch the URL type after.

Agree. 3d600c1e387b13e5a8d31208a50431c4bcb9bca0

PR #7424, opened 2021-01-24
Relocate structure methods from category to structure service

timkelty:
Since nothing in here is specific for categories, this make reuse easier for a plugin wanting to use these methods for a non-category structure.


timkelty:
Opening new PR for 3.6 branch

PR #7426, opened 2021-01-24
Move structure methods from category services

timkelty:
Since nothing in here is specific for categories, this make reuse easier for a plugin wanting to use these methods for a non-category structure.


brandonkelly:
Thanks!

Issue #7360, opened 2021-01-07
Limit pending image transforms in GeneratePendingTransforms job


SHoogland:
### Description
When adding webp transforms to an existing site, it will trigger an enormous amount of pending image transforms (2000+)
When this job tries to transform all of them at once the job will most likely hit its timeout (which leaves imagemagick tmp files around)

Is it an idea to limit the query in this job to a maximum? or an option to disable generating images in the background as a temporary solution?

https://github.com/craftcms/cms/blob/35eafe4e66fa3de95fc1ed86d235e9a7dc290fcd/src/queue/jobs/GeneratePendingTransforms.php#L24
For example:
https://github.com/craftcms/cms/blob/35eafe4e66fa3de95fc1ed86d235e9a7dc290fcd/src/services/AssetTransforms.php#L953-L959
with a `->limit(50)`

### Steps to reproduce

1. Change your code to use new image sizes or formats (webp)
2. publish that code to an existing site with a lot of images and using srcsets
3. this will trigger 100s of pending image transforms
4. which the generate pending transforms jobs tries to generate all at once.

### Additional info

- Craft version: Craft Pro 3.5.17.1
- PHP version: 7.4.3
- Database driver & version: MySQL 8.0.22
- Image driver & version	Imagick 3.4.4 (ImageMagick 6.9.10-23)
- Plugins & versions:
CP Field Inspect	1.2.3
Entry Instructions	1.0.7
Environment protection	1.1.1
Field Manager	2.2.2
Position Fieldtype	1.0.16
Redactor	2.8.5
SEO	3.6.7
Super Table	2.6.5

brandonkelly:
Where are you seeing those ImageMagick temp files?

SHoogland:
![image](https://user-images.githubusercontent.com/2057831/104299315-5cf85080-54c5-11eb-8ecb-d9f0ed7d352c.png)

In the /tmp folder on our server

SHoogland:
But that is kind of besides the point... do we expect the transformJob to resize 2000+ images in (the default) 300 seconds.

brandonkelly:
Well typically there wouldn’t be… transforms only get added to that queue when their URL has been requested by a template. Do you know what is triggering these? Just a bunch of front-end web requests?

SHoogland:
Well, we do graphql queries like these to generate images for all the different screensizes including a webp copy:
[...]

PR #7430, opened 2021-01-25
 Added a limit to the pending image transform.

martijnvanonz:
### Description
Seeing the thread in #7360, there will be an option to stop the pending transform queue job to be fired. This will give problems with caching in twig templates. Because the cache is not saved of there are pending transforms. This will give problems if you are using source sets in your images or picture tags.

> Tip: The {% cache %} tag will detect if there are any ungenerated image transform URLs within it. If there are, it will hold off on caching the template until the next request, so those temporary image URLs won’t get cached. 

By adding the following code to the components in the config/app.php in combination with this pull request. The Job will only transform 10 images. If there are more images left the job will add a new job, until all the images are transformed.

`
 'assets' => [
   'generatePendingTransformsLimit' => 10,
 ],
`

### Related issues
#7360

### DIFF
--- a/src/queue/jobs/GeneratePendingTransforms.php
+++ b/src/queue/jobs/GeneratePendingTransforms.php
@@ -8,6 +8,7 @@
 namespace craft\queue\jobs;
 
 use Craft;
+use craft\helpers\Queue;
 use craft\queue\BaseJob;
 
 /**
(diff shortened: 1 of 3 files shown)


brandonkelly:
I don’t really care for the idea of limiting the queue job, especially if it’s just going to go and create a new job after it with the same limit set.

PR #7438, opened 2021-01-26
Install new plugins before deleting old plugins

### LINKED ISSUE #7436: Project Config - Install new plugins before uninstalling old plugins (to preserve data)
### Description

`New Plugin` is built to replace `Old Plugin`. During the installation process of `New Plugin`, it will

1. Configure everything it needs for the `New Plugin`.
2. Migrate all of the existing data between plugins.
3. Automatically uninstall the `Old Plugin`.

**This works great when you install the `New Plugin` normally.** And by "normally", I mean in an environment in which `allowAdminChanges` is enabled. Because when that value is disabled, you _must_ use Project Config to swap the plugins.

If the `New Plugin` were to be installed first (the desired behavior), it would have a chance to run the entire process as described above.

But if the `Old Plugin` gets uninstalled first (the current behavior), it deletes all of its relevant data before `New Plugin` has had a chance to migrate it.

I hope that's clear, I apologize for speaking in the abstract. The project is currently private, so feel free to DM me for details.

### Additional info

- Craft version: 3.5.18

lindseydiloreto:
Fixes #7436.

This patch is for Craft 3.6. I submitted a separate PR for 3.5.


brandonkelly:
Thanks!

lindseydiloreto:
Excellent, thanks Brandon! 🍺

PR #7449, opened 2021-01-27
Fix ArgumentCountError

thomasgoemaere:
### Description

Fix ArgumentCountError: Too few arguments to function craft\services\Structures::applyBranchLimitToElements(), 1 passed and exactly 2 expected


brandonkelly:
Thanks!

PR #7496, opened 2021-02-04
Fix bug in HTML::_namespaceAttributes to allow multiple url() calls in one line of CSS

manud99:
I had an issue with a SVG file where a CSS rule with multiple url() calls in one line. The rule didn't add any namespace at all to this line. Here is my line of CSS, that didn't work:

```css
.st4{mask:url(#mask-4_1_);fill-rule:evenodd;clip-rule:evenodd;fill:url(#Fill-6_1_);}
```

I modified the regular expression to disallow closing brackets in the ID. As far as I know you can't use brackets in IDs. With this PR both url() calls are handled by the regex callback separately.

### DIFF
--- a/src/helpers/Html.php
+++ b/src/helpers/Html.php
@@ -618,7 +618,7 @@ class Html extends \yii\helpers\Html
 
         // ID references in url() calls
         $html = preg_replace_callback(
-            "/(?<=url\\(#)[^'\"\s]*(?=\\))/i",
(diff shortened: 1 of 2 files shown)


brandonkelly:
…and you even added a test case for it! Thanks!!

brandonkelly:
Craft 3.6.4 is out now with that fix ✨ Thanks again!

PR #7505, opened 2021-02-04
Enable tracking changes on User element

d-pollard:
### Description

```
Event::on(User::class, User::EVENT_AFTER_SAVE, function (ModelEvent $event) {
    /** @var User $user */
    $user = $event->sender;

    Craft::info($user->getDirtyAttributes());
});
```

`getDirtyAttributes` now returns the dirty attributes; before it returned an empty array

### Related issues

https://craftcms.stackexchange.com/questions/37127/getdirtyattributes-not-showing-changes

### DIFF
--- a/src/elements/User.php
+++ b/src/elements/User.php
@@ -1388,7 +1388,7 @@ class User extends Element implements IdentityInterface
      * @inheritdoc
      * @throws Exception if reasons
      */
-    public function afterSave(bool $isNew): void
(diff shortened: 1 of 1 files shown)


brandonkelly:
Thanks for the PR!

brandonkelly:
Craft 3.6.4 is out now with that change ✨ Thanks again!

PR #7506, opened 2021-02-05
Add null check to URL field

boboldehampsink:
### Description
Adds a null check to `$value` before it goes into `StringHelper::removeLeft(string $value)`

### Related issues
Fixes error when the URL field is still empty on a fresh element, showing 
```craft\helpers\StringHelper::removeLeft(): Argument #1 ($str) must be of type string, null given, called in /app/user/vendor/craftcms/cms/src/fields/Url.php on line 224```


brandonkelly:
Thanks!

brandonkelly:
Craft 3.6.4 is out now with that fix ✨

PR #7680, opened 2021-03-14
Focus style a11y

### LINKED ISSUE #6886: Add visually prominent focus styles for all focusable elements
### Description
There doesn't seem to be a standard focus style for focusable elements.

It looks like anchors/links use the browser default focus style (some type of outline), and buttons have removed the browser default focus style for a more subtle one (slight background colour change).

For example, in this image, the left button is in focus and the right one is not:
Slightly dark gray button next to slightly lighter gray button.

Related: #7289

Some areas have focusable elements with no distinguishable focus style at all. For example, the sidebar navigation links:
Sidebar navigation link with current link styled differently, but no distinguishable style on the focused link.

Also the profile dropdown in the top corner (which is also getting cut off a bit by my scrollbar):
Profile icon button.

I think this issue has 2 parts:
1. Don't completely remove outline from focus styles, even if there is a custom focus style. [If we make the outline colour transparent instead of removing the outline entirely, the outline will be preserved and recoloured for Windows High Contrast Mode](https://sarahmhigley.com/writing/whcm-quick-tips/#1.-custom-focus-styles-%2B-a-transparent-outline).

2. The custom focus styles should be more visually prominent. Having a more consistent style across all focusable elements (e.g. anchors/links, buttons, and inputs) would make it easier to keep track of where focus is. 

### Steps to reproduce

1. Use keyboard to tab through interactable elements such as anchors/links, buttons, and inputs.

### Additional info

- Craft version: 3.5.11.1

missmatsuko:
### Description
Adds visible focus style to interactive elements such as links, buttons, and inputs.

I removed containerAttributes from the radio template because I didn't think it was used (couldn't find any attr blocks in the code base). 

Requires https://github.com/craftcms/sass/pull/5 and npm package update.

### Related issues
Closes #6886


src/templates/_includes/forms/radio.html
```diff
@@ -1,16 +1,13 @@
...
-    {%- set containerAttributes = containerAttributes|merge(('<div ' ~ block('attr') ~ '>')|parseAttr, recursive=true) %}
-{% endif %}
-
 {% set inputAttributes = {
     type: 'radio',
     id: id,
     class: (class ?? [])|explodeClass,
+       class: (class ?? [])|explodeClass|merge([
+        'radio'
+    ]|filter),
```
benface, line 10: Looks like there was a bad merge or something here; `class` appears twice in the hash.

brandonkelly: @benface Thanks for pointing that out! Fixed now (647fc1b5d6fe854c2bbf95cf9a9a1591489f03e7).

PR #7684, opened 2021-03-15
Fix infinite loop in setUniqueUri for entries  with duplicated long slugs

juban:
### Description
In situations where the entry slug or the generated one based on title is really long (255+), an infinite loop will occur if one try to save another entry with the same slug.

For instance, having an existing entry with slug
`arrêté-du-24-décembre-2020-portant-modification-de-larrêté-du-4-décembre-2020-fixant-la-liste-des-personnes-autorisées-à-exercer-en-france-la-profession-de-médecin-dans-la-spécialité-gériatrie-en-application-des-dispositions-de-larti`
Trying to publish a new entry with the same slug (or in real case, auto generated slug based on the title) will end up in an infinite loop issue.
This is because the setUniqueUri does not calculate the `overage` value based on the added suffix.
The fix prevent such case by chopping the overage amount including the length of the added suffix.

### DIFF
--- a/src/helpers/ElementHelper.php
+++ b/src/helpers/ElementHelper.php
@@ -82,8 +82,10 @@ class ElementHelper
         for ($i = 0; $i < $maxSlugIncrement; $i++) {
             $testSlug = $element->slug;
 
+            $suffixLength = 0;
             if ($i > 0) {
                 $testSlug .= $slugWordSeparator . $i;
+                $suffixLength = mb_strlen($slugWordSeparator . $i);
             }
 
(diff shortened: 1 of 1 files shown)


brandonkelly:
Thanks for reporting that, and for the PR! Took the fix in a different direction, but this is now resolved for the next release.

juban:
Cool. Thank you Brandon.

brandonkelly:
Craft 3.6.11 is out now with that fix. Thanks again for the help.

PR #7776, opened 2021-04-07
4.0

brandonkelly:
Docs:

- [ ] Update Twig references for Twig 3

Upgrade guide:

- [x] Cover [breaking changes in Twig 3](https://github.com/twigphp/Twig/blob/b13513a53b1a0660c7f467f1413f82eafb7e434f/CHANGELOG#L5)
- [ ] `UserPermissions::getAllPermissions()` and `getAssignablePermissions()` now return groups as sub-arrays with `heading` and `permissions` keys, rather than an array with `heading => permissions[]` pairs (d3d20ba6e4f145d8da78989372ec1e5103678b8e)
  - [ ] Also impacts `EVENT_REGISTER_PERMISSIONS`
- [ ] Cover deprecated/removed config settings, Twig variables, methods, etc. (refer to changelog)
- [x] Pre-4.0, the event for modifying fields on GraphQL types was fired when the type was created. In 4.0, it's fired when the fields are resolved. It would be advisable for applicable plugins to follow suit (5f759a59e263986a53c97272530441c4e9f05e21)


LarkAnspach:
Woke up

PR #7812, opened 2021-04-20
Add fix for 6140

### LINKED ISSUE #6140: Assets causing n+1 queries, even with eager-loading
### Description

Accessing asset URLs causes n+1 queries, even if you are eager-loading the assets and using the `withTransforms` param.

Reason is because assets are getting their folder path through `getFolder()->path` rather than using the `$folderPath` value already set on the assets. This happens in at least two places:

- `craft\helpers\Assets::generateUrl()` – called when no transform is applied:

  https://github.com/craftcms/cms/blob/a4e4ae5ce7b067f31d00c27770bc3e7c8434ba91/src/helpers/Assets.php#L88

- `craft\services\Assets::getAssetUrl()` in Craft 3.5 (f78ee107fa72d69f81b719968a4cc33f3eb562e2).

Possibly other places as well. We should check usages of `craft\models\VolumeFolder::$path` and see if there are any other places where `craft\elements\Asset::$folderPath` could be used instead.

### Additional info

- Craft version: 3.4.20, 3.5.x-dev

lenvanessen:
### Description
Fixed an edge case where asset-queries would still query folders for their paths, even though these are available on the asset model. Fixed #6140

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 - Fixed a bug where renaming an empty volume folder would create a subfolder inside it. ([#7721](https://github.com/craftcms/cms/issues/7721))
 - Fixed a bug where simultaneously-executed MySQL backups could result in a `my.cnf` conflict. ([#7801](https://github.com/craftcms/cms/issues/7801))
 - Fixed a bug where radio button labels weren’t including custom container attributes passed to the input.
-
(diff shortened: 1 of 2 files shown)


andris-sevcenko:
Appreciate it, @lenvanessen!

brandonkelly:
Craft 3.6.12.1 is out now with this fix. Thanks for the PR!

PR #7846, opened 2021-04-27
Save one query in Elements’ `_elementById()` when `$elementType` is `null`

benface:
Except when running Craft < 3.2.6.

### DIFF
--- a/src/services/Elements.php
+++ b/src/services/Elements.php
@@ -494,16 +494,29 @@ class Elements extends Component
         if (!$elementId) {
             return null;
         }
-
-        if ($elementType === null) {
-            $elementType = $this->_elementTypeById($property, $elementId);
-
-            if ($elementType === null) {
+        
+        try {
+            $data = (new Query())
+                ->select(['draftId', 'revisionId', 'type'])
+                ->from([Table::ELEMENTS])
+                ->where([$property => $elementId])
+                ->one();
+            
+            if (!$data) {
                 return null;
+            }
+            
+            if ($elementType === null) {
+                $elementType = $data['type'];
+            }
+        } catch (DbException $e) {
+            // Not on schema 3.2.6+ yet
+            if ($elementType === null) {
+                $elementType = $this->_elementTypeById($property, $elementId);
             }
         }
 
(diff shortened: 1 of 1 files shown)


brandonkelly:
What is the context for this? Can you describe what the issue is and what you’re trying to improve?

benface:
@brandonkelly I was just looking at this code to understand something about Craft's internals and I thought I saw an optimization opportunity, but maybe it's actually not? I needed to get an element by its id, but it can be any type, so I was under the impression that this would result in one less database query in that scenario.

src/services/Elements.php
```diff
@@ -494,16 +494,29 @@ private function _elementById(string $property, $elementId, string $elementType
         if (!$elementId) {
             return null;
         }
-
-        if ($elementType === null) {
-            $elementType = $this->_elementTypeById($property, $elementId);
```
benface, line 499: First query.

src/services/Elements.php
```diff
@@ -513,16 +526,6 @@ private function _elementById(string $property, $elementId, string $elementType
         $query->anyStatus();
 
         // Is this a draft/revision?
-        try {
-            $data = (new Query())
-                ->select(['draftId', 'revisionId'])
-                ->from([Table::ELEMENTS])
-                ->where([$property => $elementId])
-                ->one();
```
benface, line 521: Second query.

src/services/Elements.php
```diff
@@ -494,16 +494,29 @@ private function _elementById(string $property, $elementId, string $elementType
...
-            $elementType = $this->_elementTypeById($property, $elementId);
-
-            if ($elementType === null) {
+        
+        try {
+            $data = (new Query())
+                ->select(['draftId', 'revisionId', 'type'])
+                ->from([Table::ELEMENTS])
+                ->where([$property => $elementId])
+                ->one();
```
benface, line 503: Now just one query that handles both things.

brandonkelly:
Ahh, gotcha. Makes sense. Thanks for the PR!

brandonkelly:
Craft 3.6.13 is out with this change.

Issue #5661, opened 2020-02-17
Where is "save as draft" button ?


LouisCuvelier:
### Description

When I create an entry, there is only the button "Save". "Save as draft" only appears when I save the entry for the first time ?

Maybe I'm not understanding the logic. 🤷‍♂️
For the moment, to save a new entry but without publishing it, I have to disable the entry then save it.

As mentioned here #1021, there is an auto-save but where I can see all the drafts ?

### Steps to reproduce

1. Create a new entry
2. Save
3. See the button "Save as draft" appear

### Additional info

- Craft version: 3.4.5
- PHP version: 7.3.14
- Database driver & version: MySQL 5.5.5

jamiewade:
When you create an entry for the first time, it is a Draft **until you hit the save button**. You will see the Draft has been saved because of the green tick:

![Screenshot_2020-02-17 Create a new entry (First draft)](https://user-images.githubusercontent.com/6672331/74650517-d387b380-5179-11ea-96df-62f2ee867c5a.png)

To see all of your Drafts, you can go to the Entries page in Craft and choose Drafts from the dropdown:

![Screenshot_2020-02-17 Entries - Hand in Hand](https://user-images.githubusercontent.com/6672331/74650579-f31edc00-5179-11ea-97e0-6b468fcf1e0c.png)

Because of Craft's autosave functionality that was introduced in 3.2, if you save a Draft using the keyboard shortcut or the save button, that will then **publish the Draft** and put it live.

So, to create a new Draft without publishing it, you will need to follow these steps:

- Create your entry and give it a name
- Set the Enabled lightswitch to false

And that's it. As I mentioned, don't save the entry as this will take it out of a Draft state and into a live entry state. As you add/change content on the entry, you wil see the green tick icon swap between a loading spinner then back again, which is confirmation your content changes are being saved to the Draft.

Hope that helps!

LouisCuvelier:
Oh ok, this is clear now. I understand the logic.
Thank you for your clarifications !

LouisCuvelier:
Maybe I'm the only one, but I found that the "draft flow" is not obvious.
When I create an entry and it's a draft, the "Save" button should be "Publish", and the button "Save as draft" should be next to it. Like when I edit a published entry.

In my opinion, it's not instinctive to just quit the page without "Save as draft" because I feel that my draft is not saved due to the "Save" button right in the top corner. This is not giving a confident feeling.

brandonkelly:
[...]

PR #7899, opened 2021-05-07
Entry authoring workflow improvements

### LINKED ISSUE #6681: Don’t autocreate a draft in Live Preview
### Description

_**The "Preview" button will move you from "Current" to a "Draft" ,**_ 
Unlike developers, the content writers focus on one version only.
On 3.5, the "Preview" button will move you from "Current" to a "Draft".
and when they click "ctrl-S", this will save the draft instead of save the "current". 

This makes sense to developers that is familiar with git and branches,
but creates a lot of confusions to the content writers.

To content creaters, if they are working on "current", whatever they do should keep them at the "current" version,
and ctrl-s should save the changes to "current".  (like working on a word doc file)
If the content writer wants to create a draft to test changes, they should do it themselves via a "create draft" button.
(like duplicate the word doc).

Suggestion: 

When you create a new entry is disabled by default. The first edit (create) will create a "Draft 1". All edits are auto-save to the "Draft 1". Clicking "preview" button will  save the "Draft 1". To create another "Draft", they should click "Create a new draft" and they get "Draft 2". To publish changes to live, they should click "Publish" button. This should push the changes to live ("current"), and create a "Revision 1" record the current live version.

### LINKED ISSUE #7728: Draft vs. entry status
It's a bit confusing that a draft can be enabled or disabled. Status doesn't appear to matter, which tripped up a couple clients on the last update. Edit screen icon also looks like a broken page and doesn't match the other status icons.

brandonkelly:
Introduces a couple entry authoring workflow improvements.

## Unpublished drafts

When creating a new entry (unpublished draft), the available actions are now modeled after a normal entry rather than a draft.

- Nothing in the UI identifies the entry as a draft.
- The submit button is labelled “Save”.
- The Save button and <kbd>⌘</kbd><kbd>S</kbd> shortcut will publish the draft.
- Secondary actions include “Save and continue editing”, “Save and add another”, and “Delete entry”.

The top portion of an Edit Entry page when editing an unpublished draft

## Provisional drafts

When an author loads the Current revision of an entry and begins to make changes (whether in the main edit form or within Live Preview), a new **provisional draft** will be created _behind the scenes_ to store the WIP content.

Provisional drafts should generally go unnoticed. All buttons/keyboard shortcuts remain the same and appear to behave the same. The only visual differences between a normal entry and a provisional draft are:

- The revision menu’s button label will and in “— Edited”.
- The “Current” option within the revision menu will show a note about how unsaved changes are being shown, along with a button to discard them.

An expanded revision menu for a provisional draft

If an author leaves the page and returns to the entry later, the provisional draft will be loaded automatically (merged with any recent changes to the Current revision), and show blue status indicators next to any fields/attributes that have been modified to remind them what they were working on. At that point they can either continue their edits, discard them (via the revision menu), or click the “Create a draft” button to store the changes more permanently as a real draft.

Resolves #6681
Resolves #7728


mmikkel:
Nice work, this is certainly a huge improvement over the current workflow!

One thing:

> If an author leaves the page and returns to the entry later, the provisional draft will be loaded automatically (merged with any recent changes to the Current revision)

Since the current revision's content will be inaccessible to the author as long as a provisional draft exists, I feel like the _Discard changes_ button should be more visible/easily accessible. Maybe it should sit next to the revision dropdown, instead of inside it?

image

brandonkelly:
@mmikkel We tried both ways and felt it was better to be directly linked to the “Edited” label, along with that “Showing your unsaved changes” message. (Maybe we chose wrong, but we’ll wait and see how well it’s received by normal authors during the beta before making the final call.)

mmikkel:
Cool. I enjoy how consistent the new workflow is and I think always loading the provisional draft (if one exists) is the right call. But, I also suspect that in a lot situations any unsaved changes will be _intentionally_ unsaved, in which case it might be frustrating to some that discarding those unsaved changes isn't easier/more streamlined.

It's definitely going to be interesting to hear how real authors respond though.

LarkAnspach:
Yes I agree

LarkAnspach:
Pants or no pants?

sunscreem:
Oh no. I can't believe this is changing again after we've just spent hours and hours explaining it all to our clients.

brandonkelly:
@sunscreem think of all the hours you will save in the future, not explaining it to future clients!

jsunsawyer:
@brandonkelly Testing 3.7 and noticed that the updated "Unpublished drafts" workflow can be confusing when intentionally making a new post with the intent of working on it in the Draft state before publishing it.

In this instance, "Save" is a confusing option, as it takes the auto-created draft and publishes it. Not having a clear option to save a draft will lead many users to simply publish their work.

It might be preferred if there were multiple options here.
- `Save Draft` (or similar) so the user has a clear way of knowing that their work is saved
- `Publish` (or similar) so the user has a clear way of knowing that clicking this will publish their content

mmikkel:
> In this instance, "Save" is a confusing option, as it takes the auto-created draft and publishes it.

I don't agree. The UI/workflow in Craft 3.7 acknowledges that most authors don't consider their "new entries" to be "drafts", but rather things that are not-quite-real-yet or ephemeral, i.e. "unsaved changes". And the word "Save" works, because it implies that this ephemeral thing that the author is working on will be persisted, which is not necessarily the same thing as "publishing" it (a word that implies that the entry will go live on the site, which may or may not be the case).

> Not having a clear option to save a draft will lead many users to simply publish their work.

I do agree with this; it'd be nice to have a "Save as draft" option (which would convert the provisional draft to a "real" draft); either as the secondary action (i.e. gray button to the left of the red "Save" button, or inside the primary action dropdown.

jsunsawyer:
@mmikkel Let me explain why I think Save is confusing.

Here's a user's steps:

1. Create new entry
2. Begin writing content
3. Decides to stop and come back to it later
4. Attempts to save current work by clicking "Save" (even though it's already auto-saved as an unpublished draft)
5. Returns to entries list to realize that they just published their draft (or may not even notice and have incomplete content showing on the website)

I think there just needs to be a clear way for the user to one of two things. Save their current progress or publish their content, and it should be crystal clear what's happening.

mmikkel:
> I think there just needs to be a clear way for the user to do one of two things. Save their current progress or publish their content, and it should be crystal clear what's happening.

@jsunsawyer I don't disagree, and I think adding a secondary "Save as draft" or "Create draft" action to provisional drafts would be the perfect solution to the scenario you describe.

Part of the reason why 3.6 as so confusing to a lot of people, was that most authors don't consider their "new entry" as anything like a "draft". It's an unsaved entry; saving it will make it a saved entry, and if it's saved in an enabled state, that will effectively make it a "published" entry. This is how Craft worked for years, and I can't remember anyone being confused.



but I think the "Save" button works perfectly in 3.7, considering that most authors don't perceive their "new entries" as "drafts" but rather an "unsaved entry" – and that "saving" an unsaved entry in an enabled state will effectively "publish" that entry, is pretty intuitive. In my opinion, the best solution to the scenario you describe would be to add a "Save as draft" or "Create draft" option to provisional drafts, in addition to the primary "Save" action.

brandonkelly:
> Testing 3.7 and noticed that the updated "Unpublished drafts" workflow can be confusing when intentionally making a new post with the intent of working on it in the Draft state before publishing it.

This is exactly the use case we were trying to improve with 3.6 (see #5661).

The challenge is, when you click “New entry”, you are _already_ editing an (unpublished) draft. That’s exposed in two ways:

- The entry status in the metadata pane shows “Draft”.
- If you were to leave the page and go back to the Entries index page, it will be listed there with a draft icon.

So it’s not really clear what the expected behavior would be for a button labeled “Save as draft” or “Create a draft”, when you’re already editing a draft.

We “[solved](https://github.com/craftcms/cms/issues/5661#issuecomment-739084173)” this in 3.6 by making unpublished drafts’ action buttons more consistent with normal entry drafts – so the primary action was to save the draft, and publishing the draft was relegated to a secondary action. That turned out to be confusing as hell for most people though, as the primary action no longer matched the likely intent, which is to fully save the new entry they’re creating.

Maybe instead of “Save as draft” or “Create a draft”, we could add a button labeled “Done” or that just takes the author back to the Entries index page, which might feel safer than navigating away from the page. But I’m worried that would still add some level of confusion, where currently it’s very clear what the next step should be once you’re finished editing.

Plus, the author can always set the entry to disabled before pressing “Save”, which would avoid it being published on the front end.

jsunsawyer:
Well, we can just wait to see how it's used (or misused) then. People will probably get used to the new behavior, but I think that it will lead to some confusion.

Something about the act of simply leaving an entry that you're working on to leave it in its auto-saved draft state just feels unintuitive to me, and I think that some users might appreciate a clear action for peace of mind. Opening up a Craft 2 project, I see that new entries present a single "Save" option, then a few dropdown options that include "Save as a draft", which gives the user a clear action to ensure that their work is saved, but not ready to publish.

brandonkelly:
> Something about the act of simply leaving an entry that you're working on to leave it in its auto-saved draft state just feels unintuitive to me, and I think that some users might appreciate a clear action for peace of mind.

The UI is driving them toward saving the entry. I don’t expect many authors to know or care that they could technically leave the page as-is and let it continue to exist as an unpublished draft. If they aren’t ready to publish an entry, they can set its status to disabled before pressing Save.

> Opening up a Craft 2 project, I see that new entries present a single "Save" option, then a few dropdown options that include "Save as a draft", which gives the user a clear action to ensure that their work is saved, but not ready to publish.

Right, because entries didn’t start out as unpublished drafts until 3.2.

mmikkel:
> The UI is driving them toward saving the entry. I don’t expect many authors to know or care that they could technically leave the page as-is and let it continue to exist as an unpublished draft.

This. I think the concept of drafts is close to a power user feature, and that the big win in 3.7 is bringing back "Save (entry)" as the primary action (which in my experience is what makes the most sense for _most_ authors and use cases) whilst retaining and improving how _provisional_ drafts work. The UI is kept clean and intuitive, and for authors who do need to work with drafts in a deliberate way, explaining how these provisional drafts function should be manageable.

That said, although the provisional draft technically covers the use case, I do think a potential "Save (as) draft" button could have merit – mostly as a reassuring feature, enabling authors to deliberately convert provisional drafts to what would basically amount to "real" drafts without a current version attached. I guess these drafts – without a current entry – could basically use the 3.6 UI (i.e. with a primary "Save draft" action and a secondary action to move the draft to a current state), with the crucial difference being that the author deliberately created the draft from the provisional.

brandonkelly:
@mmikkel Couple clarifications:

- Entries start out as _unpublished_ drafts – which is how we refer to drafts that aren’t attached to a real entry yet.
- When you have unsaved changes to an entry, those are stored in a _provisional_ draft.

The old (pre-3.2) “Save as a draft” button would save your work as a real entry (set to disabled), and also create a draft for that entry, and take you to the draft. I don’t see how adding that complexity back into 3.7+ would add any value, when the thing you’re asking for (a “‘real’ draft without a current version attached”) is already exactly what an unpublished draft is. (Also keep in mind that unpublished drafts are no longer ephemeral as of 3.6, so there’s no harm in .)

I just made a couple changes that I think will help:

- The “Save” button is now relabelled to “Create entry” for unpublished drafts.
- The “Save” context menu now includes a “Save draft” action, which saves the work in progress to the draft and returns you to the Entries index page.

See #8094 for more details on those changes.

mmikkel:
Aha, thanks for clearing that up Brandon.

brandonkelly:
3.7.0-beta.3 is out with those changes ✨

jsunsawyer:
@brandonkelly I'm giving the most recent beta another go and am in agreement with @mmikkel's [concerns](https://github.com/craftcms/cms/pull/7899#issuecomment-835375224) about the provisional drafts.

Imagine a scenario where a user is simply testing out some content, but doesn't save the entry, and doesn't really intend to ever use these test changes.

Exiting the entry now leaves the entry in a provisional draft state for this user. There is no warning when they leave the page, and it may not be clear to the user that their test edits will persist upon their next edit.

Revisiting the entry that is in a provisional draft state doesn't feel much different than viewing an unedited entry.
@mmikkel outlines some related issues in [this comment](https://github.com/craftcms/cms/pull/7899#issuecomment-835531451).

**Improve exiting from provisional draft**

It may be nice if, when leaving an edited entry, the user was given the option to:
- save as draft
- discard changes

This would prevent users from coming back to a partially edited entry at a much later date, then unintentionally updating content (from their initial test) that had not been intended to be part of their new edits.

**Improve revisiting a provisional draft**

Provisional drafts may still exist, as the user may have never been presented an option to save as draft.

Users should be presented with a clear notice that they're working from previous, auto-saved edits.

This could be as simple as alerting the user when they revisit the entry that they're working from previous auto-saved changes – potentially with a timestamp of the last auto-save.

Perhaps this would be most beneficial if Craft were to first check for more recent changes than the latest auto-save, then give a clear warning that their auto-saved changes may conflict with a more recent version, presenting the user with options on how to proceed (create draft from current auto-save or discard changes to edit the current version). If there aren't any recent changes, then it'd be as if the user left off editing the latest version (though I still think it should be more clear that they may be working with previous, unpublished edits).

**Summary**

I fear that left as is, users will over-use provisional drafts as their primary way of working on edits for an entry, and they'll run into issues where they make unintended changes, either overriding more recent content, or having more recent content get pulled it and not make sense in the context of their updates. Pushing users toward real drafts for content authoring and making provisional draft states more clear to the user seems like it'd prevent a lot of potential issues.

There are probably other things I'm not considering, and my initial suggestions for improvements aren't too thought out, but hopefully we can work toward some improvements here.

MoritzLost:
Just to provide some feedback: User clients are also having trouble with the automatically created 'provisional' drafts. In particular, the following scenario is now a common occurance:

1. One editor changes *something* in an entry, realizes they don't want to save those changes right now and leave the page. A provisional draft has been created, even if they haven't noticed that.
2. Other editors edit the entry, creating new revisions.
3. Days or weeks later, the first editor returns to the entry and doesn't understand why they're seeing massively outdated versions, even though the website shows the current versions. It's not clear to them that they have to 'discard changes' in order to see the latest version.

This problem is exacerbated if the majority of the content modeling is done with a matrix field, in which case the provisional draft and the current version can be wildly different. It also doesn't help that sometimes, provisional drafts are created even if you haven't changed anything at all (see [this post with a screenrecording demonstrating this issue](https://github.com/craftcms/cms/issues/10106#issuecomment-973904724)).

Of course, this behaviour is understandable from a technical perspective and can be explained to clients. But I still see why it's confusing if you don't know how provisional drafts works.

I would suggest the following fixes:

1. Fix the bug where provisional drafts are created if you haven't changed anything (see the link above).
2. Display a more noticable warning when editing an entry where the provisional draft is based on a previous revision that is older than the current version of the entry.

darylknight:
Seconding this. I'm working on a site right now which is going to have multiple authors in multiple regions (multi-site) and it's critical that this isn't confusing or broken before the editors start working on it. I'm already finding during the build that when I'm the _only_ author, I go to edit an entry, and my edit page is filled with blue and yellow lines with no descriptions. I work with Craft every day, and I find it confusing what's going to happen if I save or discard.

**Issue 1:** Looking at the screenshot below, I have no idea what's current, or what will happen if I "discard changes". If I hover over the blue or yellow lines, the cursor becomes a question mark, but there's no text anywhere to describe what those lines mean. Clicking when the cursor is a question mark does nothing. Perhaps this is broken? Normally the question mark cursor is meant to indicate help text.

image

**Issue 2:** It's not clear enough when you land on the page that you're not looking at the "current" version of an entry. I've had several clients already ask why, when they go to edit a page, that the edit screen doesn't reflex what's on the website. They expect to click Edit, and change what they can see on the site right now, not what Craft saved (without telling them) 2 weeks ago.

brandonkelly:
> Clicking when the cursor is a question mark does nothing. Perhaps this is broken? Normally the question mark cursor is meant to indicate help text.

There is `title` text that will show up if you keep your cursor there for a second.

> It's not clear enough when you land on the page that you're not looking at the "current" version of an entry.

Yeah we’ve heard similar feedback – that the “Edited” label isn’t obvious enough.

PR #7981, opened 2021-05-26
Add buttons param to admintable

nfourtythree:
### Description
This PR adds the ability to specify secondary actions buttons that show in the toolbar for vue admin tables.

These secondary actions are designed to be simple button links. An example of this is to have a link to create a new record for the table.


brandonkelly:
If this is just for links, we should call it `links` rather than `secondaryActions`.

nfourtythree:
This was a debate I was having with myself. Was trying to keep it in line with other Craft naming. 

Also, wanted to leave it a little open to potentially supporting true action buttons at a later time.

brandonkelly:
I don’t think we should use the word “actions” unless it works similarly to [element actions](https://craftcms.com/docs/3.x/extend/element-action-types.html) – where they don’t even appear until you’ve started selecting rows, and they send a POST request off to a controller action defined by the action, along with a list of IDs/keys of the selected rows.

Even if you do get around to something like that, it probably wouldn’t be the right way to add a link that should always be visible.

PR #8005, opened 2021-06-01
Fix Guzzle config params not merged

bencroker:
Fixes https://github.com/craftcms/cms/commit/a81de4d50da933e5eeaa50e105f0e64d9a24d3a0, which incorrectly ignores config params when merging config settings.

### DIFF
--- a/src/Craft.php
+++ b/src/Craft.php
@@ -346,7 +346,8 @@ EOD;
         $guzzleConfig = static::$app->getConfig()->getConfigFromFile('guzzle');
 
         // Merge everything together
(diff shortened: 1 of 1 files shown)


brandonkelly:
Just tested locally and it’s working as expected on my end. `$config` is passed to `ArrayHelper::merge()` as the 3rd argument.

bencroker:
My bad, I saw that `ArrayHelper::merge()` accepted two arguments, but now that I look closer I see that it can accepted multiple arguments. Please disregard and sorry for the waste of time.

PR #8006, opened 2021-06-01
Added preferSite argument to Element Arguments

cornedor:
### Description

Adds preferSite to the ElementArguments.

For example, when querying siteA and siteB for the homepage, you could prefer siteB like this:
```
query Entry {
  entry(
    uri: "__home__", 
    site: ["siteA", "siteB"], 
    unique: true,
    preferSites: ["siteB"]
  ) {
   id
  }
}

```

Only the type definition is added since the rest is already part of craft's Models.


andris-sevcenko:
Thank you! Tweaked it a little bit and changed the target for Craft 3.7.

PR #8016, opened 2021-06-03
Adding Ability for GQL to parse dates as System Timezone Defaults

michtio:
### Description

This PR adds a simple small addition:

Currently when parsing a date through a GQL query, the timezone will ALWAYS be UTC unless you specifically set the directive `@formatDateTime(timezone: "Europe/London")` which would then apply the `Europe/London` timezone.

This can get quite cumbersome when working with a lot of dates in a project and queries relying on these dates. 
Not all projects need timezone support, not all headless projects need timezone support.

So it would be nice, that the default parsing of the DateTime could be set to the system time zone when returning data.

--> in my general settings of the site I have set `"Europe/London"` , I want all my dates returned by GQL to follow that specific timezone. 

In this PR I have added a general config setting ( because it's global for GQL in this case ) `$enableGraphQlSystemTimezone`, this has by default been set to false, to not break any installs relying on UTC.

In the `FormatDateTime` directive a new method has been created that will determine if this setting is set and will return the system timezone or the default timezone set:

```
    /**
     * @inheritdoc
     */
    public static function defaultTimezone(): string
    {
        return Craft::$app->getConfig()->getGeneral()->enableGraphQlSystemTimezone ? Craft::$app->getTimezone() : self::DEFAULT_TIMEZONE;
    }
```
    
The instances using `self::DEFAULT_TIMEZONE` have been replace by `self::defaultTimezone()`
    
 Always open for any other ideas or suggestions!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Release Notes for Craft CMS 3.x
 
+## 3.7.0 - Unreleased
+- Added the `$enableGraphQlSystemTimezone` setting, when set to true GQL DateTime types will by default return dates in the set System Timezone, when set to false it will return dates as UTC
+
 ## 3.6.16 - 2021-06-01
 
 ### Added
--- a/src/config/GeneralConfig.php
+++ b/src/config/GeneralConfig.php
@@ -657,6 +657,17 @@ class GeneralConfig extends BaseObject
     public $enableGraphQlCaching = true;
 
     /**
+     * @var bool Whether Craft should use the system date for default GraphQL dates.
+     *
+     * If set to `true`, Craft will use the system timezone when returning DateTime Objects in favor of UTC. 
+     *
+     *
+     * @since 3.7.0
+     * @group GraphQL
+     */
+    public $enableGraphQlSystemTimezone = false;
+
+    /**
      * @var bool Whether to enable Craft’s template `{% cache %}` tag on a global basis.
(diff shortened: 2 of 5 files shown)


brandonkelly:
Thanks for the PR! We’ll review and get this pulled in for 3.7.

toyflish:
on craft 7.3.9  I have an date field configured as datetime that is not required, if it is left empty graphql errors. Is that connected to the update?

```
{
  "errors": [
    {
      "debugMessage": "Call to a member function setTimeZone() on null",
      "message": "Internal server error",
      "extensions": {
        "category": "internal"
      },
      "file": "/home/site/wwwroot/vendor/craftcms/cms/src/fields/Date.php",
      "line": 405,
      "trace": [
        ....
```

toyflish:
when I enable "show time zone" for that field,  it does not fail any more.

andris-sevcenko:
Connected to _a_ update in 3.7.9. Not this one, though. Lemme push a fix real quick.

andris-sevcenko:
Just pushed https://github.com/craftcms/cms/commit/befd5fc4c4844c3eb97ae793cfc6982f18746686 that fixes this.

To get the fix early, change your `craftcms/cms` requirement in composer.json to:

```json
"require": {
  "craftcms/cms": "dev-develop#befd5fc4c4844c3eb97ae793cfc6982f18746686 as 3.7.9",
  "...": "..."
}
```

Then run `composer update`.

Issue #6622, opened 2020-08-13
Consider showing UI elements/tabs in live preview / HUD modals


wsydney76:
### Description

Great feature to hide a title when the context is obvious by tab or heading, however this context is gone when using live preview;  the input is orphaned, the user lost.

In general we would like to (optionally?) see tab names and ui elements also in preview, as it can be a long unstructured form, difficult to orientate yourself, and missing useful tips and warnings.

brandonkelly:
Related: #6587

mmikkel:
Really hope to see this at some point soon-ish. 

IMO the big problem with Preview has always been that the author loses all edit form context and focus when it opens. Having tabs and UI elements inside the Preview panel would be super helpful in this regard (especially if Craft was able to retain the currently active tab and/or focused field from the edit form), but the ideal solution would be if there was a way to open the preview panel without actually replacing the actual edit form.

brandonkelly:
This is halfway resolved for 3.7 – we’ve solved it for element editors, but not Live Preview yet. See #8017.

PR #8017, opened 2021-06-04
Slideouts

### LINKED ISSUE #3448: Feature Request: Add button to open currently edited element in inline element editor in a new tab
### Description
The inline element editor when you double-click a relation field is no-doubt a huge benefit; however, it can actually be a bit of a sore spot if a client is trying edit a particularly complicated field layout in the inline editor and feeling constrained. Adding a button to the top of the editor to open the element edit page in a new tab seems like an easy and reasonable solution. Of course, the inline editor would need to close when you click this link to prevent overwriting content when they're done. Ideally, this window would automatically close when the element is saved. I'm imaging just utilizing the UI for the title bar shown when in multi-site/language mode and adding an edit icon.

### Additional info

- Craft version: 3.x

### LINKED ISSUE #4129: [Feature request] Ability to resize HUD modal for new entries
### Description

When adding a new entry in a second-level HUD modal (I don't know what to call it - see the picture), the form is capped at `max-width: 600px`. When a user has the screen real estate available, it's more comfortable to have a wider area for editing, especially when plugins have wonky styling.

Image 2019-04-11 at 4 42 03 PM

Workaround I may try in the meantime, using the Control Panel CSS plugin:
```css
.hud .elementeditor {
  max-width: 100%;
}
```

I'm playing around with `resize: horizontal`, but haven't found the right application yet. 

### Steps to reproduce

1. Open a related entry field
2. Add a new element from there

### LINKED ISSUE #5857: Making Element Link Plugin part of core?
### Description
Just a question - would you consider making this little plugin <https://github.com/marionnewlevant/craft-element_link> as part of Craft core?
https://github.com/marionnewlevant/craft-element_link

Simply it creates a link the element if clicked.
it works in conjunction with crafts native double click for in situ popup editor for element links.

Screenshot 2020-03-29 at 13 46 58

### LINKED ISSUE #6247: [FR] Include slug to entry HUD edit window
### Description

One of the most popular features in Craft's CP is to add entries from a relationshop field. However if you use this in a multi lingual environment, you can switch to another language to translate the created entry, but you can't edit (clear) the slug in the HUD window. 

image

So you have to take additional steps, go to the element index, find the entry, open it , edit the slug, save, disrupting the workflow.

So just adding the slug would avoid these additional steps.

/Mel

### Steps to reproduce

1. Set up a multilingual system
2. Add a relational field
3. Add a new entry via HUD, save => a slug is created for all sites
4. Reopen HUD, switch to another language, translate title, save => the slug from step 3 persists

### Additional info

- Craft version: 3.5.x-dev
- PHP version:
- Database driver & version:
- Plugins & versions:

### LINKED ISSUE #6587: Remove meta class from HUD
At the moment the HUD uses the 'meta' styling for the fields. Whilst this make sense in theory (it's narrow, the meta box is narrow) in reality it results in a fairly messy layout with fields much narrower than they need to be.

With a just few minor changes the usability can be much improved:
1. Remove the `.meta` class
2. Add a little padding 
```scss
.hud .elementeditor .main { 
  padding: 0 24px 24px 
}
.preview-thumb-container.editable {
  margin-bottom: 24px;
}
```
Adjusting the width to be proportional with a larger maximum (if needed at all) would solve a couple of other open issues:
```scss
.hud .elementeditor {
  width: 50vw;
  min-width: 320px;
  max-width: 800px; // Or no max?
}
```
Before
before

After
after

Would solve #4129 I think.
Solves #5948 if editing via the HUD but not asset entry.

I'm not sure if there any other contexts of the HUD that I haven't considered.

### LINKED ISSUE #6647: Expanded HUD view
### Description

The current HUD when inline editing can get a little cramped and is missing UI elements such as tabs along with sidebar meta information / fields. So would like to propose replacing the inline HUD with an expanded view (full element edit page) shamelessly ~~stealing~~ borrowing from [Statamic V3 stacks feature](https://v3.statamic.com/blog/stacks-on-stacks-on-stacks).

This would solve most of the current HUD related issues including but not limited to https://github.com/craftcms/cms/issues/6622, https://github.com/craftcms/cms/issues/6587, https://github.com/craftcms/cms/issues/4129.

### LINKED ISSUE #6918: Please add a way to view the full location of an asset in an assets-field
### Description

I find myself wanting to know the full location of assets several times now from within the assets field of an entry, but there seems to be no way to see it from there.

When hovering over an image inside an asset field we only see a filename.
When doubleclicking on the asset we still only see the filename.

**Suggestion**
Show the full location on both the hover over an asset as well as a readonly label in the editform that appears when doubleclicking the asset.

brandonkelly:
This PR replaces the element editor HUDs with slideouts:

Screenshot of a new element editor slideout

The slideouts offer the following benefits:

- They’re bigger. (#4129, #6587, #6647)
- They render the element’s full field layout form, including tabs and custom UI elements. (#6622, #6647)
- They have nested sidebars that contain meta fields (Slug, Post Date, etc.) and metadata. (#6247, #6918)
- They contain a link to open the element’s dedicated edit page in a new browser tab. (#3448, #5857)
- They’re more mobile friendly.

Slideouts can also be nested, and parent slideouts will automatically scoot over to the left, making it stay oriented.

Screenshot of a nested element editor slideout

On mobile, slideouts will take up the full viewport:

Screenshot of a slideout on iOS

Invoking an element editor is identical to before:

```js
const editor = Craft.createElementEditor('craft\\elements\\Entry', $myElement, settings);
```

Slideouts have been implemented as their own concept, and can be used for other applications as well. Working with them is  similar to modals and HUDs:

```js
const slideout = new Craft.Slideout(contents, settings);
```

The following settings are available:

- `containerElement` – The type of element that should be used for the container (default is `div`).
- `containerAttributes` – Any attributes that should be added to the container element.
- `autoOpen` – Whether the slideout should be opened automatically on instantiation (default is `true`).
- `closeOnEsc` – Whether the slideout should be closed when the <kbd>Esc</kbd> key is pressed (default is `true`).
- `closeOnShadeClick` – Whether the slideout should be closed when the surrounding area is clicked on (default is `true`).

Finally, the PR adds two new helper JS methods:

- `Craft.trapFocusWithin()` – Prevents focus from leaving a given container via the <kbd>Tab</kbd> key.
- `Craft.setFocusWithin()` – Focuses on the first focusable element within a given container.

Resolves #3448
Resolves #6587
Resolves #4129
Resolves #6647
Resolves #6247
Resolves #5857
Resolves #6918


Mosnar:
@brandonkelly I tried this out locally and it works and feels really great and refined. The only thing I noticed was that the slide-out animation missed a few frames and doesn't feel super smooth. Have you looked at adding `will-change: left;` to the slide-out panel right before it opens?

brandonkelly:
@Mosnar Just added that. Doesn’t seem like it’s making any difference for me, but YMMV. 8a1939ea1e137958e2a04a0178d353e193db16a6

Issue #8014, opened 2021-06-03
Add users/create console command


timkelty:
### Description

Would be especially helpful for project onboarding, when a dev may have a database dump and local running site, but no Craft user/CP access.

- [x] Should include a "Make them an admin?" prompt
- [x] Possibly an --admin flag to bypass
- [x] Non-interactive options?

leevigraham:
@timkelty Any chance of adding a users/delete command in this PR? I would love a create / delete combo for automated testing.

leevigraham:
@timkelty Maybe something like:

```php
<?php


namespace modules\site\console\controllers;


use Craft;
use craft\console\Controller;
use craft\helpers\Console;
use yii\console\ExitCode;

class UsersController extends Controller
{
    public ?string $transferTo = null;

    /**
     * @inheritdoc
     */
    public function options($actionID)
    {
        $options = parent::options($actionID);

        if ($actionID === 'delete') {
            $options[] = 'transferTo';
        }

        return $options;
    }

    public function actionDelete(string $usernameOrEmail): int
    {
        $user = Craft::$app->getUsers()->getUserByUsernameOrEmail($usernameOrEmail);

        if (!$user) {
            $this->stdout("No user exists with a username/email of “{$usernameOrEmail}”." . PHP_EOL);
            $this->stdout('done' . PHP_EOL, Console::FG_GREEN);
            return ExitCode::OK;
        }

        // Are we transferring the user's content to a different user?
        if($this->transferTo) {
            $transferToUser = Craft::$app->getUsers()->getUserByUsernameOrEmail($this->transferTo);
            if (!$transferToUser) {
                $this->stderr("No user (transfer to) exists with a username/email of “{$this->transferTo}”." . PHP_EOL, Console::FG_RED);
                return ExitCode::UNSPECIFIED_ERROR;
            }
            $user->inheritorOnDelete = $transferToUser;
        }

        if (!Craft::$app->getElements()->deleteElement($user)) {
            $this->stderr(Craft::t('app', 'Couldn’t delete the user.'));
            return ExitCode::UNSPECIFIED_ERROR;
        }

        $this->stdout('done' . PHP_EOL, Console::FG_GREEN);
        return ExitCode::OK;
    }
}
```

timkelty:
> I would love a create / delete combo for automated testing.

@leevigraham just to clarify, the use case you're describing is _not_ testing via unit test (eg phpunit/codeception), right?

leevigraham:
[...]

PR #8029, opened 2021-06-08
Feature/create user

timkelty:
### Description
- Similar to the `install/craft`, this creates an active/verified user. If we want to add options to make the email unverified and send an verification email, we certainly can…just figured I'd start here.
- Things get a little awkward with the inherited `interactive` option (same with `set-password`), as `passwordPrompt` doesn't check that prop the other prompt methods do… perhaps it should?
    - Alternatively, we could just remove the `interactive` option for these methods.

### Related issues
- Resolves https://github.com/craftcms/cms/issues/8014


src/console/controllers/UsersController.php
```diff
@@ -123,4 +201,28 @@ public function actionSetPassword(string $usernameOrEmail): int
...
+            if (!$model->validate([$attribute])) {
+                $error = $model->getFirstError($attribute);
+
+                return false;
+            }
+            $error = null;
+
+            return true;
+        };
+    }
```
timkelty, line 227: @brandonkelly If you're ok with this approach, we could move this somewhere reusable for [use elsewhere](https://github.com/craftcms/cms/blob/develop/src/console/controllers/InstallController.php#L272-L293).

src/console/controllers/UsersController.php
```diff
@@ -81,6 +106,66 @@ public function actionListAdmins(): int
...
+
+        $user->admin = $this->admin ?? $this->confirm('Make this user an admin?', $this->admin);
+
+        if ($this->password) {
+            $user->newPassword = $this->password;
+        } elseif ($this->interactive) {
+            $user->newPassword = $this->passwordPrompt([
+                'validator' => $this->_createInputValidator($user, 'newPassword'),
+            ]);
+        }
```
timkelty, line 155: Checking `$this->interactive`, as `$this->passwordPrompt` doesn't check for `$this->interactive` like prompt/confirm do… but maybe it should, just to avoid awkwardness?

leevigraham:
@timkelty Please consider adding an array of group ids to assign the user to when created.

timkelty:
> @timkelty Please consider adding an array of group ids to assign the user to when created.

@leevigraham Added `--groups` and --groupIds`

timkelty:
@brandonkelly quick summary:

## `users/create`
- should work as expected with `--interactive=0`
- setting a password is optional
- `--groups` and `--groupIds` are set via options only (no prompt)

## `users/delete`
- should work as expected with `--interactive=0`
- should never delete anything without a confirmation
    - In lieu of a prompt, `--interactive=0` requires `--delete-content` or `--inheritor`

src/console/controllers/UsersController.php
timkelty: Looks like `yii\console\Controller::select` isn't `interactive` aware either, so I guess we can just leave as-is.

brandonkelly (approved):

PR #8056, opened 2021-06-14
What's New HUD

brandonkelly:
Adds the ability for Craft and plugins to register new feature announcements, which should be pushed to all control panel users.

The “What’s New” HUD in Craft 3.7

Once a user views an announcement, it will stick around for one additional week, and then dissappear forever.

## Pushing New Announcements

Announcements can be pushed using `craft\services\Announcements::push()`.

**This should be called from migrations, to ensure they’re only pushed once.**

```php
Craft::$app->announcements->push('Heading', 'Body text', 'plugin-handle');
```

The heading and body will both be HTML-encoded, however the body can contain Markdown formatting, so hyperlinks can be included in it, etc.

If the heading and body are translated, callback functions can be passed instead of strings, which wrap the `Craft::t()` calls:

```php
Craft::$app->announcements->push(
    function($language) {
        return Craft::t('plugin-handle', 'Heading', [], $language);
    },
    function($language) {
        return Craft::t('plugin-handle', 'Body text with [link]({url}).', [
            'url' => 'https://example.com/blog-post-with-more-details',
        ], $language);
    },
    'plugin-handle'
);
```


khalwat:
Suggestion: maybe just like Apple's Notifications, the icon of the thing registering the new item should be displayed alongside the title of the notification?

That'd make it clear whether the item was coming from Craft, or Commerce, or some other plugin.

brandonkelly:
Considered that, but in the case of iOS, that’s a personal device where you installed and manage every app yourself. In Craft’s case, most users will not be the same as the admins that manage the installed plugins, and may not actually know where core Craft ends and plugins begin. To them it’s just one big app. So not sure it’s as relevant.

OwenMelbz:
Is there anyway to turn this system off?

My concern is that (what you've briefly mentioned) is that most users have no requirement to know about these things, like Craft updates, or uncontrollable messaging that plugin authors can push to your clients.

E.g. We build a website for a client, white labelled, branded etc. Simple.

Then suddenly plugin authors start publishing announcements to our customers without our consent etc, which could contain anything - Advertisements about upgrading to premium versions etc, (Yes I know plugin authors can include any code they want regardless) but this gives prime real estate on every page for users to be presented with information that more than likely isn't relevant to them.

Not that it is a bad feature to have, completely support that. But as Craft has the ability to brand its Pro versions, having this leakage can be undesirable for many, so some config option to turn it off would be 👌 

Thanks :)

brandonkelly:
The idea is that it will only be used for things that most CP users _will_ actually care about, such as the two announcements we push out in Craft 3.7. If a plugin were to abuse it, I suspect that would correct itself pretty quickly due to feedback or people switching to a different plugin.

khalwat:
> Considered that, but in the case of iOS, that’s a personal device where you installed and manage every app yourself. In Craft’s case, most users will not be the same as the admins that manage the installed plugins, and may not actually know where core Craft ends and plugins begin. To them it’s just one big app. So not sure it’s as relevant.

Agree with part of this; to the user it is one big app. But having just implemented it, I felt it would be really weird not to give context that it was an improvement in a specific plugin:

Screen Shot 2021-08-17 at 7 21 58 PM

So I'm adding manually what I think would look and work better if it was built-in, with the plugin icon and maybe dimmed/grayed out text.

wbrowar:
A. this is awesome that you can hook into this!
B. I agree that some sort of icon might be helpful here. For plugins, but also to call out Craft’s core updates. Maybe something like widgets or CP section that can use `icon-mask.svg`?

brandonkelly:
Can you start a [new discussion](https://github.com/craftcms/cms/discussions/new?category=ideas) about it?

khalwat:
Done -> https://github.com/craftcms/cms/discussions/9747

peimansh:
When CP is opened for the very first time after installation/update to 3.7, the "new features" icon doesn't show up. It's on the next action (another CP page or refresh) that it shows up. Is this intentional or some bug or something wrong with my installation ?
First time I see a text "Pushing announcements" in the Queue section and when it is done the icon doesn't appear.

brandonkelly:
@peimansh Announcements are pushed out to each of the users with control panel access, via a queue job. So if that job hasn’t completed yet, then it’s expected the announcement wouldn’t be available to all users yet.

PR #8066, opened 2021-06-15
Use CLI input validator method

timkelty:
### Description
This uses the `createInputValidator` method introduced by https://github.com/craftcms/cms/pull/8029, and removes a bunch of methods that were doing the same thing.

Shouldn't change anything, just tidying up.

### Related issues
https://github.com/craftcms/cms/pull/8029


src/console/ControllerTrait.php
```diff
@@ -79,4 +80,28 @@ protected function outputCommand(string $command, bool $withScriptName = true)
...
+
+    /**
+     * Creates a function for the `validator` option of `Controller::prompt`.
+     *
+     * @param Model $model
+     * @param string $attribute
+     * @param string|null $error
+     * @return callable
+     */
+    protected function createInputValidator(Model $model, string $attribute, ?string &$error = null): callable
```
brandonkelly: @timkelty Looks like this `$error` attribute is unused by the method?

timkelty: Ah, yeah. That was to satisfy the nested pass by ref weirdness, but wasn't very clear or needed. Updated PR.

brandonkelly (approved):

PR #8069, opened 2021-06-16
fix m200620_230205_field_layout_changes migration

asallay:
Safely check if the column to drop / add already exists

### Description

The m200620_230205_field_layout_changes migration checks if the titleInstructions exists, but not in titleLabel, and also doesn't check for the 'elements' column before trying to add it.

This fixes the `Column already exists: 1060 Duplicate column name 'elements'` error when trying to upgrade directly from 3.4.x to 3.6.17.

### DIFF
--- a/src/migrations/m200620_230205_field_layout_changes.php
+++ b/src/migrations/m200620_230205_field_layout_changes.php
@@ -19,9 +19,15 @@ class m200620_230205_field_layout_changes extends Migration
      */
     public function safeUp()
     {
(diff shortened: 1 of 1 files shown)


brandonkelly:
It checks the `titleInstructions` column exists because that column was only briefly added during the Craft 3.5 development cycle, but then removed before Craft 3.5 was officially released. So the only people who would actually have the column are those who happened to be running Craft off of the `3.5` branch during that time, rather than sticking with released versions.

There’s no need to go through the same effort every time we drop a column.

PR #8077, opened 2021-06-17
Add `--hard` argument to `users/delete`

leevigraham:
Related issue: #8014


brandonkelly:
3.7.0-beta.2 is out now with a `--hard` action for `users/delete`.

PR #8081, opened 2021-06-17
Update UsersController.php

leevigraham:
Don't fail if there's no user found. Deleting a non-existing user has no negative side affects.

In this scenario it's assumed that if there's no user to delete that isn't an issue. CI tools would fail if this was called and the user was not found.

### Related issues

#8014

### DIFF
--- a/src/console/controllers/UsersController.php
+++ b/src/console/controllers/UsersController.php
@@ -232,8 +232,9 @@ class UsersController extends Controller
         $user = Craft::$app->getUsers()->getUserByUsernameOrEmail($usernameOrEmail);
 
         if (!$user) {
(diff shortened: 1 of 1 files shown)


timkelty:
🤔 hmmm still feels like an error to me, but I get your point about failing CI/CD processes…

As it stands, it is consistent with other controllers (eg users/set-password), so I'm hesitant to change it.
I also think someone else could make the argument expecting an error here, just as well.

I think for now we should keep it as is…but for your CI use-case, it seems like you should be able to just redirect the output to `/dev/null` to avoid getting hung up on the error, eg:

```
craft users/delete foobar > /dev/null 2>&1
```

The other thing we can consider is instead of explicitly returning `ExitCode::UNSPECIFIED_ERROR`, if we threw `\yii\console\Exception` instead, that would allow you to control the exit code via `--silent-exit-on-exception`.

Issue #7729, opened 2021-03-25
Email errors not in logs


bossanova808:
### Description

I have been switching all our emails from Mailgun to Postmark.

This all went very well and was simple.  However, after a few days, I noticed a particular group of messages was not being delivered.

Digging in to this, the only information from Craft was `false` coming back from `$message->send()` - there were no errors ([error] or [warning] etc) - anywhere in any of the Craft logs to be found.  No way whatsoever so see/surface the error, without hacking in to Craft or even Yii code, that I could see.  I spoke with Josh Crawford about it, and he's run in to this frustration as well.  

A quick look at `Mailer` suggests it will catch exceptions, but not that it will 'bubble up' API errors.  The only feedback to the developer is that `false` and it would be really super helpful (and, I presume, not all that hard to implement), if more was made available.  Just a `$message->getErrors()` type of thing...

Postmark don't have an API log, and doesn't show anything for messages not accepted either, so in the end I had to spend a couple of hours just using good ol' trail and error to work out the actual issue.  Apparently Postmark rejects messages with an empty (`""`) body (which is another issue I will talk to them about, as there are plenty of legitimate uses for emails with only a subject, and all the other email services accept these without complaint).

Anyway, would be great if it were more possible to see errors with the email service.

brandonkelly:
There should be _something_ in the logs, beginning with `Error sending email:` followed by the error message. We don’t log the stack trace as it could contain sensitive info:

https://github.com/craftcms/cms/blob/05f280380028c062efa5acfb42d35259804c6240/src/mail/Mailer.php#L179-L182

I guess the empty body thing is tricky, as that wouldn’t have been an issue when sending the test email. Not sure what more we could do about that though.

bossanova808:
Yeah I saw that, but that is only if there is an exception raised, right?   If 

`            return parent::send($message);`

...simply returns `false`, i.e. the API at the other end has raised an error but that doesn't trigger an exception....`false` is all that comes back, I think?  I didn't dig in deeply, but I can be sure that all I got was false, and there was definitely nothing the logs (believe me I checked a _bunch_ of times ;) - and again just now).

This is not for issues with sending email in the general sense, that's for sure working - it's just if there is a specific API error in some circumstance.  Not denying it's probably an edge case of course - but the errors are getting swallowed somewhere and ideally, they wouldn't be.

brandonkelly:
[...]

PR #8091, opened 2021-06-20
Add `error` and `exception` properties to Mailer messages

engram-design:
We've hit this a number of times in the past, but figured I should do something about it this time!

It's well-known that using the `Craft::$app->getMailer()->send($email)` function only returns a boolean for fail/success, which is a bit of a pain. Particularly in [Formie's](https://github.com/verbb/formie) case where we're smashed with issues on sending email notifications, commonly due to a misconfigured mailer, network issues, or anything related to specific mailers handling of emails.

In addition, we use [craft-log-to-file](https://github.com/putyourlightson/craft-log-to-file) to store all logs for the Formie plugin in a separate log file. This makes our support way, way more efficient.

I'm aware that Craft will save the error from the mailer in it's own `web.log` log, but this is tedious to get people to fish around in that for the error, when all other error information is included in our log file.

As email notifications are run in the queue, an error message is shown to CP users. At the moment, this can only be generic "Email failed", because we don't actively have the information as to _why_ the email failed to send. It'd be great if the error-handling you do:

https://github.com/craftcms/cms/blob/5bd4d4163f81156007bedc45138fe0f387089d6f/src/mail/Mailer.php#L174-L186

Could be stored somewhere, where plugins can access this, and report back to the user in a user-friendly way. I'm really just after the error message, but maybe not a bad idea to have the full exception there too, in case we want to display a full stack trace of the error.

Now we can use:

```
if (!Craft::$app->getMailer()->send($newEmail)) {
    Craft::dd($newEmail->error);
```

Long story short, please consider this 😄

Related to https://github.com/craftcms/cms/issues/7729


brandonkelly:
Thanks for the PR! Changed the base to `3.7` since at this point 3.6 is just on maintenance mode, only receiving bug fixes.

Couple things:

- We should just set the actual exception to an `$error` property (rather than `$exception` to be consistent with `yii\queue\ExecEvent::$error`). The exception message can be fetched from `$message->error->getMessage()`.
- Need to check if `$message` is an instance of `craft\mail\Message` before assuming it actually has an `$error` property.

Also keep in mind that Swift Mailer [swallows some exceptions](https://github.com/swiftmailer/swiftmailer/blob/15f7faf8508e04471f666633addacf54c0ab5933/lib/classes/Swift/Mailer.php#L70-L76), which is why we haven’t tried to make this easier in the first place.

engram-design:
Good call on the exception/error, a bit overkill with both. And yeah, annoying about those some of those errors, here's hoping it'll help a little. Updated!

brandonkelly:
Merged for the next 3.7 release!

brandonkelly:
3.7.0-beta.3 is out now with this change ✨

PR #9060, opened 2021-06-22
Improve `backup` phrasing

mattstein:
### Description

Another excruciating one-character PR because [_backup_ is a noun, but _back up_ is a verb](https://www.merriam-webster.com/dictionary/backup).

Revised in favor of [@Mosnar’s suggestion](https://github.com/craftcms/cms/pull/9060#issuecomment-866176476).

### DIFF
--- a/src/console/controllers/BackupTrait.php
+++ b/src/console/controllers/BackupTrait.php
@@ -75,7 +75,7 @@ trait BackupTrait
             return $backupOnUpdate;
         }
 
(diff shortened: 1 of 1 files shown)


Mosnar:
@mattstein This feels like one of those things that's "technically" correct but colloquially incorrect. What if you just changed the phrasing to "Create database backup?"

mattstein:
I disagree @Mosnar, but your suggestion avoids the issue and seems easier to read so you win.

PR #9066, opened 2021-06-23
Add getter method to WidgetTrait.

manud99:
Add a getter method for the widget's colspan. This way it is possible to define a min colspan for example.


brandonkelly:
Widgets must be able to support all colspan sizes, as mobile browsers only have the width for one column.

PR #9556, opened 2021-07-13
Replace getSourceId() with getCanonicalId()

bencroker:
Replaced deprecated `getSourceId()` method with `getCanonicalId()`. I've left the `sourceId` property names as they are because I assume they also need be changed in other places.


brandonkelly:
Thanks!!

brandonkelly:
Craft 3.7.1 is out now with this fix. Thanks again!

PR #9569, opened 2021-07-14
“Guides” → “Knowledge Base”

mattstein:
This updates outdated links to craftcms.com/guides to craftcms.com/knowledge-base, along with a single UI string that appears in the support widget:

support-widget

Not urgent since each of these links already redirects properly to its new equivalent.


brandonkelly:
:pray:

PR #9597, opened 2021-07-19
fix: running in console has no session for temp folder

phpsa:
### Description
When running in console mode and uploading an image, the `getUserTemporaryUploadFolder` method is causing an exception to be thrown, this breaks the feed-me plugin.

by checking if in console mode we now use a temp folder instead of a user folder.

### Related issues

* [ ] https://github.com/craftcms/feed-me/issues/863
* [ ] https://github.com/craftcms/feed-me/issues/642

### DIFF
--- a/src/services/Assets.php
+++ b/src/services/Assets.php
@@ -1020,6 +1020,9 @@ class Assets extends Component
 
         if ($user) {
             $folderName = 'user_' . $user->id;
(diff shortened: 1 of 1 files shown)


andris-sevcenko:
Appreciate it, thank you! Merged in for the next 3.7.x release.

brandonkelly:
3.7.4 is out now with this change. Thanks again!

PR #9651, opened 2021-07-28
FindAndReplace is now compatible with columnSuffixes

niektenhoopen:
It seems that FindAndReplace ignores the columnSuffixes (that were introduced in 3.7.x ?)

Personally, I think that the logic for generating the column name should be in the Field model but that's probably a different issue and not something for my first PR.

### DIFF
--- a/src/queue/jobs/FindAndReplace.php
+++ b/src/queue/jobs/FindAndReplace.php
@@ -90,6 +90,11 @@ class FindAndReplace extends BaseJob
             return;
         }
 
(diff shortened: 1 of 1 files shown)


brandonkelly:
Thanks for pointing out that bug, and the PR!

Took the fix in a different direction, taking advantage of `craft\helpers\ElementHelper::fieldColumn()` which handles determining the actual column name based on a given prefix/handle/suffix.

brandonkelly:
Just released Craft 3.7.7 with this change. Thanks again!

PR #9655, opened 2021-07-29
Console command doc cleanup

mattstein:
### Description

Minor update to pull in some notes in from the hand-maintained [Console Commands](https://craftcms.com/docs/3.x/console-commands.html) page and improve consistency throughout.

Also fixes an options bug in the project config controller where `reverse` [should be `invert`](https://github.com/craftcms/cms/pull/9655/files#diff-8ec7ff0286928126818362751a055f71751cae46785745197b86f07f8883b9b7R66).

Relates to [DOC-141](https://linear.app/craftcms/issue/DOC-141/improve-console-command-docblocks) and moves us closer to auto-generating this page.

### DIFF
--- a/src/console/controllers/GcController.php
+++ b/src/console/controllers/GcController.php
@@ -31,7 +31,7 @@ class GcController extends Controller
     /**
      * @var bool Whether all soft-deleted items should be deleted, rather than just
      * the ones that were deleted long enough ago to be ready for hard-deletion
-     * per the `softDeleteDuration` config setting.
+     * per the <config3:softDeleteDuration> config setting.
      */
     public $deleteAllTrashed = false;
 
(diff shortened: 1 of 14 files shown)


src/console/controllers/GcController.php
```diff
@@ -31,7 +31,7 @@ class GcController extends Controller
     /**
      * @var bool Whether all soft-deleted items should be deleted, rather than just
      * the ones that were deleted long enough ago to be ready for hard-deletion
-     * per the `softDeleteDuration` config setting.
+     * per the <config3:softDeleteDuration> config setting.
```
brandonkelly, line 34: We should avoid special link syntaxes in these docblocks because they won’t get parsed in `help` command output.

mattstein: Ah, that’s right. Will revise. 👍

brandonkelly:
🚀

PR #9673, opened 2021-08-03
fix double preposition in german translation

timrosskamp:
### Description

The preposition is included in `{timestamp}` and therefore appears twice in the control panel.


brandonkelly:
Thanks!

PR #9702, opened 2021-08-11
Feature/dev 13 multi factor authentication

andris-sevcenko:
Refactor how authentication works for Craft


dsmrt:
I'm trying to prepare for our `saml-sp` plugin to be updated for Craft 4.0. This PR is getting me excited to rework some of it. Do you think this is the direction 4.0 is going? Just trying to prepare. Let me know if there is anything I can help with here?

brandonkelly:
@dsmrt Yep we should have it wrapped up in the next couple weeks :)

PR #9706, opened 2021-08-11
Simplify site `enabled` label and explain disabled field

mattstein:
## Description

This PR reduces a site’s `enabled` field label from “Enable on the front end” to “Enabled”, since “on the front end” may or may not reflect what’s actually presented, and “Enabled” is a commonly-used term that’s applicable here.

### Current
  
current

### Proposed

proposed

**Bonus**  
This also adds an instruction **only for the primary site** to explicitly reinforce that it can’t be disabled:  
proposed-tip


brandonkelly (approved):

brandonkelly:
Thanks!

PR #9740, opened 2021-08-19
Remove `isNitro`

olivierbon:
Part 2 of https://github.com/craftcms/cms/commit/d405df13aede11af617b4f5e0cd92bcef32b1ca2

### DIFF
--- a/src/templates/_special/install/db.html
+++ b/src/templates/_special/install/db.html
@@ -28,7 +28,7 @@
                 {{ forms.textField({
                     label: "Server"|t('app'),
                     id: 'db-server',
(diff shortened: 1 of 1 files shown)


knynkwl:
Would this have anything to do with this error I just got with a fresh install?

` "craftcms/cms": "^3.6.4"`


Screen Shot 2021-08-19 at 12 16 59 PM

brandonkelly:
@knynkwl Yep. You can get passed it by running the `php craft setup` command. We’ll release an update today with this fix.

brandonkelly:
3.7.10 is out now with the fix.

PR #9755, opened 2021-08-25
Collections

brandonkelly:
### Description

- Adds the [Illuminate Collections](https://github.com/illuminate/collections) package
- Adds `craft\db\Query::collect()`, which returns the query results as a Collection
- Modifies `Element::getEagerLoadedElements()` to return eager-loaded elements as a Collection

As a result, templates now have a consistent way of interacting with elements, whether they were eager-loaded or queried directly: templates can just call `collect()` rather than `all()`. `collect()` will either end up referring to `Illuminate\Support\Collection::collect()` (if eager-loaded) or `craft\db\Query::collect()` (if not).

(Templates which are already looping over eager-loaded element arrays won’t break, because `Illuminate\Support\Collection` objects are traversable.)

### Related issues & discussions

- https://github.com/craftcms/cms/discussions/8475
- https://github.com/craftcms/cms/discussions/8513


michaelrog:
Curiously... why not have `all()` return a Collection? (except when `asArray()` is used)

brandonkelly:
Because I don’t want to deviate from the `yii\db\Query::all()` return type.

PR #9798, opened 2021-09-05
Method can return `false` if the data is tampered

bencroker:
Added a return type of `false` in case the data is tampered.

This is already in the parent method: 
https://github.com/yiisoft/yii2/blob/c86cfb66d37314bf9dff39b236697d55a9f23944/framework/base/Security.php#L441


brandonkelly:
Thanks!

PR #9833, opened 2021-09-14
Re-enable text scaling and zoom on mobile by removing maximum-scale =…

gcamacho079:
… 1 from meta viewport

### Description
Removed maximum scale from viewport settings to allow text scaling and zoom.

@brandonkelly hopefully this doesn't re-introduce the issue you were trying to solve. 🤞🏼 

### Related issues
N/A

### DIFF
--- a/src/templates/_layouts/base.html
+++ b/src/templates/_layouts/base.html
@@ -29,7 +29,7 @@
     <title>{{ docTitle ~ (docTitle|length and systemName|length ? ' - ') ~ systemName }}</title>
     {{ head() }}
     <meta name="referrer" content="origin-when-cross-origin">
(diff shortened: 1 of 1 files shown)


brandonkelly:
The goal with cd6e8e8bf011eca742d3901c972761d6dc2ba6cb was to disable _auto_ scaling on text inputs, without disabling the user’s ability to zoom if they really want to (per https://stackoverflow.com/a/46254706/1688568). Seems to be working on iOS as expected for me, but just noticed the second comment on that Stack Overflow answer says it breaks user zooming on Android. Oh well…

brandonkelly (approved):

gcamacho079:
@brandonkelly will do some digging and check this on Android

PR #9859, opened 2021-09-21
Fixed getTotalSelected() function name

bencroker:
Assuming this is what it should be...


brandonkelly:
Nope, that’s a valid JS getter function.

bencroker:
Well I'll be... that was intentional after all!!

PR #9862, opened 2021-09-21
Add `EVENT_PREP_QUERY_FOR_TABLE_ATTRIBUTE` Event for customizing Element Index Queries

AugustMiller:
### Description

This came up [in Discord](https://discord.com/channels/456442477667418113/456442884258922529/889907290784804934), and although it's not a direct solution for the developer's more radical reverse-relations problem, it seems like a worthwhile + lightweight addition, for those of us who are interesting in making Element Index views as useful and performant as possible.

The goal is to let each of the defined "table attributes" on a built-in (or Plugin-owned) Element have control over the query, _in addition to_ the HTML output. I've added the `EVENT_PREP_QUERY_FOR_TABLE_ATTRIBUTE` event, just after the Element has a chance to set its own (native) query arguments, but while still in the context of a specific table attribute.

It didn't seem worthwhile to have a before _and_ after event (like many others in the class) as applying values _before_ could get squashed by native behavior, anyway—despite many using the `->andWidth()` method.

From the new Event constant's doc-block, it basically allows this:

```php
use craft\elements\Entry;
use craft\events\PrepareElementQueryForTableAttributeEvent;
use craft\events\RegisterElementTableAttributesEvent;
use craft\events\SetElementTableAttributeHtmlEvent;
use craft\helpers\Cp;
use yii\base\Event;

Event::on(
    Entry::class,
    Entry::EVENT_REGISTER_TABLE_ATTRIBUTES,
    function (RegisterElementTableAttributesEvent $e) {
        $e->attributes[] = 'authorExpertise';
    });

Event::on(
    Entry::class,
    Entry::EVENT_PREP_QUERY_FOR_TABLE_ATTRIBUTE,
    function (PrepareElementQueryForTableAttributeEvent $e) {
        $query = $e->query;
        $attr = $e->attribute;

        if ($attr === 'authorExpertise') {
            $query->andWith(['author.areasOfExpertiseCategoryField']);
        }
    });

Event::on(
    Entry::class,
    Entry::EVENT_SET_TABLE_ATTRIBUTE_HTML,
    function (SetElementTableAttributeHtmlEvent $e) {
        $attribute = $e->attribute;

        if ($attribute !== 'authorExpertise') {
            return;
        }

        // The field data is eager-loaded!
        $author = $e->sender->getAuthor();
        $categories = $author->areasOfExpertiseCategoryField;

        $e->html = Cp::elementPreviewHtml($categories);
    });
```

### DIFF
--- a/src/base/Element.php
+++ b/src/base/Element.php
@@ -226,7 +226,7 @@ abstract class Element extends Component implements ElementInterface
      * 
      * Event::on(
      *     Entry::class,
-     *     Entry::EVENT_PREP_QUERY_FOR_TABLE_ATTRIBUTE,
+     *     Entry::EVENT_SET_TABLE_ATTRIBUTE_HTML,
      *     function (SetElementTableAttributeHtmlEvent $e) {
      *         $attribute = $e->attribute;
      * 
--- a/src/base/Element.php
+++ b/src/base/Element.php
@@ -26,6 +26,7 @@ use craft\events\DefineMetadataEvent;
 use craft\events\DefineValueEvent;
 use craft\events\ElementStructureEvent;
 use craft\events\ModelEvent;
+use craft\events\PrepareElementQueryForTableAttributeEvent;
 use craft\events\RegisterElementActionsEvent;
 use craft\events\RegisterElementDefaultTableAttributesEvent;
 use craft\events\RegisterElementExportersEvent;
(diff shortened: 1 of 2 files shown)


brandonkelly:
Good idea, thanks!

brandonkelly:
Craft 3.7.14 is out now with this new event.

PR #9882, opened 2021-09-27
Fix lacking Norwegian translation

eddiedale:
### Description
"Apply Draft" was not translated correctly.


brandonkelly:
Thanks for the PR! Unfortunately our translation service doesn’t offer Norwegian Bokmal, so it’s a bit behind.

If you would like to contribute to the NB translations in the future, you may wish to use [CrowdIn](https://crowdin.com/project/craft-cms), which is where all of our translations are managed.

PR #9899, opened 2021-09-29
Clarify self relation field instruction

mattstein:
### Description

The Users field type’s **Allow self relations** field instruction recently caused some confusion:

Screen Shot 2021-09-29 at 09 42 59 AM@2x

Particularly in the case of users, it could seem like human beings or permissions somehow relate to this setting:

> Whether users should be allowed to relate themselves.

This PR clarifies that we’re referring to elements—where singular form seems more natural—and adds a _to_ in order to make the phrase “relate themselves” less ambiguous by having “themselves” be the recipient of “to relate”:

> Whether user elements should be allowed to relate to themselves.

It works just as well with other element types:

> Whether asset elements should be allowed to relate to themselves.

> Whether category elements should be allowed to relate to themselves.

> Whether entry elements should be allowed to relate to themselves.

> Whether tag elements should be allowed to relate to themselves.

> Whether product elements should be allowed to relate to themselves.

> Whether product variant elements should be allowed to relate to themselves.


michaelrog:
Just curious — Is there a first-party preference about when to capitalize the names of element types?

I feel like most of the time, when I see them referred to as a proper class, it's capitalized, but not when referring to instances (and not in this PR)...? 

e.g. `This action will delete 34 entries.` vs `Whether User elements may be related to themselves`

It's a subtle thing, but I always wonder whether I'm doing it right, or if there is even "style guidance" on this...

brandonkelly:
@michaelrog We don’t consider them proper nouns, so keep them lowercase.

PR #9919, opened 2021-10-03
Impersonate Console Command

markhuot:
### Description
When jumping in to the project it’s sometimes helpful to be able to log in from the console without needing to share passwords or make DB changes to get in to the admin. This exposes a `./craft users/impersonate --id=1` command that will generate an impersonation URL.


michaelrog:
Great idea — What about loosening up the syntax, like this:

`./craft users/impersonate xxxx`

(If _xxxx_ `is_numeric`, try `getUserById()` first. Otherwise, try `getUserByUsernameOrEmail`.)

brandonkelly (approved):

brandonkelly:
Thanks! Added user ID support to `users/delete` + `users/set-password` commands, and went with the same `$user` ID/username/email argument on `actionImpersonate()` to be consistent.

brandonkelly:
Craft 3.7.15 is out now with that change.

PR #9943, opened 2021-10-07
Remove duplicated `Asset::KIND_VIDEO` values

hahnzilla:
### Description

Such a small thing, but noticed both `mp4` and `flv` were duplicated when doing some source diving today so figured I'd open a PR to remove them.

If you look at the line immediately preceding the ones removed you'll see the duplicated values.

### Related issues

nada


brandonkelly:
Whoops, thanks for that!

PR #9950, opened 2021-10-11
Add twig tests for is_countable, is_iterable, is_object

MoritzLost:
### Description

Craft 3.7.15 [adds more tests to check variables types](https://github.com/craftcms/cms/commit/fd4e546270f14ce1412d38a8b87a6a05e011bcb3) to the built-in Twig extension, like `is array`, `is callable` etc. Before this update, I used a site module to provide those tests, including a couple that aren't included in the core yet. This PR adds the following tests to the core:

- `is countable` (using [is_countable](https://www.php.net/manual/en/function.is-countable.php) with a fallback for PHP < 7.3)
- `is iterable` (using [is_iterable](https://www.php.net/manual/en/function.is-iterable.php))
- `is object` (using [is_object](https://www.php.net/manual/en/function.is-object.php))

Pretty tame addition, but useful in some scenarios.

---

Sidenote: My custom extension also includes filters for type-casting, which I use pretty frequently in my templates. Would this be a welcome addition to the core? Something like this:

```php
public function getFilters()
{
    return [
        new TwigFilter('array', function ($obj): array {
            return (array) $obj;
        }),
        new TwigFilter('bool', function ($obj): bool {
            return (bool) $obj;
        }),
        new TwigFilter('float', function ($obj): float {
            return (float) $obj;
        }),
        new TwigFilter('integer', function ($obj): int {
            return (int) $obj;
        }),
        new TwigFilter('string', function ($obj): string {
            return (string) $obj;
        }),
    ];
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,16 +3,23 @@
 ## Unreleased
 
 ### Added
-
+- Added the `provisional` argument for the remaining draft mutations via GraphQL. ([#9946](https://github.com/craftcms/cms/issues/9946))
+- Added the `notes` and `name` arguments to the create draft mutation via GraphQL.
 - Added the `is countable` Twig test.
 - Added the `is object` Twig test.
+- Added `craft\helpers\ElementHelper::isCanonical()`.
+- Added `craft\helpers\ElementHelper::isDerivative()`.
 
 ### Changed
 - Logs now include the raw request body in place of the `$_POST` array, if the `$_POST` array is empty. ([#9941](https://github.com/craftcms/cms/discussions/9941))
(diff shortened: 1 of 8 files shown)


MoritzLost:
@brandonkelly Oops, I missed that. Thanks for the hint, done!

What about the type-casting filters? Casting to integer, for instance, is often useful:

```twig
{% set categoryFilter = craft.app.request.getQueryParam('category') | integer %}
```

I'd be happy to add those filters to the PR if they're a welcome addition.

brandonkelly:
Merged, thanks!

Not sure how I feel about typecasting filters for now.

PR #9972, opened 2021-10-18
Disable highlighting in console requests when using `dd()`

bencroker:
This PR automatically disables code highlighting in console requests when `dd()` is called. This makes it easier to debug code using `dd()` regardless of whether the request is via web or console, without having to perform checks in our code. If a value for the `$highlight` parameter is already specified then the value is not changed.


brandonkelly:
Just released 3.7.17.2 with this change.

Issue #9999, opened 2021-10-22
Database lock-up due to huge transaction rollback


mattgrayisok:
### Description

We've been seeing a sporadic issue with, so far, 6 sites hosted on Servd in which the database locks up due to a huge transaction roll back. We spoke briefly to Brad on a support@ thread but I wanted to shift the issue over here as it seems to be a rare but general problem across multiple projects.

The first symptom we see is a database running with high CPU and elevated response rates. Slower reads and often fully blocked writes. 

Running `show engine innodb status` shows a transaction rollback containing millions of undo log entries:

`ROLLING BACK 101 lock struct(s), heap size 24696, 11025 row lock(s), undo log entries 2792119`

Upon filesystem investigation it seems as though the transaction ran until it filled the undo log to capacity, at which point mysql aborts the transaction and rolls back. (The roll back itself can take anywhere from 1 hour to 1 month to complete so we respond to this by just trashing and rebuilding the database entirely).

Once the rollback starts there's no sensible way for us to figure out what queries the original transaction contained, but we did manage to catch 2 of the projects while they were still applying the transaction, rather than rolling it back. Both of those were running queries similar to:

```
UPDATE `craft_structureelements` SET `rgt`=`rgt`-0, `dateUpdated`='2021-09-10 18:23:27' WHERE (`rgt` >= 344) AND (`root`=169781)
```

A couple of weird things:

- Notice the `-0`
- Although this and similar queries ran millions of times within the transaction, the dateUpdated was always current, not a fixed value, but it's possible the output I was looking at was expanding a NOW() or something. Not 100% sure if this matters.

On the most recent project to experience this we managed to get some relatively good logs from around the period in which the event occurred. It seems as though a `Pruning extra revisions` task was running and a control panel user saved several Entries at the same time.

Speculating: It feels like there might be a race condition or some sort of logic which triggers the prune task to drop into an infinite loop if the entries it's working on are edited during its processing.

Possibly related to #9905 although we can't fix by just restarting the database - the transaction rollback just resumes which is the expected behaviour.

### **Speculative** Steps to reproduce

1. Trigger a Prune revisions task
2. Edit the entries that the prune task is working on

### Additional info

- Craft version: We've only seen it with 3.7+
- PHP version: 7.4 / 8
- Database driver & version: MariaDB 10.5
- Plugins & versions: Varied

brandonkelly:
[...]

Issue #9977, opened 2021-10-19
Prune-revisions and structureelements


Geestig:
### Description
On a website with around 700 entries with a lot of data there seems to be an issue when I try to prune the revisions. It shows to be pruning exactly as it should however while the process is running the structureelements table starts to increase in size. At one point this table alone was over 1.5GB which caused the DB to crash.

Even when I only prune 1 single revision and try to do it gradually it still increases too much in size. This is quite a big problem since we rely on cronjobs with feed-me feeds. Which automatically prune revisions after running the feed.

This problem seems to have started to become worse when the structure section that is being filled by feed me was changed into a channel. Perhaps there is still some remnant in de database that causes an issue with this?

EDIT: Have been able to complete a pruning. Pruned 1 revision of each entry - structureelements is now 450MB in size. And after completion did not seem to get cleaned. So those 450MB are now stuck in that table.

EDIT: Another thing I noticed is that row count in structureelements actually lowers whilst the size of the table grows. Very weird.

EDIT: Disabling Blitz does not help, running gc afterwards does not work. There are no errors in any log files. Structureelements just keeps on getting bigger.

EDIT: SELECT RELEASE_ALL_LOCKS(); does not help with the "Unable to acquire a lock for the structure 2" error i'm getting when trying to save an entry that no longer is a structure but now a channel.

### Steps to reproduce

1. Create section as structure.
2. Populate section with 600+ entries.
3. Change from structure to channel.
4. Trigger prune-revisions (through CLI or Dashboard is the same result)
5. Watch the structureelements database grow in size.

EDIT: When placing back a backup where the section is still a structure the same issue takes place. However it's only when deploying that it seems to increase the database. If i prune locally there do not seem to be any issues.

### Additional info

- Craft version: 3.7.16
- PHP version: 7.4
- Database driver & version: MySQL 5.7.33
- Plugins & versions:
Asset Rev 6.0.2
Blitz 3.10.4
CP Field Inspect 1.2.5
Feed Me 4.4.0
Guest Entries 2.4.0
Imager v2.4.0
Inlin 2.1.1
Notifications 1.1.4
Preparse Field 1.2.5
QRCode 1.0.5
Redactor 2.8.8
SendGrid 1.2.2
SEOMatic 3.4.16
Snaptcha 3.0.9
Style Inliner 2.3.0
Super Table 2.6.8
Umami 1.0.4

brandonkelly:
[...]

PR #10040, opened 2021-10-30
De-structure drafts and revisions

brandonkelly:
### Description

With this change, drafts and revisions no longer store structure data, with the exception of:

- Unpublished drafts (since they are canonical)
- Provisional drafts which have been assigned a new parent

Removing existing draft/revision structure data is expensive since we need to patch up the other structure elements’ nested set values, so that is left up to the `utils/repair/section-structure` command.

### Related issues

- #9977
- #9999

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,13 +2,22 @@
 
 ## Unreleased
 
+> {tip} After updating, run the `utils/repair/section-structure` command for each Structure section, to prune unneeded structure data for drafts and revisions.
+
+### Added
+- Added `craft\services\Structures::remove()`.
+
 ### Changed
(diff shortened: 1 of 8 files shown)


sjcallender:
Very cool.

> You should notice an immediate, drastic performance improvement while working with entries within your Structure sections.

Just so we're analyzing the correct details, are the performance gains with regard to loading time of the elements list table, saving entries, and/or other actions?

boboldehampsink:
@brandonkelly Postgres doesn't seem to like this:

```
craft utils/repair/section-structure members
Exception 'yii\db\Exception' with message 'SQLSTATE[42883]: Undefined function: 7 ERROR:  operator does not exist: boolean = integer
LINE 16: ..."canonicalId" IS NULL) OR (("drafts"."provisional"=1) AND (N...
                                                              ^
HINT:  No operator matches the given name and argument types. You might need to add explicit type casts.
```

brandonkelly:
@boboldehampsink Whoops, fixed now.

Issue #9909, opened 2021-10-01
Run Craft behind a reverse proxy on subpath


jan-thoma:
### Description

When running craft behind a reverse proxy in this manner:

```
/ => frontend
/backend => Craft
```

Calls to index.php always resolve to the root of the domain and therefore fail. Is there a way to solve this problem. Without placing craft in that subdirectory. Background we run Craft and SSR Vue.js Docker Containers along. To avoid CORS Issues we would like to serve the Craft container from a subpath of the domain.

### Additional info

- Craft version: 3.7
- PHP version: 7.4
- Database driver & version: Mysql 5.6

timkelty:
This seems like it would be addressed by web server. Can you provide some more info about your reverse proxy/web sever configuration?

> Without placing craft in that subdirectory

when you say "placing craft", do you mean just [`index.php` ](https://github.com/craftcms/craft/blob/main/web/index.php)?

jan-thoma:
It's docker setup with traefik in front, with Craft, Vue and MySQL running in their separate containers. Traefik routes everything except urls starting with `/backend  to the Vue container.

Urls starting with `/backend` will be routed to the craft container, traefik strips `/backend` part. The `@web` path is set to e.g. `my.domain.com/backend`.

What's working:

- Access to assets and cpressources folders
- Graphql Endpoints

What's not working:

- Login (index.php gets called from the root)
- Control Panel redirects

timkelty:
@jan-thoma are you able to share anything where I could reproduce the behavior (eg dockerfiles/docker-compose)?

timkelty:
@jan-thoma can you try setting `'baseCpUrl' => 'my.domain.com/backend',` (the same value you're setting the `@web` alias to?

jan-thoma:
```
WEB=https://viiv.podcast.nitro/backend
WEBROOT=/app/web
BASE_CP_URL='https://viiv.podcast.nitro/backend'
```

With this settings the login attempt still hits an error 404

![Bildschirmfoto 2021-11-08 um 08 42 22](https://user-images.githubusercontent.com/24450307/140702529-32361e8a-db54-4555-b239-40ea03773cc4.png)

timkelty:
@jan-thoma thanks again for your patience!
I've managed to get this all working and reproducible, here: https://github.com/timkelty/craftcms-subpath-proxy

It is a default Craft install, running with docker-compose, and a stripped down version of your Traefik rules (just the subpath stuff).

[...]

PR #10048, opened 2021-11-02
Fall back to @web alias for baseCPUrl

timkelty:
In `baseCpUrl`, we were falling back to the deprecated `baseRequestUrl` instead of using the `@web` alias (like we do in `baseSiteUrl`).

This means someone setting `@web` (including a path segment) would have to explicitly set `@web` AND `baseCpUrl`:

```
    'baseCpUrl' => 'http://mysite/backend',
    'aliases' => [
        '@web' => 'http://mysite/backend',
    ]
```

With this PR, they would only have to set `@web`, as `baseCpUrl` would appropriately fall back to that.

## Related issues:
- https://github.com/craftcms/cms/issues/9909

### DIFF
--- a/src/helpers/UrlHelper.php
+++ b/src/helpers/UrlHelper.php
@@ -474,9 +474,9 @@ class UrlHelper
         if ($generalConfig->baseCpUrl) {
             return rtrim($generalConfig->baseCpUrl, '/') . '/';
         }
-
-        // Use the request's base URL as a fallback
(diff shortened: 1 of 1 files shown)


brandonkelly:
This was already changed for Craft 4 (0f2051bc6ca598d483104d06ff2a93b15b886188). Decided not to do it in Craft 3 because it could have some unexpected side effects.

PR #10054, opened 2021-11-04
Delete annotated $site property

bencroker:
### Description

The annotated `$site` property is not available via the code. The following code results in an `UnknownPropertyException` being thrown.

```php
$query = Entry::find()->siteId(1);
Craft::dd($query->site);
```

Screenshot 2021-11-04 at 09 59 09


brandonkelly:
You can set it, you just can’t read it. So just swapped it to `@property-write`, along with all other magic setters within element query classes.

bencroker:
That makes sense, thanks!

PR #10070, opened 2021-11-08
Remove credentials method

timkelty:
Adds a method to remove credentials from a users.

We have to save the record directly to get around validation in `User::beforeSave`: https://github.com/craftcms/cms/blob/4.0/src/elements/User.php#L1582-L1602

Flagging @andris-sevcenko here, in case it has implications for https://github.com/craftcms/cms/tree/feature/dev-13-multi-factor-authentication


andris-sevcenko:
@timkelty works for me.

brandonkelly (approved):

PR #10080, opened 2021-11-11
Feature/dev 53 restrict assets fields to paths

andris-sevcenko:
### Description

Adds a field mode to Assets fields that can be one of
- Normal mode
- Restrict assets to a single folder
- Restrict assets to a subtree of a volume

The latter is an addition that allows selecting a subfolder of a volume to be used as a constraining folder. It can have dynamic tokens.

### Related issues

Resolves https://github.com/craftcms/cms/discussions/9070

Docs:
- [ ] Document the new behavior


brandonkelly:
Toned down the PR so it’s more in-line with how things currently work, at least settings-wise. Instead of a new “Field Mode” setting, we now have the following changes:

- Renamed **Restrict assets to a single folder** to **Restrict assets to a single location**
- When that’s selected, a new **Allow subfolders** checkbox is shown below it
- When that’s selected, a **Default Upload Location** setting is shown (similar to the one you’d get when **Restrict assets to a single location** is **not** selected, except this one doesn’t have a source/volume menu; only a subpath.

Here’s what the settings look like when assets are restricted:

Screenshot of a portion of an Assets field’s settings, where the “Restrict assets to a single location” checkbox is checked; the “Asset Location” setting is set to “Site Assets → {site.handle}”, the “Allow subfolders” setting is checked; and the “Default Upload Location” setting is set to “{slug}”.

And when they’re not:

Screenshot of a portion of an Assets field’s settings, where the “Restrict assets to a single location” checkbox is checked; the “Sources” setting is set to a single “Site Assets” volume; and the “Default Upload Location” setting is set to “Site Assets → {site.handle}”.

brandonkelly (approved):

Issue #7366, opened 2021-01-07
Accessibility of assets field UI in entries


ajanec01:
#### Basic UI 

- <s>The add an asset and upload files components would benefit from `role="group"` or `<fieldset> `; the name of the asset field is not announced when the focus is moved to either of the two buttons</s> Fixed
![grouping-assets](https://user-images.githubusercontent.com/47136246/103952458-5ae66880-5138-11eb-9579-701d98feaf0b.png)
- <s>Poor keyboard focus indication on all components</s> Fixed
- <s>The remove asset links are not keyboard accessible (should be buttons-) they don’t have `href`; they should also have the programmatically determined link context, or add a screen reader only text</s> Fixed
![assets-display](https://user-images.githubusercontent.com/47136246/103955558-73598180-513e-11eb-9e85-4d12c5819535.png)
- Not possible to move the assets up or down using the keyboard only
- Only one of the assets can be navigated into using the Tab- the rest can be accessed using the arrow keys which is confusing
- The assets triggers need to be buttons 
- Not possible to open the asset info dialog using the keyboard only
![asset-information-dialog](https://user-images.githubusercontent.com/47136246/103955946-5ec9b900-513f-11eb-923e-6fd797a07634.png)
- The asset info dialog has accessibility issues covered in #7297 and in #7308 
- The additional info trigger is not keyboard accessible; it also needs accessible role and state (consider `<details>` and `<summary>`)
![additional-info](https://user-images.githubusercontent.com/47136246/103956486-82d9ca00-5140-11eb-889a-69dbaf3497f3.png)

#### Assets modal dialog

- the modal dialog with assets does not have acc name and role + the focus is not trapped- details in #7297 
![assets-modal](https://user-images.githubusercontent.com/47136246/103952994-3939b100-5139-11eb-9778-2d323134457e.png)
- Couple of issues in the navigation panel:
  - Whenever the trigger button is activated using the keyboard, the focus moves away from the trigger; need to do shift + tab to come back
![expandable-assets](https://user-images.githubusercontent.com/47136246/103953050-51113500-5139-11eb-9130-3abc1010b628.png)
  - Only one of the folders receives keyboard focus, a click on either of them adds `tabindex`- in short, possible to access only one folder using the keyboard only
[...]

PR #10085, opened 2021-11-12
Entry management improvements

gcamacho079:
### Description

- Use fieldset to group asset UI components as discussed in issue [7366](https://github.com/craftcms/cms/issues/7366)
- Fix bug of fieldset legend not being read for entry fields by providing screen reader only `legend` as first element in node
- Clarify some button text using explicit ARIA labels to prevent "plus" CSS content from being read
- Fix issues of unallowed ARIA attributes on status badge elements
- Ensure "required" text is read for all field types in place of asterisk
- Darken input placeholder text to meet minimum contrast guidelines


src/fields/Assets.php
```diff
@@ -320,6 +320,14 @@ public function getElementValidationRules(): array
         return $rules;
     }
 
+    /**
+     * @inheritdoc
+     */
+    public function useFieldset(): bool
+    {
+        return true;
+    }
```
brandonkelly, line 329: This should probably happen in `BaseRelationalField` so Entries, Users, Categories, and Tags fields get updated as well.

src/templates/_components/fieldtypes/Assets/input.html
```diff
@@ -20,7 +20,7 @@
 
     {% if sources %}
         <div class="flex{% if limit and elements|length >= limit %} hidden{% endif %}">
-            <button type="button" class="btn add icon dashed">{{ selectionLabel }}</button>
+            <button type="button" class="btn add icon dashed" aria-label="{{ selectionLabel }}">{{ selectionLabel }}</button>
```
brandonkelly, line 23: This should probably be made to `_includes/forms/elementSelect.html` so it affects Entries and Users fields as well.

brandonkelly (changes requested):

src/templates/_components/fieldtypes/Assets/input.html
gcamacho079: @brandonkelly have made that change!

PR #10100, opened 2021-11-16
Custom config support

### LINKED ISSUE #10012: limitAutoSlugsToAscii do not work
### Description
I have enabled the following setting in `general.php`:

``` 
 'limitAutoSlugsToAscii ' => true,
```

When I create a new page with "ä", the slug is not converted to "ae":

image

### Steps to reproduce

1. Add `'limitAutoSlugsToAscii ' => true,` to config.php
2. Add a new entry with `ä`

### Additional info

- Craft version: 3.7.17.2
- PHP version:  8.0.11
- Plugins & versions:

Contact Form	2.2.7
Image Resizer	2.1.0
Imager X	v3.5.3
Minify	1.2.10
Redactor	2.8.8
SEO	3.7.4

brandonkelly:
Adds support for a new `config/custom.php` file, where any custom config settings should be defined (rather than `config/general.php`.

The file can optionally be a multi-environment config, like other config files.

```php
return [
    '*' => [
        // global defaults
    ],
    'dev' => [
        // dev environment settings
    ],
    'production' => [
        // prod environment settings
    ],
];
```

Custom config settings are accessible via `Craft::$app->config->custom` (PHP) or `craft.app.config.custom` (Twig).

Resolves #10012

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - It’s now possible to edit images’ focal points from their preview modals. ([#8489](https://github.com/craftcms/cms/discussions/8489))
 - Added the `assetUploaders` user query param.
 - Added the `authors` user query param.
+- Added support for setting custom config settings from `config/custom.php`, which are accessible via `Craft::$app->config->custom`. ([#10012](https://github.com/craftcms/cms/issues/10012))
 - Added `craft\base\Volume::CONFIG_MIMETYPE`.
 - Added `craft\base\Volume::CONFIG_VISIBILITY`.
 - Added `craft\base\Volume::VISIBILITY_DEFAULT`.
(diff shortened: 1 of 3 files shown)


michaelrog:
Not strongly opinionated, but, couldn't we just use [`getConfigFromFile`](https://docs.craftcms.com/api/v3/craft-services-config.html#method-getconfigfromfile) for pulling custom values out of our own `whatever.php` config files?

mmikkel:
@michaelrog Sure can, but that approach requires a custom module or plugin. The point of `config/custom.php` is to have a simpler mechanism available for custom config settings, since adding custom values to `general.php` will no longer be possible in 4.0.

mmikkel:
@michaelrog I actually didn't think of that before, and I suppose that would be a viable option too. I still think `custom.php` is a good idea though, `craft.app.config.custom.someValue` seems a bit simpler and cleaner as a convention, compared to `craft.app.config.configFromFile('whatever').someValue`.

brandonkelly:
`getConfigFromFile()` wouldn’t memoize the results though. The config file would need to be re-read each time you call it, along with the overhead of checking if it’s a multi-environment config, etc.. And you wouldn’t be able to modify it at runtime.

michaelrog:
Yeah, if you're only using it once, maybe the extra few characters are meaningful...

Also I guess being able to reach in from a module and change/override the loaded values _might_ be useful in some cases?

Maybe there's a perf gain from having the config object loaded once, rather than on each invocation?

Only thing is, in this implementation, there's only one custom config, and it's always called "custom" — and maybe that convention is a feature, but...

I wonder what a magic getter API might look like... Where calling an ad-hoc config name would try to load the corresponding custom file into a config object:

```
craft.config.foo.something
craft.config.bar.another
```

brandonkelly:
Well `general` and `db` already contain special meaning and resolve to specific classes `craft\config\GeneralConfig` and `craft\config\DbConfig` so it would be weird if it was mostly arbitrary except for those two cases. Plus I’m not sure what the benefit of more than one custom config would be, verses just storing multiple config arrays within `general.php`, indexed by those `foo`/`bar`/etc. keys.

PR #10110, opened 2021-11-18
Improve french (fr-fr) translations

nstCactus:
### Description

Improve the french translation to be more grammatically correct and make it "sound" right.

It's a lot of commits but I figured it would be easier to review if needed.

Hope this helps!


brandonkelly:
Thanks a lot for this!

PR’ing here is fine if that’s easiest for you, but you might want to check out CrowdIn, which is where we manage our translations: https://crowdin.com/project/craft-cms

PR #10111, opened 2021-11-18
Broader environment variable support

brandonkelly:
This adds new `selectize` and `selectizeField` macros to `_includes/forms`, which make it easier to create Selectize-based select menus. These support an `includeEnvVars` setting, which can be set to `true` to include any defined environment variables which are set to one of the available option values.

There’s also new `booleanMenu` and `booleanMenuField` macros, which are boolean-specific implementations of `selectize`/`selectizeField`. They have `yesLabel` and `noLabel` settings (“Yes” and “No” by default), and will include any environment variables set to a boolean-esque value such as `on`/`off`/`yes`/`no`/`true`/`false`/`1`/`0` (see [filter_var()](https://www.php.net/manual/en/function.filter-var.php)).

Environment variables are now supported by the following settings:

- General → System Status
- General → Time Zone
- Email → Encryption Method (SMTP transport type)

A screenshot of the System Status setting page, expanded to show environment variable options

Resolves #8866
Resolves #8989


Mosnar:
@brandonkelly Should this field also include the same disclaimer that other autosuggest fields have? 
image

brandonkelly:
@Mosnar Considered it but it’s a bit more immediately obvious that it’s possible, since all available options are always present in the dropdown (until you start typing), whereas autosuggest fields are limited to just the first five options, which may not be environment variables.

timeverts:
This change is revolutionary @brandonkelly! Thanks so much!

brandonkelly:
Went ahead and added tips to these new settings as well, in 3.7.23 (see https://github.com/craftcms/cms/discussions/8989#discussioncomment-1707172).

sjelfull:
@brandonkelly This is awesome! Should be added to [Plugin docs](https://craftcms.com/docs/4.x/extend/environmental-settings.html).

PR #10128, opened 2021-11-19
Webpack cleanup

timkelty:
Lots of refactor here, so here's the summary:

- removed need for empty `entry.js` files
    - this also removed the need to `removeFiles`
- no more `lib`, `asset`, 'base' types…`vue` is the only remaining `type`
- Removed the `.min` output default
    - There was already lots of inconsistency here (eg css files), so it seemed better to just drop the `.min`. 

Note: There are instances where we're using `CopyWebpackPlugin` (eg jquery) where we could simply use the module as the entry. I've left this in place, because we need the globals exposed to the window object and this is alot more straightforward than dealing with `expose-loader`, `exports-loader`, `DefinePlugin`, etc.


packages/craftcms-webpack/index.js
```diff
@@ -0,0 +1,11 @@
+const glob = require('glob');
+const path = require('path');
+const fetchConfigs = (
```
nfourtythree, line 3: Was it not possible to put this in its own file having the "base path" passed into the constructor? With there default being the `module.parent.path`?

brandonkelly (approved):

src/web/assets/jqueryui/webpack.config.js
```diff
@@ -21,9 +20,6 @@ module.exports = new CraftWebpackConfig({
                         require.resolve('jquery-ui/ui/widgets/mouse.js'),
                     ],
                 },
-                transform: {
-                    'jquery-ui.js': code => require("uglify-js").minify(code).code
-                }
```
timkelty, line 26: This is already getting minified, so this isn't needed.

lerna.json
```diff
@@ -1,5 +1,10 @@
 {
   "packages": ["packages/*"],
   "npmClient": "npm",
-  "version": "independent"
+  "version": "independent",
+  "command": {
+    "version": {
+      "allowBranch": ["main", "develop"]
+    }
+  }
```
timkelty, line 9: So we don't accidentally release on a feature branch

packages/craftcms-webpack/get-config.js
```diff
@@ -0,0 +1,309 @@
...
+};
+
+module.exports = ({
+  context,
+  type = null,
+  config = {},
+  watchPaths = [
+    path.join(rootPath, "src/templates"),
+    path.join(context, "dist"),
+  ],
```
timkelty, line 140: Replaced `templatePath` with `watchPaths`

packages/craftcms-webpack/get-config.js
```diff
@@ -0,0 +1,309 @@
...
+    client: {
+      overlay: {
+        errors: true,
+        warnings: false,
+      },
+    },
+    static: watchPaths.map(directory => ({
+      directory,
+      watch: true,
+    })),
```
timkelty, line 121: @nfourtythree I killed the `DEV_SERVER_CONTENT_BASE` env var…do we need it? Seems like you could just pass `watchPaths` for whatever you wanted to watch.

packages/craftcms-webpack/index.js
```diff
@@ -0,0 +1,347 @@
...
+ * Returns and array of webpack configs.
+ *
+ * @param {string} globPattern
+ * @param {Object} options
+ * @param {string} options.cwd Path where the glob should be executed from.
+ */
+const getConfigs = (
+  globPattern = "src/web/assets/*/webpack.config.js",
+  options = {
+    cwd: module.parent.path,
```
nfourtythree, line 44: @timkelty Should we rename this to be like the other changes you made?

```suggestion
    context: module.parent.path,
```

nfourtythree: This is not required because we are just passing the glob options straight through

nfourtythree (approved):

PR #10163, opened 2021-11-25
Update editable table options to allow finer configuration around add, reorder, delete

nfourtythree:
### Description
This PR adds finer grain control over when the add, reorder and delete controls are shown.

Currently, the `staticRows` property is the main factor in determining when the reorder handle and delete buttons. In its original state, this meant that it was not possible to have a table that simply allowed the addition and deletion of rows **but not** reordering

`staticRows` has been dropped in favour of three new properties `allowAdd`, `allowReorder` and `allowDelete`. These boolean properties are self-explanatory when it comes to the functionality they enable/disable. The defaults for these settings are `false` so when nothing is passed no additions, reordering or deletion will be allowed, this is a slight change in behaviour from version 3.

Checklist
- [x] Keep default behaviour if nothing is passed
- [x] Respect the `static` option to display as a "static" field
- [x] Check internal uses of the field macro and update with new parameters (where required)
- [x] Check internal uses of CP helper method and update with new parameters (where required)
- [x] Update changelog
- [ ] Upgrade documentation


src/templates/_includes/forms/editableTable.twig
```diff
@@ -5,10 +5,20 @@
 {%- set initJs = not static and (initJs ?? true) -%}
 {%- set minRows = minRows ?? null %}
 {%- set maxRows = maxRows ?? null %}
-{%- set staticRows = static or (staticRows ?? false) or (minRows == 1 and maxRows == 1 and rows|length == 1) %}
-{%- set fixedRows = not staticRows and (minRows and minRows == maxRows and minRows == rows|length) %}
+{%- set allowAdd = allowAdd ?? true %}
+{%- set allowDelete = allowDelete ?? true %}
+{%- set allowReorder = allowReorder ?? true %}
```
brandonkelly, line 10: These should probably be `false` by default.

nfourtythree: Happy to make that change, the reason they are `true` is because that replicates the current default behaviour. 

Let me know if you still want these changes, and if so I can go an update the relevant twig implementations as well.

PR #10169, opened 2021-11-26
Remove buttons

gcamacho079:
### Description
Darken color of remove buttons for relational field types and clear date for post/expiry fields. Add additional sr-only text to remove buttons for additional context.


brandonkelly:
That seems a little harsh:

CleanShot 2021-11-27 at 06 17 34@2x

If I’m reading [Understanding Success Criterion 1.4.11: Non-text Contrast](https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast.html) right, we just need a 3:1 contrast ratio for these. Going with `$grey400` (`#7B8793`) would give us a 3.4:1 here (and even 3.09:1 against the , so we can probably go with that?

CleanShot 2021-11-27 at 06 30 35@2x

Just pushed a commit that applies that color to all UI control icons, as well as tweaked button icon colors a bit. Let me know if that looks good to you.

gcamacho079:
Looks good @brandonkelly , that does indeed meet the Non-text Contrast criterion. Thanks for hitting all those other icon colors in the process! 🙌🏼

brandonkelly:
Sweet! Merged.

PR #10180, opened 2021-11-29
`tsconfig.json` in root of repo

nfourtythree:
Move the `tsconfig` extension to the root to avoid multiple duplicate files.


timkelty (approved):
Worked in vscode!

PR #10182, opened 2021-11-30
feat: Add logging of Twig `{% deprecated %}` notices

khalwat:
### Description

Since Twig 2.6, Twig has supported the [`{% deprecated %}` tag](https://twig.symfony.com/doc/3.x/tags/deprecated.html). Currently, it does nothing in Craft (despite it being [documented as doing something](https://craftcms.com/docs/3.x/dev/tags.html))

This PR implements it in Craft as described here:

https://twig.symfony.com/doc/3.x/recipes.html#deprecation-notices

...by passing along Twig deprecaction errors to `Craft::$app->getDeprecator()->log()`

So Twig code like this:

```twig
        {% deprecated 'The "base.twig" template is deprecated, use "layout.twig" instead.' %}
```

Will generate this:

Screen Shot 2021-11-29 at 11 33 20 PM

### Related issues

Closes: https://github.com/craftcms/cms/discussions/10181

### DIFF
--- a/src/web/View.php
+++ b/src/web/View.php
@@ -340,7 +340,7 @@ class View extends \yii\web\View
         set_error_handler(static function ($type, $msg) {
             if (E_USER_DEPRECATED === $type) {
                 $file = null;
-                $line = $null;
+                $line = null;
                 if (preg_match('/\(.*(".*") at line (\d+)\)\.$/', $msg, $matches)) {
                     $file = $matches[1] ?? null;
                     $line = $matches[2] ?? null;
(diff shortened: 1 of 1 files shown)


brandonkelly:
According to [set_error_handler()](https://www.php.net/manual/en/function.set-error-handler.php), for PHP < 8.0, the `$errno` argument would always just be `0` (rather than whatever `E_USER_DEPRECATED` is set to), when the error is prefixed with `@` (which Twig’s built-in `{% deprecated %}` tag will do). So adding the error handler wouldn’t actually do anything for anyone running PHP 7.2–7.4.

So I added a custom `{% deprecated %}` tag instead, which overrides Twig’s, and logs the deprecation error directly.

PR #10183, opened 2021-11-30
Set `aria-describedby` with explicit `describedBy` variables

brandonkelly:
This PR makes the following changes:

- Built-in input types to set their `aria-describedby` attributes based on a `describedBy` variable rather than `instructionsId`.
- `craft\helpers\Cp::fieldHtml()` now sets `describedBy` to the `instructions` `tip`, and `warning` IDs (if not already set).
- `craft\base\FieldTrait` now has a `describedBy` property, which should be used to define fields’ inputs’ `aria-describedby` attributes.
- `craft\fieldlayoutelements\CustomField::inputHtml()` now sets `describedBy` on the field.
- Added a new `_includes/forms/datetime.html` template to simplify the `datetime()` macro and `craft\helpers\Cp::dateTimeField()`.

To-do:

- [x] Include error lists (should come first, before instructions)
- [x] Add visually-hidden labels to the tips, warnings, and error lists


src/fieldlayoutelements/BaseField.php
```diff
@@ -263,6 +266,59 @@ protected function id(): string
...
+        return $this->id() . '-instructions';
+    }
+
+    /**
+     * Returns the `id` of the field tip.
+     *
+     * @return string
+     * @since 3.7.24
+     */
+    protected function tipId(): string
```
gcamacho079, line 286: How will tips and warnings be associated with particular input fields? Visually, it seems like they can be used to provide additional context about one or more fields at a time. Will we give users the option of what fields they will apply to?

src/fields/Color.php
```diff
@@ -124,10 +124,9 @@ public function getElementValidationRules(): array
     protected function inputHtml($value, ElementInterface $element = null): string
     {
         /** @var ColorData|null $value */
-        $id = Html::id($this->handle);
         return Craft::$app->getView()->renderTemplate('_includes/forms/color', [
-            'id' => $id,
-            'instructionsId' => "$id-instructions",
+            'id' => Html::id($this->handle),
```
gcamacho079, line 128: Curious about why some id values reference `Html::id($this->handle)` while others reference `$id`

src/fieldlayoutelements/BaseField.php
brandonkelly: The IDs are gathered from `BaseField::describedBy()` here:

https://github.com/craftcms/cms/blob/00c77f7d8557687d4e028f793c10988504fc9b0d/src/fieldlayoutelements/BaseField.php#L311-L320

Which is called when setting `Field::$describedBy` before calling its `getInputHtml()` method:

https://github.com/craftcms/cms/blob/00c77f7d8557687d4e028f793c10988504fc9b0d/src/fieldlayoutelements/CustomField.php#L250

It’s up to the individual field types to actually pass their `describedBy` property to their input’s `aria-describedby` attribute. For example, Email fields do that by passing it to the `_includes/forms/text` template:

https://github.com/craftcms/cms/blob/00c77f7d8557687d4e028f793c10988504fc9b0d/src/fields/Email.php#L96

…which in turn will set it here:

https://github.com/craftcms/cms/blob/00c77f7d8557687d4e028f793c10988504fc9b0d/src/templates/_includes/forms/text.html#L33-L35

src/fields/Color.php
brandonkelly: `Html::id()` returns an `id`-safe version of a string (e.g. `foo bar` --> `foo-bar`). We just call it as a safety measure, although it’s probably not ever needed for field handles, since those already must be a single alphanumeric word.

In this case, `$id` was set to `Html::id($this->handle)` to begin with, but after replacing line 130 with `'describedBy' => $this->describedBy`, there was only one remaining reference of the `$id` variable (previous line 129), so I just got rid of the variable and am now calling `Html::id()` directly when defining the `'id'` key.

PR #10222, opened 2021-12-07
Support AVIF format

davidwebca:
### Description
Hi!

Since the Avif format is landing quickly (https://caniuse.com/?search=AVIF) and the gains are considerable compared to jpeg and even webp, it would be nice to have the support when necessary. I've landed a few progressive enhanced websites since its release and the compression is out of this world, especially on larger images and gradients (they almost keep their original quality, especially compared to jpegs 😳 ).

I made those edits pretty quickly and naively, basing my code changes on the somewhat recent (2017) rollout of webp. Plus, ImageMagick and GD both already support Avif, so I thought basing my changes on webp to add progressive support for the servers would be the best solution. 

Feel free to review and have a nice day 👍

### DIFF
--- a/src/config/GeneralConfig.php
+++ b/src/config/GeneralConfig.php
@@ -153,6 +153,7 @@ class GeneralConfig extends BaseObject
         'asc',
         'asf',
         'avi',
+        'avif',
         'bmp',
         'cap',
         'cin',
--- a/src/helpers/Assets.php
+++ b/src/helpers/Assets.php
@@ -488,6 +488,7 @@ class Assets
                 Asset::KIND_IMAGE => [
                     'label' => Craft::t('app', 'Image'),
                     'extensions' => [
+                        'avif',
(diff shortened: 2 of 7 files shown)


andris-sevcenko:
The one issue I have is that Safari does not support the AVIF format yet and not sure if that can be considered a web safe format, then.

OTOH, it stands to reason that if you're transforming images to AVIF, you have safeguards in place that check if browser supports it.

Have you tested this locally?

davidwebca:
Yes, it works really well and I also tested in older versions of PHP and the conditions seem to work: no avif support nor functions detected, no option showing in the admin. I agree it's early, but people who would go out of their way to select it in the admin must know what they do. Otherwise, there's always the possibility of adding a warning when it's selected in the admin settings. 🤔

brandonkelly:
Craft 3.7.26 is out with added AVIF support ✨

PR #10234, opened 2021-12-10
Improve accessibility of BaseElementSelectorModals via Garnish updates, ARIA, and focus management

gcamacho079:
### Description
This PR focuses on relational element modal accessibility. Adds additional info such as modal name to the select modal, and makes updates to BaseElementSelectInput class to manage keyboard focus when adding and removing elements.

These changes aim to address several general modal issues (CMS-009, CMS-010, CMS-014), in addition to some more specific to the relational element modals (CMS-011, CMS-012).

### Related issues
Resolves DEV-164


src/web/assets/cp/src/js/BaseElementSelectorModal.js
```diff
@@ -24,17 +24,19 @@ Craft.BaseElementSelectorModal = Garnish.Modal.extend({
     init: function(elementType, settings) {
         this.elementType = elementType;
         this.setSettings(settings, Craft.BaseElementSelectorModal.defaults);
+        var $headingId = 'elementSelectorModalHeading-' + Date.now();
 
         // Build the modal
-        var $container = $('<div class="modal elementselectormodal"></div>').appendTo(Garnish.$bod),
+        var $container = $('<div class="modal elementselectormodal" aria-labelledby="' + $headingId + '"></div>').appendTo(Garnish.$bod),
+            $heading = $('<h2 id="' + $headingId + '" class="visually-hidden">' + this.settings.name + '</h2>').appendTo($container),
```
gcamacho079, line 31: @brandonkelly ideally the heading of each modal would reflect the field name (i.e. "Select featured image", "Select categories", etc.). I was unsure how to get the specific field names here, so I added a generic translation for the time being.

gcamacho079: @brandonkelly I've added two new translations 'Select element' and 'Select {element}' to handle this. The latter pulls the name from the fieldset legend node. Let me know if you prefer a different naming convention for these

gcamacho079:
@brandonkelly these changes (in CMS and Garnish PRs) add additional modal attributes/functionality to modals and slide-out classes. One user flow where this is in place is the adding a category to a category field:
- Opening "add a category" modal hides background content from screen readers and traps focus
- Stores initial trigger element for refocus on close, or moves focus to another logical target in the user flow (such as remove button if user cannot add another category)
- If adding another category and triggering the slide-out, hiding the previous modal from screen readers. Un-hides _only_ the category modal on slide-out close.
- The last modal that is closed triggers the removal of `aria-hidden` attribute from all other body children

Please let me know if there's a similar HUD-triggering-modal (or vice versa) place where this code should be implemented as well; I couldn't identify a HUD in the codebase that had a similar user flow.

gcamacho079:
@brandonkelly this will be my next PR once the

Issue #6013, opened 2020-05-01
Setting unknown property: craft\behaviors\CustomFieldBehavior:: - Sporadically Breaks Production Site


rossthomson89:
### Description

One of my production sites is sporadically showing an internal server error with the following error logged in the web logs:

```
Setting unknown property: craft\behaviors\CustomFieldBehavior::[CUSTOMFIELDNAME]
```

The site can will show without issue for several hours before the same error appears again without warning (or any apparent reason). Both my local and staging sites are not experiencing this issue and as a workaround I've been manually regenerating the storage folder which would temporarily fix the issue.

However as a last ditch attempt I've also regenerated the vendor folder via composer install (as well as regenerating another storage folder) and the site is currently up again.

The CUSTOMFIELDNAME that the site seems to be tripping over is just a default redactor field if that would help debug the issue.

### Stack Trace

```
2020-05-01 19:02:48 [-][-][-][error][yii\base\UnknownPropertyException] yii\base\UnknownPropertyException: Setting unknown property: craft\behaviors\CustomFieldBehavior::aboutThisMember in /home/scis/web/vendor/yiisoft/yii2/base/BaseObject.php:163
Stack trace:
#0 /home/scis/web/storage/runtime/compiled_classes/CustomFieldBehavior.php(94): yii\base\BaseObject->__set('aboutThisMember', NULL)
#1 /home/scis/web/vendor/craftcms/cms/src/base/Element.php(2170): craft\behaviors\CustomFieldBehavior->__set('aboutThisMember', NULL)
#2 /home/scis/web/vendor/craftcms/cms/src/base/Element.php(2149): craft\base\Element->setFieldValue('aboutThisMember', NULL)
#3 /home/scis/web/vendor/yiisoft/yii2/base/Component.php(180): craft\base\Element->setFieldValues(Array)
#4 /home/scis/web/vendor/craftcms/cms/src/base/Element.php(1113): yii\base\Component->__set('fieldValues', Array)
#5 /home/scis/web/vendor/yiisoft/yii2/BaseYii.php(558): craft\base\Element->__set('fieldValues', Array)
#6 /home/scis/web/vendor/yiisoft/yii2/base/BaseObject.php(107): yii\BaseYii::configure(Object(craft\elements\Entry), Array)
#7 /home/scis/web/vendor/craftcms/cms/src/elements/db/ElementQuery.php(1790): yii\base\BaseObject->__construct(Array)
[...]

PR #10281, opened 2021-12-22
Feature/clear cache compiled classes

khalwat:
### Description

I'd long assumed that `clear-caches/all` CLI/Utility would clear all of the runtime caches, but apparently it does not clear `storage/runtime/compiled_classes`

Adding this fixes that from a CI POV, and also fixes a nasty issue that can crop up because of it (see below):

### Related issues

https://github.com/craftcms/cms/issues/6013

^ clearing out the `storage/runtime/compiled_classes` dir addresses the issue above, something a client of mine ran into yesterday


brandonkelly:
This introduces new translations so I’ve changed the base to `develop-t9n` (which is where we are putting changes that are going to require a new translation order). Can you please add the new source translations to `src/translations/en/app.php` (correctly alphabetized)? I’d do it myself but looks like I’m not allowed to modify the PR.

khalwat:
Done ^

brandonkelly:
Thanks!

brandonkelly:
Finally got this released, in 3.7.39 (and 4.0.0-RC1).

PR #10298, opened 2022-01-03
Update BaseField.php

leevigraham:
Makes `instructions()` public just like `label()`.

### DIFF
--- a/src/fieldlayoutelements/BaseField.php
+++ b/src/fieldlayoutelements/BaseField.php
@@ -424,7 +424,7 @@ abstract class BaseField extends FieldLayoutElement
      * @return string|null
      * @since 3.7.24
      */
-    protected function instructions(ElementInterface $element = null, bool $static = false): ?string
+    public function instructions(ElementInterface $element = null, bool $static = false): ?string
     {
         return $this->instructions ? Craft::t('site', $this->instructions) : $this->defaultInstructions($element, $static);
     }


brandonkelly:
This could break any subclasses that override `instructions()`, so would need to wait until 4.0. And if we did change its visibility, we should follow suit with `tip()`, warning()` and maybe others. What’s the use case?

leevigraham:
I'm building out a front end entry form. 

```
    {% set formModel = formModel ?? create({
        class: 'craft\\elements\\Entry',
        sectionId: 16,
        typeId: 17,
        enabled: 1,
        authorId: currentUser.id,
        fieldValues: {
            'masterMug': [entry.id]
        }
    }) %}

    {% set fieldLayout = formModel.getFieldLayout %}
```

Then in a custom twig extension I pull the field label, instructions etc

```php
    public function formCustomField(Element $formModel, string $property, FieldLayout $fieldLayout): string
    {
        /** @var CustomField $customField */
        $customField = $fieldLayout->getField($property);
        $field = $customField->getField();

        return $this->formField($formModel, $property, [
            'form' => [
                'blockName' => $this->formBlockPrefix($field::class)
            ],
            'field' => $field,
            'label_attrs' => [
                'text' => $customField->label(),
            ],
            'hint_attrs' => [
                'text' => $customField->instructions ?: $field->instructions
            ],
            'widget_attrs' => [
                'name' => "fields[$property]"
            ]
        ], $customField->required);
    }
```

I have worked around this with: `$customField->instructions ?: $field->instructions`… and I could probably do that with `tip`.

My goal here is to provide a limited front end "admin" to users without exposing the whole backend. The field layout is pretty complex so I didn't want to duplicate labels, instructions etc hardcoded in the front end template.

brandonkelly:
Gotcha. Well given there’s actually quite a few other functions that would need to be updated for consistency, and this isn’t even a use case field layouts were intended for, I’m thinking it’s not worth the breaking changes.

leevigraham:
Fair enough on the breaking changes part.

Side note… I'm always torn between trying to make CraftCMS a custom application with heaps of additional functionality vs spinning up a Symfony App. Front end entry forms are part of that consideration. Have you thought about that use case?

brandonkelly:
I think the first step is we would need to build out a component library for various popular front-end frameworks (e.g. using [stencil.js](https://stenciljs.com/), which includes UIs for each of the built-in field types. If we had that it would start to make more sense to repurpose field layouts for the front end as well.

PR #10300, opened 2022-01-03
Conditional fields and tabs

brandonkelly:
### Description

The field layout designer has been updated to open tab and layout element settings in slideouts, which now include two new settings:

- **Current User Condition** – determines when the tab/field should be visible based on the logged-in user.
- **[Element type] Condition** (e.g. “Entry Condition”) – determines when the tab/field should be visible based on the element being edited.

Screenshot of a field layout designer, with a custom field’s settings slideout open, showing the two new condition settings

Edit Entry pages will show/hide tabs and fields as edits are being made, thanks to draft autosaves. Other forms will only factor in the conditions when they are first rendered.

### Related issues

- #8099
- #8154


mmikkel:
Took this for a quick test drive, and at first glance everything looks and works great. Thanks for finally getting this much-wanted feature into core!

That said, there's a couple of improvements I'd love to see in the final release 😅

1. Layout elements with conditions should have some kind of visual cue in the FLD – similar to how required fields get a `*` next to their label.  

    Without any sort of visual cue in the FLD that an element has conditions, editing complex layouts with multiple conditionals is probably going to feel a bit opaque and unnecessarily frustrating.

    _The Reasons plugin solved this by adding a simple, dashed outline to elements with conditions. I always thought this was a decent way to communicate that a particular element would only _conditionally display_, but I'm sure there are better/clearer ways to achieve the same thing._  

    CleanShot 2022-01-03 at 23 42 06@2x

2. The ability to create simple "Empty / Not empty" conditions for relational fields would be massively useful.

brandonkelly:
@mmikkel Both good suggestions. Can definitely do an empty/not-empty relational field condition rule.

brandonkelly:
[Just added](https://github.com/craftcms/cms/commit/f5412f205172131ff8681150bb986d5cca09a57d) an operator menu to relational field conditions, with the options “has a value”, “is empty”, and “is related to”. The first two are available in all contexts; “is related to” is only available for conditions that don’t need to be stored in the project config (so… element index filters).

brandonkelly:
Annnnd now layout elements with conditions get a special border:

Snippet of a field layout designer, showing that two of the included layout elements have a special border, indicating that they have conditions set on them

mmikkel:
> Annnnd now layout elements with conditions get a special border

That is awesome 👏  Thank you!

andrewfairlie:
I've not been able to install this yet, so apologies for asking what may be obvious, but I'm thinking ahead to a future project I have coming up. With this solution is it possible to show/hide fields based on file type? For example: if it's an image asset show "Alt Text", but if it's a PDF don't?

brandonkelly:
Yep!

jacksoncool:
i put this another place, but what about show field by site?  Like in a multisite setup?

PR #10305, opened 2022-01-04
Feature/volume filesystems

andris-sevcenko:
### Description

This PR accomplishes two things.

#### Separate Volumes form Filesystems

Volumes become very thin models that have a filesystem attached to them. This will allow the filesystem setting to be set up as an environment variable, allowing a different filesystem to be used in different environments.

#### Asset Transforms => Image Transform and Image Transform Drivers

Asset Transforms are now Image Transforms. Further, a transform driver concept is introduced. All existing image transformation capability has been rolled into the `Default` image transform driver. This includes managing the transform indexes and the like.

It is now possible to implement new image transform drivers (for example, to take advantage of Imgix or similar services).


khalwat:
Glad you got to this one! https://github.com/craftcms/cms/discussions/8471

PR #10320, opened 2022-01-06
Fixes typo in PHPDoc since tag

mmikkel:
### Description

Fixes a typo in the PHPDoc `@since` tag for the [`craft\helpers\Html::encodeInvalidTags()`](https://github.com/craftcms/cms/blob/develop/src/helpers/Html.php#L850) method.


brandonkelly:
Doh, thanks!

PR #10321, opened 2022-01-06
Move Garnish to Webpack

timkelty:
@brandonkelly this is ready to go.

As discussed, this abandons the standalone npm package for Garnish, and treats it as a normal asset bundle, using the same webpack build process as everything else.

```
npm run build -- --config-name garnish
npm run serve -- --config-name garnish
```

Notes:
- In lieu of upgrading and reworking, I removed the Karma stuff, as it didn't seem to be in use any longer
- To get HMR/live reloading working, the most straightforward thing was to refactor Garnish to ES modules (which is a good idea anyway).
- The `Garnish` global is still exposed to the browser, but `Base` no longer is.


timkelty:
@brandonkelly this is functional within Craft, just not published to npm yet.

You can treat it like any other asset bundle, so these commands should work from the Craft root:

```
npm run build
npm run build -- --config-name garnish
npm run serve -- --config-name garnish
```

Quick test of `serve` didn't seem to be live-reloading, but I can debug that with @nfourtythree.

PR #10322, opened 2022-01-06
Webpack tweaks

timkelty:
See review ↓

### DIFF
--- a/packages/craftcms-webpack/index.js
+++ b/packages/craftcms-webpack/index.js
@@ -84,7 +84,7 @@ const getConfig = ({
     );
   }
 
-  const configName = path.basename(context);
+  const configName = config.name || path.basename(context);
 
   if (!watchPaths) {
     watchPaths = [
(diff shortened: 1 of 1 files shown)


packages/craftcms-webpack/index.js
```diff
@@ -84,7 +84,7 @@ const getConfig = ({
     );
   }
 
-  const configName = path.basename(context);
```
timkelty, line 87: In case a name was passed with the config, we should use that. That will happen anyway, because we merge `config` at the end, but this seems more clear in case we do anything else with `configName` and expect it to be right.

packages/craftcms-webpack/index.js
```diff
@@ -343,7 +343,7 @@ const getConfig = ({
       config.plugins.push(new Dotenv({ path: dotenvResult }));
     }
 
-    if (!isDevServerRunning) {
+    if (!process.env.NODE_ENV === "production") {
       config.plugins.push(new CleanWebpackPlugin());
```
timkelty, line 347: Seems conceivable you'd want to do a build and not have it be minified. Since env is development when serving anyway, this allows for that flexibility.

packages/craftcms-webpack/index.js
nfourtythree: Works for me!

packages/craftcms-webpack/index.js
nfourtythree: Yup, good point!

nfourtythree (approved):

PR #10324, opened 2022-01-06
Allow unlocking users no matter their status (active, pending, suspended).

nstCactus:
Also fixes the status label on user edit page: "Locked" is always shown, when there is an ongoing cool-down period, no matter their status

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Changed
 - The content in the `resourcepaths` table is now excluded from database backups by default.
+- Allow unlocking users no matter their status
 
 ## 3.7.28 - 2022-01-05
 
--- a/src/controllers/UsersController.php
+++ b/src/controllers/UsersController.php
@@ -783,25 +783,7 @@ class UsersController extends Controller
                     }
                     break;
                 case User::STATUS_ACTIVE:
-                    if ($user->locked) {
-                        $statusLabel = Craft::t('app', 'Locked');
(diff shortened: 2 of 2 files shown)


brandonkelly:
What would the point of unlocking a pending/suspended user be? They still won’t be able to sign in.

nstCactus:
One of our clients made us design a complex user activation procedure, where users are created with a pending status and get an "account number" & "activation password" (stored in custom fields) that's mailed to them (slow mail, not email). Once they receive the mail, they can activate their account using their "account number" & "activation password" and choose their new credentials (email & password).

The user activation form may lock those pending users if they enter 3 invalid "activation password" (we do this be calling the `Craft::$app->users->handleInvalidLogin($user)` method). Once they're locked, they have to wait for the cooldown period to try again to activate their account. Our client wants to be able to unlock those users to let them try again right away.

I already found how to add the "Unlock user" action using the `UsersController::EVENT_REGISTER_USER_ACTIONS` event, but I thought this could be added in Craft itself, for such custom scenarios. 

In my opinion, it doesn't hurt to allow pending users to be unlocked but as it is possible to get them locked using custom code, it makes sense to also be able to unlock them.

brandonkelly:
Gotcha. I think I’m going to leave it as-is for now. Out of the box a user would only ever be locked if they were already active to begin with. And I don’t want to lose the visible “Locked” status for locked users, which is working as expected for typical use cases as far as I can tell.

PR #10341, opened 2022-01-09
Give user indexes group based URLs

### LINKED ISSUE #10336: URL does not change when we switch user groups on "users" page
On other element lists URL does change, for example when we switch section on entries page.

timkelty:
### Description
User indexes didn't have source based URLs like other elements.

### Related issues
Fixes #10336

### DIFF
--- a/src/config/cproutes/common.php
+++ b/src/config/cproutes/common.php
@@ -69,6 +69,7 @@ return [
     'settings/tags/new' => 'tags/edit-tag-group',
     'settings/tags/<tagGroupId:\d+>' => 'tags/edit-tag-group',
     'settings/users' => ['template' => 'settings/users/fields'],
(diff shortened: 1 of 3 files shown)


brandonkelly:
it was a bit more involved than that (72551b2513e1de55b3b29c990b4ba66f26b8a2f7).

- That `defaultSource` variable wouldn’t do anything on its own. The `users/index` template needs to pass it to JS, and the UserIndex JS class needs to have a `getDefaultSourceKey()` method that references it.
- Ideally the “All users” source has its own URL as well (`users/all`). So I added a new `data-slug` attribute to each source that defines the slug separately from its key.
- `admins`, `all`, and `new` needed to be marked as reserved user group handles, so it doesn’t get awkward when you select a group with one of those handles, reload the page, and are then routed somewhere else.

PR #10355, opened 2022-01-11
Remove extra wrapper from error container and reference error list di…

gcamacho079:
…rectly to solve issues with VO/Safari readout

### Description
Fixes a Voiceover bug - list text is only read out as part of `aria-describedby` when it's the directly-referenced node. Instead of "Errors: " text preceding the list, each list element will start with the visually-hidden text "Error: "


brandonkelly:
Merged! 🚀

PR #10356, opened 2022-01-11
set/to/if-empty options for resave/* commands

brandonkelly:
### Description

This adds new `--set`, `--to`, and `--if-empty` options to all `resave/*` commands, which can be used to bulk-set a new value on resaved elements’ attributes and custom fields.

Examples:

```sh
# set `foo` to "bar" if it’s empty:
php craft resave/entries --set foo --to "=bar" --if-empty

# copy the value of `baz` to `foo`:
php craft resave/entries --set foo --to baz

# clear out the `foo` field/attribute:
php craft resave/entries --set foo --to :empty:

# set `foo` to the result of an object template (Twig allowed):
php craft resave/entries --set foo --to "={author.name}"

# set `foo` to the result of an arrow function:
php craft resave/entries --set foo --to "fn(\$entry) => \$entry->author->name"
```

### Related issues

- #8132
- #8461

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 ### Added
+- `resave/*` commands now support bulk-setting an attribute/custom field value via new `--set`, `--to`, and `--if-empty` options.
 - Added `craft\helpers\App::parseEnv()` and `parseBooleanEnv()`, replacing their `Craft` class counterparts. ([#10319](https://github.com/craftcms/cms/discussions/10319))
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## Unreleased
 
 ### Added
-- `resave/*` commands now support bulk-setting an attribute/custom field value via new `--set`, `--to`, and `--if-empty` options.
+- `resave/*` commands now support bulk-setting an attribute/custom field value via new `--set`, `--to`, and `--if-empty` options. ([#10356](https://github.com/craftcms/cms/pull/10356))
 - Added `craft\helpers\App::parseEnv()` and `parseBooleanEnv()`, replacing their `Craft` class counterparts. ([#10319](https://github.com/craftcms/cms/discussions/10319))
 
 ### Changed
--- a/src/console/controllers/ResaveController.php
+++ b/src/console/controllers/ResaveController.php
@@ -19,7 +19,9 @@ use craft\elements\MatrixBlock;
 use craft\elements\Tag;
 use craft\elements\User;
 use craft\events\BatchElementActionEvent;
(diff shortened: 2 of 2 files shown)


samhibberd:
Nice one @brandonkelly, some really nice updates to the resave command.

One thought, with the new functionality to now update content / attribute values as a part of this, should the `dateUpdated` value not update in line with any content changes. I can understand why this was stopped when elements were just resaved as is, but with content explicitly being updated?

Apologies if i'm missing something obvious.

And if you currently needed to mark the elements as updated at that point would it be safe to set the dateUpdated attribute to now manually?

brandonkelly:
@samhibberd It was intentional to not change `dateUpdated` even if `--set` is used. In general I expect this is going to be used for things like setting default values for new fields (e.g. #8461) or migrating content from one field to another (e.g. [the new Alternative Text fields in Craft 4](https://github.com/craftcms/cms/discussions/10302#discussioncomment-1946371)).

samhibberd:
Digging deeper into the craft I can see why you have decided that makes sense.

In a situation that you would need to update the dateUpdated is updating the value set in the `elements` table sufficient?

brandonkelly:
@samhibberd Yeah that’s the place to do it.

Kashkin:
@brandonkelly I'm attempting to use `--set` and `--to` to update an Entries field, and not having any luck. Is this a supported use case, and if so, what is the syntax? If not, what are the supported field types that can be updated using the CLI?

brandonkelly:
@Kashkin For that you would need to use the PHP arrow function syntax, because relational fields need to be set to an array of related element IDs.

```sh
php craft resave/entries --set foo --to "fn(\$entry) => [1, 2, 3]"
```

Kashkin:
Thanks for the tip! That makes sense in retrospect, but it would be great to add that example to the docs!

mattstein:
[Example added](https://craftcms.com/knowledge-base/bulk-resaving-elements#resaving-with-computed-field-values) @Kashkin, thank you! We’ll also get a link to that article from the [Console Commands](https://craftcms.com/docs/3.x/console-commands.html#resave) page in the docs.

Kashkin:
@mattstein Hi Matt! Thanks for the continuously improving documentation. I can't seem to figure out the syntax for `--to` when setting a lightswitch field. Any ideas?

brandonkelly:
@Kashkin You can use `=1` or `=0`:

```php
php craft resave/entries --set myLightswitchField --to "=1"
```

Kashkin:
Thanks @brandonkelly! I feel really silly for not noticing the `=` in all the documentation. 🤦

brandonkelly:
Hah, no worries :) It’s required if you want to pass a literal value, as opposed to reference another field/attribute.

PR #10359, opened 2022-01-12
Fix a bug where `ElementQuery` would silently discard the contents of the `withQueries` inherited property when preparing the query

nstCactus:
This bug prevents using the MySQL [`WITH` clause](https://dev.mysql.com/doc/refman/8.0/en/with.html) to define Common Table Expressions.


brandonkelly:
Craft 3.7.29 is out now with this change. Thanks again!

PR #10366, opened 2022-01-14
Log fatal php errors to stream log

timkelty:
This sends fatal PHP errors to `php://stderr` when the app has opted in to `CRAFT_STREAM_LOG`, which is most likely what you want if you've set that.


timkelty:
@mattstein Probably worth a mention on:
- https://craftcms.com/knowledge-base/locating-error-logs-and-database-backups and 
- https://craftcms.com/docs/3.x/config/#craft-stream-log
- https://craftcms.com/docs/3.x/config/#craft-log-php-errors

angrybrad (approved):
Looks good to me.

angrybrad:
@timkelty we should add a changelog item for it probably as a bug fix

jasonmccallister (approved):
💯!!

brandonkelly (approved):

Issue #3844, opened 2019-02-14
Empty checkbox fields won't save


jalendport:
### Description

When you have a checkbox field on an entry and none of the option(s) are selected, saving the entry results in a `Couldn’t save entry.` error. The checkbox field is highlighted and says `CheckboxField is invalid.`

Looking in the logs I see:
```
2019-02-14 11:48:13 [-][1][-][info][craft\services\Elements::saveElement] Element not saved due to validation error: Array
(
    [testField[0].options[0].misc] => Array
        (
            [0] => CheckboxField is invalid.
        )

)
```

I just upgraded to the latest version of Craft CMS this morning so 99% sure this was introduced in one of the last couple releases (3.1.8 +).

### Steps to reproduce

1. Create a checkbox field and add it to a section
2. Try to save entry without selecting any of the checkbox field(s)

### Additional info

- Craft version: Craft CMS 3.1.10
- PHP version: PHP 7.1.25
- Database driver & version: MySQL 5.5.5
- Plugins & versions: tested on a clean install with no plugins

samhernandez:
I've replicated the error on a fresh installation of Craft 3.1.10, PHP 7.2.10, PostgreSQL 11.1

jalendport:
Thanks @samhernandez and @brandonkelly 🙌

brandonkelly:
Just released 3.1.11 with a fix for this.

PR #10373, opened 2022-01-17
Always append empty values to dropdowns with invalid data

### LINKED ISSUE #10365: Invalid dropdown values don't provide blank option if a default is set
### Description
Based on #9989 dropdowns are supposed to add a blank option if the current value is not valid. We recently changed our button types and are getting an error because the first value in a required dropdown cannot be chosen without first choosing the second, saving, and then choosing the first value. Note on these fields, the second value is marked as default. A blank option is only shown if `default` is not set on the field. 

### Steps to reproduce

1. Select a value in a REQUIRED dropdown field, save the page
2. Change the field to have different values, and set one as default which is NOT THE FIRST item. 
3. Edit the original page, and try to save without changing the value. Get a Red ⚠️ Error because the value cannot be blank.
4. Open the dropdown to see there's no blank option.
4. Removing the default does provide a blank option.

### Screenshots
First state, where a default value is selected, but the first item cannot be chosen, no blank option is provided
Screen Shot 2022-01-14 at 12 08 46 PM

After removing the default, a blank option is provided
Screen Shot 2022-01-14 at 12 12 50 PM

### Additional info

- Craft version: 3.7.28
- PHP version: 7.4.27
- Database driver & version: MySQL 5.7.33
- Plugins and versions 
  - Amazon S3 1.3.0 
  - Asset Usage 2.2.1 
  - Calendar Links 1.0.1 
  - Cloudflare 1.0.2 
  - Cookie Consent Banner 1.2.9 
  - Donkeytail 2.0.6 
  - Element API 2.8.4 
  - Feed Me 4.4.1.1 
  - Field Manager 2.2.4 
  - GeoMate v1.3.1.2 
  - Icon Picker 1.1.12 
  - ImageOptimize 1.6.39 
  - Linkit 1.1.12.1 
  - Matrix Finder 1.1.0 
  - quickbase-process 1.0.13 
  - Read Time 1.6.0 
  - Redactor 2.8.8 
  - Redactor Anchors 1.1.0 
  - Retour 3.1.68 
  - SEOmatic 3.4.21 
  - Similar 1.1.5 
  - Simple Sharing 1.0.8 
  - Smith 1.1.14 
  - Sprig 1.10.4 
  - Super Table 2.7.1 
  - Tags 1.0.9 
  - Twigpack 1.2.15 
  - Workflow 1.6.6

timkelty:
### Description
Fixes #10365, being careful not to break #3844

### DIFF
--- a/src/fields/BaseOptionsField.php
+++ b/src/fields/BaseOptionsField.php
@@ -239,11 +239,12 @@ abstract class BaseOptionsField extends Field implements PreviewableFieldInterfa
         }
 
         if (is_string($value) && (
-                $value === '' ||
                 strpos($value, '[') === 0 ||
                 strpos($value, '{') === 0
             )) {
             $value = Json::decodeIfJson($value);
+        } else if ($value === '' && $this->multi) {
+            $value = [];
         } else if ($value === null && $this->isFresh($element)) {
             $value = $this->defaultValue();
         }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@
 - The content in the `resourcepaths` table is now excluded from database backups by default.
 - User caches are now invalidated after login attempts. ([#10313](https://github.com/craftcms/cms/discussions/10313))
 - Assets fields’ “Upload files” buttons now have `aria-describedby` attributes when appropriate.
-- Color fields’ hexn inputs now have `aria-label` attributes.
+- Color fields’ hex inputs now have `aria-label` attributes.
 - Field status messages are now included in their `aria-describedby` attributes.
 - User queries will no longer exclude suspended users when `status('pending')` is passed. ([#10361](https://github.com/craftcms/cms/issues/10361))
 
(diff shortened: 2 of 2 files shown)


src/fields/BaseOptionsField.php
```diff
@@ -239,11 +239,12 @@ public function normalizeValue($value, ElementInterface $element = null)
         }
 
         if (is_string($value) && (
-                $value === '' ||
                 strpos($value, '[') === 0 ||
```
timkelty, line 242: Removing this allows `$value` to remain `''` instead of being coerced to `null`.

src/fields/BaseOptionsField.php
```diff
@@ -239,11 +239,12 @@ public function normalizeValue($value, ElementInterface $element = null)
...
                 strpos($value, '{') === 0
             )) {
             $value = Json::decodeIfJson($value);
+        } else if ($value === '' && $this->multi) {
+            $value = [];
```
timkelty, line 247: This ensures we don't brake the fix from #3844

brandonkelly (approved):

PR #10398, opened 2022-01-21
Improve global focus outline contrast

gcamacho079:
### Description
Introduces a new focus ring color, updates previous values, and updates all instances of `:focus-visible` to `.focus-visible` for compatibility with Safari

Completes DEV-211


gcamacho079:
A few examples of the new focus outlines:
Screenshot of entry panel with multiple focusable elements with forced focus states

PR #10416, opened 2022-01-27
Add Money custom field

nfourtythree:
### Description
Add a custom field type that supports the Money/Money PHP lib.


leevigraham:
I asked @lukeholder if I should add my input for this new field type and he said yes… so it's not totally unsolicited at the last moment 😀.

I've built a money field as part of [newism/craft3-fields](https://github.com/newism/craft3-fields) and here's what I've learnt / my approach / things I've considered as part of a rewrite.

### Ability to change currency

First thing I noticed about the proposed implementation is there is no way for users to change the currency:

image

NSM Fields: 

image

In my scenario on https://mytiki.life collectors set a value of their item in the purchase currency. Users are from all over the world so currency changes.

Changing the currency updates the prefix:



#### Storing the currency

Currently the amount and currency are stored as a json object. This could be improved using the new `array` syntax in `getContentColumnType`. Storing this value creates an immutable object even if the system admin changes the default currency / hides the currency drop down. More on that later.

```php
    public function getContentColumnType(): array|string
    {
        return [
            'amount' => Schema::TYPE_DOUBLE,
            'currency' => Schema::TYPE_STRING,
        ];
    }
```

#### Preferred / Excluded Currencies

Following the Symfony ChoiceType form element the NSM Fields Money Field has the concept of preferred currencies. These options are moved to the top of the list and separated by a divider. 

image

Excluded currencies are removed from the list. Given moneyphp/money has all kinds of crazy currencies I recommend some default exclusions.

There are two levels of settings, plugin and field. Field settings are appended to plugin settings. I'm not 100% convinced *merge* this is the best approach… maybe overwrite is a cleaner option.

image

image

#### Currency choices

Currency choices are pulled from https://github.com/moneyphp/money/blob/master/src/Currencies/ISOCurrencies.php which you're also doing.

### Hiding the currency dropdown

This is the approach the proposed implementation takes. I like this as an option and I've considered a few use cases.

In some cases the user may want to restrict the currency to a single choice… this is important for sorting as it's impossible to do a sort on amount across multiple currencies.

This pattern follows the addition of the timezone for the Date Time Field. 

I think in these cases it's still best to store the currency shown to the user in the data even if the user cannot change it. I'm not sure how one would handle an admin changing the field settings when the currency is set… not sure I'm explaining the issue well 🤷🏻

### Possible Events

A new money field could include an event to update the currency options via a CurrencyList: https://www.moneyphp.org/en/stable/features/currencies.html#currencylist

### Big Brain 🤯

In regards to sorting… the money field could provide a console command to normalise field data to a standard currency using the [Currency Conversion](https://www.moneyphp.org/en/stable/features/currency-conversion.html) integrations. The normalised values could be stored as part of the `getContentColumnType` data model. 

```php
    public function getContentColumnType(): array|string
    {
        return [
            'amount' => Schema::TYPE_DOUBLE,
            'currency' => Schema::TYPE_STRING,
            'normalised_amount' => Schema::TYPE_DOUBLE,
            'normalised_currency' => Schema::TYPE_STRING,
        ];
    }
```

brandonkelly (approved):

brandonkelly:
Appreciate the suggestions, @leevigraham!

Originally the plan was to make this a multi-currency field type, and we were storing the currency in a second content column as you demonstrated. However that led to complications with sorting values across multiple currencies. Your `normalized_currency` idea would technically solve that, but feels a little messy, especially as those values would need to be updated as exchange rates change.

Seems like a good opportunity for a more powerful third party field type :)

PR #10429, opened 2022-01-28
Fixes PHPDoc param errors in namespacing methods

mmikkel:
### Description

The PHPDoc blocks for a few of the namespacing methods in the `craft\helpers\Html` class implies that `null` can be passed for the `$namespace` param, but that doesn't actually seem to be the case.

This PR removes the `null` param type from the PHPDoc blocks for these methods.


brandonkelly:
Doh, thanks!

PR #10454, opened 2022-01-31
Adds support for WOFF2 files in Webpack config

kennethormandy:
### Description

This changes the default Webpack config so `.woff2` files are handled the same way as `.woff` files. Currently, they end up in a different directory, without the filename preserved.


timkelty (approved):
LGTM

nfourtythree (approved):
Yea, good spot!

PR #10460, opened 2022-02-02
Enforce permissions for suspend element action

### LINKED ISSUE #10422: Non-admin users with "Moderate users" permission can suspend admins.
### Description
I have a user group called Manager that is allowed to add new users, assign them to a group called "Editor" with less permissions, as well as suspend/un-suspend them ("Moderate users" permission). Now I have two things that are seem a little strange. But the first issue is by far the more critical one.

Issues:
- Manager (non-admin user) can suspend admin users. That way the admin is locked out of the CP.
- According to the Manager group permission they should be able to assign users to the Editor group only. Instead they are able to assign users to both Editor and their own Manager group.

Expected:
- Manager (non-admin) should be able to suspend and un-suspend non-admin users only, they should not be able to suspend admins.
- Manager should be able to create new users and assign them to the Editor group, but not to the Manager group.

I'm not sure if this is a bug or just how it is meant to work, but it seamed a little bit weird that an admin could be suspended by a non-admin.

Thanks a lot for your help :)

### Steps to reproduce

1. Create a User Group called Manager.
2. Activate the following permissions, especially "Moderate users": 
image
3. Assign a non-admin user to the Manager group.
4. Log in as that user and go to the User/Admins section of CP.
5. Select an admin user and suspend the user through the actions dropdown.
6. Try to login as the suspended admin in a private window. -> Admin is now locked out by a non-admin user.

For the second problem:
1. Add a new user using the "Manager"-user.
2. Assign user groups -> "Manager" is available and assignable even though it's not checked in the permissions list of the "Manager" group.

### Additional info

- Craft version: 3.7.30.1
- PHP version: 8.0.9
- Database driver & version: MySQL 8.0.24
- Plugins & versions:
"carlcs/craft-redactorcustomstyles": "3.0.4",
"craftcms/cms": "3.7.30.1",
"craftcms/feed-me": "4.4.1.1",
"craftcms/redactor": "2.8.8",
"ether/seo": "3.7.4",
"mmikkel/cp-field-inspect": "1.2.5",
"nystudio107/craft-twigpack": "1.2.15",
"ostark/craft-async-queue": "2.3.0",
"sebastianlenz/linkfield": "1.0.25",
"spicyweb/craft-neo": "2.12.3",
"verbb/field-manager": "2.2.4",
"verbb/knock-knock": "1.2.16",
"verbb/super-table": "2.7.1",
"wrav/oembed": "1.3.13"

timkelty:
FIxes #10422

This relocates the permissions enforcement for suspending users to the service, so it can be enforced from the controller and the element action.

Also introduces `Users::canSuspend` (similar to `canImpersonate`), so we can hide the suspend menu item entirely on the user edit screen when the user doesn't have permission.


src/services/Users.php
```diff
@@ -831,10 +845,11 @@ public function suspendUser(User $user): bool
      * Unsuspends a user.
      *
      * @param User $user The user.
+     * @param User|null $moderatingUser The user who is performing the action.
      * @return bool Whether the user was unsuspended successfully.
      * @throws \Throwable if reasons
      */
-    public function unsuspendUser(User $user): bool
+    public function unsuspendUser(User $user, ?User $moderatingUser = null): bool
     {
```
timkelty, line 853: Defaults to null to be non-breaking, and you can suspend a user outright by just not passing second arg.

src/services/Users.php
```diff
@@ -846,6 +861,17 @@ public function unsuspendUser(User $user): bool
...
         }
 
+        if ($moderatingUser) {
+            if (!$moderatingUser->can('moderateUsers')) {
+                throw new Exception('User is not permitted to perform this action');
+            }
+
+            if (!$moderatingUser->admin && $user->admin) {
+                throw new Exception('Only admins can unsuspend other admins');
+            }
```
timkelty, line 871: Something better to throw here than just `Exception`?

src/services/Users.php
```diff
@@ -801,6 +804,16 @@ public function suspendUser(User $user): bool
             return false;
         }
 
+        if ($moderatingUser) {
+            if (!$moderatingUser->can('moderateUsers')) {
+                throw new ForbiddenHttpException('User is not permitted to perform this action');
```
timkelty, line 809: Probably weird to be using `ForbiddenHttpException` in the service…but didn't seem like we had a non-http specific Forbidden exception.

brandonkelly:
I don’t think the `suspendUser()` / `unsuspendUser()` methods should have to worry about permission enforcement in any way, If something calls them, they should just do what they’re told. So moved the `canSuspend()` checks out into the code that’s calling them.

PR #10461, opened 2022-02-02
Feature/220202 search performance

harmputman:
### Description

Only search keywords that are in the pre-filtered results


timkelty:
@harmputman could you rebase your PR so only your changes are on it?

brandonkelly:
This would only affect `craft\services\Search::filterElementIdsByQuery()`, which is deprecated and not in use anywhere in Craft. Nowadays, element queries call `searchElements()` instead, where `$elementIds` will be `null`.

PR #10464, opened 2022-02-02
Removes duplicate Webpack externals config

kennethormandy:
### Description

Currently, there are two `externals` objects in the config. jQuery is in both, so this just removes the redundant one.


timkelty (approved):
good catch!

PR #10465, opened 2022-02-02
Some fixes related to GraphiQL page in RTL

vnali:
GraphiQL on RTL interface causes lots of problems like wrong direction for explorer and code mirror, missing side number when explorer is open and ...

graphql - Copy

Most parts of GraphiQL page is written code in English (explorer, code mirror, history, although code result could be mixed) so this PR suggests to always use LTR orientation for this page.

in this PR only top bar for RTL uses RTL direction.

Untitle1d - Copy

the change to _cp.scss is related to fixing external link icon position

graphqli - Copy

### DIFF
--- a/src/templates/graphql/graphiql.html
+++ b/src/templates/graphql/graphiql.html
@@ -4,7 +4,7 @@
 {% set showHeader = false %}
 
 {% block body %}
-    <div endpoint="{{ url }}" id="graphiql" schemas="{{ schemas|json_encode}}" selectedSchema="{{ {name: selectedSchema.name, schema: selectedSchema.uid, }|json_encode }}">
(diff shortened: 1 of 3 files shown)


vnali:
@andris-sevcenko it seems changes to graphiql.scss files merged but not applied on 3.7.31 release and inspecting graphiql.scss file on graphiql page shows old version. i can see changes to _mixins.scss though.
please let me know what i missed here. thanks

brandonkelly:
@vnali Doh… we needed to rebuild the dist files. Done now (be591dd55b7d60bf50bbc78d60c773800f900f3a), and we just tagged 3.7.32 with it.

PR #10484, opened 2022-02-06
feat: Allow exporters to format generator functions and callables with `::isFormattable() = true`

khalwat:
### Description

This PR allows you to return a php generator function (as the [exporter class recommends](https://github.com/craftcms/cms/blob/develop/src/base/ElementExporterInterface.php#L41)) or other callable that returns array data, and have it be formatted via `::isFormattable() = true`

This makes it easier to use minimal memory when exporting data while still having it formatted, because the generator function or callable can provide the array data row by row.


brandonkelly:
Merged!

brandonkelly:
3.7.31 is out now with that change.

PR #10507, opened 2022-02-09
Add Address Element and Field

lukeholder:
Adds a new “Addresses” field that can optionally be added to the user field layout, for managing user addresses:

The new Addresses field, with Home and Work addresses listed within it

Addresses are elements, and their field layout can be edited from **Settings** → **Users** → **Address Fields**.

In addition to any custom fields, they can have the following native fields:

- Label (mandatory)
- Country (mandatory)
- Address (mandatory)
- Full Name
- Organization
- Organization Tax ID
- Latitude/Longitude

Other element types can implement addresses as well (e.g. orders in Craft Commerce).

Eventually we’d like to add an Addresses custom field as well, but that didn’t make it into this cycle.


lindseydiloreto:
Very exciting to see this taking shape, I'm sure it's a significant undertaking @lukeholder! 👏

It's fascinating that you are creating an Address _element_ alongside the Address _field_... I think it's a great idea with lots of potential (for awhile, I actually considered adding a custom element type to Smart Map).

Mind if I make a fairly simple suggestion? The element type should be named **"Locations"** (not "Addresses").

I say this for a few reasons...

 - Prevents confusion when discussing the field & element type together.
 - A "Location" is a far broader term, encompassing more than just the physical address.

Within the context of a Location, folks will certainly add fields like "Store Hours" and "Phone Number". Those values reference other factors about the _location_, and have nothing to do with the _address_.

Would love to hear your thoughts on this relatively straightforward suggestion!

lukeholder:
@Mosnar 

> Sorry if I'm early to the party, but would you consider adding latitude and longitude columns to the address table?

This is now added to the Address element, thanks!

brandonkelly:
Merged! 🚀

@Mosnar We got latitude/longitude in :)

@lindseydiloreto Addresses are not standalone elements; they belong to users (or other element types that may implement them, such as orders in Craft Commerce), so they won’t be used to represent locations directly.

lindseydiloreto:
Copy that, looking forward to seeing how it all shakes out!

PR #10510, opened 2022-02-09
Add `assertNotPushedToQueue` to Craft Test Helpers

dhahn:
### Description
This adds a test helper to the main Craft test helper class. It is actually just a direct copy of the method `assertPushedToQueue`, but with the assertion flipped. The main reason for adding this is to cover scenarios where a developer wants to explicitly check that a job isn't pushed to the queue. It's not enough to just remove a truthy assertion. 

To solve in our project, we just added that helper to our `_support/` folder as a custom helper. This isn't that big of deal, but I figured if we needed the opposite assertion then another developer would as well. 

Example of how to use in a project:
```
<?php

use Codeception\Test\Unit;

class ExampleQueueTest extends Unit
{
    public $tester;

    public function testGivenSomeCaseJobIsNotPushedToQueue()
    {
        ArbitraryClass::arbitraryCodeIsRunThatWouldConditionallyPushAJob()
        
        $this->tester->assertNotPushedToQueue('Arbitrary Job That Is Conditionally Pushed to a Queue');
    }
} 
```

### Related issues
no related issues

### DIFF
--- a/src/test/Craft.php
+++ b/src/test/Craft.php
@@ -564,6 +564,21 @@ class Craft extends Yii2
             );
         }
     }
+    
+    /**
+     * @param string $description
+     */
+    public function assertNotPushedToQueue(string $description)
+    {
+        if (\Craft::$app->getQueue() instanceof Queue) {
+            $this->assertFalse((new Query())
+                ->select(['id'])
+                ->where(['description' => $description])
+                ->from([Table::QUEUE])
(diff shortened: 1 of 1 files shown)


brandonkelly:
👍🏻

brandonkelly:
Craft 3.7.38 is out now with this method. Thanks for the PR!

PR #10514, opened 2022-02-10
Add en-AU translation file

leevigraham:
### Description

Added an `en-AU` translation file. This `en-AU` translation file imports the `en-GB` translation file.

The purpose of this PR is to add `English (Australia)` to both Language and Formatting user preferences.

image

This is important becuase the `src/i18n/Locale.php` class uses the users locale when formatting numbers and **currency** (new Money Field).

Ideally the `Formatting Locale` drop down options would be independent of the language and include a more exhaustive list of locales.

Example OSX settings:

Screen Shot 2022-02-10 at 10 12 37 pm

### DIFF
--- a/src/translations/en-AU/app.php
+++ b/src/translations/en-AU/app.php
@@ -1,0 +1,3 @@
+<?php
+
+return require __DIR__ . '/../en-GB/app.php';


leevigraham:
@engram-design In that case wouldn't the CP language still default to `en-US` given there's no language file for `en-AU`.

leevigraham:
@engram-design A slightly better solution is to:

1. add `'extraAppLocales' => ['en-AU'],` to `config.general.php`
2. add a new translation file in `/translations/en-AU/app.php`

```php
<?php
# /translations/en-AU/app.php
return require CRAFT_VENDOR_PATH . '/craftcms/cms/src/translations/en-GB/app.php';
```

engram-design:
@leevigraham Well, no idea how it works then! 🤷‍♂️

PR #10525, opened 2022-02-10
Stream log tweaks

timkelty:
Changes:
- Adds and uses an `App::isStreamLog` helper (anticipating supporting an env var check)
- When `isStreamLog`, we _only_ log to `stderr`/`stdout`, not in addition to file like we were
- Previously, we didn't log to `stderr`/`stdout` at all for console requests, as to not pollute the console output. We still don't log info, but we will log errors (otherwise they'd have nowhere to go)

### DIFF
--- a/src/helpers/App.php
+++ b/src/helpers/App.php
@@ -702,34 +702,29 @@ class App
             ],
         ];
 
-        // Console requests should always log to file, as to not pollute console output.
-        if (self::isStreamLog() && !$isConsoleRequest) {
+        if (self::isStreamLog()) {
             $targets[Dispatcher::TARGET_STDERR] = Craft::createObject(array_merge($baseTargetConfig, [
                 'class' => StreamLogTarget::class,
                 'url' => 'php://stderr',
                 'levels' => Logger::LEVEL_ERROR | Logger::LEVEL_WARNING,
             ]));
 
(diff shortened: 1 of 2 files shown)


src/helpers/App.php
```diff
@@ -675,52 +686,45 @@ public static function defaultLogTargets(): array
...
-
-            if ($isConsoleRequest) {
-                $fileTargetConfig['logFile'] = '@storage/logs/console.log';
-            } else {
-                $fileTargetConfig['logFile'] = '@storage/logs/web.log';
-            }
+        // Only log console requests and web requests that aren't getAuthTimeout requests
+        if (!$isConsoleRequest && !Craft::$app->getUser()->enableSession) {
+            return [];
+        }
```
timkelty, line 694: I'm not 💯 what the purpose of this check is, but keeping it in place…I dug up the comment from https://github.com/craftcms/cms/commit/5a3f2781190ae0122dc9f46550161c3480a5c260#diff-8ddcc488c3dc025ab9e7d97a17dedba912ac12119f5f5f7a010e07b4bf981688L559

angrybrad:
@timkelty can you update changelog for the change in behavior as well as the new `App::isStreamLog` method?

PR #10532, opened 2022-02-10
Add prefersReducedMotion fxn to Garnish, update Lightswitch component

gcamacho079:
### Description
Turns off the non-essential animation for lightswitch components when a users has enabled their prefers-reduced-motion setting.

### Related issues
Resolves DEV-248, DEV-197


brandonkelly:
Brilliant! Thanks!

PR #10534, opened 2022-02-10
Configure all constants via env var, stream log tweaks

timkelty:
- Adds `App::constant` to get the value of a constant, falling back to an environment var
- [All constants](https://craftcms.com/docs/3.x/config/#php-constants) now support retrieval from env var.
 
Stream log Changes:

- Adds and uses an `App::isStreamLog` helper
- When `isStreamLog`, we only log to stderr/stdout, not in addition to file like we were
- Previously, we didn't log to stderr/stdout at all for console requests, as to not pollute the console output. We still don't log - info, but we will log errors (otherwise they'd have nowhere to go)

### DIFF
--- a/bootstrap/bootstrap.php
+++ b/bootstrap/bootstrap.php
@@ -7,6 +7,7 @@
  * @license https://craftcms.github.io/license/
  */
 
+use craft\helpers\App;
 use craft\helpers\ArrayHelper;
 use craft\services\Config;
 use yii\base\ErrorException;
@@ -24,8 +25,10 @@ if (!isset($appType) || ($appType !== 'web' && $appType !== 'console')) {
 }
 
 $findConfig = function($constName, $argName) {
-    if (defined($constName)) {
-        return constant($constName);
+    $constant = App::constant($constName);
+
+    if ($constant) {
+        return $constant;
     }
 
     if (!empty($_SERVER['argv'])) {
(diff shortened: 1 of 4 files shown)


src/helpers/App.php
```diff
@@ -65,11 +65,35 @@ class App
...
+
+        $envVal = static::env($envName ?? $constName);
+
+        if ($envVal === false) {
+            return null;
+        }
+
+        return $bool ? filter_var($envVal, FILTER_VALIDATE_BOOL, FILTER_NULL_ON_FAILURE) : $envVal;
+    }
+
```
timkelty, line 96: I considered calling this `App::config` or something, and rolling in the `argv` checks we're doing in bootstrap…

src/helpers/App.php
```diff
@@ -96,15 +106,15 @@ public static function parseEnv(string $str = null)
...
+            if (is_string($str)) {
+                switch (strtolower($str)) {
                     case 'true':
                         return true;
                     case 'false':
                         return false;
                 }
-                $str = $value;
             }
```
timkelty, line 118: Changed this check, as `App::env` will now return `null` when it can't find a var, not `false`.

bootstrap/bootstrap.php
```diff
@@ -139,25 +144,26 @@
     }
 }
 
-if (!defined('CRAFT_EPHEMERAL') || CRAFT_EPHEMERAL === false) {
-    $ensureFolderIsReadable($storagePath, true);
```
timkelty, line 143: It doesn't really make sense for `CRAFT_EPHEMERAL` to be a condition here…
Even in an ephemeral env, craft needs to be able to write to the storage dir.

PR #10542, opened 2022-02-11
Fix UI quirk with editable table

lukeholder:
### Description

Editable table had a small quirk on corner rounding, when the add button is below it:

CleanShot 2022-02-11 at 23 48 36

- Adds a new 'addable' class to modify editable table rounded corners:

CleanShot 2022-02-11 at 23 50 22

Feel free to implement in a better way. I couldn't figure out a css selector that could accomplish this without the need for a css class & javascript.


brandonkelly:
That shouldn't have been getting a dark hover BG to begin with. Just fixed!

PR #10545, opened 2022-02-11
Sidebar updates

gcamacho079:
### Description
@brandonkelly for some reason the sidebar focus wasn't updating based on the `--focus-hsl` variable, but does when the `--focus-ring` variable is re-implemented in the same rule.

Sidebar transition is also removed for users who prefer reduced motion.

### Related issues
Resolves DEV-211


brandonkelly:
Doh, I guess CSS variables that reference other CSS variables don’t get updated if the referenced vars are changed down the cascade.

PR #10554, opened 2022-02-14
Allow relatedTo for fields

timkelty:
### Description
Adds an autosuggest input for entering `relatedTo` IDs in a field context.

Did some rough adjustment for display, but might need to think about guidelines for other rules that might be multi-line. For now I just aligned the row to top and put the label in placeholder.

We also talked about having some kind of warning about the danger of putting in IDs directly.

CleanShot 2022-02-13 at 22 55 29@2x


brandonkelly:
Not working locally for me:

Screenshot of a relation condition rule, where the input is not visible

Screenshot of JavaScript error output

timkelty:
@brandonkelly should be fixed now.

timkelty:
@brandonkelly This is technically working now, but I'm switching it to draft until @lukeholder has a chance to ask the HTMX team about what exactly is going on and what event we should be using.

timkelty:
…or not. Apparently I can't switch it to back to a draft. At any rate – don't merge yet :)

brandonkelly:
Converted the PR to a draft for the time being.

timkelty:
After some back and forth with @lukeholder and the HTMX folks, the best way to inject our JS dependencies prior to `htmx:load` is to use the `transformResponse` method.

The are considering adding a `htmx:newContent` event, which when implemented, we would be able to use for a cleaner approach.

brandonkelly:
Merged!

src/web/assets/htmx/src/htmx.js
```diff
@@ -9,19 +9,22 @@ htmx.defineExtension('craft-cp', {
...
             return;
         }
 
-        const allHeadHtml = evt.detail.elt.querySelectorAll("template.hx-head-html");
-        const allBodyHtml = evt.detail.elt.querySelectorAll("template.hx-body-html");
+        const allHeadHtml = doc.querySelectorAll("template.hx-head-html");
+        const allBodyHtml = doc.querySelectorAll("template.hx-body-html");
 
         for (let i = 0; i < allHeadHtml.length; i++) {
             const headHtml = allHeadHtml[i].innerHTML;
```
lukeholder, line 30: Maybe we should be using https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent

as: 

> textContent gets the content of all elements, including [<script>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script) and [<style>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/style) elements. In contrast, innerText only shows "human-readable" elements.

src/web/assets/htmx/src/htmx.js
```diff
@@ -37,8 +40,11 @@ htmx.defineExtension('craft-cp', {
             }
         }
 
+        return text;
```
lukeholder, line 43: Maybe we should remove the template tags from the `doc` node and return `doc.textContent` here?

lukeholder (commented):
Can discuss on slack.

src/web/assets/htmx/src/htmx.js
lukeholder: Sorry just realized `innerHTML` was being used, not innerText. Ignore.

PR #10558, opened 2022-02-14
Serve files resources directly without publishing for ephemeral

timkelty:
As discussed, this bypasses publishing of resources when `isEphemeral`.

@brandonkelly Alternatively, we could do this in `craft\web\AssetManager`…
It seems like we should at least throw an exception there if you try and use a publish method and `isEphemeral`?


src/helpers/App.php
```diff
@@ -853,7 +853,8 @@ public static function projectConfigConfig(): array
     {
         return [
             'class' => ProjectConfigService::class,
-            'readOnly' => Craft::$app->getIsInstalled() && !Craft::$app->getConfig()->getGeneral()->allowAdminChanges,
+            'readOnly' => self::isEphemeral() || (Craft::$app->getIsInstalled() && !Craft::$app->getConfig()->getGeneral()->allowAdminChanges),
+            'writeYamlAutomatically' => !self::isEphemeral()
         ];
     }
```
timkelty, line 860: I'm not 💯 clear on the interaction between `readOnly` and `writeYamlAutomatically`…

Either way, we don't ever want it to write when `isEphermeral`, so toggling both.

PR #10562, opened 2022-02-14
Attempt to create paths dirs before checking realpath

timkelty:
With an env like:

```
CRAFT_STORAGE_PATH=/tmp/storage/
```

We need to create that dir before doing `realpath`, or they will always come back `false` if they don't exist yet.


brandonkelly:

PR #10569, opened 2022-02-15
Fixed issue where testing queue jobs was broken

Zwartpet:
### Description
I was trying to test my queue jobs and ran into the issue that it results in an exception:
```
[Error] Access to undeclared static property: craft\test\Craft::$app
```

Concluded it used the wrong Craft class.


brandonkelly:
Thanks!

PR #10573, opened 2022-02-15
Config Overrides

timkelty:
**What it does:**

- Implements an abstract `craft\base\Config` for `DbConfig`, `GeneralConfig` to extend.
- On `init`, it checks all of it's public properties and:
    - Checks for an env var override, if `ENV_PREFIX` is set (snake_cased and uppercase).
    - If the value is a string, and the prop is typed as an array, it will split it for you (eg `CRAFT_DISABLED_PLUGINS=foo,bar`)
        - As we were doing this manually for a few props, I was able to remove some redundant code

**Notes:**
- An alternative approach might be to make all props `protected` and implement a `__get`. I did it this way so it it only ever happens once, on `init` (or if you explicilty call `normalize`).
- I set the `ENV_PREFIX` for `GeneralConfig` to `CRAFT_`. Since that is shared with a few constants (and `App::env` now checks constants) we just need to make sure we don't have a constant and a general config prop named the same thing…_or_ we could set the prefix to `CRAFT_GENERAL_`, but that seemed a bit lame.
- I set the `DbConfig` prefix to `DB_`, since that is what our starter uses, but it could just as easily be `CRAFT_DB_`.
- Now that we have a base config class, it might make sense to move the `renamedSettings` logic there.


timkelty:
Added tests 🏳️‍🌈

timkelty:
Was curious how expensive the reflection was, so I did a quick xdebug profile. Seems negligible:

`4.0` branch
no-overrides

`feature/config-overrides` branch:
overrides

brandonkelly:
> * If the value is a string, and the prop is typed as an array, it will split it for you (eg `CRAFT_DISABLED_PLUGINS=foo,bar`)

How would this handle properties that can be either a string or an array, such as [loginPath](https://craftcms.com/docs/3.x/config/config-settings.html#loginpath), where there may be different values for each site?

Also, seems like plugin settings should be factored in, somehow.

timkelty:
> How would this handle properties that can be either a string or an array, such as [loginPath](https://craftcms.com/docs/3.x/config/config-settings.html#loginpath), where there may be different values for each site?

Because those are just typed as `mixed` and don't have `array` in their declared type, they would always just get parsed as string or bool scalars.

I was considering anything beyond a simple csv string to array out of the scope of this, as that comes with more baggage, and there are other tools that already do it: https://github.com/josegonzalez/php-dotenv, https://packagist.org/packages/symfony/dotenv.

If someone needed to set a complex var like via env like this, I would just suggest they do it themselves in general:

```php
[
  'loginPath' => ['siteA' => App::env('SITE_A_LOGIN_PATH'), 'siteB' => App::env('SITE_B_LOGIN_PATH')]
]
```

We could go down the road of parsing these, or implement env dot notation, but that seems like a whole other concern.

PR #10605, opened 2022-02-17
Feature/better json responses

timkelty:
- Fixes #8631
- Removes `Craft.postActionRequest` in favor of Craft.sendActionRequest
- Commerce PR: https://github.com/craftcms/commerce/pull/2708
- Redactor PR: https://github.com/craftcms/redactor/pull/369


src/web/assets/updater/src/Updater.js
```diff
@@ -77,7 +75,7 @@ import './update.scss';
             }
 
             if (state.nextAction) {
-                this.postActionRequest(state.nextAction);
+                this.sendActionRequest('POST', state.nextAction);
```
lukeholder: There is no internal `sendActionRequest` method. Do we want to use the internal postActionRequest method or

Issue #7541, opened 2021-02-10
Register new user- accessibility


ajanec01:
#### Account

- The Account and Permissions tabs need appropriate role, state  and properties- see [aria example with automatic activation](https://www.w3.org/TR/wai-aria-practices-1.2/examples/tabs/tabs-1/tabs.html) and [aria example with manual activation](https://www.w3.org/TR/wai-aria-practices-1.2/examples/tabs/tabs-2/tabs.html)
![tabs](https://user-images.githubusercontent.com/47136246/107581543-21e65b80-6bf0-11eb-8753-ad36b4a0e301.png)
- Status update for not being able to save a user not picked up by screen reader software, although it wouldn’t need to be picked up if the following point was implemented
![status](https://user-images.githubusercontent.com/47136246/107581471-04b18d00-6bf0-11eb-8a28-b3832cd89863.png)
- It would be good if the error messages were associated with input fields using `aria-describedby` and the keyboard focus was moved to  the first incorrect field 
- It would be good to either move the save button below the form or to make sure it is part of  the `<form>` element  so that the user can just hit enter to save changes

#### Permissions

- Checkboxes should be grouped to provide context
- The select All elements are not keyboard accessible
- There is 1.47:1 contrast ratio between the dark body text and the Select All elements. Technically, bumping contrast ratio to at least 3:1 would work for WCAG compliance but I’d recommend using another visual clue, e.g. underline, or border with 3:1 contrast ratio against the background
- Information buttons need aria-expanded and aria-controls; I also think it would be better if they didn’t receive keyboard focus when the checkbox is disabled and their name was more descriptive (e.g. More about moderate users)
![pop-ups](https://user-images.githubusercontent.com/47136246/107581583-332f6800-6bf0-11eb-9724-9646c4459b63.png)

### Additional info

- Craft version: 3.6.5

PR #10606, opened 2022-02-17
Tabs a11y

gcamacho079:
### Description
Updates tab JS and markup to use [Tabs with Manual Activation pattern](https://www.w3.org/TR/wai-aria-practices/examples/tabs/tabs-2/tabs.html)

### Related issues
Addresses one role/state/property issue raised in #7541 

Resolves DEV-208


src/templates/_includes/tabs.twig
```diff
@@ -8,38 +8,37 @@
         class: (tab.class ?? [])|explodeClass,
 })) %}
 
-{% tag 'nav' with containerAttributes %}
-    <ul class="scrollable">
+{% tag 'div' with containerAttributes %}
+    <div class="scrollable" role="tablist" aria-label="{{ 'Fields'|t('app') }}">
```
gcamacho079, line 12: @brandonkelly targeted this for the Entry page, but where's would one pass an appropriate label depending on the template?

src/models/FieldLayoutForm.php
```diff
@@ -69,6 +69,11 @@ public function render(bool $showFirst = true): string
                     'id' => $id,
                     'layout-tab' => $tab->getUid(),
                 ],
+                'role' => 'tabpanel',
+                'tabindex' => '0',
+                'aria' => [
+                    'labelledBy' => 'tab-' . $tab->getId(),
```
gcamacho079, line 75: This seems hacky - better way to specify the ID of the tab button that activates the tabpanel?
Screenshot showing accessibility information of tab button

src/templates/_includes/tabs.twig
brandonkelly: If we want this to be configurable, you could use a new `tabContainerLabel` variable:

```twig
aria-label="{{ tabContainerLabel ?? 'Fields'|t('app')"
```

src/models/FieldLayoutForm.php
brandonkelly: This is maybe a little bit cleaner: d14fa747d38fb0613545dab5cd2d0ab21863014f

PR #10612, opened 2022-02-18
Setup `{% d %}` Twig tag to dump using symfony/var-dumper

aaronbushnell:
Now that `symfony/var-dumper` is installed in the Craft 4 core it might be useful to expose a Twig tag for dumping a variable, but not die out. 

```twig
{# Dump the request out and die #}
{% dd craft.app.request %}

{# Dump the request out using Symfony's VarDumper #}
{% d craft.app.request %}
```

Just a thought!

### DIFF
--- a/src/web/twig/Extension.php
+++ b/src/web/twig/Extension.php
@@ -36,6 +36,7 @@ use craft\web\twig\nodevisitors\Profiler;
 use craft\web\twig\tokenparsers\CacheTokenParser;
 use craft\web\twig\tokenparsers\DdTokenParser;
 use craft\web\twig\tokenparsers\DeprecatedTokenParser;
+use craft\web\twig\tokenparsers\DTokenParser;
 use craft\web\twig\tokenparsers\ExitTokenParser;
 use craft\web\twig\tokenparsers\HeaderTokenParser;
 use craft\web\twig\tokenparsers\HookTokenParser;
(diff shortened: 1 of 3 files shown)


brandonkelly:
Thanks for the PR! This feels like it’s already covered by the `dump()` function though?

(We only added `symfony/var-dumper` to avoid an error when calling `Collection::dump()`.)

aaronbushnell:
You bet, @brandonkelly! This is a _bit_ nicer than just `{{ dump() }}` in that VarDumper will tidy up the output, collapse the data, apply some minor syntax highlighting, etc. Here's a quick screenshot of the two outputting some request data.

And, of course, if you all feel this isn't worth the new tag, more documentation, etc. and it's better to allow folks to leverage a module/plugin for this that's certainly okay!

Screen Shot 2022-02-18 at 3 33 56 PM

Screen Shot 2022-02-18 at 3 34 07 PM

brandonkelly:
I think I’d prefer to just move `Craft::dd()`, the `dump()` function and `{% dd %}` over to use `symfony/var-dumper`, to be consistent.

aaronbushnell:
That makes sense, @brandonkelly! I know the `dump()` function is baked into Twig directly so I wasn't sure if that was better to preserve as-is or move over. I'll close this one out!

PR #10615, opened 2022-02-18
Add return type

Seldaek:
### Description

Forward compatibility with upcoming Composer versions.

Note that as of Composer 2.3, you should be able to delete this entirely as I am making the ArchiveManager initialization conditional based on which classes exist in https://github.com/composer/composer/pull/10547 - not released yet tho!


brandonkelly:
Thanks @Seldaek!

PR #10624, opened 2022-02-22
Full names for users

brandonkelly:
### Description

Replaces the ºFirst Name” and “Last Name” fields on Edit User pages with a new “Full Name” field.

When a user is saved, its First/Last Name values will be automatically parsed from its Full Name, if set, and those values are still stored so they can be shown/sorted from user indexes.

There’s also a new `fullName` user query param which maps to the new `fullName` values.

The `users/save-user` action still supports posting separate `firstName`/`lastName` values; they will be automatically merged into `fullName` on save.

### Related issues

- #10405


carolineboeykens:
Hi, we are having an issue with a user who has two first names (in two different words).
But the `user.firstName` now only shows the first word. Is there a work around here?

brandonkelly:
@carolineboeykens @hassanmah Can one of you please post a new issue about this?

kevinmu17:
We have also cases where there are 2 firstnames and a last name with an infix. for example Jan Vennegoor of Hesselink.
How to handle this when you want to present the values separated?

brandonkelly:
As of Craft 4.6, there’s a new [showFirstAndLastNameFields](https://craftcms.com/docs/4.x/config/general.html#showfirstandlastnamefields) config setting which you can enable to go back to showing separate “First Name” and “Last Name” fields.

PR #10629, opened 2022-02-23
Tables

gcamacho079:
### Description
Updates to make table sorting functionality keyboard and screen reader accessible

### Related issues
Resolves DEV-214


src/web/assets/cp/src/js/TableElementIndexView.js
```diff
@@ -19,9 +21,20 @@ Craft.TableElementIndexView = Craft.BaseElementIndexView.extend({
...
+        // Set table caption
+        this.$tableCaption = this.$table.find('caption');
+
+        this.$statusMessage = this.$table.parent().find('[data-status-message]');
+
         // Set the sort header
         this.initTableHeaders();
 
+        // Add callback for after elements are updated
+        this.elementIndex.settings.onUpdateElements = () => {
```
gcamacho079, line 33: Feel iffy about setting this directly as there might already be something assigned. Is there a better way you can recommend @brandonkelly ?

src/templates/_elements/tableview/container.html
```diff
@@ -1,14 +1,24 @@
+{% set tableName = tableName ?? '' %}
+{% set instructionId = instructionId ?? "sort-instructions-#{random()}" %}
+
 <div class="tableview{% if context == 'index' %} tablepane{% endif %}">
-    <table class="data fullwidth"{% if structureEditable is defined and structureEditable %} data-structure-id="{{ structure.id }}" data-max-levels="{{ structure.maxLevels }}"{% endif %}>
-        <thead>
+    <span class="visually-hidden" data-status-message aria-live="assertive"></span>
+    <table class="data fullwidth"{% if structureEditable is defined and structureEditable %} data-structure-id="{{ structure.id }}" data-max-levels="{{ structure.maxLevels }}"{% endif %} data-name="{{ tableName }}">
+      <caption class="visually-hidden">{{ tableName }}<span id="{{ instructionId }}" data-sort-instructions>, column headers with buttons are sortable</span></caption>
```
gcamacho079, line 7: Do all tables have at least one sortable column? If not, the sort instructions may need to be conditional.

src/web/assets/cp/src/js/TableElementIndexView.js
```diff
@@ -19,9 +21,20 @@ Craft.TableElementIndexView = Craft.BaseElementIndexView.extend({
...
+            this.$table.find('th [aria-pressed="true"]').focus();
```
gcamacho079, line 34: `onUpdateElements` seems to run the first time an element select modal is opened as well. Hoping to focus on the pressed button only after a user explicitly sorts by pressing a button - any suggestions? Tried adding to the `handleSelectedSortHeaderClick` and `handleUnselectedSortHeaderClick` functions, but couldn't time them to run after the element update.

brandonkelly: Fixed by just registering an `updateElements` event:

https://github.com/craftcms/cms/blob/0d32e98121cb819d5b1eb1d52955d66c2cf55bb5/src/web/assets/cp/src/js/TableElementIndexView.js#L33

src/templates/_elements/tableview/container.html
brandonkelly: Unfortunately it’s not known by the template, but in general, yes they will be.

src/web/assets/cp/src/js/TableElementIndexView.js
brandonkelly: Fixed by just registering an `updateElements` event:

https://github.com/craftcms/cms/blob/0d32e98121cb819d5b1eb1d52955d66c2cf55bb5/src/web/assets/cp/src/js/TableElementIndexView.js#L33

src/web/assets/cp/src/js/TableElementIndexView.js
brandonkelly: Fixed this from the base element index class, which now captures the active element before it begins swapping out the element index view with the new one:

https://github.com/craftcms/cms/blob/0d32e98121cb819d5b1eb1d52955d66c2cf55bb5/src/web/assets/cp/src/js/BaseElementIndex.js#L1762-L1763

Then checks if that element is no longer part of the DOM, and if so, brings focus back to a new element with the same ID (if one exists):

https://github.com/craftcms/cms/blob/0d32e98121cb819d5b1eb1d52955d66c2cf55bb5/src/web/assets/cp/src/js/BaseElementIndex.js#L1920-L1922

PR #10642, opened 2022-02-24
Cache asset path hashes.

timkelty:
This allows cpresources to work in the web installer, when CRAFT_EPHEMERAL is true.


brandonkelly:

PR #10653, opened 2022-02-26
Make properties private

bencroker:
Changed the visibility of these properties to private, which the underscores and getter methods would seem to suggest they should be.


brandonkelly:
Thanks Ben!

PR #10655, opened 2022-02-27
Return new `static`

bencroker:
By replacing `self` with `static`, it makes it possible for extending classes to use this method. 

An example is in Sprig's `PaginateVariable` class:
https://github.com/putyourlightson/craft-sprig-core/blob/0f181fee2012c9a629c509ee5341f4037ca96a19/src/variables/PaginateVariable.php#L17-L23

With `return new self`, a `Paginate` class is returned. With `return new static`, a `PaginateVariable` class is returned, which is the desired beaviour.


bencroker:
Closing this in favour of https://github.com/craftcms/cms/pull/10656.

PR #10656, opened 2022-02-27
Revert `new self()`

bencroker:
This PR reverts incorrect changes made in https://github.com/craftcms/cms/commit/f61a082277f5c4b88c332232deeaff635c9b54c6, in which `new static()` was changed to `new self()`, breaking calls from extending classes that use these methods, such as Sprig's `PaginateVariable` class:
https://github.com/putyourlightson/craft-sprig-core/blob/0f181fee2012c9a629c509ee5341f4037ca96a19/src/variables/PaginateVariable.php#L17-L23


brandonkelly:
Thanks! Looks like PHPStan isn’t even complaining about those lines anymore.

PR #10677, opened 2022-03-03
4.0.0-alpha.1 — Quick Post Widget patch

AugustMiller:
### Description
Replaces a call to `FieldLayout::getFields()` with the renamed `FieldLayout::getCustomFields()`.

### Related issues
None, sorry! Figured I'd just patch and not clutter up the issues tab. 😉


brandonkelly:
Why thank you!

PR #10678, opened 2022-03-03
4.0.0-alpha.1 — Expanded element exporter patch

AugustMiller:
### Description
Same as #10677, this just replaces a couple calls to `FieldLayout::getFields()` in the Expanded element export class.

### Related issues
- #10677


brandonkelly:
🙏

Issue #9835, opened 2021-09-15
Email has already been taken


jerome2710:
### Description
After receiving some questions about users being unable to save their account, we started digging what could be wrong. They were reporting messages of "Email X has already been taken", which would be triggered if there already is an account for that e-mail address. Picking some e-mail addresses of these reports, all of them are having more than 1 account.

Running the following query:

```sql
SELECT email, COUNT(*) AS count FROM craft_users GROUP BY email HAVING count > 1
```

we found out that 75 e-mail addresses have more than 1 account. Some even having up to 6. Some of these accounts were not yet activated through the confirmation e-mail and having the status "pending", some were (all) already activated. Some were all created within the hour, some over the course of multiple days/months. Some were created back in 2019, some were created this week.

We are only having one user registration page. When trying to register myself twice before clicking the first confirmation email, I already receive the error message. Our registration page (minus some styling) looks like this:

```twig
<form method="post">
	{{ actionInput('users/save-user') }}
	{{ redirectInput('register/check-email') }}
	
	<input id="email" type="email" name="email" required="required"{%- if user is defined %} value="{{ user.email }}"{% endif %}>
	
	<input id="password" type="password" name="password" required="required">
	
	<input id="confirm-password" type="password" name="confirm-password" required="required" data-bouncer-match="#password">
	
	<input type="checkbox" id="termsConditions" name="termsConditions" required="required">
	
	<input type="checkbox" id="newsletter" name="newsletter">
	
	<input type="checkbox" id="partners" name="partners">
	
	<button type="submit">
		Register
	</button>
</form>
```

I am unsure where to start. We are having a few plugins and custom modules enabled which are narrowly integrated in the website. I am unable to reproduce this issue at the moment, which is essential before I could start stripping them out to narrow down the cause. Does anyone have a clue? Does some user-status allow the user to register under the same e-mail address twice?

### Additional info

- Craft version: Craft Pro 3.7.10
- PHP version: 7.2.34
- Database driver & version: MySQL 5.5.68
- Plugins & versions:
    - Advanced Url Field	1.0.0
    - Asset Rev	6.0.2
    - Commerce Product Bundles	dev-develop-v2
    - Control Panel CSS	2.3.1
    - Craft Commerce	3.4.2
    - Expanded Singles	1.1.4
    - Gift Voucher	2.6.1
    - Image Resizer	2.1.0
    - Inventory	2.1.1
    - Mollie for Craft Commerce	3.0.0
    - Neo	2.11.7
[...]

PR #10680, opened 2022-03-03
Fix logic flaw in new user save

### LINKED ISSUE #10661: Double-click on register form creates multiple users with the same email address
### Description

We noticed some duplicate user accounts appearing that were being created at the same time and managed to have the same email address.

If I use our public registration form to create a new account using an email address that _already exists_, the form validation returns false and I'm told a user already exists with that email.

However, if I try to create a new account using an email address that doesn't exist, and I double click the form's submit button, two accounts are created.

Screenshot 2022-03-02 at 10 03 06

The only other Github issue I've found that might be related is #9835

### Steps to reproduce

1. Follow the steps in [this knowledge base guide](https://craftcms.com/knowledge-base/front-end-user-accounts) to allow public registrations and create a frontend registration form
2. Double-click the form's submit button to submit the form twice successively

### Additional info

- Craft version: 3.7.35
- PHP version: 7.4.26
- Database driver & version: mysql 5.7

timkelty:
### Related issues
Fixes #10661, #9835

### DIFF
--- a/src/controllers/UsersController.php
+++ b/src/controllers/UsersController.php
@@ -1245,10 +1245,9 @@ JS;
         }
 
         // Manually validate the user so we can pass $clearErrors=false
-        if (
-            !$user->validate(null, false) ||
-            !Craft::$app->getElements()->saveElement($user, false)
-        ) {
+        $success = $user->validate(null, false) && Craft::$app->getElements()->saveElement($user, false);
+
+        if (!$success) {
             Craft::info('User not saved due to validation error.', __METHOD__);
 
             if ($isPublicRegistration) {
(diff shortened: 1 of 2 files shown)


src/controllers/UsersController.php
```diff
@@ -1245,10 +1245,9 @@ public function actionSaveUser()
         }
 
         // Manually validate the user so we can pass $clearErrors=false
-        if (
-            !$user->validate(null, false) ||
-            !Craft::$app->getElements()->saveElement($user, false)
-        ) {
```
timkelty, line 1251: The way it was, it would call saveElement (w/ validation) when `validate` failed.

PR #10691, opened 2022-03-05
Add defineBehaviors() to base Model

bencroker:
This PR adds a `defineBehaviors()` to the base `Model` class, making it easier for models to attach behaviors without having to call the `parent` method, similar to `defineRules()`. It also uses the `behaviors` attribute of the `DefineBehaviorsEvent` event so that plugins can modify behaviors already defined.

Usage before:
```php
    public function behaviors(): array
    {
        $behaviors = parent::behaviors();
        $behaviors['parser'] = [
            'class' => EnvAttributeParserBehavior::class,
            'attributes' => ['apiKey'],
        ];

        return $behaviors;
    }
```

Usage after:
```php
    public function defineBehaviors(): array
    {
        return [
            'parser' => [
                'class' => EnvAttributeParserBehavior::class,
                'attributes' => ['apiKey'],
            ],
        ];
    }
```

### DIFF
--- a/src/base/Model.php
+++ b/src/base/Model.php
@@ -87,9 +87,14 @@ abstract class Model extends \yii\base\Model
      */
     public function behaviors(): array
     {
-        // Fire a 'defineBehaviors' event
-        $event = new DefineBehaviorsEvent();
+        $behaviors = $this->defineBehaviors();
+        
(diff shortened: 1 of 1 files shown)


brandonkelly:
Like it, thanks!

PR #10695, opened 2022-03-07
Don't assume every field layout type is an element

engram-design:
For Vizy, we use a model for the owner of field layouts. This is because when configuring the field settings, you can have multiple blocktypes, each with their own field layout. Currently, this will assume the `type` of a field layout is an element, and try to run conditionals based on that. Being a model, it's not an element.

I'm not sure if I'm abusing field layouts (probably?), but this would be great.

### DIFF
--- a/src/base/FieldLayoutComponent.php
+++ b/src/base/FieldLayoutComponent.php
@@ -204,7 +204,7 @@ abstract class FieldLayoutComponent extends Model
             /** @var ElementInterface|string|null $elementType */
             $elementType = $this->getLayout()->type;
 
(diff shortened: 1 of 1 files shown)


brandonkelly:
I’ll take it, but field layouts are currently definitely expected to only be used for element types. I’m not 100% sure what other side effects there might be when used elsewhere.

engram-design:
Yeah I figured I was pushing it with this. I believe the type is a non-null value, so I need to put _something_ in there. I'm not sure if other plugins out there might do this as well. But thanks!

PR #10696, opened 2022-03-07
Fix `$_parentId` for elements

engram-design:
Looks like the typehints are correct, just not the code:

```
/**
 * @var int|false|null Parent ID
 * @see getParentId()
 * @see setParentId()
 */
private self|false|null $_parentId = null;
```

This will error when calling `setParent()` which will try and set the `$_parentId` to an int, but also of course when calling `setParentId()`.


brandonkelly:
Whoops, thanks!

PR #10698, opened 2022-03-07
Transform Filesystems

andris-sevcenko:
### Description
- This PR adds an additional transform FS to Volumes that defaults to the FS used for Volume. This allows Volumes with Filesystems that have no public URLs to expose transforms, anyway, which, in turn, allows Craft to use transforms for thumbnails.

- It also shuffles some logic around - transforms are more centered on Asset elements, instead of the Asset service and some events are moved around to reflect that.

- It adds `ImageEditorTransformerInterface`, which a transformer can choose to implement if it supports the functionality required by the Image Editor.


src/controllers/AssetsController.php
```diff
@@ -822,54 +821,39 @@ public function actionSaveImage(): Response
                 throw new BadRequestHttpException('Invalid cropping parameters passed');
             }
 
+            // TODO Fire an event for any other image editing takers.
+            $transformer = new ImageTransformer();
```
brandonkelly, line 825: @andris-sevcenko Did you forget to add this event?

andris-sevcenko: 50/50. There are no other image editing takers yet, so I skipped this part.

Issue #9738, opened 2021-08-18
Broken cpresources behaviour in load balanced environments without a shared filesystem


meestaben:
### Description

This change in 3.6.18:

https://github.com/craftcms/cms/commit/9f13b36828e5dc57573aae234b6b45096b5f645d

breaks cpresources URLs in load balanced environments where code is deployed independently to each box, and cpresources directory is not on a shared filesystem. 

The change was originally discussed on #9140, with the aim to prevent old assets from being incorrectly cached downstream during a rolling deployment. 

In versions prior to 3.6.18 everything worked fine, even though you'd in theory end up with as many copies of each logical asset as you have application servers running (all with different hashes in their path). 

Reverting the above change fixes the issue for us.

### Steps to reproduce

1. Be running a load balanced environment with Craft >=3.6.18 and no shared filesystem except for `./storage`
2. Visit the admin CP
3. Reload the page repeatedly and observe that on each load several cpresources assets will 404, depending on which server is being hit.

### Additional info

- Craft version: >= 3.6.18
- PHP version: 7.4.15
- Database driver & version: MySQL 5.7 (AWS Aurora)
- Plugins & versions:

brandonkelly:
Thanks for reporting that. Just released Craft 3.7.10 which reverts the changes made for #9140.

andygrn:
@brandonkelly Is there a way to fix this without losing the cache headers? Our control panels on 3.7.10+ have become unacceptably slow again, for the same reason outlined in #9140.

brandonkelly:
@andygrn I’ll have to think about it some more. We can’t blindly add the cache headers because that risks browsers caching an outdated file if the deployment hadn’t hit the current server yet (https://github.com/craftcms/cms/issues/9140#issuecomment-877521916), and we can’t validate based on the hash because that can change from server to server (as pointed out by this issue). So it’s tricky. Ideally we could start allowing you to use an S3 bucket or something.

andygrn:
@brandonkelly Presumably you could force browsers to re-request changed assets using a change to the ?v= string between Craft versions, although I'm not certain that would prevent jmortlock's problem from #9140.

Some sort of native cpresources volume would be very welcome! I suspect we'll have to figure out a workaround in the meantime, I've seen some people using custom AssetManager components to handle it...

Issue #9140, opened 2021-07-08
Streamed cpresources are not browser cached


andygrn:
### Description

We're running Craft in a load-balanced environment, where cpresources are not publically accessible, and therefore Craft is streaming them on-demand (as outlined in https://craftcms.com/knowledge-base/configuring-load-balanced-environments#cpresources-folder).

However these files are being served with anti-cache headers (e.g. Expires: 0), so every file is being re-downloaded on every CP request. Obviously this leads to very slow CP page loads compared to a directly-served copy of the site.

Cpresource URLs are already versioned so presumably this isn't intended behaviour.

### Steps to reproduce

1. Run Craft in an environment without public /cpresources access.
2. Load CP pages with a network inspector, see each cpresource downloaded every time.

### Additional info

- Craft version: 3.6.17
- PHP version: 7.4.14
- Database driver & version: Mysql 5.5.5
- Plugins & versions: craftcms/aws-s3 1.2.11, craftcms/redactor 2.8.7

brandonkelly:
Thanks for reporting that! Just fixed for the next release.

jmortlock:
Hi @brandonkelly 

I'm not entirely sure this is a safe change as its likely for those that are running in a Load Balanced environment are also deploying with a rolling type mechanism, i.e. During a craft update there could be 2 different versions of Craft serving assets at the same time.

Due to the ability for assets to be looked up in the database its conceivable for an old node during this deployment phase to serve up its old (outdated) version of an asset via the newly published hash, which after this change will get cached at the client.

This is not a new issue but due to the non-client cached behaviour it is relatively short lived (minutes) until all nodes settle on the new version.

brandonkelly:
@jmortlock Any time any assets change, they get republished to a new URL, so this change won’t have any effect on updates.

jmortlock:
Maybe i misunderstand the whole cpresources caching mechanisim, but from my understanding old nodes will still happily publish assets that match that new URL, as the Hash / Source is stored in the database as per changes made in this PR https://github.com/craftcms/cms/issues/2500

For example The new server say Craft 3.8 publishes a new jquery Asset, this is now the hash stored in the database, however during a rolling update with say Craft 3.7 / 3.8 running concurrently the older version will happily accept the new hash value, read the source from the database and publish up its version of the asset using the newer hash value.

brandonkelly:
[...]

PR #10705, opened 2022-03-07
Use `buildId` to determine caching headers for a load-balanced/rolling deployment

timkelty:
### Description
This allows us to send CP resources with cache headers, by introducing a new config setting `GeneralConfig::$buildId`.

When set, CP resource URLs will be generated with a `buildId=xxx` query param.

Only if the `buildId` from the request matches the config setting, or `GeneralConfig::$buildId` is left as `null`, will we serve cache headers.

A load-balanced environment should set `GeneralConfig::$buildId` to something unique to the deployment, such as a `GIT_SHA` env, or a deployment timestamp.

### Related issues
- https://github.com/craftcms/cms/issues/9140
- https://github.com/craftcms/cms/issues/9738


timkelty:
Docs note for @mattstein:
- https://craftcms.com/knowledge-base/configuring-load-balanced-environments
- https://craftcms.com/docs/3.x/config/config-settings.html

brandonkelly:

mattstein:
Already [in the 4.x docs](https://craftcms.com/docs/4.x/config/config-settings.html#buildid), and just added a tip here: https://craftcms.com/knowledge-base/configuring-load-balanced-environments#cpresources-folder

PR #10706, opened 2022-03-07
Auto-property typecasting

brandonkelly:
### Description

Adds a new `craft\helpers\Typecast::properties()` method, which can be used to typecast a given array of property names/values, based on the type declarations of the given class.

Used in the following places:

- `craft\base\Model::__construct()`
- `craft\base\Model::setAttributes()`
- `craft\helpers\Component::createComponent()`

`craft\base\Model::datetimeAttributes()` is deprecated now, and date/time attributes are recommended to use a `DateTime` type declaration instead. We can remove `datetimeAttributes()` entirely in Craft 5.

### Related issues

- #10689

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -246,6 +246,7 @@
 - Added `craft\helpers\Number::toIntOrFloat()`.
 - Added `craft\helpers\ProjectConfig::encodeValueAsString()`.
 - Added `craft\helpers\ProjectConfig::ensureAllSectionsProcessed()`.
+- Added `craft\helpers\Typecast`. ([#10706](https://github.com/craftcms/cms/pull/10706))
 - Added `craft\i18n\Translation`.
 - Added `craft\imagetransforms\ImageTransformer`.
 - Added `craft\models\AssetIndexingSession`.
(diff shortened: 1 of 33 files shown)


khalwat:
Feels like perhaps this should be done via the existing `AttributeTypecastBehavior`? https://www.yiiframework.com/doc/api/2.0/yii-behaviors-attributetypecastbehavior

Or are you eschewing that because it doesn't handle `array` types?

Given that I've already added this behavior to a number of my models, it'd be nice if it were extended or such instead of a new way to do the same (but slightly better) thing?

khalwat:
Further thought... the point of typing is to catch errors where you're setting, say, a `string` on an` int` -- all this really does is bring things back to PHP pre-7, where variables are auto-coalesced into the appropriate type. So you'll never catch subtle errors.

I realize it's more convenient but aren't we side-stepping the reason for typing to begin with?

I feel like the real issue here is that the value coming to the controller shouldn't be an empty string, it should be an empty array.

brandonkelly:
@khalwat Responded over at #10689

michaelrog:
Very cool! Curiously, what's the perf impact of spinning up a Reflection class on every single model instantiation? I would expect it to be negligible for the handfuls of objects in typical requests... but is it something we can/should try to address for bulk cases where we're instantiating hundreds or thousands of models at a time?

brandonkelly:
@michaelrog Yeah it’s negligible, and we’re caching the type results for each unique class that needs to be typecasted.

brandonkelly:
Also worth mentioning there’s already a fair amount of reflection, for `fields()` and other things.

PR #10707, opened 2022-03-08
Don't override title when duplicating element

### LINKED ISSUE #10702: Duplicate doesn't duplicate titles of multi-site entries and products
### Description
When on the overview page of entries or products, and you select an entry, you click on the gear icon on top and select "Duplicate". The entry is duplicated and all fields are duplicated in the correct language if you have a multi-site. Only the title is in the "Default site" title with "copy" added to the end. It is not the translated title for all languages.

For example and entry with these titles:
- en-US: Craft CMS EN
- nl-NL: Craft CMS NL
- de-DE: Craft CMS DE
becomes after duplicate in all languages "Craft CMS EN copy".

But if you are on an entry page and you select "Save as new entry", the title are saved in all languages. 

- Craft version: 3.7.36

timkelty:
### Description
Overriding the title seems unnecessary, and has [worse consequences with multi-site elements](https://github.com/craftcms/cms/issues/10702).

### Related issues
Fixes #10702


jordikroon:
Hi @brandonkelly. One of our customers reported to us that they find it confusing that Copy is not added anymore after an article after duplicating.

The change feels less than a bug-fix, more as a removal of a feature to us. Would it be possible to have a fix for this issue that adds "Copy" correctly over multi-site elements.

Thank you :)

brandonkelly:
@jordikroon Please start a new discussion about it.

PR #10717, opened 2022-03-09
4.0 Fix sendmail command being incorrectly formatted

engram-design:
Getting an error with the default Sendmail mailer, trying to send a test email on 4.0.

```
Symfony\Component\Mailer\Exception\InvalidArgumentException
The "/usr/sbin/sendmail" mailer DSN must contain a scheme.
```

I'm pretty sure it's because the DSN is being sent incorrectly, due to incorrect grouping with the ternary operator here 

https://github.com/craftcms/cms/blob/7594767f9f326e7c6810037311dfe1046c2e3323/src/mail/transportadapters/Sendmail.php#L121

This results in:

```
[
    'dsn' => '/usr/sbin/sendmail -bs'
]
```

Where is should be:

```
[
    'dsn' => 'sendmail://default?command=/usr/sbin/sendmail -bs'
]
```

This PR just fixes that so that the default command isn't used without the `sendmail://default?command=` part.

screencapture-craft4-test-admin-settings-email-2022-03-09-14_43_12

### DIFF
--- a/src/mail/transportadapters/Sendmail.php
+++ b/src/mail/transportadapters/Sendmail.php
@@ -118,7 +118,7 @@ class Sendmail extends BaseTransportAdapter
     public function defineTransport(): array|AbstractTransport
     {
         return [
-            'dsn' => 'sendmail://default?command=' . $this->command ? App::parseEnv($this->command) : self::DEFAULT_COMMAND,
+            'dsn' => 'sendmail://default?command=' . ($this->command ? App::parseEnv($this->command) : self::DEFAULT_COMMAND),
         ];
     }
 


brandonkelly:
Thanks! Fixed it in a slightly different way, which accounts for `$this->command` being set to an environment variable that is set to an empty value.

engram-design:
Thanks!

engram-design:
Ah, so sorry to raise this from the dead, new error now (attached)
.

`'sendmail://default?command=/usr/sbin/sendmail'` is missing the ` -bs` command which is being stripped out. Debugging at https://github.com/symfony/mailer/blob/f6e927ec95c957131e6b2c78790e1a6d4c576447/Transport.php#L161 show the above command is being fired. I assume this is the mailer's issue, but have even tried hard-coding that to `sendmail://default?command=/usr/sbin/sendmail%20-bs` without luck.

screencapture-craft4-test-admin-settings-email-2022-03-09-17_02_20

brandonkelly:
@engram-design suggestion on a fix?

engram-design:
Will dive in further for another PR, or see where in the chain it's going wrong (Craft, Yii-Symfony-Mailer, Symfony Mailer). Is it working for you? I'm just using the defaults.

PR #10718, opened 2022-03-09
Escape spaces in Sendmail commands for emails

engram-design:
A fix for https://github.com/craftcms/cms/pull/10717 - I was going to use `urlencode()` but feel wrong because it's not really a URL. The command won't work without this, and looks correct according to the docs https://symfony.com/doc/current/mailer.html#other-options


brandonkelly:
Thanks for figuring it out!!

PR #10735, opened 2022-03-11
Set visibility to protected

bencroker:
Sets the visibility of `defineFieldLayouts()` to `protected`, to match the visibility of the parent method: 
https://github.com/craftcms/cms/blob/6f5d4df5eebe733ecb8b7bb4eb7147d0198cc31c/src/base/Element.php#L703

This PR should ideally be merged into the 4.0 branch as well.


brandonkelly:
Thanks!

PR #10736, opened 2022-03-11
Fix @var type hint

bencroker:
Makes PhpStorm happy.


brandonkelly:
🙏🏻

brandonkelly:
Not sure how this didn’t break the build, but this introduces a PHP error because the `*/` closes the actual multiline comment prematurely.

bencroker:
Oh, my bad.

PR #10761, opened 2022-03-16
Wrong type for Config::$_dotEnvPath

ostark:
`Craft::$app->getConfig()->getDotEnvPath()` returns `'1'` when using strict types.

The reason is a wrong type of `craft\services\Config::$_dotEnvPath`;

---

The [ConfigTest](https://github.com/craftcms/cms/pull/10761/commits/69fc7dbbf37e9b02c33da0a59f9b24d1bba7aa6d#diff-3026b0038b5ca9e43db9de4a59ca0ad37f1d57722c133f525bb31e2092a1561f) fails unless the type is changed from `boolean` to `string`

### DIFF
--- a/src/services/Config.php
+++ b/src/services/Config.php
@@ -72,9 +72,9 @@ class Config extends Component
     private array $_configSettings = [];
 
     /**
-     * @var bool|null
+     * @var string|null
      */
-    private ?bool $_dotEnvPath = null;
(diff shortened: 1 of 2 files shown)


brandonkelly:
Craft 4.0.0-beta.2 is out now with this. Thanks again!

PR #10768, opened 2022-03-17
phpstan level 2

ostark:
@brandonkelly 
Please let me know if I should waste more time of my life with this :-)


brandonkelly:
Wow, thanks Oliver!

I’ve been meaning to get back on this, but there seem to be some PHPStan bugs that need to be addressed, and haven’t found the time to report them yet. Very much appreciate you making this effort though, and will happily merge these changes in even if we can’t fully get to level 2 for the time being.

ostark:
```/vendor/bin/phpstan analyze src --level=2 --memory-limit=1G```
Current status: 297 errors

(Obviously, the  ~50% achievement was wrong)

ostark:
By returning `static` instead of `self` a lot of issues with ElementQueries would disappear.
I'm afraid to apply it everywhere, but in theory, it should be possible: https://wiki.php.net/rfc/static_return_type#variance_and_subtyping

ostark:
**Currently: 90 errors (from 646)** 🥳 (in src)
I think I should stop here. The remaining issues are easier for you @brandonkelly

brandonkelly:
Thanks @ostark! Mostly AFK this week but I will review next week.

ostark:
< 50 🥳

brandonkelly:
That’s a wrap!! Thanks so much @ostark!

PR #10795, opened 2022-03-21
Fix widget method comments

ccchapman:
### Description

Widgets are hard deleted, not soft deleted.

Also, the `enabled` column on the `widgets` table does not seem to do anything.
Not sure if anything needs to be fixed there.


brandonkelly:
Thanks!

Issue #6172, opened 2020-06-02
[FR] Option to automatically trigger migrations from a module


BenParizek:
I have a group of plugins that use shared modules and those modules define schema in one or more of the plugins where they are used. This presents a problem in that I do not know which plugin may be updated first and need to trigger the migrations in all of the plugins. This is necessary right now because migration folders are registered automatically by Craft and expect migration folders to have a one-to-one relationship with plugins.

Currently, for modules that manage shared schema, this requires at least 3 migrations, where the migrations in any plugins can just import and call `safeUp` on the shared migration in the module but leads to several extra steps and room for errors when managing releases.

```
migrations/m000000_000000_base_module_migration.php
migrations/m000000_000000_base_module_migration_plugin1.php
migrations/m000000_000000_base_module_migration_plugin2.php
```

The acute problem in my situation is that modules that define shared schema for us in multiple plugins have no way of triggering migrations without multiple steps or duplicate code, and more generally, modules have no way of automatically communicating to Craft to trigger schema updates like plugins do using the `schemaVersion` attribute.

## Register Migration Roots Event

One potential approach to solving this could be something similar to `EVENT_REGISTER_CP_TEMPLATE_ROOTS` Event, where Craft provides an API where modules could register new migration folders to be checked when it is resolving migrations:

```
Event::on(View::class, View::EVENT_REGISTER_MIGRATION_ROOTS, function(RegisterMigrationRootsEvent $e) {
    $e->roots['sprout-base'] = $this->getBasePath().DIRECTORY_SEPARATOR.'migrations';
});
```

brandonkelly:
This has (sortof) been added for the next 3.5 beta release.

Instead of doing something module-specific, I just added support for arbitrary migration tracks, via a new `--track` option on `migrate` commands.

You can add support for it from your Sprout plugins like so:

```php
use craft\console\controllers\MigrateController;
use craft\db\MigrationManager;
use craft\events\RegisterMigratorEvent;
use yii\base\Event;

Event::on(
    MigrateController::class,
    MigrateController::EVENT_REGISTER_MIGRATOR,
    function(RegisterMigratorEvent $event) {
        if ($event->track === 'sprout-base') {
            $event->migrator = Craft::createObject([
                'class' => MigrationManager::class,
                'track' => 'sprout-base',
                'migrationNamespace' => 'barrelstrength\sproutbase\migrations',
                'migrationPath' => '@vendor/barrelstrength/sprout-base/src/migrations,
            ]);
[...]

PR #10817, opened 2022-03-24
Allow modules to register custom migration tracks to run on `migrate/all`

michaelrog:
Currently, [custom migration tracks](https://github.com/craftcms/cms/issues/6172) exposed via `MigrateController::EVENT_REGISTER_MIGRATOR` must be explicitly run via a `migrate/up --track=my-custom-track` command.

This PR adds `EVENT_REGISTER_MIGRATION_TRACKS` event to `MigrateController`, with an accompanying `MigrationTracksEvent` class for registering a list of custom migration tracks.

Now, modules can add their custom migration tracks to be run automatically on `migrate/all` (similar to how Craft automatically runs plugin migrations), which lets me keep my deploy scripts nice and tidy:

```php
use craft\console\controllers\MigrateController;
use craft\events\MigrationTracksEvent;
use yii\base\Event;

Event::on(
    MigrateController::class,
    MigrateController::EVENT_REGISTER_MIGRATION_TRACKS,
    function(MigrationTracksEvent $event) {
        $event->tracks[] = 'my-custom-track';
    }
);
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Added
 - Added `craft\helpers\ArrayHelper::onlyContains()`.
 - Added `craft\test\Craft::assertNotPushedToQueue()`. ([#10510](https://github.com/craftcms/cms/pull/10510))
+- Added `craft\console\controllers\MigrateController::EVENT_REGISTER_MIGRATION_TRACKS`.
 
 ### Changed
 - Duplicated elements no longer have “copy” appended to the end of their titles. ([#10707](https://github.com/craftcms/cms/pull/10707))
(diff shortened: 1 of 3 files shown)


brandonkelly:
I think I would prefer to change the way custom migration tracks are registered, so they can work for both of these cases with a single event.

michaelrog:
> I think I would prefer to change the way custom migration tracks are registered, so they can work for both of these cases with a single event.

100% agree. This is tricky for Craft 3, since `RegisterMigratorEvent` is already out in the wild.

What if we made `RegisterMigratorEvent` extend `DefineComponentsEvent`, and have modules tack onto the `$components` array in the `migrate/all` scenario, i.e. when `RegisterMigratorEvent::$track` is `'all'`... (`MigrationManager` is a `Component`, so it fits.)

Then in Craft 4, we'd remove `RegisterMigratorEvent::$migrator`, standardize on the array setup, and maybe rename the class to `RegisterMigrationsEvent`, to be consistent.

Thoughts?

PR #10824, opened 2022-03-25
Add read/write methods to FS

timkelty:
Relies on https://github.com/craftcms/craft-flysystem/pull/1


andris-sevcenko (changes requested):
All good, just missing the changelog entries.

andris-sevcenko (approved):

PR #10825, opened 2022-03-25
Allow debug module to use Craft FS storage

timkelty:
## Relies on:
- https://github.com/craftcms/craft-flysystem/pull/1
- https://github.com/craftcms/cms/pull/10824

### Description
This makes it possible for the debug module/toolbar to use storage from a Craft FS, allowing it to be centrally stored (eg S3), and therefore can work in load-balanced/distributed environments.

With no configuration, there should be not behavior change.
If configured via `config/debug.php`, you can configure it to use an fs:

This will use a `runtime` FS, and put it in the `debug` folder.
```php
 <?php

return [
    'fs' => Craft::$app->getFs()->getFilesystemByHandle('runtime'),
    'dataPath' => 'debug',
];
```

Note: Most of the code is just lifted directly from the parent methods, with the spots where file read/writes swapped out.


andris-sevcenko:
- https://github.com/craftcms/craft-flysystem/pull/1
- https://github.com/craftcms/cms/pull/10824

Both are approved and awaiting merge. The `craft-flysystem` will require the next beta (Craft 4.0.0-beta.4).

src/debug/controllers/UserController.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
+namespace craft\debug\controllers;
+
+class UserController extends \yii\debug\controllers\UserController
+{
+}
```
brandonkelly, line 7: @timkelty Is this necessary?

timkelty: I think so, unfortunately. Because we're overriding the `$controllerNamespace` so that we can tweak `DefaultController`, we have to also bring this along, otherwise the controller won't get recognized. Unless you know another way?

brandonkelly:

PR #10851, opened 2022-03-31
Bypass erroneous console warnings

timkelty:
### Description
With this PR, commands like `craft db/restore` will not longer throw a ``Craft can’t fetch the `info` table row.`` error to the console when it doesn't actually need an `info` table.

When Craft is not installed, any `craft help/*`, `craft db/*`, `craft install/*` will bypass the check for the `info` table. Note: `install/plugin` won't, it just runs `plugin/install`.


brandonkelly:

PR #10866, opened 2022-04-06
Fix a `$expectedType must not be accessed before initialisation` error with `MissingComponentTrait`

engram-design:
I'm getting the following issue with [Formie](https://github.com/verbb/formie/issues/820) which uses this trait. I'm not _quite_ sure if this shouldn't be an issue, but I've also found that anything strictly-typed should have a default value.

### DIFF
--- a/src/base/MissingComponentTrait.php
+++ b/src/base/MissingComponentTrait.php
@@ -24,7 +24,7 @@ trait MissingComponentTrait
      * @var string The expected component class name.
      * @phpstan-var class-string<ComponentInterface>
      */
-    public string $expectedType;
+    public string $expectedType = '';
 
     /**
      * @var string|null The exception message that explains why the component class was invalid


brandonkelly:
Missing components should always be configured with `expectedType` set on initialization. For example:

https://github.com/craftcms/cms/blob/38e06ebaa04a829ccc77d0a3db065dc4dc6e85dc/src/services/Fields.php#L590-L594

If you’re not setting that, then this is “breaking as expected”.

engram-design:
Strange, that's exactly what I'm doing https://github.com/verbb/formie/blob/87cf1282b555b5f0e7a87243a73aa3dc7420f1e4/src/services/Fields.php#L865-L869

But this error is cropping up when bringing up the GQLi explorer (without querying), so something else must be going on...

engram-design:
Yep, all good - something on my end.

PR #10875, opened 2022-04-07
Upgrade d3

timkelty:


brandonkelly:
The New Users widget doesn’t work on this branch.

Also, it’s too late in the game to be swapping D3 with a breaking update. So this will need to wait for Craft 5.

Issue #10888, opened 2022-04-08
[4.x]: Can't edit/open slide out for addresses anymore after update to beta.2


Nellyaa:
### Description

I updated to Commerce 4 beta.2 on Craft CMS 4 beta.4 and editing addresses in orders or carts results in a red error message 'a server error occurred'. The console logs a 500 Bad Request. 
This worked fine before the update on Commerce beta.1 (beta.3 on Craft CMS)

### Steps to reproduce

1. Open an order
2. Click on the address to open the slide out to edit it
3. Get empty slide out with red error message at the top

### Expected behavior

I can edit addresses and save them :)

### Actual behavior

![Unbenannt-3](https://user-images.githubusercontent.com/38379317/162436406-6595cd09-ed76-43bb-bd25-687dd4ff9f34.jpg)

The following error shows up after opening the error url:

`````
An Error occurred while handling another error:
TypeError: craft\services\Addresses::getAdministrativeAreaTypeLabel(): Argument craftcms/commerce#1 ($type) must be of type string, null given, called in /var/www/vendor/craftcms/cms/src/elements/Address.php on line 155 and defined in /var/www/vendor/craftcms/cms/src/services/Addresses.php:162
Stack trace:
#0 /var/www/vendor/craftcms/cms/src/elements/Address.php(155): craft\services\Addresses->getAdministrativeAreaTypeLabel(NULL)
craftcms/commerce#1 /var/www/vendor/craftcms/cms/src/elements/Address.php(279): craft\elements\Address::addressAttributeLabel('administrativeA...', 'DE')
craftcms/commerce#2 /var/www/vendor/craftcms/cms/src/helpers/Cp.php(1338): craft\elements\Address->getAttributeLabel('administrativeA...')
craftcms/commerce#3 /var/www/vendor/craftcms/cms/src/helpers/Cp.php(1232): craft\helpers\Cp::_subdivisionField(Object(craft\elements\Address), 'administrativeA...', false, false, Array, true)
craftcms/commerce#4 /var/www/vendor/craftcms/cms/src/fieldlayoutelements/addresses/AddressField.php(154): craft\helpers\Cp::addressFieldsHtml(Object(craft\elements\Address))
craftcms/commerce#5 /var/www/vendor/craftcms/cms/src/fieldlayoutelements/BaseField.php(215): craft\fieldlayoutelements\addresses\AddressField->inputHtml(Object(craft\elements\Address), false)
[...]

PR #10889, opened 2022-04-08
[4.x] Address slideout type error

nfourtythree:
### Description
Fix an error that could occur when opening an address slideout

### Related issues
#10888


brandonkelly:
Took a different approach and just made the Addresses service methods accept `null` (6b16a39342439ca851b27d2270415a1b76e7f47a).

PR #10941, opened 2022-04-18
Prettier linting

timkelty:
This add [Prettier](https://prettier.io/docs/en/options.html) linting to all file types Prettier knows about, with the default prettier config.

Prettier is intentionally opinionated and has few options, but the obvious ones, we can [configure](https://prettier.io/docs/en/options.html) (tab width, single/double quotes, etc).

We can also limit the scope to specific filetypes, but to start I did everything Prettier knows about, with the default configuration.

We would be able to, for instance, say that we wanted `.yaml,.yml` files to have 4 spaces, while everything else has 2.

Our pipeline would run this check like so:

```
npx prettier --check .
```

[A companion PR](https://github.com/craftcms/cms/pull/10940), shows how this can be run locally (along with ecs, phpstan) on staged files when committing.

Probably makes sense to start by excluding Markdown /cc @mattstein


brandonkelly:
Looks like `lib/` needs to be ignored, and let’s have it stop looking at `.md` files as well, unless we can tell it to stop forcing a blank line after headings.

timkelty:
> Looks like `lib/` needs to be ignored, and let’s have it stop looking at `.md` files as well, unless we can tell it to stop forcing a blank line after headings.

Done.

brandonkelly:
Possible to change the double quote enforcement to single quotes, for `.js` and `.ts` files?

timkelty:
Yep, one of the few settings that are configurable.

timkelty:
Now doing `"singleQuote": true` for everything. If we want to do it by type, we can, but starting here.

PR #10982, opened 2022-04-22
Use getGeneral() method

jamesmacwhite:
### Description

Updates usage of getConfig() to also use the getGeneral() method. Mostly for consistency/guidelines, shouldn't have any functional difference.


brandonkelly:
Thanks!

PR #11010, opened 2022-04-25
Default to `true` when creating new migrations

AugustMiller:
Another unsolicited PR from yours truly; this time to change the default behavior when creating a new migration via the CLI.

The existing `false` default doesn't make a lot of sense, insofar as I've already manually tapped out the command, a name, scope, etc. _and_ have a built-in opportunity via the confirmation itself to cancel the operation.


brandonkelly:
Agree! Current behavior was just based on Yii’s, but this is better.

PR #11035, opened 2022-04-28
Add `Element::getElementHtml()` to allow elements to define their own HTML for the element icon

engram-design:
For [Comments](https://github.com/verbb/comments) and [Navigation](https://github.com/verbb/navigation) plugins we want to show a little extra information against the title/status of an element, in their element index. Shown as below:

For Navigation, we actually just want to append some icons to the element's icon/pill (there must be a name for that?)
image

For Comments, we actually want to be a bit more hands-on, by adding a snippet of their comment.
image

Right now, the only way to achieve this is through the `cp.elements.element`, checking against the element type is the one we want, and modifying it. It's not amazing, but it works.

One thing that doesn't work is when modifying the element with the new element side-out pane. For instance, with Navigation, there's a new-window icon that shows up when you enable that for the element. This won't change when the slide-out is closed, because the Twig hook isn't run, instead `Cp::elementHtml()` is.

I've also opted to modify _just_ the `$innerHtml` of the element icon thing - because of the way `Craft.BaseElementIndex` works, it'll be looking for a `.element` `div`, to swap out with JS. If we append things outside of that (like I am currently doing with Navigation), then it gets messy. As such, anything modifiable, should really be contained within the `.element` `div`.

As such, I'm using a modify-style handling here, instead of straight replacing of what `Cp::elementHtml()` does. So I can still grab the stock HTML generated and add my own, or replace it entirely.

Keen to get your thoughts.


engram-design:
@brandonkelly After feedback, I've swapped this to a `RegisterCpElementHtmlEvent` event. I was going to name it `ModifyCpElementHtmlEvent` but that doesn't seem to follow your naming convention. It's also made on the `Cp` helper class, as opposed to the element. It makes it not as nice for developers of your own element class, but it's also no worse that the current hook-registration at the plugin/module level.

(If you registered this event in your element class, you'd get duplicate content being called every instance of the element).

engram-design:
Thanks @brandonkelly sorry for the extra work!

brandonkelly:
No worries! Appreciate the PR :)

brandonkelly:
4.0.0-RC2 is out now with this change :)

engram-design:
Thanks so much @brandonkelly 🎉

PR #11053, opened 2022-05-01
Added option to add the attribute 'hidden' to the select options

MarcusGaius:
### Description

Allows for the option to have attribute 'hidden' (for example, lead/description of select), and not appear in the list of options. Works the same way as the existing ability to have the attribute 'disabled'.

### DIFF
--- a/src/templates/_includes/forms/select.html
+++ b/src/templates/_includes/forms/select.html
@@ -49,6 +49,7 @@
                     value: optionValue ?? '',
                     selected: (optionValue~'') is same as (value~''),
                     disabled: option.disabled ?? false,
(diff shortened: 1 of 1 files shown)


brandonkelly:
`hidden` doesn’t appear to be an [accepted `<option>` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/option).

brandonkelly:
My mistake… didn’t realize that `hidden` is a [global attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden).

brandonkelly:
3.7.40 is out now with that change.

PR #11062, opened 2022-05-02
Add col.width to heading and html cells

leevigraham:


brandonkelly:
Looks like your changes are missing a couple `{% if %}` tags.

leevigraham:
Derp… 12am… PR's… fixed.

leevigraham:
@brandonkelly Would you be opposed to also adding info buttons to row cells?

image

PR #11065, opened 2022-05-02
Rely on GeneralConfig::$revAssetUrls to add versioning params

timkelty:
Adding params to URLs can break when someone is overriding the thumb/preview URL, or using their own transformer.

This removes adding any versioning query strings from the service level, since we already have `\craft\config\GeneralConfig::$revAssetUrls`.

This also deprecates `\craft\helpers\Assets::urlAppendix`, as it made assumptions about URLs not being query string based.


brandonkelly:
4.0.0-RC3 is out now with this change.

Issue #11063, opened 2022-05-02
[3.7.38][GraphQL] orderBy RAND() regression


adrienpicard:
### Description
I understand  cb59bde049e82c6d5a029f60dd4dc396ec2a9a3b fixes a vulnerability issue but it also means we can't orderBy RAND() anymore with GQL queries. 

Would be great to know if there's any work around.

### Steps to reproduce
Write a GQL query with orderBy: "RAND()"

### Expected behavior
It returns random results

### Actual behavior
It throws an `Illegal value for  orderBy argument` error

### Craft CMS version

3.7.38

### PHP version

7.4.3

### Installed plugins and versions

-

brandonkelly:
Sorry, we overlooked that, but it should be allowed.

Just fixed this for the next release.

To get the fix early, change your `craftcms/cms` requirement in composer.json to `"dev-release/3.7.40 as 3.7.39"` and run `composer update`.

brandonkelly:
3.7.40 is out now with that fix.

adrienpicard:
Amazing. Thanks for the quick fix @brandonkelly

davidwebca:
I just got reported a bug about that and found this issue. Glad I just need to update to have the fix already 😂  Thanks for the quick fix!

katiasmet:
It looks broken again in 3.7.44.

brandonkelly:
@katiasmet Just tested and it’s still working on my end. That’s a SQL error though – so `RAND()` is making it into the query but it’s being used incorrectly. What’s your full GraphQL query?

PR #11072, opened 2022-05-03
Allow `random()` in GraphQL for PostgreSQL #11063

kennethormandy:
### Description

To my knowledge, if you are using PostgreSQL, you need to use `random()` for `orderBy` [rather than `RAND()`](https://www.postgresql.org/message-id/Pine.LNX.4.21.0007182150230.1545-100000@localhost.localdomain).

This PR allows `random()` in GraphQL, in the same way #11063 was fixed.

### Related issues

#11063
https://github.com/craftcms/cms/commit/aca7413d3bca5aabc265f3aea981d778f5075299


brandonkelly:
The `rand()` fix was initially made on the `release/3.7.39` branch, so made a similar fix on that branch, and merged into `4.0`.

Issue #11085, opened 2022-05-04
[3.7.38]: OOM when resaving entries


BrandonJamesBorders:
### Description
Craft installation that has a large number of entries ~200,000. Nightly import jobs have updating the search index disabled and then after all imports are complete we run a reindex job.  

We've recently started getting OOM errors. Discovered that Yii's BatchQueryResult has this issue due to MySQL config. https://github.com/yiisoft/yii2/issues/9269

Examined how another plugin resolves this issue here: https://github.com/nystudio107/craft-imageoptimize/issues/243 and implemented it into the core.  

### Steps to reproduce

1. Load up a Craft install with 200,000+ entries. 
2. Trigger a resave/entries --update-search-index command
3. Observe excessively high memory usage due to Yii and MySQL limitatations

### Expected behavior
Expect that BatchQueryResult would be memory optimized. 

### Actual behavior
Discover it is not. 

### Craft CMS version

3.7.38

### PHP version

7.4.28

### Database type and version

MySQL 5.7.23

### Installed plugins and versions

-

brandonkelly:
Query batching comes with some [caveats](https://www.yiiframework.com/doc/guide/2.0/en/db-query-builder#batch-query-mysql) when using MySQL, so Craft doesn’t do it by default. However you can opt into it using the [useUnbufferedConnections](https://craftcms.com/docs/4.x/config/db-settings.html#unixsocket) database connection setting.

```php
// config/db.php
return [
    // ...

    'useUnbufferedConnections' => true,
];
```

PR #11086, opened 2022-05-04
Update resaveElements to use craft\db\Paginator

BrandonJamesBorders:
### Description
Update resaveElements to use craft\db\Paginator to resolve OOM 

### Related issues
https://github.com/craftcms/cms/issues/11085


brandonkelly:
Craft already supports query batching out of the box; you just need to enable the [ useUnbufferedConnections](https://craftcms.com/docs/4.x/config/db-settings.html#useunbufferedconnections) database connection setting. (see https://github.com/craftcms/cms/issues/11085#issuecomment-1118041629)

PR #11090, opened 2022-05-04
Upgrade d3

timkelty:
@brianjhanson Assigning this to you, just because you're closer to it at the moment, but happy to take it back, or go through it together.


steverowling:
Is there a chance that this will get accepted and a more recent version of d3 will be included with Craft? I have a plugin idea that requires d3 ^6.6.1, which conflicts with the current version that ships with Craft (4.13.0). To get it to work, I have to render on a page outside of the Craft CP, whereas I would love to be able to get it to work in a CP page. Alternatively, is there a way of forcing a newer version of d3 to load on a specific CP page?

brandonkelly:
@steverowling It would be a breaking change so will happen for Craft 5 instead.

brandonkelly:
@timkelty May be the fault of my [merge commit](https://github.com/craftcms/cms/pull/11090/commits/c7707fa6085c833762286e49134664534008c99c) but I’m getting `d3 is not defined` JS errors on this branch.

timkelty:
@brandonkelly not seeing any errors in my console 🤔 

FWIW - `d3` _shouldn't_ be defined (globally at least) since we're just including as a proper module in this branch.

timkelty:
Closed in favor of #14124

PR #11124, opened 2022-05-06
Change `TypeConditionRule` to allow multiple selections of entry types

nfourtythree:
### Description
Currently, the entry type condition will only allow the ability to define one entry type for its value.

This PR adjusts the functionality so that the `TypeConditionRule` becomes multi selectable allow the ability to specify `in`/`not in` a list of entry types. This gives a little more power to the rule.


brandonkelly:
There’s no way we can know all of the places an entry condition may be used, so rather than trying to update existing configs with a migration, the condition rule should just be able to handle an outdated configuration, e.g. by normalizing the config array from the constructor.

nfourtythree:
@brandonkelly have done the following:

- Removed the migration
- Handle the old settings in the init
- Which required an update to `Conditions::createConditionRule()` to pass the attributes during the creation of the object.

I also merged `develop` into this branch to make sure it was up to date, but looks like `develop` needs merging into `4.1`.

brandonkelly:
If you move the config normalization to the constructor, you should be able to remove the deprecated properties entirely, right?

nfourtythree:
> If you move the config normalization to the constructor, you should be able to remove the deprecated properties entirely, right?

If we remove those and someone was accessing those directly you would get an error? I was trying to be overly cautious and make sure it wasn't a breaking change.

Granted I probably still could have done it in the constructor which is where I started with it until I realised we need [this fix](https://github.com/craftcms/cms/pull/11124/files#diff-28fc6dcf9d8ce08817d65bf403364715e0734979a4232104d29222d12111fa74).

PR #11130, opened 2022-05-06
Time travel

markhuot:
Adds a "timetravel" feature to the `DateTimeHelper`. I'm using something like this in some unit tests where I set a post date to `now` and then run tests at points in time like `-1 month` and `+1 month` to make sure things appear/disappear as they should.

Notably, this affects and `$entry->getStatus()` return and flips around the `PENDING` or `EXPIRED` status of that return.

Not sure if this is something you're looking to implement. It's ripe for abuse from developers and I'd be cautious to use it anywhere other than a unit test. But for unit testing it's _really_ helpful.

### DIFF
--- a/src/base/ApplicationTrait.php
+++ b/src/base/ApplicationTrait.php
@@ -227,6 +227,12 @@ trait ApplicationTrait
      * @see saveInfoAfterRequest()
      */
     private $_waitingToSaveInfo = false;
+    
+    /**
+     * @var string The current time that Craft's time helpers will respond with. Usefull for time traveling
+     *             during unit tests
+     */
+    private $_now = 'now';
 
     /**
      * Sets the target application language.
@@ -841,6 +847,20 @@ trait ApplicationTrait
         }
 
         return true;
+    }
+    
+    public function timetravel($now = 'now'): self
+    {
+        $this->_now = $now;
+        
+        return $this;
+    }
+    
+    public function timetravelBack(): self
+    {
+        $this->_now = 'now';
+        
+        return $this;
     }
 
     // Service Getters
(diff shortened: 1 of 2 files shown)


markhuot:
FYI, this mimics `Carbon`'s `setTestNow` so it may make sense to rename this to the same method name if you like the idea, https://carbon.nesbot.com/docs/#api-testing

markhuot:
FYI, this mimics `Carbon`'s `setTestNow` so it may make sense to rename this to the same method name if you like the idea, https://carbon.nesbot.com/docs/#api-testing

brandonkelly:
Just had a need for this (#11272) so added something similar. Went with adding `pause()`/`resume()`/`now()` methods to `DateTimeHelper` instead of methods on `ApplicationTrait`, and all other `DateTimeHelper` methods respect paused time as well.

markhuot:
`pause()` is nice! Sometimes I want to jump forward/backwards though and your change doesn’t account for that. If I could somehow set `DateTimeHelper::$_now` then I could,

1. Create a new entry with a post date +1 week
2. Confirm the default logic (future entry) works as is
3. Set `_now` +2 weeks, confirm the entry now shows where past entries should

I think the underpinnings are in place for this, I just don’t see a way to set `_now` with an explicit `DateTime`. I’d be happy to PR in a `DateTimeHelper::setNow(new DateTime(‘+2 weeks’))` if you’d like.

brandonkelly:
Sorry, overlooked that aspect. Just added a `$now` arg to `DateTimeHelper::pause()`, so you can override which time should be considered “now” while time is paused (16d58080da272dda79aa49dfdf53dde56a714384).

brandonkelly:
Craft 4.1 is out with those changes!

Issue #11128, opened 2022-05-06
Garnish menu btn still active even with `disabled` attribute in admin table


BenParizek:
### What happened?

Just a minor bug to note in the latest Commerce 4.0.0:

## /admin/commerce/promotions/sales

![CleanShot 2022-05-06 at 11 33 13@2x](https://user-images.githubusercontent.com/309735/167165196-eb94d127-2bc8-4a2f-9add-af54f47a58e3.png)

### Craft CMS version

4

### Craft Commerce version

4

### PHP version

8

### Installed plugins and versions

-

brandonkelly:
@nfourtythree There aren’t any references to `disabled` in `Garnish.CustomSelect` or `Garnish.DisclosureMenu`, so this is probably a case of Commerce setting the `disabled` class on the button, which will give it the reduced-opacity styling we apply across the board via `.disabled`, though the component itself has no support for it. So probably more of a Commerce bug, or you can PR support for `disabled` to the component in use.

PR #11140, opened 2022-05-07
Garnish Menu Buttons now watch for the presence and status of the `disabled` attribute

nfourtythree:
### Description
When setting the `disabled` attribute on a garnish menu button the button is still able to be interacted with.

This PR makes the menu buttons aware of the `disabled` attribute and updates the button's interactivity as required.

### Related issues
#11128


nfourtythree:
@brandonkelly the only thing I thought but hadn't done would be to change the `enable()` and `disable()` methods to just simply add and remove the `disabled` attribute (and the CSS class) and then let the `MutationObserver` handle the rest. I think that would tidy things up and then wouldn't require any other outside coding if someone is using the component.

Thoughts? Let me know if you would like me to update the PR with that.

nfourtythree:
Have updated the PR based on my last comment

src/web/assets/garnish/src/MenuBtn.js
```diff
@@ -63,7 +65,21 @@ export default Base.extend(
...
+
+      this.observer = new MutationObserver((mutations) => {
+        mutations.forEach((mutation) => {
+          if (
+            mutation.type === 'attributes' &&
+            mutation.attributeName === 'disabled'
+          ) {
+            this.handleStatusChange();
+          }
+        });
```
brandonkelly, line 77: Could this be replaced with a `for...of` loop so you can `break` out when you find the attribute you’re looking for?

src/web/assets/garnish/src/MenuBtn.js
```diff
@@ -255,11 +271,33 @@ export default Base.extend(
...
+
+      this.$btn.removeAttr('disabled');
     },
 
     disable: function () {
-      this.disabled = true;
+      if (!this.$btn) {
+        return;
+      }
+      console.log('disable this');
```
brandonkelly, line 285: Should be removed

src/web/assets/garnish/src/MenuBtn.js
```diff
@@ -255,11 +271,33 @@ export default Base.extend(
...
+      this.$btn.attr('disabled', 'disabled');
+    },
+
+    handleStatusChange: function () {
+      if (!this.$btn) {
+        return;
+      }
+
+      if (this.$btn.attr('disabled') === 'disabled') {
```
brandonkelly, line 294: Should probably go with `Garnish.hasAttr()` here rather than worrying about the specific value.

src/web/assets/garnish/src/MenuBtn.js
```diff
@@ -268,6 +306,7 @@ export default Base.extend(
     destroy: function () {
       this.menu.destroy();
       this.$btn.removeData('menubtn');
+      this.observer = null;
```
brandonkelly, line 309: Is this enough to actually kill the observer? Guessing you’ll need to call `this.observer.disconnect()` for it to actually stop observing changes.

brandonkelly (changes requested):

src/web/assets/garnish/src/MenuBtn.js
nfourtythree: 🤦

src/web/assets/garnish/src/MenuBtn.js
nfourtythree: Updated.

src/web/assets/garnish/src/MenuBtn.js
nfourtythree: I didn't know that method existed in Garnish. Have updated.

src/web/assets/garnish/src/MenuBtn.js
nfourtythree: Have added the disconnect before setting it to `null`.

PR #11169, opened 2022-05-09
Improve accessibility of sidebar toggle on mobile element index pages, darken sidebar headings

gcamacho079:
### Description
Darkens sidebar headings to meet minimum contrast guidelines. Updates toggle element to use disclosure attributes and convey purpose more accurately.

Adds a second-level heading about the element view to aid in screen reader usability.

### Related issues
Resolves DEV-417
Resolves DEV-368


src/web/assets/cp/src/js/BaseElementIndex.js
```diff
@@ -1138,7 +1138,7 @@ Craft.BaseElementIndex = Garnish.Base.extend(
       this.setInstanceState('selectedSource', this.sourceKey);
       this.sourceSelect.selectItem($source);
 
-      Craft.cp.updateSidebarMenuLabel();
```
brandonkelly, line 1141: Probably worth leaving this around as an alias for `updateContentHeading()`, just in case a plugin is calling it.

src/web/assets/cp/src/js/CP.js
```diff
@@ -524,7 +522,10 @@ Craft.CP = Garnish.Base.extend(
       );
     },
 
-    toggleSidebar: function () {
+    toggleSidebar: function ({target}) {
```
brandonkelly, line 525: If we need to accept a target here, it should just be a regular `target` argument, rather than an object. Then update the event handler to extract the target from the event data.

```js
this.addListener($('#sidebar-toggle'), 'click', ({target}) => {
  this.toggleSidebar(target);
});
```

Also, the function should continue to work if `target` isn’t passed in, for backwards compatibility.

brandonkelly (changes requested):

src/web/assets/cp/src/js/CP.js
gcamacho079: For the `target` variable, I thought I'd use a destructuring assignment to pull it out of the event object, to cut down on code inside the `toggleSidebar` function. I can re-work that though 👍🏼 

Out of curiosity, what browsers/versions don't support `target` in this context? I'm not really versed on issues that would arise there. Appreciate the feedback!

src/web/assets/cp/src/js/BaseElementIndex.js
gcamacho079: I've re-added the `updateSidebarMenuLabel` function inside that object and called the new function inside of it.

gcamacho079:
@brandonkelly I've made the requested updates to this PR

src/web/assets/cp/src/js/CP.js
brandonkelly: Not a browser support issue; just an API clarity issue. `toggleSidebar()` should be something that can be called directly, without needing to pass an event object into.

gcamacho079: Makes sense! I'll keep that in mind going forward

PR #11172, opened 2022-05-09
Center customer support logos, update Partners alt text, and add disabled state to submit button

gcamacho079:
### Description
Fixes bugs where support logos were not centered, and alt text referred to Craft Support instead of Craft Partners

Screenshot showing off-center logos

Adds `aria-disabled` attribute to disabled buttons

### Related issues
Resolves DEV-286


brandonkelly:
Thanks!

PR #11181, opened 2022-05-10
Vertically center condition rule lightswitch

mattstein:
### Description

Lightswitch inputs in condition rules currently align with the top of the rule panel, which looks like a mistake given that every other bit of UI is vertically center-aligned—like the _Status_ rule and remove _x_’s here:

Screen Shot 2022-05-10 at 07 24 36 AM@2x

This adds an `.items-center` class to the base condition rule’s `getHtml()`, because it fixes that case and it _seems_ like you’d always want these elements vertically centered so they don’t have to resort to margin or padding for optical alignment:

Screen Shot 2022-05-10 at 11 39 56 AM@2x


brandonkelly:
Lightswitches are already vertically centered via dadaecb9b92305dab259baebafeb0790e1ff3c24. And if I remove that, this change doesn’t seem to have any effect. (Not seeing any `.items-center` class defined.)

mattstein:
I’m adding it to the list [here](https://github.com/craftcms/cms/pull/11181/files#diff-afd4039d878b5496a6646f314931ad1bb1aeee0f3eff4bf319e6b706e5309675R179)—wasn’t sure if I should build assets or if that’d make it harder to see what I did.

mattstein:
Welp, deleted my `vendor/` folder and that condition rule lightswitch is still center-aligned without my change. If you can find it in your heart to pretend this PR never happened, I would appreciate it.

Issue #11209, opened 2022-05-11
[3.x]: Unable to eager load multiple transforms with srcsets


Wiejeben:
### Description
Update: I made a pull request https://github.com/craftcms/cms/pull/11210.

Essentially what I'm trying to do is to eager load 3 different transforms with a `2x` srcset:
* `cardIndexLarge` with `2x`
* `cardIndexMedium` with `2x`
* `cardIndexSmall` with `2x`

In order to eager load the base transforms I wrote the following:
```twig
{% set query = craft.entries.section('blogArticles').orderBy('postDate').limit(8).with([
    ['coverImage', {withTransforms: ['cardIndexLarge', 'cardIndexMedium', 'cardIndexSmall']}],
]) %}
```

As far as I'm aware, the current transform eager loading implementation only supports a single transform when combined with eager loading srcsets. This is to us the last step to achieving full eager loading!

When I add `2x` as the last item in the transforms array it will only eager load `cardIndexSmall` with 2x. However if I add the srcset after `cardIndexLarge` it will actually overwrite `cardIndexMedium` and `cardIndexSmall` to the `cardIndexLarge 2x` variant.

The overwrites happen because the `$sizeValue, $sizeUnit` at 461 are being remembered from the previous loop:
https://github.com/craftcms/cms/blob/a7b735321468a6cf7d73a0eb582aa610eb5fc390/src/services/AssetTransforms.php#L

In order to fix this I added `unset($sizeValue, $sizeUnit);` to the end of the foreach loop, this would allow me to apply srcset eager loading to multiple transforms like this:
```twig
{% set query = craft.entries.section('blogArticles').orderBy('postDate').limit(8).with([
    ['primaryImage', {withTransforms: ['cardIndexLarge', '2x', 'cardIndexMedium', '2x', 'cardIndexSmall', '2x']}],
]) %}
```

Apart from the `format` it would eager load all srcsets correctly (that might be a different issue?):
<img width="418" alt="Screenshot 2022-05-11 at 20 58 17" src="https://user-images.githubusercontent.com/2922378/167925706-9a5d6f19-95ae-447c-a7e2-1ad38bccbbf4.png">

Unless the format is applied from the base transform during eager loading the srcset transforms will still be requested in every loop:
<img width="401" alt="Screenshot 2022-05-11 at 20 58 55" src="https://user-images.githubusercontent.com/2922378/167925817-7dc28558-016e-475b-a1f8-4778db3b51dd.png">

[...]

PR #11210, opened 2022-05-11
[3.x] Fix srcset eager loading when specific format is defined and allow srcset eager loading on multiple transforms

Wiejeben:
This pull request would fix srcset eager loading by maintaining the `format` provided by the the base transform. 

Also this pull request would add support for allowing srcset transforms on multiple transforms like this:
```twig
{% set query = craft.entries.section('blogArticles').orderBy('postDate').limit(8).with([
    ['primaryImage', {withTransforms: ['cardIndexLarge', '2x', 'cardIndexMedium', '2x', 'cardIndexSmall', '2x']}],
]) %}
```

### Related issues
https://github.com/craftcms/cms/issues/11209

### DIFF
--- a/src/services/AssetTransforms.php
+++ b/src/services/AssetTransforms.php
@@ -483,6 +483,11 @@ class AssetTransforms extends Component
                         $transform['height'] = (int)ceil($refTransform->height * $sizeValue);
                     }
                 }
+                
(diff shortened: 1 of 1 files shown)


brandonkelly:
Guessing this isn’t necessary since #11209 is already possible. Let me know if I missed something, though.

PR #11235, opened 2022-05-16
Add description property to Section and EntryTypes

leevigraham:
### Description

At [Newism](https://github.com/newism/) we've been discussing ways to better self document the CraftCMS control panel.

I noticed that `UserGroups` have descriptions but `Sections` and `EntryTypes` do not so I added them.

There's no front end display of these values yet – open to discussion. The values could be used in custom UI elements at the discretion of the developer.

One thing that could throw a spanner in the works… what happens when a new property is added to a model and there's an existing custom field with the same handle?

- Added `craft\models\Section::description`
- Added `craft\models\EntryTypes::description`

Screen Shot 2022-05-16 at 11 58 34 pm

image


brandonkelly:
There would need to be an obvious place to show the description in the core UI for this to be seriously considered.

PR #11251, opened 2022-05-17
Add accessible text to disclosure menu on Categories/Entries screens

gcamacho079:
### Description
- Adds accessible text to disclosure menus that specify new entry/category type (Resolves DEV-443)
- Adds `ariaLabel` to `Craft.ui.createButton` config


brandonkelly:
“Specify new entry type” could be confusing because Craft already has a concept called “entry types”.

Can we change the wording to “Choose a category group” and “Choose a section”?

gcamacho079:
@brandonkelly I've updated the translations to match your suggestions! I did realize while making that update that having different button text for the disclosure button depending on whether the user is in a group/section or on the index page is a failure under [3.2.4 Consistent Identification](https://www.w3.org/WAI/WCAG21/Understanding/consistent-identification.html).

So I streamlined the button text on the index page so that it's consistent with the split button's accessible text:
Screenshot showing a disclosure button with the text 'Choose a section' on the All entries page

gcamacho079:
@brandonkelly that makes complete sense. Let me think through this a little more and I'll see what I can come up with. 🤔

brandonkelly (approved):

PR #11256, opened 2022-05-18
Language mistake corrected

honsa:
### Description
Wrong kasus in test email translation.

Just a language kasus error in the test email translation. 

"Vom Person" is wrong in German instead you use "von Person".


brandonkelly:
Thanks for the PR! I’ve made the same changes at the source in [Crowdin](https://crowdin.com/project/craft-cms).

(In the future if you want to make any translation adjustments, that’s the best place to do it.)

PR #11257, opened 2022-05-18
Bug fix create new entry: "Entry is missing its section ID"

ardin0546:
I created a section (posts) which has only a basic field `title`. The type of the section was channel. When I wanted to create a new entry through the GraphQL api I got the exception `Entry is missing its section ID`. 

I found out that a new entry is created without a `sectionId`. So before the `sectionId` is set the method `entry->getTypeId()` is called. Deeper in the in code it calls the function `getSection()` inside the namespace `craft\elements\Entry` which throws that error.

### DIFF
--- a/src/gql/resolvers/mutations/Entry.php
+++ b/src/gql/resolvers/mutations/Entry.php
@@ -224,12 +224,13 @@ class Entry extends ElementMutationResolver
             throw new Error('Impossible to change the section of an existing entry');
         }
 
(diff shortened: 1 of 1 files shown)


brandonkelly (approved):

brandonkelly:
Thanks for the PR!

brandonkelly:
Craft 4.0.3 is out now with that fix. Thanks again!

PR #11260, opened 2022-05-18
User Group condition rule `isSelectable()` implementation

### LINKED ISSUE #11252: [4.x]: User Group condition rule is available when there are no user groups
### Description

In the early days of a new Craft 4 project, I might add a Current User Condition to a field layout element with a _User Group_ rule—but if my site doesn’t have any user groups, it’s unclear what I’m supposed to do with the rule’s last field:

Screen Shot 2022-05-17 at 01 20 40 PM@2x

The text input makes it seem like maybe I can enter something, but without any suggestions or other hints it may not be clear there’s just nothing I _can_ enter.

To help reinforce the expected usage, the condition rule should **either**...

- a) include a note that no user groups are available to select (ideally with a link to create one)
- b) not appear in the first place if it’s not a relevant option

### Steps to reproduce

1. Start with a Craft site that does not have any user groups.
2. Create some fields and a field layout.
3. Pick a field and add a User Condition with a User Group condition rule.
4. Make sensible choices for each field in the User Group condition rule. (Hint: you can’t.)  

### Expected behavior

The condition rule would be unavailable if it doesn’t apply, or some UI hint would help me understand there’s nothing to select.

### Actual behavior

The rule is available even though I can’t provide any meaningful input.

### Craft CMS version

4.0.2

### PHP version

8.0.15

nfourtythree:
### Description
Fixes #11252


brandonkelly:
Thanks Nate!

PR #11286, opened 2022-05-20
Add data-name to the .matrixblock

davidhellmann:
### Description
We use CPSS very much to extend/style our Matrix Builders.
The data-attr would be very helpful to do things like: content: attr(data-name); for custom labels on the blocks. It's a bit sugar and has no side effects I think.

Here is an image what we do:
image
it's a bit easier to scan on huge content pages.

Created the same PR on the NEO Field.


davidhellmann:
Nice Brandon,
I saw it afterwards that I forgot something. Thanks for finishing!

PR #11288, opened 2022-05-20
Add db/clean command

timkelty:
### Description
Adds `craft db/clean` and a `--clean` flag to `craft db/restore`.

Includes a confirm dialog, and a big angry "this is destructive action" warning.


jannisborgers:
@brandonkelly 😅 That was it! I exported it again with `--add-drop-table` checked (using TablePlus) and `db/restore` did its thing. Thanks!

PR #11296, opened 2022-05-23
Adds a highly visible active tab indicator and removes tabpanel attributes if no tablist exists

gcamacho079:
### Description

- Adds more visible active tab indicator in the form of a top border (Resolves DEV-573)
- Removes tab attributes being set on the panel element when there is only a single tab (Resolves DEV-391)
- Removes tabindex from being set on tabpanels, since there will be focusable elements inside in the majority of cases

A 4px tall blue top border on the active tab button


mattstein:
I think there are two challenges with the active tab indicator as it is here:

1. The tab’s color strip falsely implies a relationship to the field change indicator.
2. Tabs throughout the control panel lack the ability to convey more subtle information, like whether there are changes to contained field content or [whether they’re conditional](https://github.com/craftcms/cms/discussions/10953). Ideally this highly-visible indicator can carefully support those things later without changing visual treatment the content author has to re-orient to.

(1) field change indicator, (2) tab with changes that _aren’t_ indicated:


Tabs _do_ have a way of indicating that they contain validation errors:


If we knew changed fields or conditionality were indicated with a little bubble or icon like validation errors, for example, the active tab indicator could skip problem 1 with a simple color change:


(Though I don’t like how the strong horizontal tab highlight and vertical field edit stripe clash there.)

But if we use that top tab highlight to convey other information (changed fields, conditionality, etc.), the visual active indicator has to use some other treatment for emphasis.

brandonkelly:
I don’t really want to do anything that is going to change the width of a tab (or nav item label) when selected, so that rules out making the text bold.

Maybe we ditch the skeuomorphic tab style altogether and add a bottom border to the selected tab, instead of a top border? And make it dark gray instead of blue, to avoid possible confusion with the dirty field indicators.

> Tabs throughout the control panel lack the ability to convey more subtle information, like whether there are changes to contained field content or [whether they’re conditional](https://github.com/craftcms/cms/discussions/10953). Ideally this highly-visible indicator can carefully support those things later without changing visual treatment the content author has to re-orient to.

Maybe tabs that contain dirty fields can get a light blue bottom border when unselected, and a dark blue bottom border when selected (matching the dirty field indicator color) when they contain any dirty fields.

brandonkelly:
@gcamacho079 Does this satisfy the success criteria?

Two content tabs, showing a dark bottom border on the selected tab

gcamacho079:
> @gcamacho079 Does this satisfy the success criteria?

Yes, the dark bottom border does indeed satisfy the SC! However...

> Maybe tabs that contain dirty fields can get a light blue bottom border when unselected, and a dark blue bottom border when selected (matching the dirty field indicator color) when they contain any dirty fields.

This would not satisfy the success criterion for [Use of Color](https://www.w3.org/WAI/WCAG21/Understanding/use-of-color.html) because the color change of the bottom border would be the only indicator of a dirty field. 🤔 And because the bottom borders for all tabs w/ dirty fields would all have to meet 3:1 contrast ratio against the background, it may be hard to tell which tab is selected.

@mattstein's suggestion of adding an icon or bubble to the tab would do the trick. Another possibility is using indicator lines with different patterns (i.e. dotted, dashed, double lines, etc.)

brandonkelly:
Ah right. Well then let’s just say that’s out of scope for this PR and play around with ideas for that separately.

PR #11297, opened 2022-05-23
Quick a11y fixes

gcamacho079:
### Description
- Adds visually-hidden "Notification" text to badges in navigation for added context (Resolves DEV-287)
- Add `aria-label` to primary navigation


src/templates/_elements/sources.twig
```diff
@@ -38,7 +38,13 @@
...
     <span class="label">{{ source.label|trim is not same as('') ? source.label : '(blank)'|t('app') }}</span>
     {% if source.badgeCount is defined %}
-        <span class="badge">{{ source.badgeCount|number(decimals=0) }}</span>
+        <span class="badge">
+            {{ source.badgeCount|number(decimals=0) }}
+            {{ tag('span', {
+                class: 'visually-hidden',
+                text: source.badgeCount == 1 ? 'Notification'|t('app') : 'Notifications'|t('app')
+            }) }}
+        </span>
```
brandonkelly, line 47: Ideally this should be done in a way where the visually-hidden text completely overrides the visible text, so that the number and “notification(s)” can be combined into a single translated string, using the following syntax:

```twig
{{ tag('span', {
    class: 'visually-hidden',
    text: '{num, number} {num, plural, =1{notification} other{notifications}}'|t('app', {
        num: source.badgeCount,
    }),
}) }}
```

Maybe we can add a new `no-speak` class with `speak: none` (if there’s not something already like that – I forget), and then wrap the initial `{{ source.badgeCount|number(decimals=0) }}` tag in a `<span class="no-speak">`.

src/templates/_layouts/components/global-sidebar.twig
```diff
@@ -38,7 +39,13 @@
...
 
                         {%- if not item.sel and item.badgeCount -%}
-                            <span class="badge">{{ item.badgeCount|number(decimals=0) }}</span>
+                            <span class="badge">
+                                {{ item.badgeCount|number(decimals=0) }}
+                                {{ tag('span', {
+                                    class: 'visually-hidden',
+                                    text: item.badgeCount == 1 ? 'Notification'|t('app') : 'Notifications'|t('app')
+                                }) }}
+                            </span>
```
brandonkelly, line 48: (Same deal as the other badge.)

src/templates/_layouts/components/global-sidebar.twig
```diff
@@ -55,14 +62,21 @@
                                         (item.external ?? false) ? 'external',
                                     ],
                                     target: (item.external ?? false) ? '_blank',
+                                    'aria-current' : itemIsSelected ? 'page' : false,
```
brandonkelly, line 65: Extra space before `:` should be removed.

src/templates/_layouts/components/global-sidebar.twig
```diff
@@ -21,7 +21,8 @@
                 } %}
                 {% set linkAttributes = {
                     href: url(item.url),
-                    class: item.sel ? 'sel'
+                    class: item.sel ? 'sel',
+                    'aria-current' : (item.sel and not item.subnav) ? 'page' : false,
```
brandonkelly, line 25: This should be replaced with:
```suggestion
                    aria: {
                        current: item.sel and not item.subnav ? 'page' : false,
                    },
```

src/templates/_layouts/components/global-sidebar.twig
brandonkelly, line 65: This should be replaced with:
```suggestion
                                    aria: {
                                        itemIsSelected ? 'page' : false,
                                    }
```

brandonkelly (changes requested):
See comments

brandonkelly:
`aria-hidden="true"` 🤦🏻  I knew we were already doing this but couldn’t remember how :)

brandonkelly (approved):

PR #11306, opened 2022-05-24
Matrix field a11y improvements

gcamacho079:
### Description
- Darken UI control color for improved contrast (Resolves DEV-590)
- Add ordered list markup to matrix blocks (Resolves DEV-539)
- Focus on first focusable element when adding a block, not just text-type fields (Resolves DEV-541)
- Adds additional context to Matrix block “add” buttons so users will know what field they are associated with (Resolves DEV-526)


src/templates/_components/fieldtypes/Matrix/input.twig
```diff
@@ -2,7 +2,7 @@
 
 <div class="matrix matrix-field" id="{{ id }}">
     <span aria-live="assertive" class="visually-hidden" data-status-message></span>
-    <div class="blocks">
+    <ol class="blocks">
```
brandonkelly, line 5: Potentially-breaking DOM changes like this should be postponed until the next major release (Craft 5).

brandonkelly (changes requested):

brandonkelly (approved):

PR #11338, opened 2022-05-27
Update incorrect translations

MoritzLost:
### Description

This PR updates two incorrect translations in German and Swiss German.

> Create a new filesystem

This is currently incorrectly translated as `Das Dateisystem konnte nicht erstellt werden`, which actually means `Couldn’t save filesystem.` Looks like the string was just misassigned. This one is already correct in Swiss German.

> Try

This translation is only used for the `Try` button in the plugin store for commercial plugins. This is currently translated as `Versuch`, which is the German **noun** meaning try. This needs to be translated as a verb, which would be `Testen`. Another option would be `Ausprobieren`. `Testen` is also the phrase used in the translation for `Try for free` – `Kostenlos testen`.


brandonkelly:
Thanks! I pulled the same changes into [Crowdin](https://crowdin.com/project/craft-cms).

PR #11341, opened 2022-05-28
Add PC manipulation commands

timkelty:


brandonkelly:
I think the `set` value syntax should be identical to the `--to` argument values supported by `resave/*` commands.

timkelty:
> I think the `set` value syntax should be identical to the `--to` argument values supported by `resave/*` commands.

As discussed, since this is yaml-in/yaml-out, it probably make sense to keep the value as yaml.

src/console/controllers/ProjectConfigController.php
```diff
@@ -64,11 +83,82 @@ public function options($actionID): array
                 break;
             case 'diff':
                 $options[] = 'invert';
+                // no break
```
brandonkelly, line 86: Why would `diff` need the `fromFile`, `message`, `updateTimestamp`, and `force` options?

src/console/controllers/ProjectConfigController.php
```diff
@@ -64,11 +83,82 @@ public function options($actionID): array
...
+            case 'get':
+                $options[] = 'fromFile';
+                // no break
```
brandonkelly, line 89: Why does `get` need `message`, `updateTimestamp`, and `force` options?

src/console/controllers/ProjectConfigController.php
```diff
@@ -35,6 +37,23 @@ class ProjectConfigController extends Controller
      */
     public bool $invert = false;
 
+    /**
+     * @var bool Whether to get values from the loaded config, or directly from file.
+     */
+    public bool $fromFile = false;
```
brandonkelly, line 43: This should probably be `external` to match wording from `craft\services\ProjectConfig::get()`’s `$getFromExternalConfig` argument.

brandonkelly: Also needs a `@since 4.1.0`

src/console/controllers/ProjectConfigController.php
```diff
@@ -35,6 +37,23 @@ class ProjectConfigController extends Controller
...
+
+    /**
+     * @var string|null A message describing the changes.
+     * @see \craft\services\ProjectConfig::set
+     */
+    public ?string $message = null;
```
brandonkelly, line 49: Missing `@since 4.1.0`

src/console/controllers/ProjectConfigController.php
```diff
@@ -35,6 +37,23 @@ class ProjectConfigController extends Controller
...
+
+    /**
+     * @var bool Whether the `dateModified` value should be updated
+     * @see \craft\services\ProjectConfig::set
+     */
+    public bool $updateTimestamp = false;
```
brandonkelly, line 55: Missing `@since 4.1.0`

src/console/controllers/ProjectConfigController.php
```diff
@@ -64,11 +83,82 @@ public function options($actionID): array
...
         }
 
         return $options;
     }
 
+    /**
+     * @param string|null $path The config item path
+     * @return int
+     */
+    public function actionGet(?string $path = null): int
```
brandonkelly, line 103: Missing `@since 4.1.0`

src/console/controllers/ProjectConfigController.php
```diff
@@ -64,11 +83,82 @@ public function options($actionID): array
...
+        $this->stdout(PHP_EOL);
+        return ExitCode::OK;
+    }
+
+    /**
+     * @param string $path The config item path
+     * @param string $value The config item value as a valid YAML string
+     * @return int
+     */
+    public function actionSet(string $path, string $value): int
```
brandonkelly, line 117: Missing `@since 4.1.0`

src/console/controllers/ProjectConfigController.php
```diff
@@ -64,11 +83,82 @@ public function options($actionID): array
...
+        $this->stdout(($multiline ? '' : '.') . PHP_EOL);
+
+        return ExitCode::OK;
+    }
+
+    /**
+     * @param string $path The config item path
+     * @return int
+     */
+    public function actionRemove(string $path): int
```
brandonkelly, line 157: Missing `@since 4.1.0`

brandonkelly (changes requested):

src/console/controllers/ProjectConfigController.php
timkelty: I just mistakenly left out the break, but ecs added the `no break` comment, making it seem intentional 🤦

src/console/controllers/ProjectConfigController.php
timkelty: Same as above

src/console/controllers/ProjectConfigController.php
timkelty: Yeah, that was my initial though too…I went with `fromFile` because I didn't think `external` would really make sense to anyone not intimately familiar with the PC service…but alas, changed it.

timkelty:
@brandonkelly changes made and review re-requested.

PR #11371, opened 2022-06-01
Fix bug blocking keyboard users from accessing video controls

gcamacho079:
### Description
Adds bumper buttons to preview modals to address focus trap bug caused by video controls being in the Shadow DOM (Resolves DEV-413)

In addition:
- Adds a line of code to prevent the addition of a focus trap if there are no focusable elements found in the container; this prevents blocking focus traps caused when elements like `iframe` and `video` are the only ones present in a modal container
- Adjusts top position of skip link to be `0` instead of `auto`
- Modifies `setFocusWithin` function to set focus on modal container if no focusable elements are inside


src/web/assets/cp/src/js/PreviewFileModal.js
```diff
@@ -60,8 +64,21 @@ Craft.PreviewFileModal = Garnish.Modal.extend(
 
         this.$shade.velocity('stop');
         this.$shade.show().css('opacity', 1);
+
+        Garnish.setFocusWithin(this.$container);
       }
 
+      // Add bumper elements to maintain focus trap
+      this.$bumperButtonStart = Craft.ui.createButton({
+        html: Craft.t('app', 'Close preview'),
```
brandonkelly, line 73: We have an existing `Close Preview` translation (capital `P`), and `Close`, but no `Close preview` yet. Let’s go with one of the existing ones here.

src/web/assets/cp/src/js/PreviewFileModal.js
```diff
@@ -73,13 +90,21 @@ Craft.PreviewFileModal = Garnish.Modal.extend(
       Craft.PreviewFileModal.openInstance = null;
       if (this.elementSelect) {
         this.elementSelect.focusItem(this.elementSelect.$focusedItem);
+      } else if (this.$triggerElement) {
```
brandonkelly, line 93: This could be set to an empty jQuery object, which would still evaluate to `true`, so should probably be:

```suggestion
      } else if (this.$triggerElement && this.$triggerElement.length) {
```

brandonkelly (changes requested):

brandonkelly (approved):

PR #11384, opened 2022-06-04
Prevent hidden folders in volumes from being indexed

### LINKED ISSUE #11362: [4.x]: Asset Update Indexes broken - indexing hidden transform subfolders
### Description

Asset indexing seems to be trying to index all images and transforms

### Steps to reproduce

1. Utilities > Asset Indexes > Update asset indexes

### Expected behavior

Index all images, not hidden transforms directories

### Actual behavior

Index count is in thousands, clearly trying to index everything in the asset volumes, including transform sub-directories

### Craft CMS version

4.0.3

### PHP version

8.0.8

OSX 12.4

### Database type and version

MySQL 5.7.34

### Image driver and version

	Imagick @PACKAGE_VERSION@ (ImageMagick 6.9.6-2)

- CKEditor 2.0.0
- Environment Label 4.0.0
- Field Manager 3.0.0-beta.1
- Freeform Pro 4.0.0-beta.9
- ImageOptimize 4.0.0-beta.3
- Mailgun 3.0.0
- Navigation 2.0.1
- Retour 4.0.1
- SEOmatic 4.0.1
- Sherlock 4.0.1
- Sprig 2.0.1
Screen Shot 2022-06-03 at 3 48 43 pm

mmikkel:
### Description

I don't know if this is the way you want to go with it, but if merged this PR should resolve #11362.

This PR changes the way the Asset Indexes utility compiles the list of files to index, by making the `AssetIndexer::startIndexingSession()` method pull the indexing file list via the `AssetIndexer::getIndexListOnVolume()` method, instead of getting it directly from the volume's filesystem instance. 

This makes sure that any files in folders beginning with an underscore are completely ignored during the indexing process, which is consistent with [how asset indexing currently works in 4.x via CLI](https://github.com/craftcms/cms/blob/develop/src/console/controllers/IndexAssetsController.php#L151), as well as how indexing via the Asset Indexes utility worked in 3.x.

This PR also fixes a wrongly typed exception handler in the AssetIndexer service.

### Related issues

#11362

### DIFF
--- a/src/services/AssetIndexer.php
+++ b/src/services/AssetIndexer.php
@@ -50,19 +50,19 @@ use yii\base\InvalidConfigException;
  */
 class AssetIndexer extends Component
 {
+
     /**
      * Returns a sorted list of files on a volume.
      *
      * @param Volume $volume The Volume to perform indexing on.
      * @param string $directory Optional path to get index list on a subfolder.
      * @return Generator
-     * @throws FsException
      */
     public function getIndexListOnVolume(Volume $volume, string $directory = ''): Generator
     {
         try {
             $fileList = $volume->getFs()->getFileList($directory);
-        } catch (VolumeException $exception) {
+        } catch (InvalidConfigException|FsException $exception) {
             Craft::$app->getErrorHandler()->logException($exception);
             return;
         }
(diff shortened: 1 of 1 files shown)


brandonkelly:
Thanks so much @mmikkel!

brandonkelly:
Craft 4.0.5 is out now with that fix.

PR #11386, opened 2022-06-05
Redesign condition builders’ “Add” buttons

brandonkelly:
This PR replaces condition the builder’s “Add” button with a menu, making it possible to immediately select the desired new condition rule type.

Previously adding a new condition rule was a two-step process: first press the “Add” button, then choose the desired condition rule type once the new rule was added. Now adding and choosing the type are a single action.

Before:

A condition builder with a single “Has URL” rule, and an “Add a filter” button below it

After:

A condition builder with a single “Has URL” rule, and an “Add a filter” menu button below it


gcamacho079:
There are a few open tickets about the a11y issues related to the Garnish custom select component (DEV-385, DEV-428, DEV-435), so switching from the previous UI (the "Add a filter" semantic `select` elements for the condition type selector) will render most of the condition builder inaccessible to screen reader users. I've been working on a new Garnish component that follows the [Select-Only Combobox pattern](https://www.w3.org/WAI/ARIA/apg/example-index/combobox/combobox-select-only.html), but with all the existing (and new) UI components that rely on that code, I'm wondering whether that's the best approach or whether we should just retrofit the existing component (with potentially breaking changes).

A couple other things:

Status Messages: The loading spinner is not announced - it should be given role=“status” and text should be injected when the loading is in progress:
`<div id="condition-spinner" class="spinner" role="status"><span class="visually-hidden">Loading...</span></div>`
When the loading is complete, the "Loading..." text should be replaced with a visually-hidden "Loading complete" message, then the `span` should be emptied of all text after a couple seconds so it isn't still available to screen reader users.

Info & Relationships: Instead of using `hr` and headings to create groups of options, either semantic `optgroup` or [ARIA groups](https://w3c.github.io/aria-practices/examples/listbox/listbox-grouped.html) should be implemented inside the custom selects. The first group of options (URI, Slug, etc.) should also be grouped - maybe “Default” and “Custom Fields” can be used as the option group labels?

Focus order:  Focus is moved to the “Add filter” button, but should remain on the dropdown for the new condition
When tabbing backwards from the “Add a filter” button, focus should return to the toggle button in the search bar

gcamacho079 (changes requested):
See comment

PR #11387, opened 2022-06-05
Label search for custom selects

brandonkelly:
Adds search support to Garnish custom selects, based on native `<select>` behavior.

- A search string will be built from any alphanumeric/space characters typed while the menu button is focused.
- The search string is reset after one second of inactivity, and when the menu is opened/closed.
- Whenever the search string is updated, the first menu option whose (lowercased & trimmed) text label begins the search string is identified. If the menu is visible, the option will be highlighted; otherwise it will be immediately selected.
- The <kbd>space</kbd> key can no longer be used to close the menu, as it had before.

Video of the new behavior in action:

https://user-images.githubusercontent.com/47792/172060595-1f7d0d15-cb02-49b5-abe1-2736f3b1305a.mp4


gcamacho079 (changes requested):
This overlaps with the discussion on the other PR (#11386) regarding the [Select-Only Combobox authoring pattern](https://www.w3.org/WAI/ARIA/apg/example-index/combobox/combobox-select-only.html).

Based on the [Keyboard Support documentation](https://www.w3.org/WAI/ARIA/apg/example-index/combobox/combobox-select-only#kbd_label_combobox) for these types of components, any printable characters typed when focus is on the combobox and the listbox is collapsed should:
1. Open the listbox
2. Set visual focus (based on `aria-activedescendant`) on the first option that matches the search string

A user must explicitly select the option using either spacebar, enter, or tab keys - the selection change shouldn't be able to be triggered by typing when focus is on the combobox itself.

PR #11394, opened 2022-06-07
Fix input of localized value in number field

nfourtythree:
### Description
If you have your user's formatting locale set to something different than the app and try to enter a localized number the value is saved incorrectly.

### DIFF
--- a/src/templates/_components/fieldtypes/Number/input.twig
+++ b/src/templates/_components/fieldtypes/Number/input.twig
@@ -8,7 +8,7 @@
 {%- set hasSuffix = suffix is not same as(false) %}
 
 {% if formatNumber %}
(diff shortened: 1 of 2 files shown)


src/templates/_components/fieldtypes/Number/input.twig
```diff
@@ -8,7 +8,7 @@
 {%- set hasSuffix = suffix is not same as(false) %}
 
 {% if formatNumber %}
-    {{ hiddenInput("#{field.handle}[locale]", craft.app.formattingLocale.id) }}
+    {{ hiddenInput("#{field.handle}[locale]", currentUser.getPreferredLocale()) }}
```
brandonkelly, line 11: If these are different (and the user does in fact have a formatting locale set), that’s the bug.

PR #11396, opened 2022-06-07
Tweak alignment of account photo change/delete buttons on small screens

gcamacho079:
### Description
On smaller screen sizes, the buttons for changing, deleting, and editing the user photo were not aligned with the image.

**Before:**
Action buttons in the photo section of account preferences are touching the photo when stacked vertically. They also have a left margin so are not vertically aligned with the user photo.

**After:**
Action buttons with a small gap between the photo and the top borders. The buttons are now vertically aligned with the user photo on the left-hand side.


brandonkelly:
Cherry-picked this over to `v3` and pushed there. (And tweaked it a bit to move the two button rows to flexbox as well, rather than using `<br>`s.)

PR #11400, opened 2022-06-08
Update incorrect German translation

MoritzLost:
### Description

Found another incorrect translation in German. The action on the user edit page `Deactivate…` is translated as `Wird deaktiviert…`, which means `Deactivating…`, as in `[User] is being deactivated`. Correct translation would be `Deaktivieren…` which is the imperative mood.

The PR fixes this translation in both `de` and `de-CH`.


brandonkelly:
Thanks! Pulled those changes into [Crowdin](https://crowdin.com/project/craft-cms).

FFR you might find it easier to make translation changes directly there.

MoritzLost:
@brandonkelly Thanks! Alright, I just signed up for Crowdin in case I find anything else 👍

PR #11409, opened 2022-06-09
Correct isSystemLive config override docblock

rynpsc:
### Description

Corrects the docblock for the `isSystemLive` config override that previously incorrectly referenced the `ipHeaders` override.

### Related issues

N/A


brandonkelly:
Thanks!

PR #11430, opened 2022-06-13
Make asset preview/edit buttons accessible to mobile users

gcamacho079:
### Description
A follow-up to #11266. Instead of blocking access to preview/edit buttons on mobile devices, moves button container below image.

View on image edit screen:
Screenshot showing equal-width Preview and Edit buttons stacked below the preview thumbnail. Buttons horizontally aligned with the image frame.

View within slideout editor:
Screenshot showing equal-width Preview and Edit buttons stacked below the preview thumbnail inside of a slideout editor. Buttons are not aligned with the image frame on the left and right, but instead respect the container spacing.


src/elements/Asset.php
```diff
@@ -2031,14 +2031,22 @@ public function getPreviewHtml(): string
                 ]) .
                 Html::tag('div', $this->getPreviewThumbImg(350, 190), [
                     'class' => 'preview-thumb',
-                ]);
+                ]) .
+                Html::endTag('div'); // .preview-thumb-container;
 
             if ($previewable || $editable) {
-                $html .= Html::beginTag('div', ['class' => 'asset-actions buttons']);
+                $isMobile = Craft::$app->getRequest()->isMobileBrowser(true);
```
gcamacho079, line 2038: 👏🏼

PR #11445, opened 2022-06-15
Make keyed data accessible with `ActiveFixture`

nfourtythree:
### Description
Craft's implementation of `ActiveFixture` overrides the `load` method. In doing so it is not populating the `data` property during the process.

This causes an issue as it means if you are using this class you are unable to retrieve the data from the fixture. Using `$this->tester->grabFixture('my-fixture', 'alias-key')` returns `null`. This PR fixes that issue and will now return the data for that fixture.

This all gives great flexibility when writing tests that use fixtures as you will be able to write assertions based on the actual data without having to do any extra queries.


src/test/ActiveFixture.php
```diff
@@ -63,7 +63,8 @@ public function load(): void
                 throw new InvalidArgumentException('Unable to save fixture data');
             }
 
-            $this->ids[] = $arInstance->id;
+            $this->data[$key] = array_merge($row, $arInstance->toArray());
```
brandonkelly, line 66: Should this be just `$this->data[$key] = $arInstance;` ?

nfourtythree: Yes probably, I think I got a bit tunnel-visioned trying to keep it close to the base method in Yii's `ActiveFixture class: https://github.com/yiisoft/yii2/blob/master/framework/test/ActiveFixture.php#L86

Issue #10621, opened 2022-02-21
Accessibility of categories modal


Leeroy-agency:
### Description
When selecting some categories in an entry, the ones already selected are harder to read, as the contrast ratio is poor.

![acc-structure](https://user-images.githubusercontent.com/65175280/155019972-773e7e58-405f-4117-9517-82e0e6d8e1ed.jpg)
![acc-structure-color](https://user-images.githubusercontent.com/65175280/155020306-11404250-de16-4638-9577-e7afea3102d2.jpg)

### Additional info

- Craft version: 3.7.33

gcamacho079:
@Leeroy-agency thanks for bringing this to our attention! Because these elements are inactive, there’s no hard requirement for them to meet the minimum contrast ratio of 4.5:1 against their background (see Incidental note in [Understanding 1.4.3 Contrast: Minimum](https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum)). However, I do agree it’s a valid usability and a11y issue.

Right now, increasing the text opacity will make it harder to differentiate which items are active versus not, unless we added another visual indicator like text style. I think that might be too subtle of an indicator for most users:

<img width="716" alt="Screenshot showing higher-opacity disabled elements with italic text, next to selectable elements with non-italic text" src="https://user-images.githubusercontent.com/17525979/166253357-3516cfa9-eef1-4bd9-b6dc-0cc9807a8add.png">

@brandonkelly we’ve discussed refactoring these tables to show the “Selected” column in order to support keyboard and screen reader users. This column would have `readonly` checkboxes to indicate their selected/disabled states, and editable checkboxes.   I think at that point, we could increase the opacity of inactive elements to at least 3:1 to make them easier to read. Thoughts?

brandonkelly:
@gcamacho079 I think it would be a little confusing why the checkboxes aren’t interactive, especially since they are within a modal that is used to select elements.

Maybe we can introduce a new accessibility preference, “Use italics to indicate disabled statuses”, which users could opt into. Then they’d be expecting to see the altered styling, and know what it means.

gcamacho079:
@brandonkelly I’m all for that extra accessibility feature regarding italic text.

  However, it’s still necessary to provide either a checkbox or toggle button for screen reader users. This has the benefit of: 
- Providing a natively focusable element. We can’t do that with the table row, as it would need a button/checkbox role (and we can’t do that, as it would override the table row semantics)
[...]

PR #11448, opened 2022-06-15
User preference for altered styling of selected elements in element selector modals

gcamacho079:
### Description
Adds a new user preference for improved contrast of selected elements in element selector modals.

### Related issues
#10621


gcamacho079:
@brandonkelly not quite ready for review yet, just wanted to run something by you. Just applying italicized text makes it difficult to distinguish which are selected, but italics _and_ bold makes them stand out (especially when selecting a category, where the name is the only text visible by default). Thoughts?

Screen Shot 2022-06-15 at 10 58 19 AM

PR #11451, opened 2022-06-16
Fix type comparison in element condition rules

lukeholder:
Type mismatch on the element ID prevented element condition rules from matching correctly.


src/base/conditions/BaseElementSelectConditionRule.php
```diff
@@ -190,14 +190,14 @@ protected function matchValue(mixed $value): bool
         }
 
         if (is_numeric($value)) {
-            return (int)$value === $elementId;
+            return (int)$value === (int)$elementId;
```
lukeholder, line 193: `$elementId` can be a string.

PR #11502, opened 2022-06-24
Add title default to elementindex template

timkelty:
For custom element types using `_layouts/elementindex.twig`


brandonkelly:
Good idea

PR #11534, opened 2022-06-30
A11y/use fieldsets for prefs

gcamacho079:
### Description
Use fieldsets for accessibility and dev preferences (Resolves DEV-333).

Also fixes an issue where `fieldset` elements are nested inside `fieldset` elements for both radio groups and checkbox groups.


src/templates/_includes/forms/checkboxGroup.twig
```diff
@@ -4,6 +4,7 @@
 
 {% set fieldsetAttributes = {
     class: ['checkbox-group'],
+    role: 'presentation',
```
gcamacho079, line 7: @brandonkelly I added this for both the `checkboxGroup` and the `radioGroup` as I wasn't sure if changing the `fieldset` element to a `div` would be problematic; if not let me know and I can modify the tag itself.

brandonkelly: Probably safe to just change the tag.

gcamacho079: Will do!

brandonkelly:
👍🏻 Replaced “Location” with “General”, and also split the Debug Toolbar-related development settings into their own checkbox group.

The new user preferences form, with 2nd level headings “General”, “Accessibility”, and “Development”

PR #11556, opened 2022-07-06
Allow `$dateCreated` and `$dateUpdated` to be `DateTime` objects on `ActiveRecord`

nfourtythree:
### Description
Due to the fact that [this line](https://github.com/craftcms/cms/blob/develop/src/db/ActiveRecord.php#L140) will take care of preparing the data for the DB, we can allow these properties to be set to `DateTime` objects.


brandonkelly:
🙏🏻

PR #11566, opened 2022-07-07
Conditionally show require `*` for address fields based on validation rules

nfourtythree:
### Description
Now that the validation for addresses is all part of the rules we can try and help the content editor by showing the `*` only when the field is actually required.

This helps when a developer may have changed the default validation rules using the define rules event.


brandonkelly:
We should not be factoring in the scenario, as it’s helpful to know which fields are required regardless of whether they will actually be validated on save.

PR #11589, opened 2022-07-12
Add high-contrast indicator to active sidebar item

gcamacho079:
### Description
Resolves DEV-372. Also fixes left sidebar edge going offscreen on small screen sizes.

Shows global and index sidebar navigation with a bar on the left side of the item to show the active page


brandonkelly:
Wasn’t loving the added indicator strips—they felt a little too noisy—so went with making the entire active nav items high-contrast instead:

Screenshot of a control panel page, showing that the active global nav item and the active subnav item each have a high-contrast background color, relative to their containers’ background colors.

PR #11604, opened 2022-07-14
Expose element status to screen readers

gcamacho079:
### Description
Expose element status to screen readers user on element index pages (Resolves DEV-439, DEV-488)


brandonkelly:
@gcamacho079 What was the reason for moving that code to a new `getStatusIconHtml()` method?

gcamacho079:
@brandonkelly I thought it might be helpful in case we needed to use it elsewhere in the code (think I may have been playing around with something like that on a local branch 🤔 ) and for readability.

brandonkelly:
Alright, well I brought the code back inline into `Cp::elementHtml()` for now. Feel like there should be known additional use cases to justify adding new methods to `ElementInterface`, as it’s already pretty massive.

Issue #7142, opened 2020-11-16
Notifications accessibility


missmatsuko:
### Description

Regarding the notification messages that appear at the top of the screen. For example, when entries are saved.

<img width="158" alt="Screenshot of 2 notifications, both reading 'test'. First is blue and second is red." src="https://user-images.githubusercontent.com/17884198/99232971-7885fa00-27a7-11eb-8f75-93ae75d99582.png">

- Use aria live regions/roles such as `status` and `alert` to announce this text to screen readers. (Related: #7139)
- [WCAG use of color](https://www.w3.org/WAI/WCAG21/Understanding/use-of-color.html) failure: background color is used to distinguish between notification types (blue for notice, red for error). 
Recommend adding text or an icon with label to fix.
- [Color contrast of the white text on blue background is insufficient for normal text](https://webaim.org/resources/contrastchecker/?fcolor=FFFFFF&bcolor=127FBF).
Recommend using a darker blue for the background color.
- [Timing Adjustable](https://www.w3.org/WAI/WCAG21/Understanding/timing-adjustable.html) failure: the notification messages display for a few seconds (looking at the code, I think it's 2 seconds for notices, 4 seconds for errors). 
This may not be enough for some users to read all the text. Looking at the linked page, as well as [the related discussion on the wcag github repo](https://github.com/w3c/wcag/issues/976), I think it makes sense to have user settings so users can set a duration up to 10 times the default duration, or to disable the duration altogether and have the user manually dismiss the notifications. I think it's probably not ideal to have this longer duration or manual dismissal for all users since they could stack up pretty quickly. 

### Additional info
- Craft version: 3.5.15.1

tomkiss:
A related point which may be interesting for you to hear.

One of the ways in which I work with Craft CMS, is that for multiple editing I may open many edit tabs for different entries at a time - sometimes 10 or 15 or more tabs.
Depending on which environment I am working on, the response time for saving an entry might take some time. Because of this, I may edit an entry, hit `cmd + s` to save, and switch to the next tab to edit that one, and repeat that flow.

A problem with the timer on the "_Entry saved_" notification is that when I go back to another tab, because it was displayed with a timer, there's no clear indication as to whether I saved it or not (unless I check the timestamp, and even that may not be obvious enough). So quite often I end up saving again, just to be sure.

It's no biggie, but this particular use case occurs so frequently so it can be a bit annoying not knowing if I recently saved or not. A solution to the accessibility issue may help this use case too. 
[...]

PR #11612, opened 2022-07-15
Notification redesign

brandonkelly:
### Description

Control panel user notifications have been redesigned to stack up from the bottom-left of the window, with a cleaner and more accessible design.

An “Entry saved” notification, with a link to the entry.

There are several additional improvements:

- The duration is now 5 seconds by default, and there’s a new “Notification Duration” user preference where users can customize that to their liking.

The new “Notification Duration” user preference, with the options “2 seconds”, “5 seconds”, “10 seconds”, and “Show them indefinitely”.

- If the browser tab isn’t visible when a notification is first shown, the timeout won’t start until the tab becomes visible.
- If a notification is hovered over by the cursor, the timeout will pause and restart on mouse-out.
- If a notification is clicked on, the timeout will be cancelled altogether.
- Notifications can be manually closed via the Close button at any time.
- Notifications can include additional details, links, or buttons, depending on the context.
- Success notifications can now distinguish themselves from neutral “notices” by calling `Craft.cp.displaySuccess()` instead of `Craft.cp.displayNotice()`.

We are taking advantage of the additional details in two ways currently:

- When elements are saved, a link to the element will be included.
- Dashboard widget delete notifications now include an “Undo” button, which will create a new widget with the same settings as the deleted widget. (This replaces the need for a delete confirmation.)

A “Widget deleted” notification, with an “Undo” button

### Related issues

- #7142
- #11573


gcamacho079 (changes requested):
I made a couple smaller tweaks after testing on VoiceOver. Some other changes that should be made are:

### Focus Management
- If there's an interactive element in the (like undo button), `tabindex="-1"` should be added to the new `.notification` container, focus should be moved to it, and the duration timeout should pause while focus is on or inside the container.
- If the text content overflows and adds scrollbars, `tabindex="-1"` should be added to the `.notification` container and focus should be moved to it. This will allow keyboard users to scroll the content.
- When the notification closes via ESC keypress inside the container, interactive element click (i.e. Undo button), or close button keypress, **focus should move to the previously-focused element on the page**.

### Other
Something to look into in the future would be a place in the control panel where users can go and review previous toast messages; this would be a benefit for users with short term memory loss, or who might otherwise miss or accidentally dismiss messages.

tomkiss:
Excellent notifications update - Well done everyone! 🙌

thijskaspers:
@brandonkelly Is there an option to change the position of the notification that pops up?

We've gotten client feedback that they "Saved" a page in the CP, but don't see the changes on the front-end. What actually happened was, they forgot to fill in a required field, but it wasn't clear to them.

If you work on a larger monitor and click on "Save" top right of the screen, you can easily miss the "Couldn't save entry" notification at the bottom left of the screen, that isn't where your focus is after clicking save top-right.

We agree that bottom left is the nicest position and less intrusive, but for less tech-savy clients it would be better in our opinion to have the option to place it top-center.... or: Make it more clear that something went wrong, make the Tab with the exclamation icon 'red'? Add an additional error text? Scroll down to the first error after save? Something like that.

brandonkelly:
@thijskaspers We are looking into introducing an error summary at the top of the page, which should help. You can also point your clients to the “Notification Duration” setting in their user account preferences, which they can set to a longer duration or have them display indefinitely.

brandonkelly:
(If you still want to advocate for a “Notification Position” preference, please open a new discussion about it.)

thijskaspers:
> 

Nah, an error summary at the top of the page sounds like a good solution, to make it more clear there were validation errors for 1 or more fields.

brandonkelly:
@thijskaspers For posterity, Craft 4.5.0 was released today, with the new validation error summary pane.

thijskaspers:
> @thijskaspers For posterity, Craft 4.5.0 was released today, with the new validation error summary pane.

Awesome! Nice improvement, I think clients will like it, thanks!

PR #11636, opened 2022-07-19
Update slideout to use mobile styles under 600px viewport width

gcamacho079:
### Description
Resolves DEV-701, DEV-480


brandonkelly:
@gcamacho079 Is there a reason you went with 600px for this?

I was thinking it would be good to move `Slideout.js::useMobileStyles()` to `Craft.js` so we can use it for other things as well. But it would be good to be consistent about what we consider a “mobile width”.

Looking through the CSS media queries, looks like the smallest viewport size we check for currently is 500px, when determining whether 25/50%-width fields are allowed to be displayed side-by-side:

https://github.com/craftcms/cms/blob/1c7f4ee60f10ffc896534a126068f72b1d04eeca/src/web/assets/cp/src/css/_cp.scss#L1585-L1591

Do you think we should change that media query to start at 600px, or drop the `useMobileStyles()` check down to 500px?

gcamacho079:
> Do you think we should change that media query to start at 600px, or drop the useMobileStyles() check down to 500px?

My rationale here is that at 500px wide, the slideout is only about 250px wide. When I was testing in responsive design mode, 300px seemed like a comfortable minimum width for it.

We would always have the `useMobileStyles` function use a default parameter of `screenWidth = 600`, and just override it for the slideout in particular. 🤔

brandonkelly:
👍🏻 Went with 600px.

PR #11645, opened 2022-07-20
[Feature proposal] Event when user isn't found & ability to externally override the user (LDAP example)

bertoost:
### Description

In Craft 2, I had developed a LDAP plugin which created users when not found locally in the Craft CMS system. This plugin also had interface settings and options to configure which groups from LDAP should have which permissions in Craft etc.

When a user is not in the Craft CMS system, there is no way (as far I know of) to check an external source, create the user and continue the authentication process, since the user is immediately told the user was not found.

The idea of adding this event is to give plugins the ability to hook in and create the Craft user and continue authentication.

With this update I have managed to create the local user, validate the external password and login to Craft like a normal user.

### Explanation

- When a user isn't found in Craft, fire new event
- Plugins can hook in to check their external source, like LDAP
- When found, create the local Craft user (with all their necessary settings)
- Set the user back to the event, and tell Craft; it's handled!
- Then the "before authenticate" is triggered...
- The plugin can check the password in the external source
- When validated, update the local Craft user's password (for local fallback purposes)
- And the user will be logged in with their account.

### Further development...

Probably you have some thoughts, better ideas, better naming etc.
This is like a PoC proposal to get you involved and to work on improvements.

Like to hear again. Thanks!

Regards, Bert


src/controllers/UsersController.php
```diff
@@ -167,9 +173,22 @@ public function actionLogin(): ?Response
...
 
         if (!$user || $user->password === null) {
-            // Delay again to match $user->authenticate()'s delay
-            Craft::$app->getSecurity()->validatePassword('p@ss1w0rd', '$2y$13$nj9aiBeb7RfEfYP3Cum6Revyu14QelGGxwcnFUKXIrQUitSodEPRi');
-            return $this->_handleLoginFailure(User::AUTH_INVALID_CREDENTIALS);
+            
+            $event = new LoginUserNotFoundEvent([
+                'loginName' => $loginName,
+            ]);
+            $this->trigger(self::EVENT_LOGIN_USER_NOT_FOUND, $event);
```
brandonkelly, line 180: I don’t think this should be triggered in the event that `$user` was found but `$user->password === null`. Maybe just check for `if (!$user)` here, and then add a separate `if ($user->password === null)` check after it, keeping the original conditional block inside it.

bertoost: Yeah I understand.. Chang on it's way

brandonkelly:
Made a bunch of tweaks – there’s now two events: `EVENT_BEFORE_FIND_LOGIN_USER` and `EVENT_AFTER_FIND_LOGIN_USER`, which will each get triggered regardless of whether a user is found.

And merged for 4.2!

bertoost:
Thanks @brandonkelly ! That was the idea of my PoC and I'm thrilled this is going to be in 4.2.
Will continue to develop my LDAP plugin 😄

brandonkelly:
4.2.0 is out now with those events ✨

PR #11662, opened 2022-07-22
More small fixes

gcamacho079:
### Description
- Add context to Choose button for structured entries and categories (Resolves DEV-461)
- Lighten active button color for min text color contrast (Resolves DEV-438)
- Add accessible name to Related To select element (Resolves DEV-717, DEV-422)
- Add accessible name to file size select (Resolves DEV-550)
- Provide accessible names for rule type custom select only comboboxes (Resolves DEV-716)
- Move t9n indicator into label element (Resolves DEV-560)
- Add col and colgroup elements to editable table (Resolves DEV-536)
- Manage focus when deleting editable table rows (Resolves DEV-542)
- Update checkbox cells to remove ARIA-on-unallowed error (DEV-340, DEV-710)


src/templates/_includes/forms/editableTable.twig
```diff
@@ -53,6 +53,15 @@
 
 <span aria-live="assertive" class="visually-hidden" data-status-message></span>
 {% tag 'table' with tableAttributes %}
+    {% for col in cols %}
```
gcamacho079, line 56: @brandonkelly a little stuck on this one. For table fields, I'm trying to follow [this tutorial for including colgroup/col tags](https://www.w3.org/WAI/tutorials/tables/irregular/#table-with-headers-spanning-multiple-rows-or-columns) as direct children of the table element.

For a table with 4 columns and remove/reorder, this is what should be immediately after the opening `table` tag:

```
<col>
<col>
<col>
<col>
<colgroup span="2"></colgroup>
```

What's actually happening:
```
<colgroup>
  <col>
  <col>
  <col>
  <col>
</colgroup>
<colgroup span="2"></colgroup>
```
Any tips for getting those col elements to be direct children instead of being wrapped in colgroup?

brandonkelly: If I `view-source:` on a URL with an editable table, it is outputting the `<col>`s as you’d expect:

```html
<table id="sites" class="editable fullwidth">
  <col>
  <col>
  <col>
  <col>
  <col>
  <col>
  <col>
  <thead>
  ...
```

However the browser is automatically wrapping those with a `<colgroup>`. I’m guessing it’s like how you go straight to `<tr>` it will be auto-wrapped with a `<tbody>`, since `<tbody>` is required if there are body rows. Probably `<colgroup>` is required as well.

gcamacho079: Brilliant! I just checked the tutorial page I sent over and it's also wrapping this in a `colgroup`. Mystery solved! 🕵🏼

brandonkelly: @gcamacho079 So to be clear, this all looks as you’d expect?

An editable table divided into two colgroups – one for the content columns, and another for the two action columns (reorder and delete)

An editable table with a header column, showing that all columns including the header column live within a single colgroup

gcamacho079: @brandonkelly first one is looking good. Taking a look at the **Site Settings** table, it seems there are two columns with a `hidden` class on them. Are these ever shown?
I think there should only be enough `col` elements to correspond with non-hidden columns - so 5 in this case.

brandonkelly: @gcamacho079 I don’t think they would ever be shown within the current page load, no.

Issue #11735, opened 2022-08-03
[4.x]: Setting database config via url doesn't work


mikejpeters:
### Description

Trying to set the database config using `url` results in a database connection error,
even when the URL is specified correctly.

### Steps to reproduce

1. Create a local MySQL database named `test`
1. Create `config/db.php` with contents:
    ```php
    return [
      'url' => 'mysql://127.0.0.1/test'
    ];
    ```

### Expected behavior

Craft should be able to connect to the database given the `url`.
Database config properties like `server` and `dsn` should be calculated from the `url`.

### Actual behavior

Craft cannot connect to the database, and the resulting database config has no value for `server`.

### Craft CMS version

4.2.0.2

brandonkelly:
Thanks for reporting that. Just fixed for the next release.

To get the release early, change your `craftcms/cms` requirement in composer.json to `dev-develop as 4.2.0.2` and run `composer update`.

mikejpeters:
I'd considered that as a fix, but I think because the individual properties aren't being set from the url (only the dsn), if you use the `DbConfig` to first set `url()` and then some other property, e.g. `database`, all other properties will be cleared and you'll have an invalid DSN again.

For example, with simply removing the `_updateDns()` this test will fail:

```php
$config = DbConfig::create()
  ->url('pgsql://127.0.0.1:5432/db')
  ->database('db2');
$this->assertEquals('db2', $config->database);
$this->assertEquals("pgsql:host=127.0.0.1;dbname=db2;port=5432", $config->dsn);
```

brandonkelly:
Good point – fixed that as well. Originally with something closer to your PR (except only calling `dsn()`), but then realized it could be more performant if `url2config()` just returns the `driver`, `server`, `port`, and `database` keys in the returned config array, as it already has those parsed out of the URL.

brandonkelly:
Craft 4.2.1 is out with that fix.

PR #11736, opened 2022-08-03
Fixed database config from url

mikejpeters:
### Description

Ensure that necessary database config properties are set from the resulting config when setting `url`.

`Db::url2config($value)` returns an array with a `dsn` key, and it's important it's set using the  `DbConfig::dsn($value)` method (instead of just being assigned to the `dsn` property), because that function sets a bunch of properties from the dsn.

### Related issues

#11735

### DIFF
--- a/src/config/DbConfig.php
+++ b/src/config/DbConfig.php
@@ -610,7 +610,10 @@ class DbConfig extends BaseConfig
     public function url(?string $value): self
     {
         if ($value) {
-            Craft::configure($this, Db::url2config($value));
+            $config = Db::url2config($value);
+            foreach ($config as $configKey => $configValue) {
+                $this->{$configKey}($configValue);
+            }
             $this->_updateDsn();
         }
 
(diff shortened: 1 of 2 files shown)


brandonkelly:
Thanks for the PR! Really all that was needed was dropping the `$this->_updateDsn()` line though, as `Db::url2config()` was already returning the generated DSN string. Resolved via f862a4b653a87cb120e490cdc3208b5d14b6da6f.

mikejpeters:
I appreciate the excellent work you're doing and that you looked into this, but I find it a bit frustrating that I couldn't contribute. I spent a few hours on this trying to do it well and ensure I followed all the instructions in the contributing doc. Maybe if you don't want to deal with pull requests add a note in the contributing doc? I could have saved a lot of time if I'd just reported the issue and left it at that.

brandonkelly:
Sorry to disappoint. The PR was great, especially since it included tests. It just wasn't the right fix IMO, and implementing the right fix was just a matter of deleting a line, so not worth the trouble of going back and forth over a PR.

PR #11747, opened 2022-08-04
Bump @craftcms/vue `1.0.1`

nfourtythree:


brandonkelly:
Shouldn’t this include a new `package-log.json` and built Vue assets?

nfourtythree:
Nope, this package is simply a collection of Vue source components for inclusion within other assets.

brandonkelly:
Right but then won’t those other assets need to be rebuilt?

nfourtythree:
They were already rebuilt in the other PR. Because they use relative paths. After our decisions the other day in the future this will all be in one PR

PR #11752, opened 2022-08-09
Update UsersController.php

leevigraham:
Add additional attributes to user in `actionSessionInfo`

Only potential issue I see here is including the photo properties as they will require a few extra DB calls.

I still see this being useful on static sites where you want to show a logged in user in a header with their name and other properties.


brandonkelly:
Feels like a slippery slope to me – everything that’s currently returned is either actually session info or user-identifying info. If you want general info about the user, the GraphQL API is probably a better choice, which already allows you to fetch all of the things you’ve added here, and much more.

PR #11754, opened 2022-08-09
Add getCurrentUser convenience method

timkelty:
Seems like this gets written enough it warrants a helper, similar to how `request`, `response` are props.

I made it `currentUser` to match the twig global.


brandonkelly:
Should be safe to have the method return `?User` rather than `?IdentityInterface`.

timkelty:
I think so, yeah – that might have been to appease PHPstorm/stan…but yeah, the idea is you'd always get back the user element.

timkelty:
@brandonkelly now returning `?User`

brandonkelly:
Not really a fan of the `...$args` thing. No way of knowing what to actually pass in when calling it, without source diving into `getIdentity()`.

timkelty:
> 

Fair enough. Just matched the signature of `\yii\web\User::getIdentity` and added an `@see`

PR #11765, opened 2022-08-11
Fix issue introduced in PR #11563 that caused selectize inputs to shr…

gcamacho079:
…ink when value was deleted

### Description
Resolves DEV-965


brandonkelly:
@gcamacho079 Since this is more of a bugfix and #11563 is already merged/released, can you cherry-pick this fix onto a new branch off of `develop`?

PR #11774, opened 2022-08-12
EVENT_AFTER_CREATE_TWIG

brandonkelly:
Adds an `EVENT_AFTER_CREATE_TWIG` event to `craft\web\View::createTwig()`, which can be called by plugins if they need to fire at the moment a Twig environment is created.

Since Craft creates separate Twig environments for control panel and site templates, the event class has a `templateMode` property which will either be set to `cp` or `site` depending on the active template mode, in case it’s relevant to the handler.

```php
use craft\events\CreateTwigEvent;
use craft\web\View;
use yii\base\Event;

Event::on(
    View::class,
    View::EVENT_AFTER_CREATE_TWIG,
    function(CreateTwigEvent $event) {
        if ($event->templateMode === View::TEMPLATE_MODE_SITE) {
            $event->twig->addGlobal('myGlobalVariable', 'foo');
        }
    }
);
```

### DIFF
--- a/CHANGELOG-WIP.md
+++ b/CHANGELOG-WIP.md
@@ -1,7 +1,9 @@
 # Release Notes for Craft CMS 4.3 (WIP)
 
 ### Added
+- Added `craft\events\CreateTwigEvent`.
 - Added `craft\web\Controller::getCurrentUser()`. ([#11754](https://github.com/craftcms/cms/pull/11754))
+- Added `craft\web\View::EVENT_AFTER_CREATE_TWIG`. ([#11774](https://github.com/craftcms/cms/pull/11774))
 
 ### Changed
 - Improved the control panel accessibility. ([#11534](https://github.com/craftcms/cms/pull/11534), [#11565](https://github.com/craftcms/cms/pull/11565), [#11578](https://github.com/craftcms/cms/pull/11578), [#11589](https://github.com/craftcms/cms/pull/11589), [#11610](https://github.com/craftcms/cms/pull/11610), [#11613](https://github.com/craftcms/cms/pull/11613), [#11768](https://github.com/craftcms/cms/pull/11768))
--- a/src/events/CreateTwigEvent.php
+++ b/src/events/CreateTwigEvent.php
@@ -1,0 +1,30 @@
+<?php
+/**
+ * @link https://craftcms.com/
+ * @copyright Copyright (c) Pixel & Tonic, Inc.
+ * @license https://craftcms.github.io/license/
+ */
+
+namespace craft\events;
+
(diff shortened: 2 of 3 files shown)


khalwat:
So this is great except for one small thing... this sequence of code:

```php
        // Set our timezone
        /** @var CoreExtension $core */
        $core = $twig->getExtension(CoreExtension::class);
        $core->setTimezone(Craft::$app->getTimeZone());

        // Fire an 'afterCreateTwig' event
        if ($this->hasEventHandlers(self::EVENT_AFTER_CREATE_TWIG)) {
            $this->trigger(self::EVENT_AFTER_CREATE_TWIG, new CreateTwigEvent([
                'templateMode' => $this->_templateMode ?? self::TEMPLATE_MODE_SITE,
                'twig' => $twig,
            ]));
        }
```

...doesn't allow an event listener to add extensions to Twig, because the call to `$twig->getExtension()` initializes Twig, so an exception is thrown if you try to add a Twig extension inside of the event handler or do something else that wants to modify the Twig environment.

You might want to change the event sequence to be:

```php
        // Fire an 'afterCreateTwig' event
        if ($this->hasEventHandlers(self::EVENT_AFTER_CREATE_TWIG)) {
            $this->trigger(self::EVENT_AFTER_CREATE_TWIG, new CreateTwigEvent([
                'templateMode' => $this->_templateMode ?? self::TEMPLATE_MODE_SITE,
                'twig' => $twig,
            ]));
        // Set our timezone
        /** @var CoreExtension $core */
        $core = $twig->getExtension(CoreExtension::class);
        $core->setTimezone(Craft::$app->getTimeZone());
        }
```

brandonkelly:
To register Twig extensions, you should be using `Craft::$app->view->registerTwigExtension()`, which will ensure the extension gets included when Twig is initialized. (And if Twig is already initialized, it will discard the existing instance.)

khalwat:
Right, but even if we added the extension that way via the `EVENT_AFTER_CREATE_TWIG` event, for the reasons I mentioned above, it would fail because the Twig environment has been initialized already.

Twig lazily initializes itself if `$twig->getExtension()` is called... so moving the event up before you call that would allow an uninitialized Twig to be modified by the event.

Without that, there really isn't much that can be done in that event callback in terms of modifying the Twig environment before Craft uses it.

khalwat:
Okay I looked at the code for `View::registerTwigExtension()` and see what you mean, in that it will `null` out the appropriate Twig environment (which causes it to be recreated later).

So for this specific thing, yes, that's a workable solution. There are other ways you might want to modify the Twig environment that it still might be useful to move the event to before Twig gets initialized, IMO, but up to you.

PR #11777, opened 2022-08-13
Make nested sources actually nested under headings

brandonkelly:
Nests any element sources that come after a heading, into an `<ul>` within the heading’s `<li>`.

So a source list with the following structure:

- Source 1
- **Heading 1**
  - Source 2
  - Source 3
    - Source 4

Will now get the following DOM:

```html
<nav>
  <ul>
    <li><a>Source 1</a></li>
    <li class="heading">
      <span>Heading 1</span>
      <ul>
        <li><a>Source 2</a></li>
        <li>
          <a>Source 3</a>
          <ul class="nested">
            <li><a>Source 4</a></li>
          </ul>
        </li>
      </ul>
    </li>
  </ul>
</nav>
```

Previously any sources that were “nested” under a heading were placed at the top level of the source list `<ul>`, so the hierarchy wasn’t clear to screen reader users (DEV-365).


gcamacho079 (approved):
Looks great, @brandonkelly!

PR #11790, opened 2022-08-17
feat: Add the ability to `.select()` custom fields by their field handle

khalwat:
### Description

Currently, if you want to do something like:

```twig
    {% set result = craft.entries.section('blog').select(['plainText']).asArray().all() %}
```

...where `plainText` is a custom field handle, it won't work, because field columns in the content table have a prefix, and sometimes a suffix as well.

This makes it nigh impossible to do the above code from Twig, because you need to call `ElementHelper::fieldColumnFromField()` on the field handle to get the column, and `ElementHelper` isn't directly available in Twig.

This makes using `.select()` for optimal queries somewhat limited via Twig, as often what you care about is custom fields.

This PR leverages the existing `yii\db\Query::normalizeSelect()` method to swap in the real column name for custom fields (after Yii2 has already normalized them to an array), so the above "just works" and returns something like this:

```
[
    0 => [
        'plainText' => 'Some plain text'
    ]
]
```

It'll also "just work" via PHP as well. It only affects `ElementQuery`'s, and only when `.select()` is used, so it should be pretty unintrusive.

### DIFF
--- a/src/elements/db/ElementQuery.php
+++ b/src/elements/db/ElementQuery.php
@@ -1969,7 +1969,6 @@ class ElementQuery extends Query implements ElementQueryInterface
                 if (!empty($select[$handle])) {
                     $column = ElementHelper::fieldColumnFromField($field);
                     if ($column) {
-                        unset($select[$handle]);
                         $select[$handle] = $column;
                     }
                 }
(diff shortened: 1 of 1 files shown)


brandonkelly:
Implemented the same feature but in a different way via #11800, for Craft 4.3 🎉

PR #11792, opened 2022-08-17
feat: Add type conversion filters: `boolean`, `integer`, `float`, & `string`

khalwat:
### Description

With Craft moving to having explicitly typed APIs in Craft 4, there are times when it is useful to be able to typecast via Twig to ensure that parameters are of the expected type.

This PR adds the following filers ([taking a page from Drupal](https://www.drupal.org/docs/contributed-modules/twig-tools/conversion-filters)): `boolean`, `integer`, `float`, & `string`

Example usage:

```twig
    {% set val = "true" | boolean %}
    {% set val = "2135" | integer %}
    {% set val = "2135.1" | float %}
    {% set val = 2135.1 | string %}
```

These just map to the PHP functions `boolval`, `intval`, `floatval`, and `strval`, respectively.

Worth noting: _tests_ for all of these types exist already, just not conversion filters: https://craftcms.com/docs/4.x/dev/tests.html


brandonkelly:
Thanks!

PR #11794, opened 2022-08-17
Increase element query performance for structures

sjeng:
Replaced a filtering subquery with a left join for better query performance.

### Description
We started experiencing long load times when editing entries with pretty complex content models. I started doing some digging to see where are the bottlenecks. 

One area that we were able to improve was the element query when structures are involved. It appears that there is a `left join` with `structureelements` with a conditional that is a subquery. 

My understanding of the intent of the subquery is to filter out elements that are associated with structures that has been deleted. So I ended up recreating the same filter with a `left join` and an additional `where` condition that should do the same thing.

The change returns the same dataset across my testing, and performance was pretty dramatic in some cases. I had a query that returned 60 records that was taking 1 second to complete before update, and .02 sec after the update (our content may be overly complex...).

Please take a look at my proposed change and see if there's anything I missed.


sjeng:
I need to rethink this patch and adjust the query to better represent the original intent

PR #11800, opened 2022-08-18
Support mapped columns in element query SELECTs

brandonkelly:
### Description

Element queries’ `orderBy()` methods currently support passing in “column names” that would cause an “ambiguous column” SQL error (e.g. `dateCreated` could refer to `elements.dateCreated` or `entries.dateCreated`) as well as custom field handles rather than the full field column name (`fooBar` instead of `field_fooBar_xxxxxxxx`), by mapping the ambiguous/incomplete column names to the correct ones.

This PR extends the same column mapping support to element queries’ `select()` methods, so the following would be possible:

```php
craft\elements\Entry::find()
    ->select(['dateCreated', 'fooBar'])
    ->asArray()
    ->all()
```

### Related issues

- #11790

### DIFF
--- a/CHANGELOG-WIP.md
+++ b/CHANGELOG-WIP.md
@@ -26,6 +26,7 @@
 - `users/session-info` responses now include a `csrfTokenName` key. ([#11706](https://github.com/craftcms/cms/pull/11706), [#11767](https://github.com/craftcms/cms/pull/11767))
 - Twig templates now have `today`, `tomorrow`, and `yesterday` global variables available to them.
 - Element query date params now support passing `today`, `tomorrow`, and `yesterday`. ([#10485](https://github.com/craftcms/cms/issues/10485))
+- Element queries now support passing ambiguous column names (e.g. `dateCreated`) and field handles into `select()`. ([#11790](https://github.com/craftcms/cms/pull/11790), [#11800](https://github.com/craftcms/cms/pull/11800))
 - `craft\helpers\Component::iconSvg()` now namespaces the SVG contents, and adds `aria-hidden="true"`. ([#11703](https://github.com/craftcms/cms/pull/11703))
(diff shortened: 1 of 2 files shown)


brandonkelly:
What I’m not sure about is, in the above example, should the response contain the actual `field_fooBar_xxxxxxxx` column name, or just the field handle?

```php
=> [
     [
       "dateCreated" => "2015-02-10 19:09:38",
       "field_fooBar_xxxxxxxx" => "fooBar field value",
     ],
   ]
```
or
```php
=> [
     [
       "dateCreated" => "2015-02-10 19:09:38",
       "fooBar" => "fooBar field value",
     ],
   ]
```

The actual column name would be technically correct, but maybe less helpful. (If you already knew the full column name and knew to expect that in the response, you could have just as easily passed the full column name into `select()`.)

khalwat:
So originally I had it working as you show in the first example in https://github.com/craftcms/cms/pull/11790

The problem then is in a Twig template, you have no way of knowing what the key value should be for that field.

You'd need to expose the same `ElementHelper::fieldColumnFromField()` in Twig for people to be able to find the data they are looking for.

khalwat:
One other thing I did consider was adding a `.fieldSelect()` method instead, to be explicit about it... but I don't think that's how most people who are writing queries will be thinking about things. Probably nicer for them to have a single `.select()` where they can mix and match things like:

```php
craft\elements\Entry::find()
    ->select(['title', 'plainText'])
    ->asArray()
    ->all()
```

instead of:

```php
craft\elements\Entry::find()
    ->select(['title'])
    ->selectField(['plainText'])
    ->asArray()
    ->all()
```

brandonkelly:
Yeah main issue with the `selectField()` idea is you lose control over the order of the keys in the returned arrays.

khalwat:
I agree -- I didn't say I _liked_ it, just that I thought of it :)

brandonkelly:
Alright, updated to maintain the passed-in mapped column name as the column alias, when `asArray()` is used.

brandonkelly:
Craft 4.3 was released earlier this week with this change!

PR #11801, opened 2022-08-18
Adds the craft cli as a composer bin file

markhuot:
There are some times (specifically during plugin development) where I don't need the full `craftcms/craft` setup and all I really need is the Craft CLI so I can run a queue or execute a one-off task. Adding the `bin` to `composer.json` ships the CLI with the `craftcms/cms` dev so dependent projects can easily run Craft via `./vendor/bin/craft install`, for example.

With the new Craft 4 ENV changes, too, I don't even need any `config/` dir or anything like that.

After this change it is possible (in a new directory) to do something like this and have a full Craft CLI running.

```sh
$ composer require craftcms/cms
$ CRAFT_DB_NAME=craft ./vendor/bin/craft install
```

Optionally, you could explore shortening the `craftcms/craft` CLI even more and just pointing to this in the vendor directory.

This is a proof of concept and for your consideration. There are probably downsides (but I can't think of them) and other changes that must be made before this merges in. Just something I was working on and found helpful of late.

### DIFF
--- a/bin/craft
+++ b/bin/craft
@@ -1,0 +1,32 @@
+#!/usr/bin/env php
+<?php
+/**
+ * Craft console bootstrap file
+ */
+
+if (!empty($_composer_autoload_path)) {
+    $basePath = dirname($_composer_autoload_path) . '/../';
+}
+else {
+    $basePath = getcwd();
+}
+
+// Define path constants
+define('CRAFT_BASE_PATH', $basePath);
+define('CRAFT_VENDOR_PATH', CRAFT_BASE_PATH . '/vendor');
+
+// Load Composer's autoloader
+require_once CRAFT_VENDOR_PATH . '/autoload.php';
+
+// Load dotenv?
+if (class_exists(Dotenv\Dotenv::class)) {
+    // By default, this will allow .env file values to override environment variables
+    // with matching names. Use `createUnsafeImmutable` to disable this.
+    Dotenv\Dotenv::createUnsafeMutable(CRAFT_BASE_PATH)->safeLoad();
+}
+
+// Load and run Craft
+/** @var craft\console\Application $app */
+$app = require CRAFT_VENDOR_PATH . '/craftcms/cms/bootstrap/console.php';
+$exitCode = $app->run();
+exit($exitCode);
--- a/composer.json
+++ b/composer.json
@@ -92,6 +92,9 @@
     "ext-imagick": "Adds support for more image processing formats and options.",
     "ext-iconv": "Adds support for more character encodings than PHP’s built-in mb_convert_encoding() function, which Craft will take advantage of when converting strings to UTF-8."
   },
+  "bin": [
+    "bin/craft"
+  ],
   "autoload": {
     "psr-4": {
       "craft\\": "src/",


khalwat:
```json
{
    "config": {
        "bin-dir": "./"
    }
}
```

...would put `craft` (and any other binaries) in the project root, too

markhuot:
Baby steps. `bin-dir` seems a bit heavy handed to me, personally.

Whatever the outcome—being able to run `./craft` without cloning down all of `craftcms/craft` would be a _very_ welcome addition.

khalwat:
Also, you can do `composer exec craft` and it'll run it... no need to specify the kind of annoying to type `./vendor/bin/craft`

(just putting this here as a hint to others)

markhuot:
Yup, `composer exec craft` is 😈 (devilishly awesome)

bin/craft
```diff
@@ -0,0 +1,32 @@
...
+if (!empty($_composer_autoload_path)) {
+    $basePath = dirname($_composer_autoload_path) . '/../';
+}
+else {
+    $basePath = getcwd();
+}
+
+// Define path constants
+define('CRAFT_BASE_PATH', $basePath);
+define('CRAFT_VENDOR_PATH', CRAFT_BASE_PATH . '/vendor');
```
brandonkelly, line 16: This assumes [`vendor-dir`](https://getcomposer.org/doc/06-config.md#vendor-dir) wasn’t edited at all.

bin/craft
```diff
@@ -0,0 +1,32 @@
...
+
+// Load Composer's autoloader
+require_once CRAFT_VENDOR_PATH . '/autoload.php';
+
+// Load dotenv?
+if (class_exists(Dotenv\Dotenv::class)) {
+    // By default, this will allow .env file values to override environment variables
+    // with matching names. Use `createUnsafeImmutable` to disable this.
+    Dotenv\Dotenv::createUnsafeMutable(CRAFT_BASE_PATH)->safeLoad();
```
brandonkelly, line 25: `createUnsafeMutable()` was added in PHP Dotenv 5.0, but several Craft installs have older versions. Maybe we should use some reflection here and call `createMutable()` if `createUnsafeMutable()` doesn’t exist, or `create()` if `createMutable()` doesn’t exist?

brandonkelly (commented):
`craft` is making a few assumptions and it would be good if it wasn’t quite so rigid.

bin/craft
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env php
+<?php
+/**
+ * Craft console bootstrap file
+ */
+
+if (!empty($_composer_autoload_path)) {
+    $basePath = dirname($_composer_autoload_path) . '/../';
```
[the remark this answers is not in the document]
markhuot: One above `vendor/` seems like a safe enough assumption. If you're directly requiring `craftcms/cms` and not using `craftcms/craft` your "project root" would still be one level above.

Locally, though, whenever I bring `./craft` in to a non-standard environment I usually use `getcwd()` as my base path. This seems very reliable for console/CLI applications. Won't make sense for web/FPM installs though but this is only run via console commands so that shouldn't matter.

bin/craft
markhuot: Yea, the conditionals to get this working are pure mayhem.

https://github.com/markhuot/craft-pest/blob/35b45fa8afa097667579a41b1cdb1663ffcceb28/src/test/TestCase.php#L60-L69

I don't know if it'd be better to read the `composer.json` so you know exactly what version is being used? I haven't dug in too deeply here but happy to over the next few days.

PR #11814, opened 2022-08-23
Fix bug in DateTimeHelper::toDateInterval

bencroker:
When given an integer, the `DateTimeHelper::toDateInterval` method returns a negative interval (in the past). This can be verified using the following code.

```php
// Outputs -3600 instead of the expected 3600.
Craft::dd(
    DateTimeHelper::intervalToSeconds(
        DateTimeHelper::toDateInterval(3600)
    )
);
```

One of the consequences of this is that the "Keep me signed in" checkbox on the login page has no effect by default. Since `GeneralConfig::rememberedUserSessionDuration` is set to `1209600`, it is evaluated as `-1209600`, which is 2 weeks in the past.

This PR fixes the bug by replacing `$then->diff($now)` with `$now->diff($then)`. I've also added a unit test to validate this.

### DIFF
--- a/src/helpers/DateTimeHelper.php
+++ b/src/helpers/DateTimeHelper.php
@@ -488,8 +488,8 @@ class DateTimeHelper
         if (is_numeric($value)) {
             // Use DateTime::diff() so the years/months/days/hours/minutes values are all populated correctly
             $now = static::now();
-            $then = (clone $now)->modify("-$value seconds");
-            return $then->diff($now);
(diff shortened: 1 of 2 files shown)


brandonkelly:
Thanks!!

brandonkelly:
Craft 4.2.2 is out now with that fix. Thanks again!

khalwat:
Love that you added a unit test along with the fix 💯

PR #11825, opened 2022-08-24
Use index hints to specify index so Mysql does not select the less pe…

sjeng:
…rformant index.

### Description
Mysql is using the wrong index in a subquery for `structures` table when there are a lot of records. The result is that the overall query gets super slow to complete. This is compounded by the fact that the subquery is used in multiple places in the control panel, and sometimes multiple times in a single request.

The recent entries widget is one place where the query is used and below is an example of load time:
6iiKw (1)

And here is the difference after applying index hint:
image

To be fair, the above example is an extreme one, but I'm seeing improvement in other problematic areas as well (edit pages of complex entries).

The proposed code change should not be very disruptive, and should be very specific to the `exist` subquery and also only when using MySQL db.

### DIFF
--- a/src/elements/db/ElementQuery.php
+++ b/src/elements/db/ElementQuery.php
@@ -2315,8 +2315,16 @@ class ElementQuery extends Query implements ElementQueryInterface
                     '[[structureelements.elementId]] = [[subquery.elementsId]]',
                     '[[structureelements.structureId]] = [[subquery.structureId]]',
                 ]);
-            $existsQuery = (new Query())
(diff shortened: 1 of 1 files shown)


angrybrad:
@sjeng thanks for that... any chance you could send a database dump and the full SQL query in the screenshot so we can do some local testing?

sjeng:
I can do that! I'll package up the db dump and send it over as soon as I'm able.

sjeng:
I've sent the code and query. Please let me know if you need anymore info.

PR #11834, opened 2022-08-29
Set default interlace type

kringkaste:
### Description

If you have old transformation strings in your imagetransformationindex table, which do not include the interlace type, the transformation fails with an exception. With the default value the transformation is processed as expected.


brianjhanson:
Thanks for the PR @kringkaste! This has been merged in for the next release.

brandonkelly:
Craft 4.2.4 is tagged with that fix.

PR #11844, opened 2022-08-30
Sticky content footer + move element actions into it

brandonkelly:
### Description

Element action triggers are now inserted in the content footer, beside the “Export” button, rather than in the toolbar (where they previously replaced all other toolbar items).

The footer is also sticky now, to the bottom of the window, so it is always in view.

The Entries index page, with two entries selected, showing that the “Set status” action trigger and the action menu with a gear icon are both visible in a sticky footer at the bottom of the window.

### Related issues

- DEV-376


gcamacho079:
@brandonkelly looks great! I completed a manual audit on this and no new issues are introduced. I made a small update to resolve a missing label (since the Garnish update to custom selects, the buttons aren't inheriting accessible names from their inner text).

PR #11895, opened 2022-09-06
Feature/disable preview button during save

### LINKED ISSUE #11858: [4.x]: Wrong field layout for preview if changing entry types and opening preview before provisional draft is done saving
### Description

After changing an entry's entry type, if the author opens preview _while the provisional draft_ is still saving, the preview pane will use the previously active entry type's field layout. Also in such cases the following behavior happens:

- Editing the fields inside preview will not trigger an iframe refresh
- Any and all edits inside preview are not saved, and lost when preview is closed

There are no JS errors in the console.

_Possibly the easiest fix is disabling the preview button whenever a provisional draft is currently being saved or applied?_

### Steps to reproduce

1. Create a section with two different entry types 
2. Edit an entry in that section
3. Change the entry's entry type
4. While the provisional draft is saving the entry type change (i.e. the spinner is visible); click the Preview button

### Expected behavior

The preview should use the field layout for the entry type the author just switched to, and editing fields inside preview should trigger auto-saves and iframe refresh as normal.

### Actual behavior

The preview uses the field layout for the previously active type, because preview was opened before the provisional draft had a chance to save the new type ID. Auto-save and -refresh doesn't work.

### Craft CMS version

4.2.3

### PHP version

_No response_

_No response_

### Database type and version

_No response_

### Image driver and version

_No response_

- None

brianjhanson:
### Description
Disables the preview button while a draft is saving in order to prevent the wrong field layout from showing when the preview opens. 

### Related issues
- Resolves #11858


gcamacho079 (changes requested):
@brianjhanson nice update! Just one change I'd like to suggest:

- For screen reader accessibility, the `aria-disabled` attribute is preferred over `disabled`. It requires the extra step of preventing clicks programmatically, but is otherwise a usability improvement since screen reader users are alerted that the button is there, just inactive at the moment.

I was torn about whether the disabled state should be applied to the button immediately upon saving instead of on Preview click — and having the `disablePreviewButton` called with the `beforeRun` event — but I do see the benefit of keeping it clickable while the new layout is loading.

brianjhanson:
Thanks @gcamacho079! I ended up removing the event listeners when the preview button is disabled but I'm not sure if that's going too far. Let me know if there's a better way to disable programatic clicks.

gcamacho079 (approved):
Looks good, @brianjhanson. An alternative to adding/removing the click listener would be to immediately exit the `openPreview` function if the button already has the `aria-disabled` attribute.

brianjhanson:
@brandonkelly just pushed up that change

PR #11902, opened 2022-09-08
Rename Date Range helper to be consistent with other helper classes

nfourtythree:


brandonkelly:
The vast majority of Craft’s helper classes don’t have a `Helper` suffix.

PR #11906, opened 2022-09-08
Source action menu

brandonkelly:
### Description

Replaces the “Customize” buttons in element index sidebars with an action menu, to be consistent with the Fields and Sites index pages.

An expanded action menu from within an element index page’s sidebar, showing a “Customize sources” option within it

Element indexes can register additional actions for the menu via a new `getSourceActions()` method.

```js
getSourceActions: function() {
  const actions = this.base();

  actions.push({
    label: 'Option label',
    onSelect: () => {
      // ...
    },
  });

  return actions;
},
```

Each action definition must have `label` and `onSelect` keys, and optionally, either `destructive: true` or `administrative: true`, if the action should be considered destructive or administrative.

The Assets index page has been updated to include “New subfolder”, “Rename folder”, and “Delete folder” actions, as well:

An expanded action menu from the Assets index page, showing “New subfolder”, “Rename folder”, “Delete folder”, and “Customize sources” actions.

### Related

- #11310


gcamacho079:
Just a few things:

1. The button name should be more specific, something along the lines of "Source settings" to differentiate it from "Settings" buttons on other screens
2. Right now you can’t close menu by clicking the button again; you have to click outside the menu or hit `ESC`
3. When the parent asset folder is collapsed, the user doesn’t get any feedback that a child folder has been created. Some potential solutions are to:
    - Show a notification for ‘Subfolder added successfully’, or
    - Activate the new subfolder source and move focus to its button in the source sidebar; assuming that the average user flow would be to create a folder and immediately add assets to it, this would create a logical keyboard focus order and also give screen reader users feedback that the action was successful.

Regarding the 3rd item, I think a notification would be preferred since it's less opinionated about what the user's next action might be.

gcamacho079 (changes requested):
See comment

brandonkelly:
Thanks @gcamacho079! Addressed those and merged.

> Right now you can’t close menu by clicking the button again; you have to click outside the menu or hit ESC

That ended up being a DisclosureMenu bug – fixed on `develop` via c8ba0b3f3cb8b3d284c803729e66ae839420670d.

> 3\. Activate the new subfolder source and move focus to its button in the source sidebar; assuming that the average user flow would be to create a folder and immediately add assets to it, this would create a logical keyboard focus order and also give screen reader users feedback that the action was successful.

Not opposed to this but the current behavior _could_ be desired if the plan is to relocate assets from another source into the new source. So went with option 1 for now; we can revisit option 2 as a separate PR down the road if we decide it’s definitely the better behavior.

Issue #6916, opened 2020-09-28
Status accessibility, UX


missmatsuko:
### Description

- Definitely need to expose status to screen readers. Currently the icons do not have accessible text applied (e.g. aria-label).
  
  I think it would be beneficial to display the full text, though, especially since there isn't really a legend for the icons (aside from opening the status filter) and the same icons are used for slightly different meanings (e.g. **expired** entries and **suspended** users statuses are both red squares).

  This can be similar to how the status is displayed when the entry/user is being edited:
![Orange triangle with 'Pending' text to its right.](https://user-images.githubusercontent.com/17884198/94489083-b5b31000-0198-11eb-857c-2c2dcb6c8fd8.png)

- Status should have its own column in the listing tables.

- I wonder if using the same style icons for statuses and boolean fields might be confusing since they have different meanings... Maybe the boolean option can be represented with a checkmark instead?
![Screen Shot 2020-09-28 at 2 57 56 PM](https://user-images.githubusercontent.com/17884198/94490375-0297e600-019b-11eb-89c7-77fbe92685b5.png)

  I was also thinking about a different type of icon for status. For example, circles with different amounts filled? This seems like it would represent the statuses more intuitively, but might be harder to distinguish than the different shapes.

- I think it'd be good to remove the icon next to 'All' in the status dropdown, since it looks like it might filter to show items with the same icon as the disabled entries:
![Screenshot of status dropdown. There is the same empty circle icon next to 'All' and 'Disabled'.](https://user-images.githubusercontent.com/17884198/94491125-78508180-019c-11eb-9068-5f94f4664139.png)

### Additional info

- Craft version: 3.5.11.1
- PHP version:
- Database driver & version:
- Plugins & versions:

gcamacho079:
Most of the issues in this list have been addressed:

- Icons have alternative text when they are not adjacent to equivalent text
- In #14968, status label text was added to element index tables and element cards, but there are still some areas where adjacent text isn’t shown ([see PR reply for exclusions](https://github.com/craftcms/cms/pull/14968#issuecomment-2125949311)).
- Boolean fields have been updated to use a checkmark icon (#12213)
- The “All” icon has been updated to use a different color and shape than the disabled icon

PR #11911, opened 2022-09-09
Status icon updates

gcamacho079:
### Description
[Issue 6916](https://github.com/craftcms/cms/issues/6916) mentioned the confusing UX surrounding the "All" and "Disabled" status icons being identical; in addition, icons in Lightswitch custom field columns for "Enabled"/"Not Enabled" were identical to "Live"/"Disabled"

This PR adds new styles to the "All" icon, and updates the presentation of values for Lightswitch fields to use checkboxes instead.

Status filter dropdown showing unique icons for all statuses; the All icon is now a circle with a dot in the middle.

The 'Show author email' field uses either a checkbox or empty cell to show the enabled and disabled states


brandonkelly:
Not sure how I feel about that “All” icon – it looks like a radio button now, or maybe the image focal point indicator, neither of which really convey “all inclusive” to me.

Would this gradient BG work?

An expanded Status menu

When “Use shapes to represent statuses” is enabled, we could use a diamond:

An expanded Status menu, shown with the “Use shapes to represent statuses” user preference enabled.

gcamacho079:
> Would this gradient BG work?

@brandonkelly Way better 👍🏼 That would certainly do the trick

PR #11912, opened 2022-09-10
Allow @web and @webroot to be set via env

timkelty:
### Description
As it is best practice to manually set the `@web` and `@webroot` aliases, and they are most often already defined via ad-hoc env (like `PRIMARY_SITE_URL` or `DEFAULT_SITE_URL`), this lets you set it with the `CRAFT_WEB_URL` and `CRAFT_WEB_ROOT` env vars directly.

### Follow-up checklist:

- [ ] Changelog
- [ ] Docs
    - [x] ~~Update anything that references `DEFAULT_SITE_URL`, `PRIMARY_SITE_URL` or anything similar.~~
    - [ ] https://craftcms.com/docs/4.x/config/#environmental-configuration (screenshot)
    - [ ] https://craftcms.com/docs/4.x/config/#aliases
- [x] ~~Update env in https://github.com/craftcms/craft~~
- [x] ~~[Update DDEV](https://github.com/khalwat/ddev/blob/feature/craftcms-project-type/pkg/ddevapp/craftcms.go#L125-L143) so `CRAFT_WEB_URL` is automatically set~~


bencroker:
> [Update DDEV](https://github.com/khalwat/ddev/blob/feature/craftcms-project-type/pkg/ddevapp/craftcms.go#L125-L143) so CRAFT_WEB_URL is automatically set

Would it make sense to try to squeeze this into DDEV before the new `craftcms` project type is release, and if so then how?

timkelty:
> > [Update DDEV](https://github.com/khalwat/ddev/blob/feature/craftcms-project-type/pkg/ddevapp/craftcms.go#L125-L143) so CRAFT_WEB_URL is automatically set
> 
> Would it make sense to try to squeeze this into DDEV before the new `craftcms` project type is released, and if so then would it best be implemented by simply replacing `PRIMARY_SITE_URL` with `CRAFT_WEB_URL`?



> > [Update DDEV](https://github.com/khalwat/ddev/blob/feature/craftcms-project-type/pkg/ddevapp/craftcms.go#L125-L143) so CRAFT_WEB_URL is automatically set 

Probably a good idea!

> Would it make sense to try to squeeze this into DDEV before the new `craftcms` project type is released, and if so then would it best be implemented by simply replacing `PRIMARY_SITE_URL` with `CRAFT_WEB_URL`?

🤔 I wonder if it would be better to support both, for backward compatibility. Even though Craft itself doesn't recognize `PRIMARY_SITE_URL`, the installer does write it out to `.env` and save it as your site url. They're also inherently different (esp. considering multi-site), in that one is intended for the `@web` alias, while one is intended for the site url.

Seems like we should be able to detect if you're already using `PRIMARY_SITE_URL` and update that, otherwise use `CRAFT_WEB_URL`?

khalwat:
>  I wonder if it would be better to support both, for backward compatibility. Even though Craft itself doesn't recognize PRIMARY_SITE_URL, the installer does write it out to .env and save it as your site url. They're also inherently different (esp. considering multi-site), in that one is intended for the @web alias, while one is intended for the site URL.
> Seems like we should be able to detect if you're already using PRIMARY_SITE_URL and update that, otherwise use CRAFT_WEB_URL?

PRs are welcome to change/fix it however you decide to in the DDEV repo.

khalwat:
Odd this change wasn't made when the recent refactoring of the base `craftcms/craft` setup was done.

Do you plan to allow _all_ aliases to be set via magic `CRAFT_` env vars, or just certain select ones?

timkelty:
@khalwat just these for now, as they are extremely common and recommended addition to nearly all installs.

timkelty:
> Would it make sense to try to squeeze this into DDEV before the new `craftcms` project type is released, and if so then would it best be implemented by simply replacing `PRIMARY_SITE_URL` with `CRAFT_WEB_URL`?

After some discussion, automatically setting this via DDEV may be presumptive/undesirable, especially when considering DDEV aliases and Craft multi-site. It makes sense to continue using the site-specific `PRIMARY_SITE_URL` (that the installer still sets) for automatic configuration.

tests/_data/project.yaml
```diff
@@ -84,7 +84,7 @@ siteGroups:
     name: 'Craft CMS Test Site Group'
 sites:
   2fa19d28-c359-4955-8b0d-35a35e3f1128:
-    baseUrl: $DEFAULT_SITE_URL
```
brandonkelly, line 87: @timkelty is there a reason we’re still making this change for tests?

timkelty: eh, just because it is the sole remaining reference to `$DEFAULT_SITE_URL` in the codebase. Seems like it should be either `@web`, `$PRIMARY_SITE_URL`, or `CRAFT_WEB_URL`.

PR #11916, opened 2022-09-12
Make getCurrentUser static

timkelty:
Method should be static


timkelty:
@brandonkelly if we're renaming this from `getCurrentUser` to `currentUser`, we need to remove [the docblock](https://github.com/craftcms/cms/blob/f11df40d5406eb0b846422ff7dc26e3c04af6f2c/src/web/Controller.php#L30):

```
@property-read ?User $currentUser
```

Issue #11904, opened 2022-09-08
[4.x]: \craft\services\Drafts#createDraft should not require integer value for $creatorId argument


ivanBereznev:
### Description
When modifying a draft via GraphQL API, there can be no current user assigned, i.e. `Craft::$app->getUser()->getId()` returns `null`. In this case there is no way to assign any valid `creatorId` so requiring an integer for `$creatorId` argument does not look completely correct.
In particular this causes failures when modifying [Neo block](https://github.com/spicywebau/craft-neo) fields with GraphQL.
Probably the issue is on the plugin side, but since there seems to be no way for it to get ID of the current user it seems to me that API from Craft side is inconsistent.

### Steps to reproduce

1. Create an entry with a Neo field
2. Create a draft for an entry
3. Try to modify the value of Neo field for the draft using GraphQL API.

### Expected behavior
Changes are applied to the draft.

### Actual behavior
An exception is thrown:
```
[TypeError] craft\services\Drafts::createDraft(): Argument #2 ($creatorId) must be of type int, null given, called in /app/vendor/spicyweb/craft-neo/src/Field.php on line 1197
```
The exception originates from [this line](https://github.com/spicywebau/craft-neo/blob/main/src/Field.php#L1188) where plugin tries to create draft for a block providing ID of the current user which is `null`.

### Craft CMS version

4.2.3

### PHP version

8.0

### Operating system and version

Debian GNU/Linux 11

### Database type and version

PostgreSQL 13.6

### Installed plugins and versions

- spicyweb/craft-neo : 3.3.7

ttempleton:
It's worth noting that Neo fields do this the same way as [Matrix](https://github.com/craftcms/cms/blob/2ee1ed6d8f413744b4c4e1455f5313a1103ef29c/src/fields/Matrix.php#L1311) and [Super Table](https://github.com/verbb/super-table/blob/4c9397e5d22cec0e5fdca919533a838d54dc7bd2/src/fields/SuperTableField.php#L1340) fields, so I assume this would affect all three field types.

brandonkelly:
This should be resolved for Craft 4.3 via #11923.

If you’d like to test, change your `craftcms/cms` constraint in composer.json to `4.3.x-dev#7c232ea3613c779fdaf986a9b90a64f426b1faad as 4.3.0`

ivanBereznev:
[...]

PR #11923, opened 2022-09-12
Don't require `creatorId`

brianjhanson:
### Description
Stops requiring `creatorId` when creating a draft. 

@brandonkelly this might be considered a breaking change because the signature of `\craft\services\Drafts::createDraft()` and `craft\events\DraftEvent` would change. I'm also not sure if this would have any unintended consequences that I'm not aware of, as far as I can tell the system is okay with `creatorId` being `null` but let me know if there are specific areas I should test around this. 

### Related issues
#11904


brandonkelly:
Seems pretty safe since it looks like everything else surrounding `creatorId` is expecting it could be `null`. But since it is a signature change, I retargeted the PR for 4.3.

PR #11932, opened 2022-09-14
Move svg function to HTML helper

leevigraham:
This allows php functions to call HTML::svg() directly.


brandonkelly:
Just merged for 4.3 via #12040 🎉

PR #11942, opened 2022-09-14
Refactor Listbox to use exclusive button group pattern

gcamacho079:
### Description
Follows [internal resources for implementing an exclusive button group](https://www.notion.so/pixelandtonic/exclusive-button-group-ed2321f5ffe34bbea5e42e2a93ac3624), and re-implements the button pattern for Preview device type toggles


gcamacho079:
@brandonkelly proposing the `Craft.Listbox` class gets renamed to something like `Craft.ExclusiveBtnGroup`, but wanted to get your eyes on the changes first 👀 🙏🏼

PR #11944, opened 2022-09-14
Darken hover and focus color for Forgot/Remember your password buttons on login

gcamacho079:
### Description
Slight color change for these states


brandonkelly:
Seems weird to be making this change just for the Login page. Why not all links?

PR #11950, opened 2022-09-15
Added “Between” option to number condition rules

nfourtythree:
### Description
Currently, condition rules based on `BaseNumberConditionRule` only allow one value making it difficult to create a range.

This PR adds the "Between…" option allowing you to specify a min and a max.

----

This would work well with #11947 and also help with some of the condition rules that will be added in Commerce 4.2


src/base/conditions/BaseTextConditionRule.php
```diff
@@ -22,9 +22,9 @@ abstract class BaseTextConditionRule extends BaseConditionRule
     public string $operator = self::OPERATOR_EQ;
 
     /**
-     * @var string The input value.
+     * @var string|array The input value.
      */
-    public string $value = '';
+    public string|array $value = '';
```
brandonkelly, line 27: This feels like a breaking change. Think maybe Number rules should gain a `maxValue` property, which can be null when `between` isn’t used.

brandonkelly, line 27: This feels like a breaking change. Think maybe Number rules should gain a `maxValue` property, which can be null when `between` isn’t used.

brandonkelly (changes requested):

src/base/conditions/BaseTextConditionRule.php
nfourtythree: Sounds sensible. Have made that changed and pushed it up to the PR.

The only thing I kept along those lines was the change to `paramValue()` on the `BaseTextConditionRule` class (https://github.com/craftcms/cms/pull/11950/files#diff-8a311ba8c5628ce8272fb459115be6ecc8941cbed8ef90f79c0157fa0fcb70c5R96). So that we can easily use the `'and'` array syntax. 

Would you consider this a breaking change as well? Would you prefer that it be put back and `BaseNumberConditionRule` simply return a string expression of the `paramValue`?

brandonkelly: Yeah same thing. Returning a string would work.

PR #11955, opened 2022-09-15
Fix clipped content in Craft Support widgets when text is zoomed to 200%

gcamacho079:


brandonkelly:
@gcamacho079 Sorry, merged and then reverted it, because I noticed the button styling change was affecting a couple smaller buttons as well. Even if we fix these, plugins could be affected as well. So if we have to change `height` to `min-height`, we’ll need to hold off until Craft 5 and do a full scan of the control panel to ensure everything looks good.


Screenshot of a portion of an Edit Entry page, highlighting two buttons that weren’t styled properly

gcamacho079:
@brandonkelly I'll mark this as a 5.0 change in Linear for now 👍🏼

PR #11961, opened 2022-09-16
Fix post date/time and expiry date/time accessible names

gcamacho079:
### Description
Fixes an issue in the meta fields for post and expiry date and time. Previously, `aria-labelledby` attributes pointing toward "Post Date" and "Expiry Date" were overriding the Date and Time labels set on each individual field via `aria-label`.

This PR enables a fieldset wrapper around each group of fields, and sets `aria-labelledby` to null so that the Date and Time labels are read in addition to the legend text.

### DIFF
--- a/src/templates/_includes/forms/date.twig
+++ b/src/templates/_includes/forms/date.twig
@@ -5,6 +5,7 @@
 {% set outputTzParam = outputTzParam ?? true %}
 {% set isMobile = craft.app.request.isMobileBrowser %}
 {% set isDateTime = isDateTime ?? false %}
+{% set labelledBy = fieldset ?? null %}
 
 {% set containerAttributes = {
     class: ['datewrapper']|merge((class ?? [])|explodeClass),
--- a/src/elements/Entry.php
+++ b/src/elements/Entry.php
@@ -1699,6 +1699,7 @@ EOD;
                 'value' => $this->_userPostDate(),
                 'errors' => $this->getErrors('postDate'),
                 'disabled' => $static,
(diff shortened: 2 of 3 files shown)


src/templates/_includes/forms/date.twig
```diff
@@ -5,6 +5,7 @@
 {% set outputTzParam = outputTzParam ?? true %}
 {% set isMobile = craft.app.request.isMobileBrowser %}
 {% set isDateTime = isDateTime ?? false %}
+{% set labelledBy = fieldset ?? null %}
```
brandonkelly, line 8: Not following this – `fields` is going to be `true` or `false`, and `labelledBy` is supposed to be set to a DOM element ID. (Ditto for `time.twig`.)

brandonkelly (changes requested):

src/templates/_includes/forms/date.twig
gcamacho079: @brandonkelly I think I meant `{% set labelledBy = fieldset ? null : labelledBy %}`; on second thought, that’s not exactly what I wanted to accomplish.

The issue I’m having is that just for the metafields, I want the `aria-labelledby` value to go away since it overwrites the explicit Date and Time labels being set by `aria-label`.

Screenshot 2023-01-19 at 12 02 35 PM

I was looking at other instances of Datetime, Date, and Time fields in the field layout, and `aria-labelledby` isn’t being set anywhere else, so I wasn’t sure the best place to make this change to have it be specific to just the Post Date and Post Time.

PR #11975, opened 2022-09-20
Bugfix/gh 11974 resave element queue

### LINKED ISSUE #11974: [3.x]: Resaving an asset with queue fails to set field data but works when run directly.
### Description

I want to resave a bunch of assets to populate a new field called `alt` added to several image asset volumes.

In order to do this I performed an asset element query in a content migration to obtain the required element IDs and then pass these to the `./craft resave/assets` command with some additional paramaters.

When testing locally, I've tested these commands, in this case I've just used a single asset ID.

**This doesn't seem to work (it runs successfully but the title is not copied to the alt field**

```
/app $ ./craft resave/assets --element-id=602522 --queue=true --set=alt --to=title --update-search-index=true
Assets queued to be resaved.
```

**This does work when the command is not queued**

```
/app $ ./craft resave/assets --element-id=602522 --set=alt --to=title --update-search-index=true
Resaving 1 asset ...
    - [1/1] Resaving Graham basten (602522) ... done
Done resaving asset.
```

When having queue set to true the field data on `alt` is not set, when running directly without the queue set to true the field data is set accordingly.

The queue job successfully completes and returns no error, but the data is not set.

### Steps to reproduce

1. Run craft resave/assets using the properties `set` `to` and `queue`
2. The field data specified in to is not written to the field defined on the set value but it is if you omit `queue`

### Expected behavior

The data to be saved when run as a queue job

### Actual behavior

The data does not get written to the defined field when queued.

### Craft CMS version

3.7.55

### PHP version

8.0

Ubuntu 20.04 LTS (WSL2)

### Database type and version

MySQL 8

brianjhanson:
### Description
Fixes an issue where resaving values didn't work when using the queue

### Related issues
Fixes #11974


jamesmacwhite:
@brianjhanson Thanks for looking into this, can the touch parameter be supported through the queue job as well? This was recently added in Craft CMS 3.7.54.

brianjhanson:
@jamesmacwhite good catch! Just updated

jamesmacwhite:
@brianjhanson Awesome. Thank you!

jamesmacwhite:
@brianjhanson @brandonkelly  Any ideas when this can be reviewed for merge? We have 1500 assets that need to be resaved with data being copied from one field to another, I don't want to run this without queuing on our live environment due to the performance impact.

PR #12017, opened 2022-09-28
feat: Craft 3: Add craft\console\controllers\HelpController with --asJson flag

khalwat:
### Description

PR for Craft 3

This adds a `craft\console\controllers\HelpController` that adds a `--asJson` flag, to allow for external API integration, such as with [Fig](https://fig.io/) for CLI autocomplete.

Usage:

All commands:
```
php craft help --asJson
```

Individual commands: 
```
php craft help <command/name> --asJson
```

Alternately, the `-j` flag can be used as an option alias

If the `--asJson` flag isn't set, it calls through to the parent `yii\console\controllers\HelpController` to do its normal thing.

### Related issues

n/a

### DIFF
--- a/src/console/Application.php
+++ b/src/console/Application.php
@@ -9,6 +9,7 @@ namespace craft\console;
 
 use Craft;
 use craft\base\ApplicationTrait;
+use craft\console\controllers\HelpController;
 use craft\db\Query;
 use craft\db\Table;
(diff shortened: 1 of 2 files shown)


brandonkelly:
Craft 3.7.56 and 4.2.6 are out with this! (via #12074)

PR #12035, opened 2022-09-30
Prevent lazy loading relationships

markhuot:
For your consideration, this adds a new general config named `->preventLazyLoading(true)` which is copied from Laravel's similarly named feature: https://twitter.com/taylorotwell/status/1395087054254526474.

I've personally seen a number of projects ship that could be _so_ much faster if the developers had been better about eager loading their relationships. For small sites it doesn't matter much because the DB is faster than they really need. But for larger sites with more complex relationships not eager loading is a dealbreaker for me. This is an opt-in feature that developers could choose to enable per-environment depending on their needs/site.

The implementation happens in `craft\base\Element` since most everything is an element. I don't know that you can eager load non-element items so this seemed like a safe place to do it. But, if there's a higher class that this should be in, let me know.

The actual logic is relatively simple. Since `__get()` already early-exits on eager loaded values I put the check below that and `throw` if the field _should have been_ eager loaded. The logic for determining if a field should have been lazy loaded is checking if it's a pre-defined field (like entry authors) or an extension of the `BaseRelationField` (plus `Matrix`). There may be more eager loadable fields, but I couldn't find a good comprehensive list so I started here. For example, I'm actually not sure if you can eager load asset volumes or element ancestors… so fields like that may need to be added if they're eligible for eager loading.

Happy to iterate on this more if you find it useful. I assume it'd need a more comprehensive list of eagerloadable fields as well as some unit tests.

### DIFF
--- a/src/config/GeneralConfig.php
+++ b/src/config/GeneralConfig.php
@@ -5199,7 +5199,7 @@ class GeneralConfig extends BaseConfig
      * @see $preventLazyLoading
      * @since 4.3.0
      */
-    public function preventLazyLoading(bool $value = false): self
+    public function preventLazyLoading(bool $value = true): self
     {
         $this->preventLazyLoading = $value;
         return $this;
--- a/src/config/GeneralConfig.php
+++ b/src/config/GeneralConfig.php
@@ -1998,6 +1998,26 @@ class GeneralConfig extends BaseConfig
     public bool $preserveImageColorProfiles = true;
 
     /**
+     * @var bool When set to `true`, Craft will not allow relationship fields to be lazy loaded. If a relationship field is accessed lazily while
+     * set to `true` an exception will be thrown. When set to `false` relationships may be loaded lazily.
+     *
+     * Commonly you will set this based on your environment to disable lazy loading during development but allow it in production so that production
+     * doesn't error out if one slips through.
+     *
+     * ::: code
+     * ```php Static Config
+     * ->preventLazyLoading(true)
+     * ```
+     * ```shell Environment Override
+     * CRAFT_PREVENT_LAZY_LOADING=1
+     * ```
+     * :::
+     *
+     * @group System
+     */
+    public bool $preventLazyLoading = false;
+
+    /**
      * @var bool When `true`, Craft will always return a successful response in the “forgot password” flow, making it difficult to enumerate users.
      *
      * When set to `false` and you go through the “forgot password” flow from the control panel login page, you’ll get distinct messages indicating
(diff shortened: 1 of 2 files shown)


michaelrog:
This is a neat idea. I could see myself using it at least during development (in dev mode?), to keep myself accountable.

Spitballing here... what if...
 - the `preventLazyLoading` config item could be a boolean, _or_ a closure that consumes a Field instance and returns a boolean, so the enforcement could be dynamic/per-field
 - the query object also had a `preventLazyLoading` property, so I can override the config in cases where I have explicitly decided eager-loading isn't to my advantage

markhuot:
>the preventLazyLoading config item could be a boolean, or a closure that consumes a Field instance and returns a boolean, so the enforcement could be dynamic/per-field

I'm torn on this, as it seems like a bit of extra complexity when you could probably do something pretty similar via your next suggestion, which I think fits existing APIs better.

>the query object also had a preventLazyLoading property, so I can override the config in cases where I have explicitly decided eager-loading isn't to my advantage

Yup, love this. If you require eager loading in your config via `->preventLazyLoading(true)` then you could do `craft.entries.preventLazyLoading(false).all()` to specifically allow lazy loading on that one query. The rough part of this is that currently models don't have a reference back to the initiating query. So at the point of the exception we don't know if the default was overridden. This would take some more code to accomplish (and some additional memory on each model to track). Personally, I think it's worth it, but I'd love to hear from the P&T team when they're back and recovered from Dot All if they agree.

src/config/GeneralConfig.php
```diff
@@ -5162,6 +5182,29 @@ public function preserveImageColorProfiles(bool $value = true): self
...
+     * ->preventLazyLoading(true)
+     * ```
+     *
+     * @group System
+     * @param bool $value
+     * @return self
+     * @see $preventLazyLoading
+     * @since 4.3.0
+     */
+    public function preventLazyLoading(bool $value = false): self
```
AugustMiller, line 5202: I'll wait on the team’s feedback regarding the broader concept, but wanted to flag this default value—to match other fluent methods, it ought to be `true` so that `->preventLazyLoading()` (without arguments) _enables_ the feature.

Fluent method argument defaults don't have to agree with initial property values!

markhuot: Good catch. This is resolved in https://github.com/craftcms/cms/pull/12035/commits/fff2caaa605e3e7a858e975d6da9e86b5aee645f.

brandonkelly:
I like the concept, but not the implementation. `shouldHaveEagerLoaded()` feels like it’s doing too much, and should probably be renamed to `isEagerLoadable()`, and only focus on whether a given attribute _can_ be eager-loaded. The`preventLazyLoading` and `templateMode` checks could just happen within `__get()`:

```php
if (
    App::devMode() && // <-- probably also worth having this
    Craft::$app->getConfig()->getGeneral()->preventLazyLoading &&
    Craft::$app->getView()->templateMode !== \craft\web\View::TEMPLATE_MODE_SITE &&
    $this->isEagerLoadable($name)
) {
    throw new Exception('...');
}
```

That said, I also don’t really care for everything that supports eager-loading having one extra step, besides hooking into `Element::eagerLoadingMap()` to provide the actual source/target ID mapping.

Instead I think we should tackle this in Craft 5, where we can make `eagerLoadingMap()` expected to return _closure_ that returns the mapping array, rather than the mapping array directly. That way we could start calling `eagerLoadingMap()` instead of adding a new `isEagerLoadable()` method. If it returns a closure, you know the attribute is eager-loadable, but the work of establishing the mapping array won’t happen unless the closure is actually called.

brandonkelly:
> > the query object also had a `preventLazyLoading` property, so I can override the config in cases where I have explicitly decided eager-loading isn't to my advantage
> Yup, love this. If you require eager loading in your config via `->preventLazyLoading(true)` then you could do `craft.entries.preventLazyLoading(false).all()` to specifically allow lazy loading on that one query. The rough part of this is that currently models don't have a reference back to the initiating query. So at the point of the exception we don't know if the default was overridden. This would take some more code to accomplish (and some additional memory on each model to track). Personally, I think it's worth it, but I'd love to hear from the P&T team when they're back and recovered from Dot All if they agree.

Not a bad idea. Rather than keeping a reference to the original query, elements could get a new `?bool $preventLazyLoading` property, which defaults to the general config setting, and element queries could start setting those to `true`/`false` when populating the resulting elements, if their `preventLazyLoading` params were set.

PR #12036, opened 2022-09-30
Improve translation of 'Level'

rostockahoi:
When creating a field of type entry in the option 'Selectable Entries Condition' dropdown the word 'Level' should rather be translated to 'Ebene'. Currently it's called 'Höhe' which rather refers to 'Height'. 'Höhe' is actually confusing at that point.


brandonkelly:
Thanks! Made the same change [in Crowdin](https://translate.craftcms.com/) and pulled the updated translations in for the next release.

Issue #12001, opened 2022-09-23
[3.7.55.2]: `focus-visible` class is added to table cells when using the table plugin of the Redactor field type


kruncher:
### Description
See https://github.com/craftcms/redactor/issues/424 for further information

While this manifests as a bug within the redactor field it originates from the following file of the CMS: craftcms/cms/src/web/assets/focusvisible/webpack.config.js

### Steps to reproduce

1. Insert a table into a Redactor field.
2. Click a cell.
3. Save the entry.
4. Click another cell.
5. Save the entry.
6. Observe the following in the control panel:
![image](https://user-images.githubusercontent.com/554215/191932805-092e9e7d-12bb-41b0-82bc-303998fa351e.png)
7. Observe field output also includes the `focus-visible` class.

### Expected behavior

The `focus-visible` class should not be persisted onto Redactor field values and perhaps shouldn't be visible in tables at all within the control panel.

### Actual behavior

The `focus-visible` class is persisted onto Redactor field values.

### Craft CMS version

3.7.55.2

### PHP version

7.2.34

### Operating system and version

Linux 5.10.25-linuxkit

### Database type and version

MySQL 8.0.25

### Image driver and version

Imagick 3.4.4 (ImageMagick 7.0.10-48)

### Installed plugins and versions

- Google Cloud Storage v1.4.1
- Redactor v2.10.10
- Webhooks v2.4.1

brandonkelly:
Thanks for reporting that! We’ve fixed this for the next Craft 3 and 4 releases.

brandonkelly:
Craft 3.7.56 and 4.2.6 are out with the fix for this.

kruncher:
Fantastic! thank you :)

PR #12037, opened 2022-09-30
Fix focus-visible styles being duplicated across all table cells in Redactor

gcamacho079:
### Description
The Redactor scripts did not seem to mesh with the `focus-visible` scripts; Redactor was persisting classes that should have been removed, and was instead duplicating them across table cells.

Since `:focus-visible` is now supported on all major browsers, I think the best solution is to remove the polyfill and move forward with the pseudo-class implementation.

### Related issues
#12001


brandonkelly:
Backported this to Craft 3 as well (1afc6777787ee0cdcce2d27b82f1f1bf08fc8313).

PR #12041, opened 2022-10-02
Fix serialization error when ElementQuery instances are passed into relations

### LINKED ISSUE #11981: [4.2.5]: relatedTo with multiply Categories Error: Serialization of 'Closure' is not allowed
### Description
After updating to Craft 4 this query cause an error at query

```twig
{% paginate craft.entries
    .section(['blog'])
    .relatedTo(block.category)
    .with(['ovImage'])
    .limit(block.limit)
  as pageInfo, pageEntries %}
```

`block.category` is a Categories field and can include multiple categories. If just one category is selected no error appears but also no data, with more categories selected: `Serialization of 'Closure' is not allowed`

### Craft CMS version

4.2.5

### PHP version

_No response_

_No response_

### Database type and version

_No response_

### Image driver and version

_No response_

-

brianjhanson:
@brandonkelly up to you on if we want to support this or not. When an `ElementQuery` is passed into a `relatedTo` parameter, the check for the glue would result in a serialization error. As far as I can tell passing an `ElementQuery` into `relatedTo` was never explicitly supported, but happened to work in Craft 3. 

### Related issues
Fixes #11981

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixed
 - Fixed an error that could occur when editing a draft of an element type that didn’t have change tracking enabled.
+- Fixed an error that caused `ElementQuery` instances to throw a serialization exception. ([#11981](https://github.com/craftcms/cms/issues/11981))
 
 ## 4.2.5.1 - 2022-09-21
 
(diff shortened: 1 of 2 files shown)


brandonkelly:
Closing as this isn’t the right fix, as discussed offline.

PR #12046, opened 2022-10-03
Update image transforms

### LINKED ISSUE #12023: [3.7.55.2]: setWidth() and setHeight() returning transform setting values, not transformed image values
### Description
We have a production site running Craft 3.7.42 and a template containing the following code:

`<a href="{{ image.getUrl('noCrop2400x2400') }}" class="lightbox-trigger" itemprop="contentUrl" data-size="{{ image.getWidth('noCrop2400x2400') }}x{{ image.getHeight('noCrop2400x2400') }}">`

The noCrop2400x2400 transform is set to 'Fit' and I've set the width and height to 2400 each, so that the longest edge doesn't exceed that value.

In Craft 3.7.42, image.getWidth() and image.getHeight() are returning the correct dimensions of the _transformed_ image. However, in the my dev env, I've updated to Craft 3.7.55.2 and now the getWidth and getHeight are returning the values I have set within the image transform (in this case 2400 x 2400). The actual transform itself is working fine. 

I should note that it's not doing this consistently. For some images, it is returning the correct values. I can't see a pattern. All I do know is that it's working 100% consistently in Craft 3.7.42 but not in Craft 3.7.55.2.

Any idea what might have changed? or what I might need to do to get it working as expected?

### Expected behavior

image.getWidth('transform') and image.getHeight('transform') should return the width and height of the _transformed_ image.

### Actual behavior

image.getWidth('transform') and image.getHeight('transform') are returning the width and height of the values within the transform settings in Craft.

### Craft CMS version

3.7.55.2

### PHP version

8.0.23

Darwin 21.6.0

### Database type and version

MySQL 5.7.39

### Image driver and version

Imagick 3.6.0 (ImageMagick 7.1.0-48)

- Amazon S3: 1.3.0
- Element API: 2.8.6.1
- Formie: 1.6.14
- Mailgun: 2.0.1
- Redactor: 2.10.10
- SEO: 3.4.38
- String Urldecode: 1.0.0
- Super Table: 2.7.3
- Typed link field: 1.0.25

brianjhanson:
### Description
Fix an issue where image transforms were off in situations where `upscaleImages` was set to `false` and the transform size was smaller than then source size.

### Related issues
Fixes #12023

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 - Fixed an error that could occur when passing an element query to a `relatedTo` param, if the parent element query contained any closures. ([#11981](https://github.com/craftcms/cms/issues/11981))
 - Fixed a bug where unsaved drafts could be unintentionally deleted when saved, if a plugin or module was blocking the save via `EVENT_BEFORE_SAVE`. ([#12015](https://github.com/craftcms/cms/issues/12015))
 - Fixed a bug where “Propagating tags” jobs would fail if two tags had similar titles.
+- Fixed a bug where image transforms weren’t getting sized correctly in some cases when `upscaleImages` was disabled. ([#12023](https://github.com/craftcms/cms/issues/12023))
 
 ### Security
 - Reduced the amount of system information that’s available to guest users.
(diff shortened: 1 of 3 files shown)


brandonkelly:
Thanks! This has been merged in for Craft 4.3 via #12076.

PR #12049, opened 2022-10-04
Rename folder with duplicate name shows wrong folder in alert message

jeroenlammerts:
A folder with the name “{folderName}” already exists in the folder.
This displays the name of the current folder not the duplicate one.


brandonkelly:
Thanks for the PR! Looks like the same bug also existed on Craft 3, so I backported your fix there (da5de9eaea8fde6195c23ac9f3b803272616abf4) and have merged it into Craft 4 as well.

brandonkelly:
Craft 3.7.56 and 4.2.6 are out with that fix. Thanks again!

PR #12072, opened 2022-10-06
Updates relatedTo to respect site of query

### LINKED ISSUE #12000: [3.x]: Using a multi-site and GraphQL with relatedToCategories - unexpected entries returned based on field selected in a different site
### Description
Using a multi-site and GraphQL with `relatedToCategories` in the query, entries are returned when they don't have the passed category associated with them and it seems to be because the same entry on another site has that category selected.

### Steps to reproduce

1. Add multiple sites
2. Create a section with a Propagation Method of "Let each entry choose which sites it should be saved to"
3. Create a category group + field for the category (manage relations on a per-site basis) and add it to the entry field layout
4. Create an entry and save it to both sites
5. Create a category and set the category field on the entry on one of the sites
6. Run GraphQL query as below against the site that doesn't have the category selected

#### Query

```
query(
  $categories: CategoryCriteriaInput) {
  projectsList: entries(
    siteId: 2
    section: "Projects"
    relatedToCategories: [$category]
    enabledForSite: true
  ) {
    id
    title
  }
}
```

#### Variables

```
{
  "category": {"slug": "category-slug", "siteId": 2, "enabledForSite": true},
}
```

### Expected behavior

Empty response

### Actual behavior

Returns entry that isn't related to the category

I've tried a few different variations on the query and even just querying the field instead of using related to but coming up with the same issue. Any help would be appreciated.

### Craft CMS version

3.55.2

### PHP version

8.1.7

_No response_

### Database type and version

MySQL 5.7.33

### Image driver and version

_No response_

<html>
<body>

Name | Version
-- | --
Amazon S3 | 1.3.0
GeoMate | v1.3.1.2
Redactor | 2.10.10
Super Table | 2.7.3
Tags | 1.0.9

</body>
</html>

brianjhanson:
### Description
Updates the `relateTo` parameter of element queries to respect the current site. 

### Related issues
Fixes #12000


src/elements/db/ElementRelationParamParser.php
```diff
@@ -90,7 +92,9 @@ public static function normalizeRelatedToParam(mixed $relatedToParam): array
             $relatedToParam = [$relatedToParam];
         }
 
-        $relatedToParam = array_map([static::class, 'normalizeRelatedToCriteria'], $relatedToParam);
+        $relatedToParam = Collection::make($relatedToParam)->map(function($relatedToParam) use ($siteId) {
+            return static::normalizeRelatedToCriteria($relatedToParam, $siteId);
+        })->toArray();
```
brianjhanson, line 97: @brandonkelly since we're going into `4.3` I went ahead and used a collection here, but happy to change it back to `array_map` if that's better for a reason I'm not thinking of.

PR #12077, opened 2022-10-08
Prevent TypeError due to `null` values in new elements

AugustMiller:
### Description

Adding a rule that extends `BaseTextConditionRule` to a field layout element causes the edit pages to throw a `TypeError` after creation, due to titles and field values being initialized as `null`.

This adds checks to the three problematic operators (`BaseTextConditionRule::OPERATOR_BEGINS_WITH`, `BaseTextConditionRule::OPERATOR_ENDS_WITH`, and `BaseTextConditionRule::OPERATOR_CONTAINS`) to ensure the values are strings, prior to calling the appropriate `StringHelper` methods.

Edit: Today I'm feeling less confident about the `is_string()` check… initially, I thought just an emptiness check would be fine, but `null` and `''` are still different things, as far as the database is concerned. 😩 Empty strings _do_ behave as expected.


brandonkelly:
Looks good to me, thanks!

PR #12078, opened 2022-10-10
Type error on image transform save

timkelty:
When saving with "auto" quality, the template refactoring in 4.3 results in `quality` being posted as `''` vs `'0'`, which results in a `TypeError`.

### DIFF
--- a/src/controllers/ImageTransformsController.php
+++ b/src/controllers/ImageTransformsController.php
@@ -121,13 +121,9 @@ class ImageTransformsController extends Controller
             $errors = true;
         }
 
-        if (!empty($transform->quality) && (!is_numeric($transform->quality) || $transform->quality > 100 || $transform->quality < 1)) {
+        if ($transform->quality && ($transform->quality > 100 || $transform->quality < 1)) {
             $this->setFailFlash(Craft::t('app', 'Quality must be a number between 1 and 100 (included).'));
             $errors = true;
-        }
-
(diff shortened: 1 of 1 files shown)


src/controllers/ImageTransformsController.php
```diff
@@ -121,15 +121,11 @@ public function actionSave(): ?Response
...
-        if (!empty($transform->quality) && (!is_numeric($transform->quality) || $transform->quality > 100 || $transform->quality < 1)) {
+        if ($transform->quality && ($transform->quality > 100 || $transform->quality < 1)) {
             $this->setFailFlash(Craft::t('app', 'Quality must be a number between 1 and 100 (included).'));
             $errors = true;
         }
 
-        if (empty($transform->quality)) {
-            $transform->quality = null;
-        }
-
```
timkelty, line 132: At this point we can be sure `$transform->quality` is an int between 1-100, or `null`, so this is not needed.

src/controllers/ImageTransformsController.php
```diff
@@ -121,15 +121,11 @@ public function actionSave(): ?Response
             $errors = true;
         }
 
-        if (!empty($transform->quality) && (!is_numeric($transform->quality) || $transform->quality > 100 || $transform->quality < 1)) {
```
timkelty, line 124: Due to the typing of `Transform::quality`, we can simply this condition as we know it is `?int`.

PR #12109, opened 2022-10-13
Fix custom MonologTarget queue issue

timkelty:
- [x] Fixed an issue where logs from a queue job using a custom `MonologTarget` would always log to the default queue target.


AugustMiller (approved):
Smart, fixes our demonstration. Collections are cool. 😎

[We aren't currently advertising](https://craftcms.com/docs/4.x/logging.html#targets) to devs the ability to give targets a handle/ID; instead, they should be pushed with numeric keys, which won't get gobbled up.

Screen Shot 2022-10-13 at 09 01 59

---

Slow on the uptake: I think I finally understand the purpose/necessity of this; the default targets are [disabled by default](https://github.com/craftcms/cms/blob/8653a9890f77e6d3e2f127d1fe6dbd3299bf0422/src/log/Dispatcher.php#L71), but only `console` and `web` are [turned on](https://github.com/craftcms/cms/blob/8653a9890f77e6d3e2f127d1fe6dbd3299bf0422/src/log/Dispatcher.php#L80-L85) for their respective request types. `queue` isn't an entry point / request type (and can be run in either a CLI or HTTP context), so it needs to fend for itself while warming up.

timkelty:
@AugustMiller you got it.

I played around with explicitly reserving `web` `console` and `queue` for keys and/or `MonologTarget::name`, but I couldn't find a great non-breaking way to do it.

Since there generally isn't a reason to key your targets, and if you did, likely wouldn't name them those, I think we should be safe until we can refactor a bit for 5.x

niektenhoopen:
Thanks for fixing the issue @timkelty!

brandonkelly:
Craft 4.2.8 is out now with the fix.

PR #12119, opened 2022-10-13
feat: v4 -Add Twigfield integration for fields that render as object temp…

khalwat:
This PR is for Craft CMS v4

This PR adds integration with [craft-twigfield](https://github.com/nystudio107/craft-twigfield) to provide a Twig editor for fields in Craft CMS that are rendered as Twig object templates.

### Description

Autocomplete is provided for:

* Twig syntax
* Twig filters, functions, and tags
* The Craft API, and any installed plugins
* Section layout shorthand syntax (e.g.: `{title}`

It is provided in 3 places:

#### Editing Sections

In the **Site Settings** table field:

https://user-images.githubusercontent.com/7570798/195686357-db18d9ad-91f9-4931-b174-ea3457e65531.mov

#### Editing Entry Types

In the **Title Format** field:

https://user-images.githubusercontent.com/7570798/195686665-e59c54bc-56e4-4c28-aac8-c3d1b96c0db8.mov

#### Assets field settings

In the **Default Upload Location**, ** Asset Location**, & **Allow Sub-folders** fields:

https://user-images.githubusercontent.com/7570798/195688351-8141ad3b-7b4d-4a37-9d20-d48f4397e88d.mov

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -48,6 +48,7 @@
     "mikehaertl/php-shellcommand": "^1.6.3",
     "moneyphp/money": "^4.0",
     "monolog/monolog": "^2.3",
+    "nystudio107/craft-twigfield": "^1.0.13",
     "pixelandtonic/imagine": "~1.2.4.1",
     "samdark/yii2-psr-log-target": "^1.1",
     "seld/cli-prompt": "^1.0.4",
--- a/src/templates/_components/fieldtypes/Assets/settings.twig
+++ b/src/templates/_components/fieldtypes/Assets/settings.twig
@@ -1,8 +1,11 @@
 {% extends "_components/fieldtypes/elementfieldsettings" %}
 {% import "_includes/forms" as forms %}
 
+{% import "twigfield/twigfield" as twigfield %}
+
 {% macro uploadLocationField(config) %}
     {% embed '_includes/forms/field' with config %}
+        {% import "twigfield/twigfield" as twigfield %}
         {% block input %}
             {% import '_includes/forms' as forms %}
             <div class="flex flex-nowrap">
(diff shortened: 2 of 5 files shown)


khalwat:
I realize that since this is in the CP, P&T may have stringent requirements about the look & feel. @engram-design for example pointed out that he didn't like the Twig logo in the field, for example.

My thought process was that I should use the actual Twig logo, even though I don't like how it looks. To me, it looks like creepy blades of grass, but it is their official logo. If I was listing Craft CMS on a web page, even if I didn't like the Craft CMS logo, probably I use should that rather than inventing my own logo.

However, the idea of a less obtrustive `{}` for a field indicator or such is something I'm perfectly open to. I just want it looking the same everywhere a Twigfield is used, so people know what to expect from that field.

Also, things like the font size, field background, etc. all are customizable via Twigfield on a per-field basis, so wouldn't require any core changes to do them.

lindseydiloreto:
I agree with @khalwat in the comment above. 👍 

The Twig logo looks pretty weird in that format, but certainly there should be _some_ indication that the field has special behaviors. I would agree that a simple `{}` is all that's necessary to convey that message.

khalwat:
Twigfield has been deprecated, and replaced by [Code Editor](https://github.com/nystudio107/craft-code-editor), which is a general-purpose code editor that _also_ does Twig & autocomplete.

I pushed the changes necessary to convert this PR over to use it here (it's mostly just renaming the namespaces, along with some minor API changes):

https://github.com/khalwat/cms/tree/feature/v4-twigfield-integration

Code Editor:

* The JavaScript portion has been rewritten in TypeScript
* The tab handler mentioned above has been addressed
* Supports any language that the Monaco editor supports; examples: [JavaScript](https://github.com/doublesecretagency/craft-cpjs/pull/6), [CSS](https://github.com/doublesecretagency/craft-cpcss/pull/20)

I've already PR'd it into SEOmatic, Sprig, Preparse, CP CSS, and CP JS.

The only remaining issue is translation, which I leave to you to handle.

Issue #11569, opened 2022-07-07
[4.x]: Publishing entries: validation errors are only checked for currently selected site


wsydney76:
### Description

In a multi-site environment it is possible to publish entries that contain validation errors on sites that are not currently selected.

### Steps to reproduce

1. Set up two sites
2. Create a section with a translatable, required field
3. Create/open an entry, leave the field blank
4. Switch to second site
5. Enter a value for that field
6. Press 'Create entry/Save'

### Expected behavior

Don't publish, display errors.

### Actual behavior

Gets published.

### Workaround

We use a workaround in a custom module like this, but not sure whether this will or will not work under any circumstances

```php
// Validate entries on all sites
Event::on(
	Entry::class,
	Entry::EVENT_BEFORE_SAVE, function($event) {

	/** @var Entry $entry */
	$entry = $event->sender;

	// TODO: Check conditionals

	if ($entry->scenario != Entry::STATUS_LIVE) {
		return;
	}

	$entry->validate();

	if ($entry->hasErrors()) {
		return;
	}

	foreach ($entry->getLocalized()->all() as $localizedEntry) {
		$localizedEntry->scenario = Entry::SCENARIO_LIVE;

		if (!$localizedEntry->validate()) {
			$entry->addError(
				$entry->type->hasTitleField ? 'title' : 'slug',
				Craft::t('site', 'Error validating entry in') .
				' "' . $localizedEntry->site->name . '". ' .
				implode(' ', $localizedEntry->getErrorSummary(false)));
			$event->isValid = false;
		}
	}
});
```

/Aylin

### Craft CMS version

4.1.2 (should be the same in Craft 3)

### PHP version

8.0.15

### Operating system and version

Windows 10 Pro

### Database type and version

MySQL 5.7.14

### Image driver and version

Imagick 3.7.0 (ImageMagick 7.1.0-18)

### Installed plugins and versions

-

brandonkelly:
We just released Craft 4.5.0 with a solution to this (#12125). Now all sites are validated, and if any other sites contain validation errors, the site will be mentioned in the new validation error summary.

(We ended up only implementing this for full edit pages, not slideouts, since there’s no way to actually switch to editing other sites from within slideouts.)

PR #12125, opened 2022-10-14
Feature/dev 212 errors summary

i-just:
Work for [dev-212](https://linear.app/craftcms/issue/DEV-212/summarize-errors-on-top-of-entry-page) and #11569 

### Description
for dev-212:
- added errorsSummary to CpScreenResponseBehavior
- added errorsSummary to the cp layout
- added errorsSummary to the CpScreenSlideout

for #11569: 
- added crossSiteValidate param to saving elements; it gets triggered when applying draft and saving element for multi-sites; it then subsequently checks if the element we're saving is supposed to be enabled, live or pending and only validates against Live scenario if that's the case
- cross site show validation errors using the new errorsSummary section with a link to that element for the site that failed validation (with prevalidate param)

TODO:
- styling for the new panel

### Related issues
#11569 
#11612


src/services/Elements.php
```diff
@@ -3241,17 +3248,71 @@ private function _propagateElement(
 
         // Save it
         $siteElement->setScenario(Element::SCENARIO_ESSENTIALS);
+
+        // for Entries and Categories: validate element against "live" scenario across all sites,
+        // if element is to be set to live/pending/enabled
+        if ($crossSiteValidate && ($siteElement instanceof Entry || $siteElement instanceof Category)) {
```
brandonkelly, line 3254: We shouldn’t be making a special case for entries or categories here. The feature should be extended to all element types.

src/services/Elements.php
```diff
@@ -3241,17 +3248,71 @@ private function _propagateElement(
...
+            $siteElementStatus = $siteElement->getStatus();
+            if (($siteElement instanceof Entry &&
+                    ($siteElementStatus === Entry::STATUS_LIVE || $siteElementStatus === Entry::STATUS_PENDING)
+                ) ||
+                ($siteElement instanceof Category && $siteElementStatus === Category::STATUS_ENABLED)
+            ) {
+                $siteElement->setScenario(Element::SCENARIO_LIVE);
+            }
```
brandonkelly, line 3262: We should be respecting the original element’s validation scenario, regardless of status.

brandonkelly (changes requested):

src/services/Elements.php
i-just: ok, pushing an adjustment

src/services/Elements.php
i-just: This was added to mitigate issue: #11569, so the whole point was to validate againt a Live scenario, if the Element is set to be live or pending or enabled (in case of e.g. categories), right? Or am I missing something here?

Without this, it's validated against only essentials, unless someone changes the scenario via e.g. `Entry::EVENT_BEFORE_SAVE` (like in the example from that issue).

brandonkelly: Since this is happening in a private method (`_propagateElement()`) and we have control over all of the places it’s called, we should remove this line entirely from it:

```php
$siteElement->setScenario(Element::SCENARIO_ESSENTIALS);
```

And start calling `setScenario()` on the passed-in `$siteElement` in each of the places that `_propagateElement()` is called.

Generally that will still be `Element::SCENARIO_ESSENTIALS`, except for `_saveElementInternal()`. In that latter case, **_if_** `$crossSiteValidate` is `true`, set it to the same scenario as `$element` instead:

```php
if ($crossSiteValidate) {
    $siteElement->setScenario($element->getScenario());
} else {
    $siteElement->setScenario(Element::SCENARIO_ESSENTIALS);
}
```

i-just: I don't think we can do `$siteElement->setScenario($element->getScenario());` because what if you have 3 sites (propagation set to 'all'), you start creating an entry on Site 1, you trigger cross site validation and it returns an error for e.g. Site 2; you go to that entry for Site 2 and disable it - in that case `$siteElement->setScenario($element->getScenario());` will still insist on validating against live scenario even though you want `$siteElement` to be disabled. I still think it needs to check the status of the `$siteElement` and only use live scenario if that element is set to be live, pending or enabled.

I could move scenario setting to each place we call `_propagateElement()`, but `_propagateElement()` is the only place where we can be sure `$siteElement` is actually set, so having `Element::SCENARIO_ESSENTIALS` in there is a nice safety net - or am I overthinking this part?

src/web/assets/cp/src/css/_cp.scss
```diff
@@ -1485,6 +1485,20 @@ li.breadcrumb-toggle-wrapper {
   }
 }
 
+.errors-summary {
+  margin-bottom: var(--s);
+  padding: var(--s) var(--xl);
+  box-sizing: border-box;
+  box-shadow: inset 0 -1px 0 0 rgba(154, 165, 177, 0.25);
+  border-radius: var(--large-border-radius);
+  background-color: var(--gray-050);
```
gcamacho079, line 1494:
```suggestion
  background-color: var(--white);
```

src/web/assets/cp/src/css/_cp.scss
```diff
@@ -1485,6 +1485,20 @@ li.breadcrumb-toggle-wrapper {
   }
 }
 
+.errors-summary {
+  margin-bottom: var(--s);
+  padding: var(--s) var(--xl);
+  box-sizing: border-box;
+  box-shadow: inset 0 -1px 0 0 rgba(154, 165, 177, 0.25);
```
gcamacho079, line 1492:
```suggestion
  box-shadow: 0 0 0 1px #cdd8e4,0 2px 12px rgba(205,216,228,.5);
```

gcamacho079: I think matching the box-shadow being used by the `.content-pane` and `.meta` containers will help make this container more cohesive. Perhaps create a new CSS variable for this box-shadow?

src/web/assets/cp/src/css/_cp.scss
```diff
@@ -1485,6 +1485,20 @@ li.breadcrumb-toggle-wrapper {
   }
 }
 
+.errors-summary {
+  margin-bottom: var(--s);
```
gcamacho079, line 1489:
```suggestion
  margin-bottom: var(--l);
```

src/web/assets/cp/src/css/_cp.scss
```diff
@@ -1485,6 +1485,20 @@ li.breadcrumb-toggle-wrapper {
...
+  padding: var(--s) var(--xl);
```
gcamacho079, line 1490:
```suggestion
  padding: var(--l) var(--xl);
```

gcamacho079:
Awesome work on this feature @i-just
### Accessibility
The focus management on this page when the page loads is a little unexpected. Focus seems to either be placed on the title field or the active tab button. However, when there’s an error summary present, focus should be placed on the error container. This will require adding `tabindex=“-1”` to the `.errors-summary` container so that focus can be programmatically placed on it. Without explicitly placing focus on the summary, a screen reader user might miss it entirely.

I also noticed on entries with a lot of fields and/or multiple tabs or Matrix blocks, it can be difficult for users to track down the field in question. It would be nice if the list item text was wrapped in an anchor tag. When clicked, focus is moved to the field in question; if it’s in a different tab than the currently-active one, the new tab is activated and then focus is moved. [See an example of this in the 'Error on top' recovery pattern](https://webaim.org/techniques/formvalidation/#error)

### Styles
One styling suggestion that's also a11y-related - a large chunk of red text in the error summary might be anxiety-inducing (especially when paired with the error notification and the red tabs). I think having the summary heading and list text inherit the body text color would work well
A container that reads 'Found 2 errors' along with a list of fields containing errors; the text color is black
Then maybe adding a red warning icon to the left of the heading (like what's being used in the notification container):
A notification container that has a red warning icon, followed by the message 'Couldn't create entry'

i-just:
Thanks Lupe! All clear, makes sense. I've actioned the following:

- summary heading and list text have the body text color,
- added a red warning icon to the left of the heading,
- programatically placing focus on errors container,
- entire cross site validation message is a link (removed "view"),
- list items in the errors summary are wrapped in an anchor tag and link to their corresponding errors within the form - this is still a work in progress and doesn't currently work in the slideouts;

gcamacho079 (changes requested):
Testing the latest changes and I'm running into a couple blockers:

- Clicking on the anchor should move focus to the related input, but the focus is remaining in place.
- The error links don't seem to be in the page's tab order as they don't have `href` attributes, so I'm unable to move keyboard focus to them (see screenshot below, where keyboard focus moves directly from the error container to the discard button)

Screenshot of an entry page with an error message at the top. Keyboard focus moves from the error container, to the Discard button, to the layout tab, but misses the anchors inside the error container.

For reference, I tested with an empty required field in the current tab, in both the CP layout and inside of a slideout.

gcamacho079 (changes requested):
@i-just, jumping from the error to the `.field` container is a really nice experience! 👌🏼 I did notice that it's not working for errors on the title field though.

Making some code suggestions to the "external" icon span so that it allows for an `aria-label` to be used ([more info in the pattern guide for informative icons](https://www.notion.so/pixelandtonic/icons-informative-7aed79caa8d7405cb00aa4790dc5a0d2)).

Also, wondering your thoughts @brandonkelly: regardless of the user setting for “underline links”, I feel like we should add underlines to the anchors here. Even testing with the default blue link color, I feel like this feature is just too easy to miss unless we add an affordance like underlines that makes it fairly obvious that clicking jumps you directly to the field in question.

src/services/Elements.php
```diff
@@ -3242,17 +3255,73 @@ private function _propagateElement(
...
+            $url = UrlHelper::url($siteElement->getCpEditUrl(), $queryParams + ['prevalidate' => 1]);
+            $message = Html::beginTag('a', [
+                'href' => $url,
+                'class' => 'cross-site-validate',
+                'target' => '_blank',
+            ]) .
+                $message .
+                Html::tag('span', '', [
+                    'data-icon' => 'external',
+                    'arial-label' => Craft::t('app', 'Open the full edit page in a new tab'),
```
gcamacho079, line 3317:
```suggestion
                    'aria-label' => Craft::t('app', 'Open the full edit page in a new tab'),
                    'role' => 'img',
```

i-just:
Good catch on the title. Thanks :) That's fixed now.

Regarding making it obvious that errors in the summary are links, if not underlines then would an icon work? Maybe something as simple as an arrow?

Screenshot 2022-12-01 at 12 07 06

That's not committed yet as I'm not sure if it will do the job well - please LMK.

gcamacho079:
### Feedback
1. Only custom fields have their field names wrapped. Is there a way we can achieve this for native fields as well?
2. The left padding on the `span` may need to be removed (see screenshot). In addition, we should use the `--link-color` variable to set the hover color.
3. When there’s a validation error with related assets, the error summary pulls a top-level error for the field and error messages on the asset itself. (See the image below, where an error on a single matrix block results in two bullet points that link to the same field.) I wonder if it makes sense only to show the top-level error since it may be confusing if there are _visually_ only three error messages on the form but four error bullets. 🤔 Interested to hear your thoughts. _To recreate: create a Matrix field with an Image block that has “Validate related assets” checked, upload an image that requires an Alternative Text value, and leave it empty._

### Screenshots
Screenshot of an error summary block with four error messages. Only one error bullet contains underlined text.

An Image matrix block, where the selected image has a validation error

gcamacho079 (changes requested):
See comment!

i-just:
Thanks, @gcamacho079!

Re 1 => thanks, I didn’t check the native fields again; I know what the problem is - on it!
Re 2 => doh... I just typed “blue” to check something quickly and was planning to look for the correct colour/variable afterwards, and then it flew out of my head.
Re 3 => let me see what I can do 🤔

i-just:
That’s all actioned now. In terms of point 3, only the “primary” error will show (e.g. “Fix validation errors on the related asset”). 

I have noticed one more oversight - when an error is, e.g. a field in a matrix field, the name of the field doesn’t get highlighted. I’ll have a look at that shortly.

i-just:
OK, this is ready for another review.

I played with different options and would appreciate some feedback on which one I should go with:
1. go with what I have at the moment - wrapping field label in an `<em>` tag by using markdown; @gcamacho079 I googled accessibility implications of going down this road; I read some articles, and there seems to be no clear answer, but it seems like this should be okay?

2. use actual `<span>` to wrap around the field; an additional replace filter would be needed for the `src/templates/_includes/forms/errorList.twig` file and then in `src/web/assets/cp/src/js/UI.js` in [`addErrorsToList()`](https://github.com/craftcms/cms/blob/ebe7027e072f7139488881bc79e8a6b507bc9053/src/web/assets/cp/src/js/UI.js#L1147) we would either have to change the `text()` to `html()` and that brings in additional XSS risk, or process the error in JS so that we only add the span we want and discard all other markup?

3. add spans differently (e.g. via `src/controllers/ElementsController.php->_errorsSummary()`); I started down that path, and it seems like a lot of processing for something that can be done with a one-liner for options 1 and 2, plus, I still haven’t figured out how to apply this for nested fields (like Matrix, Neo, SuperTable etc.).

gcamacho079 (approved):
Other than the couple of lines of CSS I suggested for the error summary in the context of slideouts, I think this looks great. 👍🏼 💯 

@i-just I agree that the `em` markup seems like the best solution of all the bunch. The underline is indeed there to emphasize the field name, so it seems to convey identical semantics.

jakobhaerter:
Thank you to everyone involved! This helps a lot 🙏

MoritzLost:
> When there’s a validation error with related assets, the error summary pulls a top-level error for the field and error messages on the asset itself. (See the image below, where an error on a single matrix block results in two bullet points that link to the same field.) I wonder if it makes sense only to show the top-level error since it may be confusing if there are visually only three error messages on the form but four error bullets. 🤔 Interested to hear your thoughts. To recreate: create a Matrix field with an Image block that has “Validate related assets” checked, upload an image that requires an Alternative Text value, and leave it empty.

@gcamacho079 @i-just I've raised this issue in a discussion quite a while ago, but it hasn't received much attention: https://github.com/craftcms/cms/discussions/9960

This is a _huge_ pain point for our projects. All our clients encounter this problem – nobody knows what to do with this error message, it's way to technical and they believe it's an issue that we (as developers) have to fix. We've set the `alt` field for images to be required for accessibility, so the editor needs to add an alt-text to every newly uploaded image.

> That’s all actioned now. In terms of point 3, only the “primary” error will show (e.g. “Fix validation errors on the related asset”).

I think this is a step in the wrong direction. The error is still very obscure and leads to misunderstanding. Instead, I would recommend showing all the actual validation errors of the related asset. Show them both inline under the field _and_ in the new summary. Maybe as a nested list so the error 'hierarchy' is clear:

> - Headline cannot be blank
> - Fix validation errors on the related asset
>     - Alternative text cannot be blank

Literally _all_ our clients have encountered this issue when they started working with Craft. For some it's left a bad impression, because the first time they tried to edit some content on their new page, they were presented with an obscure error message that didn't tell them how to fix it. Nobody knew how to 'fix validation errors on the related asset', both the wording and the fact that the actual errors aren't shown makes this very inaccessible.

---

On a related note, it doesn't help that asset fields show no indicator that you can edit their associated fields in a modal by double-clicking. There should be a clearly visible link or context menu, both for accessibility and discoverability. But that's another discussion, see this discussion for details: https://github.com/craftcms/cms/discussions/10763

MoritzLost:
Wasn't this PR supposed to be part of the 4.4 release? We were _really_ looking forward to this one, sad to see it didn't make it. Error messages have way too little visibility right now, multiple of our clients have run into problems because they missed the tiny error message at the bottom left that the entry didn't save correctly.

This PR would've helped, though it still doesn't address the problem regarding validation errors on related assets I mentioned in [my previous comment](https://github.com/craftcms/cms/pull/12125#issuecomment-1400524166).

@i-just @gcamacho079 Any insight on why this didn't make it? Also, I would really appreciate a response to my previous comment as well as the discussion linked there. This is the largest issue we have with Craft, and our clients constantly stumble over this. Regarding the low visibility of error messages, we had to humbly apologize to a client who lost a lot of work because they didn't notice their entries hadn't been saved. There's some user error involved here, but due to the way that errors aren't clearly presented, we couldn't in good conscience tell them it was just their fault.

I _feel_ this should get a higher priority due to the accessibility implications and frustration potential related to this issue.

brandonkelly:
@MoritzLost I didn’t get a chance to review in time for 4.4, but it’s planned for 4.5 :)

i-just:
@MoritzLost, regarding your previous comment, we have decided to make the validation message for related elements more user-friendly, which we believe should help.

The related element validation will read “Validation errors in <field name> <element type> - please fix them.”

MoritzLost:
> @MoritzLost I didn’t get a chance to review in time for 4.4, but it’s planned for 4.5 :)

Thanks @brandonkelly! We're looking forward to it.

> @MoritzLost, regarding your previous comment, we have decided to make the validation message for related elements more user-friendly, which we believe should help.
> The related element validation will read “Validation errors in - please fix them.”

@i-just Thanks, that will be an improvement over the current error message. Though I would still like to see the actual validation errors that occurred somewhere in the context of the field. I guess we'll see if this change helps our new clients understand the problem better without us having to explain it.

PR #12135, opened 2022-10-17
More textual condition rule operators

brandonkelly:
### Description

Adds “has a value” and “is empty” operators to all text, number, and date range condition rules.

### Related issues

- #11070


michaelrog:
Is it possible that this implementation might yield some weird corner cases, since the PHP parts of the implementation (`matchValue()` etc.) are matching on truthy/falsy values, whereas the database parts are filtering on the presence of any value?

e.g. What happens if a text field has the value `"0"`? Querying `not :empty:` should probably still return such a record, but `matchValue()` would fail to match it...?

brandonkelly:
@michaelrog PHP considers `0` to be “empty” (`empty(0) === true`) so I’m not sure if there’s a clear expected behavior there.

PR #12148, opened 2022-10-18
Fix typo in `ImageHelper::targetDimensions()`

### LINKED ISSUE #12137: [3.x]: asset @transforms won't calculate missing width / height
### Description
When using the `@transform` directive in graphql, using "fit" mode and omitting either the `width` or `height` argument, the missing argument is not automatically returned based on the assets aspect ratio. From the [craft 3.x docs](https://craftcms.com/docs/3.x/image-transforms.html#defining-transforms-from-the-control-panel):

> If you leave Width or Height blank, that dimension will be set to whatever maintains the image’s aspect ratio. For example, if you have an image that is 600 by 400 pixels, and you set a transform’s Width to 60, but leave Height blank, the resulting height will be 40.

### Steps to reproduce

(0. Configure craft to store assets in S3)
1. Upload an asset
2. query the asset through graphql and omit 1 of the width / height arguments, i.e: `height @transform(width: 780, mode: "fit")`
3. The query returns a `null` value for the missing argument

### Expected behavior
The query automatically returns a value for the omitted argument, based on the assets aspect ration, as described in the docs.

### Actual behavior
The query returns a null value for the omitted argument. 

Downgrading to Craft 3.7.51 and clearing all the caches fixed the problem for me

### Craft CMS version

3.7.56

### PHP version

7.4.13

Linux 5.10.104-linuxkit

### Database type and version

MySQL 8.0.29

### Image driver and version

GD 7.4.13

- Amazon S3 1.3.0
- Bugsnag 2.1.2
- Freeform 3.13.17
- Redactor 2.10.10
- Redactor custom styles 3.0.4
- SEOMatic 3.4.37
- Super table 2.7.2
- Trigger 0.3.3

brianjhanson:
### Description
Fix error from #11962 where I used the `$transformWidth` and `$transformHeight` instead of `$width` / `$height`. At this point in the code `$width` & `$height` have been returned from `Image::calculateMissingDimension()` which has filled in the missing value with a calculation based on the transform. 

### Related issues
- Fixes #12137

### DIFF
--- a/src/helpers/Image.php
+++ b/src/helpers/Image.php
@@ -10,6 +10,9 @@ namespace craft\helpers;
 use Craft;
 use craft\errors\ImageException;
 use craft\image\Svg;
+use Imagick;
+use Throwable;
+use TypeError;
 use yii\base\InvalidArgumentException;
 
 /**
@@ -102,8 +105,8 @@ class Image
         $imageRatio = $sourceWidth / $sourceHeight;
 
         if ($mode === 'fit' || $imageRatio === $transformRatio) {
-            $targetWidth = min($sourceWidth, $transformWidth, (int)round($sourceWidth / $factor));
-            $targetHeight = min($sourceHeight, $transformHeight, (int)round($sourceHeight / $factor));
+            $targetWidth = min($sourceWidth, $width, (int)round($sourceWidth / $factor));
+            $targetHeight = min($sourceHeight, $height, (int)round($sourceHeight / $factor));
             return [$targetWidth, $targetHeight];
         }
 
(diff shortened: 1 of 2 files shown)


src/helpers/Image.php
```diff
@@ -10,6 +10,9 @@
 use Craft;
 use craft\errors\ImageException;
 use craft\image\Svg;
+use Imagick;
+use Throwable;
+use TypeError;
```
brianjhanson, line 15: PHPStorm added these for me but let me know if the `\Imagick`, etc. syntax is preferred.

brandonkelly: Import all the things!

Issue #11959, opened 2022-09-16
[3.x]: Problem with Save as new entry feature


jlachancekffein:
### Description
In my opinion, there are some issues with the Save as new entry feature.

1) If I create a new entry from an existing entry by clicking on Save as new entry, and the current entry does not pass validation, Craft displays an error message for the current entry but still create a new entry with the errors of the current entry.

2) The status of the new entry is live. According to me, the entry should be either Disabled or Draft. Otherwise, the entry is returned directly via queries, which doesn't make sense to me. Unless I misunderstand how it works.

### Steps to reproduce
1) saves an entry correctly (without validation errors)
2) delete the value of a required field (an asset in my case)
3) duplicates the entry by clicking Save as new entry

### Expected behavior
Prevent entry creation if current contains errors or change status for Disabled or Draft

### Actual behavior
The current entry is not saved but the copy is saved with status Live

### Craft CMS version

3.7.54

### PHP version

8.1

### Installed plugins and versions

-

brandonkelly:
> 1. If I create a new entry from an existing entry by clicking on Save as new entry, and the current entry does not pass validation, Craft displays an error message for the current entry but still create a new entry with the errors of the current entry.

Yikes, that’s definitely a bug. It also loads the edit page for the duplicated entry, despite the URL still showing the original entry’s ID.

I’ve fixed this so the new entry never gets committed to the database, and the error messages are displayed back on the original entry, for the next Craft 3 release (f822ec142861298e10554350a0bae24df0bb7bfb).



> 2\. The status of the new entry is live. According to me, the entry should be either Disabled or Draft. Otherwise, the entry is returned directly via queries, which doesn't make sense to me. Unless I misunderstand how it works.

That’s debatable. The current behavior is respecting what you’ve submitted in the form, including the Enabled setting. It would be weird if it respected most of the form settings (e.g. changes to the Title), but not the Enabled setting. If you want the duplicated entry to be saved as disabled, you can mark the entry as disabled, then press “Save as a new entry”, and the posted status will be respected.

That said, I do see the other side of the argument. From a workflow perspective, most people are going to duplicate the entry and _then_ start making changes to it. We had a plan to address this for Craft 4, but unfortunately we couldn’t implement it in time for the 4.0 release. You can subscribe to #10244 for updates on it.

> The same issue also occurs on Craft 4, without any visible errors.

@jeroenverfallie Craft 4 will duplicate the entry as-is, without validating it.

[...]

Issue #9577, opened 2021-07-15
"Save as new entry" only available if user has publish rights


nickdunn:
### Description
I have some users who don't have "Publish live change" access in a section.

For these users the "Save as new entry" option isn't available in a dropdown, so they are unable to duplicate entries. I've had to tick "publish to live" so they can see the option.

This feels like a bug — users should be able to duplicate (save as new) without having "publish to live" access.

<img width="246" alt="Screenshot 2021-07-15 at 11 54 07" src="https://user-images.githubusercontent.com/39560/125777050-b6f56001-d26c-4be8-842b-4e831498f423.png">

<img width="251" alt="Screenshot 2021-07-15 at 11 56 26" src="https://user-images.githubusercontent.com/39560/125777110-daa6fb76-8371-43e8-ac94-7433af8ca92e.png">

### Additional info

- Craft version: 3.7.0
- PHP version: 8.0.2

brandonkelly:
Agree, any user with “Create” permissions should be able to duplicate an entry into a new unpublished draft. Unfortunately this could result in a breaking change for some plugins, so will need to make the change in 4.0.

MoritzLost:
@brandonkelly This was originally on the list for 4.0, but it has been removed from the 4.0 milestone. Was this feature more difficult to implement than anticipated, or is there a reason it was delayed? Is there a new ETA for this change?

This is still a pain-point for one of my clients. See the related discussion for details: #10244

MoritzLost:
@brandonkelly That's too bad, but thanks for the info! This definitely has my vote for a highly anticipated feature 🙃

brandonkelly:
We just got to this for Craft 5 (#12303) 🎉  Going forward, any user with the “Create entry” permission will have the option to “Save as a new entry” for any entry they have permission to view, and the newly-created entry will start off in an unpublished draft state rather than a fully-saved entry.

anchovy:
We've had issues with this as well on v4 - a user saved an entry as a new entry and then mistakenly updated and modified the original entry rather than the 'new' entry which caused a lot of issues.

Is there anything on the EVENT_BEFORE_SAVE (or any other events) hook that indicates that the 'new' entry is being generated by clicking on Save > 'Save as a new entry' or 'Duplicating' (from the list view)?
![image](https://github.com/user-attachments/assets/99f504ce-c367-46d4-bfb9-d60b99d1bf26)

brandonkelly:
[...]

PR #12152, opened 2022-10-19
Feature/dev 69 save as new changes

i-just:
### Description

- “Save as a new {type}” creates an unpublished draft
- "Save as a new {type}" is available to users with "Create" permission too, (not as currently with "Save" permission)
- If you edit existing e.g. entry (create a provisional draft) and then click “Save as a new {type}” it applies your changes to the new element and discard them from the existing one.

No changes made to the index pages duplicate functions.

### Related issues

- [Dev-69](https://linear.app/craftcms/issue/DEV-69/save-as-a-new-entry-should-create-an-unpublished-draft)
- #10244 
- #9577 
- #11959

### DIFF
--- a/src/controllers/ElementsController.php
+++ b/src/controllers/ElementsController.php
@@ -677,6 +677,18 @@ class ElementsController extends Controller
                     ],
                 ]);
             } else {
+                // save as a new is now available to people who can create drafts
+                $components[] = Html::beginForm() .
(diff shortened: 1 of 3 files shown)


i-just:
"Save as a new {type}" creating an unpublished draft is on hold until C5 as it introduces breaking changes for e.g. Neo and Super Table plugins.

PR #12156, opened 2022-10-19
Add content type meta tag for system messages

kennethormandy:
### Description

This pull request adds the content type meta tag to the default HTML email for system messages:

```html
<meta http-equiv="Content-Type" content="text/html charset=UTF-8" />
```

[Litmus](https://help.litmus.com/article/253-accessibility-in-litmus-checklist) and [Email on Acid](https://www.emailonacid.com/blog/article/email-development/the_importance_of_content-type_character_encoding_in_html_emails/) both suggest this is required to properly support screen readers.

Apparently, email clients will only use the `Content-Type:` already set in message header sent as part of the email itself, and ignore this in the HTML. Screen readers, however, do need it.

There were no changes in the visual tests, either.

### Related issues

Ideally, we’d also set the `lang` attribute on the root HTML tag for screen readers, but I’m not sure if we can set this reliably:

> Without a [lang] attribute, screen readers will assume the email is in the default language the subscriber chose when setting up their screen reader. If your email is not in that user’s language, the screen reader may not accurately transcribe your message.

The system messages have to be manually translated and might not be the same as the system language, correct?


brandonkelly:
> Ideally, we’d also set the `lang` attribute on the root HTML tag for screen readers, but I’m not sure if we can set this reliably

Done (85e00b53e1d6abb362cb9178f1d947cadc2f6375). Should be pretty reliable since the default system messages are all translated, and they’ve overridden on a per-language basis, so in theory no one will override them for a language they haven’t translated yet.

Issue #12165, opened 2022-10-20
[4.x]: discard provisional draft from the element you're duplicating with "save as a new {type}"


i-just:
### Description
If you edit existing entry (create a provisional draft) and then click “Save as a new entry”, it should apply your changes to the new entry and discard them from the existing one

### Craft CMS version

4.x

### Installed plugins and versions

-

### Related issues
#11959 
#12102 
#10204

brandonkelly:
#12167 has been merged for 4.3 🎉

PR #12167, opened 2022-10-20
discard original provisional draft with save as a new

i-just:
### Description
If you edit existing element (create a provisional draft) and then click “Save as a new {type}”, it applies your changes to the new element and discards them from the existing one.

### Related issues
#12165


src/services/Elements.php
```diff
@@ -1376,6 +1378,11 @@ public function duplicateElement(
         // Clean up our tracks
         $mainClone->duplicateOf = null;
 
+        // discard provisional draft from the original element
+        if ($originalProvisionalDraftId) {
+            $this->deleteElementById($originalProvisionalDraftId);
+        }
+
```
brandonkelly, line 1385: This doesn’t feel worth the extra argument on `duplicateElement()`. Could have just as easily been a check within `ElementsController::actionDuplicate()` after the duplication is successful.

```php
if ($element->isProvisionalDraft) {
    Craft::$app->getElements()->deleteElement($element);
}
```

i-just: agreed; I've pushed an update

PR #12175, opened 2022-10-21
Return photoId from actionUploadUserPhoto

timkelty:
### Description
Makes it a bit more flexible for frontend use.


src/controllers/UsersController.php
```diff
@@ -1533,6 +1533,7 @@ public function actionUploadUserPhoto(): ?Response
 
             return $this->asJson([
                 'html' => $this->_renderPhotoTemplate($user),
+                'assetId' => $user->photoId,
```
brandonkelly, line 1536: Shouldn’t it be `photoId` to be consistent with the property?

timkelty: yes, yes it shuold

PR #12184, opened 2022-10-25
Pass ID to field layout tab configs

### LINKED ISSUE #11925: [4.x]: Dynamic Entry Titles shortcut syntax causing InvalidArgumentException error
### Description
When using Dynamic Entry Titles on entries / sections / entry types an error gets thrown in the logs.

yii\base\InvalidArgumentException: Invalid field layout element class: craft\fieldlayoutelements\EntryTitleField in /PATHTOPROJECT/vendor/craftcms/cms/src/services/Fields.php:1273

### Steps to reproduce

1. create a section + entry type and use shorthand field handle in the Title Format (uncheck Show Title Field): {fieldName} 
2. using {{object.fieldName}} solves the issue

### Expected behavior
No errors

### Actual behavior
An error that takes hours of searching to find fix for ;-)

### Craft CMS version

4.2.3

### PHP version

8.0.8

_No response_

### Database type and version

_No response_

### Image driver and version

_No response_

-

brianjhanson:
### Description
When saving field layout tabs, [the ID is checked](https://github.com/craftcms/cms/blob/cc58802be988d2311722fb0dbf390d8c10653bf0/src/services/Fields.php#L1376-L1380) to see if we have an existing record that needs to update or not. However, the `id` wasn't getting passed up within the POST request.

The result is that DB records in the `fieldlayouttabs` table could drift away from the data stored in project config. As far as I know this only resulted in a lot of error logs, but it's annoying nonetheless.

Running `project-config/apply --force` clears things up as well, but this fix should prevent it from happening in the future.

### Related issues
Fixes #11925

### DIFF
--- a/src/models/FieldLayoutTab.php
+++ b/src/models/FieldLayoutTab.php
@@ -203,7 +203,7 @@ class FieldLayoutTab extends FieldLayoutComponent
             $this->uid = StringHelper::UUID();
         }
 
-        $config = $this->toArray(['name', 'uid', 'userCondition', 'elementCondition']);
+        $config = $this->toArray(['name', 'id', 'uid', 'userCondition', 'elementCondition']);
         $config['elements'] = $this->getElementConfigs();
         return $config;
     }


brandonkelly:
`getConfig()` is primarily meant for defining the tab’s project config data – which should never include IDs since they can’t be consistent across environments. So we’ll need to come up with a different solution here.

Issue #12205, opened 2022-10-26
[4.x]: Numbers in a plain text field on an element index page get numeric formatting applied


angrybrad:
### Steps to reproduce

1. Use a fresh install of the latest 4.3.0 release.
2. Create a Plain Text field... let's call it `SKU`
3. Create a channel section... let's call it `Product`
4. Add `SKU` to `Product`'s field layout.
5. Create an entry in `Product` and give the `SKU` a numeric value: `356041`
6. Save the entry
7. Add `SKU` to the entry index view via a custom source
8. Notice local-specific numeric formatting is applied

![CleanShot 2022-10-26 at 16 35 29](https://user-images.githubusercontent.com/61869/198158347-c546d55e-38f1-4844-9682-82eb5280f36f.png)

If you add a letter (i.e. 356a041) numeric formatting doesn't get applied.

This doesn't happen on the latest Craft 3.7 release, so this is a regression.

### Expected behavior

Numeric formatting does not get applied to plain text fields on element index pages for plain text fields.

### Actual behavior

It does get applied.

### Craft CMS version

4.3.0

### PHP version

8.1

### Operating system and version

macOS

### Database type and version

MySQL 8.1

### Image driver and version

n/a

### Installed plugins and versions

None

brandonkelly:
Fixed in 4.3.1

yingban:
@brandonkelly Thanks!

PR #12209, opened 2022-10-27
number formatting in plain text field - index page

i-just:
### Description
Numbers in a plain text field on an element index page shouldn't get numeric formatting applied.

For C5 we should consider changing `src/base/Field.php` `getTableAttributeHtml()` (and other implementations of it) to accept the field and pass it along, so that we can act on that information in places like `ElementHelper::attributeHtml()`. Doing it now would introduce breaking changes for some plugins (e.g. besteadfast/craft-preparse-field).

### Related issues
#12205

### DIFF
--- a/src/fields/PlainText.php
+++ b/src/fields/PlainText.php
@@ -14,6 +14,7 @@ use craft\base\PreviewableFieldInterface;
 use craft\base\SortableFieldInterface;
 use craft\fields\conditions\TextFieldConditionRule;
 use craft\helpers\Db;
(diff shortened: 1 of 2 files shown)


brandonkelly:
I just reverted the commit that added number formatting (5b55aabbe7284de3232d95204f0a05fd2bfddbb5), as I noticed it was also affecting things like ID, which definitely shouldn’t get formatted, despite being an actual integer. Number fields are already doing their own formatting, so and it’s easy enough for any other numeric attributes to do it as well when it makes sense.

PR #12213, opened 2022-10-27
Update text colors for hovered/selected elements and add adjacent text to checkbox icons

gcamacho079:
### Description
Makes the following changes:
- Adds a new mixin for updating the focus ring with custom colors
- Updates the output for boolean custom field values to also include adjacent "Enabled" text. This fixes an issue of the icon not contrasting enough in the default/hover/selected element states (note: this is a solution that will likely need to be used for status icons as well).

### Related issues
Resolves DEV-1034


brandonkelly:
For DEV-1034, why can’t we just darken the checkmark color?

gcamacho079:
> For [DEV-1034](https://linear.app/craftcms/issue/DEV-1034), why can’t we just darken the checkmark color?

We certainly could. I was torn between this solution and just updating the color to ensure it's 3:1 against each background color; I think it would require a lighter green or white color in the "selected" state.

I opted for this solution to have some visual consistency between this and the [status indicator change in this PR](https://github.com/craftcms/cms/pull/12306) (i.e. an icon that _doesn't_ change color based the element state, next to text that always meets the minimum contrast ratio for text).

brandonkelly:
Ah true, it would be good to stay consistent with the “enabled” green color. I’d be OK with a white checkbox over a green background, if that’s acceptable.

brandonkelly:
I don’t really care for the “Enabled” label since that doens’t always make sense for the context. For example “Enabled” would be a confusing label for a Lightswitch field labeled “Featured entry?”. We could maybe add a setting for it, but if there’s a way to solve this without the label, I’d prefer that.

gcamacho079:
> Ah true, it would be good to stay consistent with the “enabled” green color. I’d be OK with a white checkbox over a green background, if that’s acceptable.

Do you mean something along these lines?
In a cell with the column titled 'Show author email', there's a white checkbox icon that's enclosed in a medium green, rounded box.

This ensures that when the element row is selected and has a dark gray background, the white checkmark still remains visible over its green background:
The same checkbox icon, but over a dark gray background

If so, yes this is completely acceptable!

gcamacho079:
@brandonkelly, this is ready for review again. I’ve made the stylistic change and ensured the custom “On” label is used as the text alternative when there is one set.

brandonkelly:
Lightened up the selected BG color a bit, to be consistent with the selected source item, the selected source path item (#12564), and hovered disclosure menu items.

gcamacho079:
@brandonkelly good call, the lighter color does indeed provide enough contrast!

brandonkelly:
Sweet! Merged 🎉

PR #12214, opened 2022-10-27
Fill image transform setting

brianjhanson:
### Description
Adds an additional `fill` parameter to image transforms.

The value of `fill` must be either a hex value (3 or 6 digits) or the string `transparent`. 

When the transform mode is set to `fit` and a fill color is provided, the fill color will be used to fill any empty pixels surrounding the image so the final image size is always equal to the transform sized passed in. 

You can also use `position` in order to control where within the transform space the image should appear. 

Within the UI there's a new "Fill empty pixels" checkbox that will show or hide the color field for the fill. 

https://user-images.githubusercontent.com/1843073/198377492-ad4793a4-9cb5-4a9e-9469-8fb6df80e369.mp4

### Related issues
#8848

### To-Do
- [x] Support for animated files (gif, webp)
- [x] Make sure non-raster images don't break


src/controllers/ImageTransformsController.php
```diff
@@ -156,6 +158,13 @@ public function actionSave(): ?Response
             $errors = true;
         }
 
+        $useFill = $this->request->getBodyParam('useFill');
+        if ($useFill && $transform->mode === 'fit') {
+            $transform->fill = $transform->fill ? ColorValidator::normalizeColor($transform->fill) : 'transparent';
+        } else {
+            $transform->fill = null;
+        }
+
```
brianjhanson, line 167: I'm not wild about this, but I couldn't think of a better way to allow users to specify if they wanted the current behavior (image is sized proportionately to fit within a window) or the new behavior (image is placed within an area specified by the transform). 

In code, the fact that `fill` exists tells Craft which behavior to us, but in the UI a fill is always returned, and currently there's not a difference between no fill and a transparent one.

src/migrations/Install.php
```diff
@@ -197,6 +197,7 @@ public function createTables(): void
             'format' => $this->string(),
             'quality' => $this->integer(),
             'interlace' => $this->enum('interlace', ['none', 'line', 'plane', 'partition'])->notNull()->defaultValue('none'),
+            'fill' => $this->string(11)->null(),
```
brianjhanson, line 200: length is 11 so `transparent` fits

src/base/Image.php
```diff
@@ -74,7 +74,7 @@ abstract public function crop(int $x1, int $x2, int $y1, int $y2): self;
      * @param bool $scaleIfSmaller
      * @return static Self reference
      */
-    abstract public function scaleToFit(?int $targetWidth, ?int $targetHeight, bool $scaleIfSmaller = true): self;
+    abstract public function scaleToFit(?int $targetWidth, ?int $targetHeight, bool $scaleIfSmaller = true, string $position = 'center-center'): self;
```
brandonkelly, line 77: Missing `@param` for `$position`

src/image/Raster.php
```diff
@@ -234,7 +240,7 @@ public function crop(int $x1, int $x2, int $y1, int $y2): self
     /**
      * @inheritdoc
      */
-    public function scaleToFit(?int $targetWidth, ?int $targetHeight, bool $scaleIfSmaller = true): self
+    public function scaleToFit(?int $targetWidth, ?int $targetHeight, bool $scaleIfSmaller = true, array|string $position = 'center-center'): self
```
brandonkelly, line 243: `$position` has a different type declaration here than in the base class (`array|string` vs `string`).

src/base/Image.php
brandonkelly: Also if the fill stuff is exclusive to `Raster`, it feels weird to have `$position` defined here in the base class. Maybe `Raster` should just be defining its own `scaleToFitAndFill()` method that combines `scaleToFit()` and `setFill()`?

brandonkelly (changes requested):

brandonkelly:
@brianjhanson This doesn’t seem to be working for me.

I created a transform with the following settings:

- **Mode:** Fit
- **Fill empty pixels:** ✅ 
- **Fill Color:** `#ff0000`
- **Width:** 300
- **Height:** 200
- **Quality:** Auto
- **Interlacing:** None
- **Image Format:** Auto

However when I output a 1200x394 image using this transform, the result is 300x99 without any red pixels. I would have expected it to end up as 300x200, with ~50 red pixels above and below the original image.

(`upscaleImages` is `false`.)

And separately, looks like there’s a bug when **Fill Color** is left blank: when you go back to the transform edit page, the input value is set to `#transparent`.

brandonkelly (changes requested):
As discussed,

- [ ] Let’s introduce a new dedicated “Letterbox” transform mode for this, with “Fill Color” and “Gravity” settings. (Gravity should default to `center-center`.)
- [ ] All transforms should get a new “Upscale” setting, and the `upscaleImages` config setting should be used by default.

brianjhanson:
Just pushed some code to address the first half of this and wanted to jot down some notes before I forget. 

### Positioning
I went with "Position" instead of "Gravity" because that felt more straightforward. In Cloudinary world, Gravity is how they refer to a focal point, but we have established language for that. We could keep it as "Focal Point" here, but given how it behaves, I think "position" is a better description (and aligns with how to determine it programmatically). 

CleanShot 2023-02-02 at 14 29 52

I duplicated the field rather than just messing with it via JS because JS introduced some complications (translations, accessibility, etc.). It felt like just duplicating the field was better all around. 

### UI Update
CleanShot 2023-02-02 at 14 18 21

### Focal Points
If an image has a focal point set, we will do our best to put that point into the center. Some screenshots of how this behaves are below. 

**Focal point**
CleanShot 2023-02-02 at 14 22 17

**Results in**
CleanShot 2023-02-02 at 14 22 29

PR #12221, opened 2022-10-28
BaseOptionsField::EVENT_SET_INPUT_OPTIONS

friartuck6000:
Adds an event, and a slight change to surrounding code, to option fields. This allows plugins and modules to modify the options that are rendered for the input.

**Related discussion:** #12212

### DIFF
--- a/src/events/InputOptionsEvent.php
+++ b/src/events/InputOptionsEvent.php
@@ -1,0 +1,25 @@
+<?php
+
+namespace craft\events;
+
+use craft\base\ElementInterface;
+use craft\elements\MatrixBlock;
+use yii\base\Event;
+
+class InputOptionsEvent extends Event
+{
+    /**
+     * @var array The options that will be available for the current field
+     */
+    public array $options;
+
+    /**
+     * @var ElementInterface|null The element that the field is generating an input for.
+     */
+    public ?ElementInterface $element = null;
+
+    /**
+     * @var mixed The current value of the field.
+     */
+    public mixed $value;
+}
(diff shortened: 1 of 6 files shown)


brandonkelly:
Seems like this could have been simplified if the event was added to `BaseOptionsField::options()` instead?

friartuck6000:
That was my first thought, but `BaseOptionsField::options()` is used for other things besides generating the input HTML. My goal was to keep this event scoped to only affecting input markup, the way [`EVENT_SET_FIELD_BLOCK_TYPES`](https://github.com/craftcms/cms/blob/main/src/fields/Matrix.php#L692) is for Matrix.

brandonkelly:
Ah ok. In that case it could be added to `translatedOptions()` which is already effectively the same as `inputOptions()`. And we can rename `translatedOptions()` to `inputOptions()` in Craft 5.

friartuck6000:
The difference is that `translatedOptions()` doesn't get passed the `$value` or `$element` args from `inputHtml()`, which are a key part of making the event useful. We would have to change the signature of `translatedOptions()` to accept those. I guess that wouldn't _technically_ be a breaking change since it's not a public method, but it still kinda feels like one, since plugins and user modules could still be using or overriding it.

brandonkelly:
Adding a new `$options` argument to the beginning of `translatedOptions()` would have been a worse breaking change than adding new `$value` and `$element` arguments _after_ `$encode`.

So just dropped `inputOptions()` and the `$option` argument from `translatedOptions()`, and added `$value` and `$element`, and merged for Craft 4.4 🎉

friartuck6000:
Heck of a point 🤦‍♂️ thanks!

PR #12224, opened 2022-10-28
Animated PNG support

brianjhanson:
Animated PNGs (APNG) intentionally [disguise themselves as a regular PNG](https://stackoverflow.com/a/4525194) which makes it a little hard to determine an APNG from a boring old PNG.

The main change made here is to add `FileHelper::isAnimatedPng` which will look for the specific characters needed for an animated PNG. When those are present we avoid running `Images::cleanImage` on it as that process will flatten the image.

One note, when uploading an `.apng` file, the file name will automatically get `.png` appended to it. This seems to be something to do with `input[type=file]` itself as it's happening even in a [reduced test case](https://codepen.io/brianjhanson/pen/MWXwPXz?editors=1011)

### Related issues
#12178

### DIFF
--- a/src/config/GeneralConfig.php
+++ b/src/config/GeneralConfig.php
@@ -241,6 +241,7 @@ class GeneralConfig extends BaseConfig
         'aiff',
         'asc',
         'asf',
+        'apng',
         'avi',
         'avif',
         'bmp',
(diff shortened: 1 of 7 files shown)


brandonkelly:
@brianjhanson, Considering we just added HEIC support to Craft 3 + Craft 4, it seems like this should initially be done for Craft 3 and then we can port the changes to Craft 4.

Issue #12250, opened 2022-11-02
In Field Layout Editor, UI Elements with Entry Conditions lack a dashed border [4.x]: 


mikeatmostlys:
### Description
Seeing that fields with conditions have a dashed border, I expected UI Elements with conditions (a template in this case) to have a similar border as well.

### Steps to reproduce

1. Add a template UI element with an entry condition

### Expected behavior

Expected design to indicate UI element was conditional.

### Actual behavior

Appears like other UI elements.

<img width="820" alt="Screen Shot 2022-11-02 at 9 46 53 AM" src="https://user-images.githubusercontent.com/92817780/199523201-8f017d28-5987-4d77-9169-2a839a20878a.png">
<img width="511" alt="Screen Shot 2022-11-02 at 9 47 00 AM" src="https://user-images.githubusercontent.com/92817780/199523210-5e144f24-10c4-4451-8096-7006d6f01acf.png">

### Craft CMS version

4.3.1

### PHP version

8.1

### Operating system and version

Mac OS

### Database type and version

MySQL 5.7

### Installed plugins and versions

- Neo

mikeatmostlys:
I forced a black background and hmm, actually looks like it has the border, it's just white-on-white so not visible.

<img width="288" alt="Screen Shot 2022-11-02 at 10 08 11 AM" src="https://user-images.githubusercontent.com/92817780/199526450-58307c1f-9687-460d-abc7-5f70374fb18a.png">

brandonkelly:
Resolved this for Craft 4.4: the dashed border has been replaced with a new condition icon, which will sit beside layout component labels (similar to the field-required icon). Advantage of going this route is that we can also use it to identify conditional tabs:

![Two configured tabs within a field layout designer, highlighting the new condition icon on a UI element, a field, and a tab.](https://user-images.githubusercontent.com/47792/200218529-1f527a0a-7a96-4604-93af-ff4dc8611908.png)

mikeatmostlys:
Great, thank you!

brandonkelly:
4.4 is out with that change!

mike-moreau:
Thanks!

PR #12254, opened 2022-11-03
make the dashed border visible for ui elements

i-just:
### Description
If UI element added to the field layout has a condition on it, the dashed border indicating that it's a conditional element doesn't show (white on white). Adjusted the styling so that conditional UI elements have a gray dashed border around them.

Screenshot 2022-11-03 at 09 59 20

### Related issues
#12250


brandonkelly:
Thanks! I’ve been meaning to come up with a better way to identify conditional layout elements that would also work for tabs, which would look really awkward with a dashed border. Spent some time playing with this today and came up with a condition icon that does the trick: https://github.com/craftcms/cms/issues/12250#issuecomment-1305025364

PR #12271, opened 2022-11-07
Allow `array` as return type for `prepareElementQuery`

engram-design:
There's an issue with GQL queries where [many](https://github.com/craftcms/cms/blob/b87d49b8060014d594b98819ec60bc146a471ea0/src/gql/resolvers/elements/Address.php#L48) of the [resolvers](https://github.com/craftcms/cms/blob/b87d49b8060014d594b98819ec60bc146a471ea0/src/gql/resolvers/elements/Entry.php#L49) return an array.

Which shown as per the below
https://github.com/craftcms/cms/blob/b87d49b8060014d594b98819ec60bc146a471ea0/src/gql/base/ElementResolver.php#L87-L93

If `prepareQuery()` doesn't return an `ElementQuery` then return it.

Otherwise, you'll get the following error: `craft\\gql\\base\\ElementResolver::prepareElementQuery(): Return value must be of type craft\\elements\\db\\ElementQuery|Illuminate\\Support\\Collection, array returned`

I suppose you could either enforce the resolvers to return a collection, or tighten up the `prepareQuery()` return types - take your pick!

### DIFF
--- a/src/gql/base/ElementResolver.php
+++ b/src/gql/base/ElementResolver.php
@@ -76,7 +76,7 @@ abstract class ElementResolver extends Resolver
      * @param ResolveInfo $resolveInfo
      * @return ElementQuery|Collection
      */
-    protected static function prepareElementQuery(mixed $source, array $arguments, ?array $context, ResolveInfo $resolveInfo): ElementQuery|Collection
+    protected static function prepareElementQuery(mixed $source, array $arguments, ?array $context, ResolveInfo $resolveInfo): ElementQuery|Collection|array
     {
         /** @var ArgumentManager $argumentManager */
         $argumentManager = empty($context['argumentManager']) ? Craft::createObject(['class' => ArgumentManager::class]) : $context['argumentManager'];


brandonkelly:
@brianjhanson Changing the return type would be a breaking change to any resolvers that override that method, so we need to tread carefully here.

My understanding is that the only time these would not be element queries is if they were eager-loaded, in which case they should be a collection, not an array.  So we should look into why arrays are cropping up—maybe a bug further up the chain?

Even if there ends up being a very good reason for them being arrays, it might make more sense to wrap them in a collection rather than changing the return type.

engram-design:
More than happy to go with what you recommend, but as it is, Craft itself will throw this error, due to your resolvers returning an array.  

https://github.com/craftcms/cms/blob/b87d49b8060014d594b98819ec60bc146a471ea0/src/gql/resolvers/elements/Address.php#L47-L49

https://github.com/craftcms/cms/blob/b87d49b8060014d594b98819ec60bc146a471ea0/src/gql/resolvers/elements/Asset.php#L48-L50

https://github.com/craftcms/cms/blob/b87d49b8060014d594b98819ec60bc146a471ea0/src/gql/resolvers/elements/Category.php#L48-L50

Might I suggest changing these to return `Collection`'s instead?

brandonkelly:
Thanks for the pointer! Fixed via #12275.

brandonkelly:
Craft 4.3.2 is out with that fix.

PR #12276, opened 2022-11-08
Improve some Dutch (NL) app translations

7ochem:
### Description

While working with the Craft CMS admin in Dutch, I noticed that some of the translations are off. I'm a native Dutch speaker.

```diff
-    'Create {type}' => '{type} maken',
-    'Create' => 'Maken',
+    'Create {type}' => 'Maak {type} aan',
+    'Create' => 'Aanmaken',
```
The actual verb is "Aanmaken" or "Maak aan", so "{type} maken" sounds off without "aan" being present. This form is more often being used when physically making something (like cooking a dish). Also with `{type}` at the start, this results in a string text starting with a lower case character.

```diff
-    'Current' => 'Actueel',
+    'Current' => 'Huidige',
```
"Actueel" is more like the _newest_ thing (being up to date) where "Huidige" is the _current_ thing or state.

### Related issues

None


7ochem:
> Thanks! I’ve made the same changes in CrowdIn over at [translate.craftcms.com](https://translate.craftcms.com/). (You can make edits there in the future if you’d like as well.)

Aha, good to know! Thanks! 👍🏻

PR #12286, opened 2022-11-09
Element index "sort updated" message via live region

gcamacho079:
### Description
Aims to fix an issue where every update of the index page (source change, new page, etc.) triggered the live region message. In addition, only the table view was triggering this message, leaving changes in the Asset thumbnail view unannounced.

Functionality has been moved to `BaseElementIndex` instead.

### Related issues
Resolves DEV-411, DEV-1057


src/web/assets/cp/src/js/BaseElementIndex.js
```diff
@@ -861,7 +869,11 @@ Craft.BaseElementIndex = Garnish.Base.extend(
       return params;
     },
 
-    updateElements: function (preservePagination, pageChanged) {
+    updateElements: function ({
+      preservePagination,
+      pageChanged,
+      sortUpdated,
+    } = {}) {
```
gcamacho079, line 876: @brandonkelly wondering if this is the way you would approach this. I had also considered adding additional events, like `sortUpdated`, `sourceChanged`, but this seemed like it might be more straightforward.

brandonkelly: We can’t make a backwards-incompatible change to `updateElements` like this, as some plugins may be calling it themselves.

We could add the additional parameter, but it seems like would just start announcing the message directly from the 3 different places you are passing `sortUpdated: true`?

brandonkelly (changes requested):

src/web/assets/cp/src/js/BaseElementIndex.js
gcamacho079: That makes sense. So instead of bundling this functionality into the `updateElements` function, just calling the `_updateScreenReaderStatus` function directly after `updateElements`?

I can give that a try!

gcamacho079:
@brandonkelly I've made the requested change 👍🏼

gcamacho079:
@brandonkelly I'm going to lump in another change with this, please hold on reviewing 🙏🏼

src/web/assets/cp/src/js/BaseElementIndex.js
```diff
@@ -909,11 +938,44 @@ Craft.BaseElementIndex = Garnish.Base.extend(
...
+                    first: first,
+                    last: this.getLastItemNumber(first, total),
+                    total: total,
+                    item: itemLabel,
+                    items: itemsLabel,
+                  }
+                );
+              }
 
-          if (pageChanged) {
```
gcamacho079, line 914: @brandonkelly would it be safe to remove this from the function parameters since it's not being used anymore?

brandonkelly: Sure!

src/web/assets/cp/src/js/BaseElementIndex.js
```diff
@@ -1323,6 +1385,38 @@ Craft.BaseElementIndex = Garnish.Base.extend(
...
 
+    criteriaHasChanged: function () {
+      if (!this._previousViewParams) {
+        return false;
+      }
+
+      return !_.isEqual(
+        this._viewParams.criteria,
+        this._previousViewParams.criteria
+      );
```
brandonkelly, line 1396: @gcamacho079 Is this the only reason for including Lodash? If so, it looks like `Craft.compare()` would work just as well for this purpose.

gcamacho079: @brandonkelly good call, I didn't realize that existed. I was just trying this locally, and I'm getting a couple of errors trying to implement it.

First, when I try it with these objects as parameters, I'm getting an error at some point recursively where `obj1` is null and the compare is failing. If I add a return at the beginning of the function (`if (obj1 === null || obj2 === null) return`), the `compare` returns the opposite of what I would expect. 🤔

brandonkelly: @gcamacho079 Doh, was able to reproduce that with `Craft.compare({foo: null}, {foo: 'foo'})`. Fixed now, including in this branch.

Issue #11538, opened 2022-07-01
[3.x]: `error: array_merge(): Expected parameter 1 to be an array, bool given` on composer update


romainpoirier:
### Description
On update, I get this composer error:
`error: array_merge(): Expected parameter 1 to be an array, bool given`
This happens even if I try to update all, or if I try to update craft/plugins indivually.

### Steps to reproduce

1. run `composer update` or click the Update button in the CP

### Craft CMS version

3.7.45.2

### PHP version

7.4.25

### Operating system and version

Mac OS 12.3.1

### Database type and version

MySQL 5.7.34

### Image driver and version

Imagick 3.4.4 (ImageMagick 6.9.6-2)

### Installed plugins and versions

- Colour Swatches	1.6.0
- Control Panel Body Classes	2.2.1
- CP Field Inspect	1.4.4
- Dumper	2.0.0
- Embedded Assets	2.11.2
- Empty Coalesce	1.0.6
- Expanded Singles	1.2.0
- Image Resizer	2.1.1
- Imager X	v3.6.1
- Mailchimp Transactional	1.0.3
- Minify	1.2.11
- Neo	2.13.11
- Obfuscator	1.0.4
- Reasons	2.3.0.1
- Redactor	2.10.8
- Retcon	2.4.2
- Sprig	1.13.0
- SuperSort	3.1.0
- Typogrify	1.1.19

brandonkelly:
Search for that error in `storage/logs/`. Does it include a stack trace? If so please post it.

romainpoirier:
Sorry for not being much clearer. 

After deleting the project, downloading it again from the repository, reset my MAMP, and I can now run `composer update` or `./craft update/update` or `composer install` all without any error.

But `./craft update/composer-install` is still returning this: 

> Performing Composer install ... error: array_merge(): Expected parameter 1 to be an array, bool given
> 
> Output:
> 
> 
> 

And here is the stack trace:
```
2022-07-04 09:19:22 [-][-][-][trace][yii\base\Application::bootstrap] Bootstrap with yii\shell\Bootstrap::bootstrap()
2022-07-04 09:19:22 [-][-][-][trace][yii\base\Application::bootstrap] Bootstrap with craft\queue\Queue::bootstrap()
2022-07-04 09:19:22 [-][-][-][trace][yii\base\Module::getModule] Loading module: utilities-module
[...]

PR #12288, opened 2022-11-10
Updated composer/composer version constraint to be more relaxed

markhuot:
### Description

I was having some install difficulties with some shared Symfony components. Specifically, I needed version 6 of `symfony/console` but because `composer/composer` was pinned to `2.2.15` and capped out at `^5.0` it couldn't be installed. Pulling this version constraint back to `^2.2` (allowing it to update all the way to the latest version 2.4.4) is maybe a bit _too_ relaxed, but figured I'd start there.

### Related issues

I looked at the `git blame` and the dep seems to be updating manually when needed. See: https://github.com/craftcms/cms/issues/11538. It'd be nice to allow it to update on its own _or_ bump this to the latest version, `2.4.4`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -35,7 +35,7 @@
     "ext-pdo": "*",
     "ext-zip": "*",
     "commerceguys/addressing": "^1.2",
-    "composer/composer": "2.2.15",
+    "composer/composer": "^2.2",
     "craftcms/oauth2-craftid": "~1.0.0",
     "craftcms/plugin-installer": "~1.5.6",
     "craftcms/server-check": "~2.1.2",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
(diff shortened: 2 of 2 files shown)


brianjhanson:
This is something I've been looking into because we'd like to be on the latest Composer as well. We bumped it up to [`2.3.5` back in April](https://github.com/craftcms/cms/commit/83e91724c77204004f96068233a80043f6b73727) but unfortunately something between 2.2 and 2.3 changed which causes errors when updating/installing/uninstalling plugins so [we had to revert the change](https://github.com/craftcms/cms/commit/4a63cd5132d43350257b3b2f64f5ff2d25be5356). I've looked into it a bit but still haven't tracked down what's causing the issue. 

Did you happen to run into any issues on the newer version? 

The error we saw was:
```
Error: Composer\Util\Filesystem::filePutContentsIfModified(): Argument #2 ($content) must be of type string, bool given, called in /var/www/html/vendor-root/composer/composer/src/Composer/Repository/FilesystemRepository.php on line 166
```

Which seemed to be because the [`InstalledVersions.php` file](https://github.com/composer/composer/blob/5b28754602d832f3487d92c5021f879007aece9e/src/Composer/Repository/FilesystemRepository.php#L166) would disappear for some reason.

PR #12294, opened 2022-11-11
Allow dynamic buttons to have copy-on-click enabled

engram-design:
For some fields ([Vizy](https://github.com/verbb/vizy), [Icon Picker](https://github.com/verbb/icon-picker)) we use Vue to render the field. However, this causes issues with jQuery due to the virtual DOM.

One thing I've noticed is that the click-to-copy of the field handle doesn't work as soon as Vue takes over rendering. This would likely be due to jQuery's event binding - while correctly bound - has bound an event to the real DOM, not the VDOM from Vue.

As such, this fix changes event-binding for these to be bound on the document, and scoped to the selector. You're using IDs so there should be no collisions, and it's [common practice](https://api.jquery.com/on/#direct-and-delegated-events) for dynamically-created elements to bind events in this way.

### DIFF
--- a/src/templates/_includes/forms/copytext.twig
+++ b/src/templates/_includes/forms/copytext.twig
@@ -22,7 +22,7 @@
 
 {% js %}
     {% block js %}
-        $('#{{ buttonId|namespaceInputId|e('js') }}').on('click', function() {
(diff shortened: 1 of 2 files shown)


engram-design:
Going to can this actually, I'll handle it myself with Vue.

PR #12296, opened 2022-11-11
Fixes: Graphql Mutation can't move element to structure

DTesch-Reem:


brianjhanson:
Thank you! I implemented something extremely similar in #12302

Issue #12289, opened 2022-11-10
[4.x]: Status condition rule has no effect for entry type field also missing on entry index page


vnali:
### What happened?

Status rule has no effect on results
**Steps to reproduce**
1. create an entry type (type1) with status rule set to enabled entries
2. add that field to a field layout for another entry type (type2)
3. try to create a type2 entry
4. open that entry type field 
5. you can see all type1 entries  (disabled, enabled, ...)

also maybe related issue:
Status condition rule is missing on entry index page

### Craft CMS version

4.3.1

### Installed plugins and versions

-

i-just:
Hi, thanks for reaching out. I'm able to replicate this and I've created a PR to for this.

brandonkelly:
It shouldn’t have been possible to select “Status” there to begin with, just like it’s not when filtering elements within an element index, or when defining custom sources. Fixed that for the next release.

brandonkelly:
Craft 4.3.2 is out with that fix.

PR #12298, opened 2022-11-11
fix for status condition being overwritten by criteria

i-just:
### Description
Status condition was being overwritten by criteria.

Example: 
If you have an entries selector field set to only accept "live" entries, the selector modal was showing all entries, not just "live" ones. It was happening because after condition was applied, `$criteria['status']` was overwriting it.

This change checks if there's a `StatusConditionRule` in the conditions, and if there is processes `$criteria['status']` in the following way:
- if `$criteria['status']` is `null` - show all conditioned results
- if `$criteria['status']` is not in the selected conditions - don't return any results as none match our criteria
- otherwise allow the `$criteria['status']` to affect things further

Takes "is not one of" condition into consideration.

### Related issues
#12289

### DIFF
--- a/src/controllers/ElementIndexesController.php
+++ b/src/controllers/ElementIndexesController.php
@@ -16,6 +16,7 @@ use craft\elements\actions\DeleteActionInterface;
 use craft\elements\actions\Restore;
 use craft\elements\conditions\ElementConditionInterface;
 use craft\elements\conditions\ElementConditionRuleInterface;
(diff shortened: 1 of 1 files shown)


brandonkelly:
Fixed this by preventing “Status” from being selectable in the first place.

If we want to bring Status rules back, we will need to come up with a way to update the Status toolbar menu, so it only includes status options that are allowed based on the Status condition rule.

PR #12302, opened 2022-11-11
Respect siteId in graphql structure mutations

### LINKED ISSUE #12291: [4.3.1]: Unable to move element in a structure - Mutation
### Description
When trying to move an Element with a mutation this error occurs.

Seems to happen on multisite-Setup because the siteId is not passed to this function:

```
/**
     * Get the related element.
     *
   * @param int $elementId
   * @return ElementInterface
   */
  protected function getRelatedElement(int $elementId): ElementInterface
  {
      $relatedElement = Craft::$app->getElements()->getElementById($elementId);

      if (!$relatedElement) {
          throw new Error('Unable to move element in a structure');
      }

      return $relatedElement;
  }
```

### Steps to reproduce

Querry:

```graphql
mutation saveEntry($id: ID = "", $title: String, $slug: String, $authorId: ID, $appendTo: ID) {
  save_bueroseiten_ch_default_Entry(id: $id, title: $title, slug: $slug, authorId: $authorId, appendTo: $appendTo, siteId: 4, appendToRoot: true) {
    title
    slug
    authorId
    parent {
      id
    }
  }
}
```

Variables: 

```json
{
  "title": "Gin and Tonic",
  "slug": "gin-tonic",
  "appendTo": 158427
}
```

### Expected behavior
Move specified element to structure position

### Actual behavior
Error Message: Unable to move element in a structure

### Craft CMS version

4.3.1

### PHP version

8.0.11

Unix

### Database type and version

mysql 8

### Image driver and version

_No response_

- Embedded Assets       embeddedassets        spicyweb/craft-embedded-assets   3.1.0   
    Feed Me               feed-me               craftcms/feed-me                 5.0.4   
    Incognito Field       incognito-field       mmikkel/incognito-field          1.3.0    
    Isolate               isolate               trendyminds/isolate              dev-craft-4  Yes        Yes    
    Matrix Field Preview  matrix-field-preview  weareferal/matrix-field-preview  dev-main   
    MatrixMate            matrixmate            vaersaagod/matrixmate            2.1.1      
    Redactor              redactor              craftcms/redactor                3.0.2     
    Redactor Anchors      redactor-anchors      utakka/redactor-anchors          1.4.0     
    Smith                 smith                 verbb/smith                      2.0.0         
    Super Table           super-table           verbb/super-table                3.0.5        
    Table Maker           tablemaker            verbb/tablemaker                 4.0.3        
    Typed link field      typedlinkfield        sebastianlenz/linkfield          2.1.4       
    Webhooks              webhooks              craftcms/webhooks                3.0.4

brianjhanson:
### Description
Respects `siteId` when performing structure mutations via graphql. One quirk: due to the order of operations new elements are saved and then "moved" when performing structure operations. This means if your structure is set to a propagation mode where elements are not directly propagated, and you supply a structure operation with a related element ID (`appendTo`, `prependTo`, `insertBefore` or `insertAfter`), an error will be shown but the element will have been created. 

So for example, I have two sites:
| Site ID | Element ID |
|:---|:-----|
| 1 | 150 |
| 2 | 200 |

If I supply a query with a mismatch:

```json
{
  "siteId": 1,
  "appendTo": 200
}
```

The element will get created in site 2 at the end of the list. This is consistent with the current behavior but can be a little confusing.

### Related issues
Fixes #12291

### DIFF
--- a/src/gql/base/StructureMutationTrait.php
+++ b/src/gql/base/StructureMutationTrait.php
@@ -30,19 +30,20 @@ trait StructureMutationTrait
         }
 
         $structureService = Craft::$app->getStructures();
+        $siteId = $arguments['siteId'] ?? null;
 
         if (!empty($arguments['prependTo'])) {
-            $structureService->prepend($structureId, $element, $this->getRelatedElement($arguments['prependTo']));
+            $structureService->prepend($structureId, $element, $this->getRelatedElement($arguments['prependTo'], $siteId));
         } elseif (!empty($arguments['appendTo'])) {
-            $structureService->append($structureId, $element, $this->getRelatedElement($arguments['appendTo']));
+            $structureService->append($structureId, $element, $this->getRelatedElement($arguments['appendTo'], $siteId));
         } elseif (!empty($arguments['prependToRoot'])) {
             $structureService->prependToRoot($structureId, $element);
         } elseif (!empty($arguments['appendToRoot'])) {
             $structureService->appendToRoot($structureId, $element);
         } elseif (!empty($arguments['insertBefore'])) {
-            $structureService->moveBefore($structureId, $element, $this->getRelatedElement($arguments['insertBefore']));
+            $structureService->moveBefore($structureId, $element, $this->getRelatedElement($arguments['insertBefore'], $siteId));
         } elseif (!empty($arguments['insertAfter'])) {
-            $structureService->moveAfter($structureId, $element, $this->getRelatedElement($arguments['insertAfter']));
+            $structureService->moveAfter($structureId, $element, $this->getRelatedElement($arguments['insertAfter'], $siteId));
         }
     }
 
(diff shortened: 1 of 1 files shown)


brandonkelly:
What error would be shown? Where in the control panel can you append an element to another element across two different sites?

brianjhanson:
You'll get the "Unable to move element in a structure" error thrown [from over here](https://github.com/craftcms/cms/blob/77cd870b9e8d7742222234f35c62430ba1751887/src/gql/base/StructureMutationTrait.php#L62). It's not consistent with the current CP behavior, but rather with the current graphql behavior because we [complete the save operation before attempting to update the position](https://github.com/craftcms/cms/blob/ca503dda81ffb61aba7540d86fbbcf581e668d07/src/gql/resolvers/mutations/Entry.php#L71-L72) of the element.

I think ideally we would test the entire operation and throw an error if it's not possible. I figured that would have to be a 5 change since 4 will currently create the element where it can and just fail to move it to the desired location. I can also see an argument for this being the desired behavior since graphql is properly returning an error for the move operation that failed, but not for the create operation that was successful.

brandonkelly:
When you change an entry’s parent, we’ll look for the new parent entry across all sites:

https://github.com/craftcms/cms/blob/83ecb4b7faf449f348f3e23e1981d72001a6ea4e/src/elements/Entry.php#L1601-L1605

So it seems like a better fix for #12291 would be to just make `getRelatedElement()` consistent with that, and start passing `'*'` to `getElementById()`’s `$siteId` argument. Then the error will go away entirely.

PR #12313, opened 2022-11-14
Darken input borders to meet minimum contrast for UI components

gcamacho079:
### Description
Inputs (text, radio, checkbox, etc.) should meet 3:1 color contrast against page background per [1.4.11: Non-Text Contrast](https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast.html). This PR introduces a new CSS variables for input border and input border color to manage field borders throughout the control panel.

### Related issues
Resolves DEV-515, DEV-322, DEV-317


brandonkelly:
That feels really harsh to my eyes. Would it be OK to only do this for [`prefers-contrast`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-contrast)? Otherwise, maybe via a new accessibility preference?

gcamacho079:
> That feels really harsh to my eyes. Would it be OK to only do this for [`prefers-contrast`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-contrast)? Otherwise, maybe via a new accessibility preference?

After some additional research, a new accessibility preference sounds like a good idea for this change, as high contrast modes can have usability implications for some users. However, we do have to ensure that the path to enable this setting does meet minimum contrast requirements. This path includes:

1. Login page
2. Opening the My Account menu
3. Clicking the link to go to preferences
4. Going to the Preferences tab
5. Enabling the high contrast setting

I’ve created #12517 to ensure the Account menu meets the minimum contrast for UI components; however, if we’re using a checkbox-style option (i.e., “Enable high contrast mode”), checkbox borders would still need to be darkened. Thoughts?

brandonkelly:
Think this goes back to coming up with some sort of account setup wizard, which could be presented to users when they first log into the control panel, which presents them with all the accessibility preferences. Accounts can default to having any relevant accessibility settings enabled by default to ensure the wizard meets the necessary criteria.

brandonkelly:
So for the purposes of this PR, let’s just add the preference and not worry about whether the screens leading up to it are compliant.

gcamacho079:
@brandonkelly user preference added! For simplicity, I reverted some changes to remove the extra `--input-border` variable. I thought having too many stylistic/layout changes tied to the higher contrast view might be difficult to maintain, so I retained the changes I made in the `.meta` container for all users, so there’s now a faint border around (and spacing between) inputs in this area for low-contrast users.

gcamacho079:
@brandonkelly I was revisiting [this PR about disabled element contrast inside indexes](https://github.com/craftcms/cms/issues/10621). I wonder if this “Increase contrast” setting would be a good candidate for looping in the "disabled elements get italicized text” update. To clarify, we can create some accessibility documentation in the docs that outline all the UI changes bundled with that user setting; this would also help us achieve the ATAG requirement for documenting accessibility features.

Issue #12338, opened 2022-11-17
[4.3.2]: Sort filter no longer accepts closure


johndwells:
### Description

Given this snippet of twig:

```
{% for e in exhibitionsQuery|sort((a, b) => b.startDate|date('Ymd') <=> a.startDate|date('Ymd')) %}
    {{ basicExhibitionLink( e, 'full' ) }}
{% endfor %}
```

Results in a "TypeError" being thrown:

> strtolower(): Argument #1 ($string) must be of type string, Closure given

And it points to this line (https://github.com/craftcms/cms/blob/develop/src/web/twig/Extension.php#L492):

```
if (strtolower($arrow) === 'system') {
...
```

The entire sortFilter function appears to have been added in 4.3.2: if (strtolower($arrow) === 'system') {

### Craft CMS version

4.3.2

### PHP version

8.1

### Installed plugins and versions

-

michaelrog:
Related: https://github.com/craftcms/cms/pull/12335

i-just:
Thanks for reporting! There's a PR for it for both v3 and v4 now.

brandonkelly:
Sorry about this! Craft 3.7.61 and 4.3.3 are out with a fix for this.

PR #12339, opened 2022-11-17
twig sort filter check if arrow is a string

i-just:
### Description
Check if `$arrow` passed to the `sortFilter()` is a string before trying to lower-case it.

### Related issues
#12338


brandonkelly:
Thanks! The Craft 3 PR resolved this for Craft 4 as well.

PR #12355, opened 2022-11-22
Feature/component system

brianjhanson:
Introduces a UI component system heavily inspired by [Symfony's UX Twig Components](https://symfony.com/bundles/ux-twig-component/current/index.html) and work by @leevigraham. 

The system provides twig templates backed by PHP classes which will primarily allow us to generate better documentation about how they should be used. It's also my hope that we'll be able to provide accessible components that are easy to use for the community. Ideally providing ways to extend without harming accessibility. 

This is a fairly revised draft, but still a draft. I've added a handful of examples (mostly input elements) but am hoping to add more over time and eventually hope to replace internal rendering of components with this new method.

### A quick example

Let's walk through building a basic button component. To create a button, you'll need two files. First, we'll create a nearly empty PHP class for the button. 

```php
use craft\ui\attributes\AsTwigComponent;  
use craft\ui\components\BaseComponent;  
  
#[AsTwigComponent('button')]  
class Button extends BaseComponent  
{  
}
```

A few things to call out here. The first is our `AsTwigComponent` attribute. That attribute is how we tell Craft what we want to call this component (`button` in this case). You can also use it to set some metadata associated with the component. 

Next, we need to create a twig file to render the button. In this draft Craft will look for components in the `templates/_ui` directory, so we'll create a `templates/_ui/button.twig` file (Note: that the filename should match our attribute name).  We'll keep that twig file nice and simple for now. 

```twig
<button>This is a button</button>
```

The last thing we need to do is tell Craft that this component exists now. That's done by adding the PHP class to the array in the `craft\services\Ui` class. 

```php
    private function getAllComponentTypes()
    {
        return [
            Button::class
        ];
    }

```

With those pieces in place, you can now call `{{ component('button') }}` within the CP  to render our new component. 

When `component` is called, we'll retrieve the `Button` class, map the passed data onto the class properties and pass all that data to the twig template. 

That's all well and good, but admittedly the component isn't terribly useful just yet. Let's add some properties so we can change the output based on options we pass in. 

Components will always get a somewhat magical `attributes` property which is comprised of all the data passed in to the component that wasn't consumed as a property. That means we can update our button just a little bit to start changing the output.

First we need to update our component slightly

```twig
<button {{ attributes }}>{{ label }}</button>
```

Next, we need to update our PHP class to be aware of the label[1]. 

```php
use craft\ui\attributes\AsTwigComponent;  
use craft\ui\components\BaseComponent;  
  
#[AsTwigComponent('button')]  
class Button extends BaseComponent  
{  
  public ?string $label = null;
}
```

Then we can call our component

```twig
{{ component('button', {label: "A test button", "data-test": "a-property-name"}) }}
```

And the HTML output will be

```html
<button data-test="a-property-name">A test button</button>
```

It's worth checking out the Symfony UX docs as many of the features there are also supported in this implementation (lifecycle methods like `preMount`, `mount` and `postMount` as well as the embedded component syntax). 

This draft also includes a bit of an idea about how validation could be added to components in the future (by having `BaseUiComponent` extend `Model`) but that support isn't extremely fleshed out. Right now I'm thinking the validation would be called after `mount` but before `postMount` so any errors could be dealt with in a `postMount` method. 

There are a few deviations from the Symfony approach that I'll need to document:

- how the `class` attributes & defaults are handled
- Passing parent `props` down to the child. This was done primarily to help with composition but I'm not totally happy with it yet. 

---

[1] It might be tempting to add  a property to our PHP class for every variant we want to support, but since properties will "fall" through to the `attributes` variable, in many cases fewer options is better. However if you want to see how a button may actually be implemented there's an example within this PR.


michtio:
Even tho it's written out nicely as explanation in the PR, but I'm wondering if this will add a real actual benefit to the CMS?
I currently feel that the way Craft decouples front-end from the actual structure is very nice to have. It allows us when we create plugins or extend the CP to use the technologies we are comfortable with (e.g. Vue / Tailwind CSS)

If I would need a button in Vue that looks the same like the Craft CP offers me, I wouldn't be able to use this component approach either way. Also I'd be a bit stuck with my custom styling or applying tailwind css classes where I see fit (e.g. an override where really necessary). 

Correct me if I'm wrong tho. But to go and create PHP classes for components, which actually could be macros or twig templates that are easy to include, even tho you'd still need to create a twig template for them... aren't we adding a layer of complexity into it?

leevigraham:
I don't think components have to be used for every ui element. I also thought there could be some kind of fallback directly to a twig template if the Class doesn't exist.

In your button example, if the button was a vue component I probably wouldn't create a PHP component for it and just use the vue component directly.

michtio:
Thanks @leevigraham. I mean I am definitely not opposed to the idea, I'm just trying to understand it well enough, if it's optional and the fallback is in place then great. Just am thinking to avoid additional complexity for creating UI elements.

brianjhanson:
Thanks for the feedback @michtio! These are all valid points, but just wanted to respond to a few specific things. 

> I'm wondering if this will add a real actual benefit to the CMS?	

The primary benefit I'm going for with this is more around documentation at the moment. Right now, the macros we have are pretty opaque. The only way to determine what arguments they accept, or what they will output is by reading through the code. This makes it difficult for plugin authors to use them easily. Twig doesn't have anything like PHPDoc (that I'm aware of) so the only way to document how exactly they work would be manually, which is both a big upfront investment, and hard to maintain long term. 

Backing twig templates with PHP allows us to use PHPDoc to better document, and maintain documentation around these components. It also allows us to move some business logic into PHP which is especially nice when things get complicated.

> It allows us when we create plugins or extend the CP to use the technologies we are comfortable with (e.g. Vue / Tailwind CSS)

We're not planning to take away abilities to customize the CP, or to use your own technologies to do so. We're hoping that by providing good, accessible components of our own, plugin authors will be able to extend the CP more easily, while maintaining more visual and functional consistency. 

> If I would need a button in Vue that looks the same like the Craft CP offers me, I wouldn't be able to use this component approach either way.

This is correct at the moment. We do also provide a JS method to create buttons, but not currently anything powered by Vue. 

> Also I'd be a bit stuck with my custom styling or applying tailwind css classes where I see fit

This approach still allows you to add classes however you'd like, but you're correct in that you're not able to remove classes that are critical to the appearance of the button (the `btn`, `disabled`, etc. classes). The main reason for that is to keep visual consistency. 

> aren't we adding a layer of complexity into it?

Yeah, that's  a fair point, but I'm still hopeful that the outcome will be to make it easier for plugin authors to extend the CP while maintaining visual consistency. That being said, there will always be a point where plugins are too complex to be supported by something like this, especially in the 1.0 iteration.

michtio:
Thank you for the very detailed reply @brianjhanson , much appreciated. I definitely see some benefits to it as how you explain it. Just still a bit worried about the added possible complexity of it. 👍🏻

khalwat:
Rather than inventing your own component system, what about leveraging an existing system like the Symfony UX Twig Components that you reference?

Basing things on standards that are not unique to Craft has a variety of benefits. I realize that Symfony UX Twig Components may not currently be a perfect fit, but wouldn't contributions to that project be a useful direction for everyone?

PR #12386, opened 2022-11-28
Announce "signing in" state on login screen

gcamacho079:
### Description
Adds a new Garnish component for a [Multi-Function Button](https://adrianroselli.com/2021/01/multi-function-button.html) that:

- Announces loading state via a live region located before the `button` element
- Updates live region for success and failure states via `data-` attributes attached to the button element

### Related issues
Resolves DEV-1047


gcamacho079:
Back to in-progress

Issue #12364, opened 2022-11-23
[4.x]: 'Show unpermitted volumes' not working when Sources are individually selected on an Asset field


tommysvr:
### Description
Volumes a User doesn't have permissions for are still shown with 'Show unpermitted volumes' unchecked when Sources are individually selected on an Asset field.

### Steps to reproduce
1. Create three Asset volumes (Volume 1, Volume 2, Volume 3)
2. Give a User access to only Volume 2
3. Create an Asset field with 'Sources' set to Volume 2 and Volume 3
4. Leave 'Show unpermitted volumes' unchecked
5. Click 'Add an Asset' on an Entry and see the User can see both Volume 2 and 3, even though they only have access to Volume 2
6. Now change 'Sources' on the Asset field to 'All', and only Volume 2 will show

For further unexpected behaviour, now check 'Restrict assets to a single location' and set the Asset Location to Volume 1, and the Add an Asset / Upload Files buttons will disappear completely, with no feedback as to why.

### Expected behavior
The User should only see the volume they have access to (Volume 2) in each case.

### Actual behavior
The User can see Volumes they don't have access

### Craft CMS version

4.3.3

### PHP version

8.1

### Installed plugins and versions

-

brandonkelly:
Fixed for the next release via #12366.

PR #12388, opened 2022-11-29
explanation why user can't see the add & upload buttons

i-just:
### Description
If Assets field sources are set to volumes for which user doesn't have permissions, make sure a message shows instead of just not showing any buttons at all.

### Related issues
#12364


brandonkelly:
I don’t think this change is necessary, and removed it from the Craft 4 PR as well (009a64bb822816c6bbefc653c910a9d2946fc4c5).

PR #12396, opened 2022-11-29
[4.x] Update `Asset::getMimeType()` to respect transforms

brianjhanson:
### Description
`Asset::getMimeType()` will now return the mime type of the transform rather than the image when a transform is set. 

### Related issues
#12269


brandonkelly:
Merged `3.8` --> `4.4` via 15c43eb2b0db07b1f91f17ea2e58147f7292ff74

PR #12399, opened 2022-11-29
Add search query param

brianjhanson:
### Description
Stores a query param in the URL when a search is performed. 

### Related issues
#8942


src/web/assets/cp/src/js/BaseElementIndex.js
```diff
@@ -276,7 +276,7 @@ Craft.BaseElementIndex = Garnish.Base.extend(
       });
 
       // Auto-focus the Search box
-      if (!Garnish.isMobileBrowser(true)) {
+      if (!Garnish.isMobileBrowser(true) && !Craft.getQueryParam('search')) {
```
brianjhanson, line 279: @gcamacho079 when I was testing it felt a little weird to auto-focus the search box if it already had a query in it, but let me know if not autofocusing is weirder

gcamacho079: Great catch @brianjhanson; it's definitely a strange behavior as it makes too many assumptions about the user's intent when landing on the page. We actually have a [PR open to remove the auto-focus](https://github.com/craftcms/cms/pull/12332/files#diff-04a6e5588e11e64220e76ed0bc020ccd3692cc26288157ad1450d997cde96f7eL278) from most of the CP, including this very spot

PR #12407, opened 2022-11-30
System info reflect current site

brianjhanson:
### Description
Updates the system info (in the top left corner) to update with the selected site. 

### Related issues
#12402


gcamacho079:
Having the system name itself link to the selected site seems like it may not adequately describe the link's purpose (see [Understanding Success Criterion 2.4.4: Link Purpose In Context](https://www.w3.org/WAI/WCAG21/Understanding/link-purpose-in-context.html)). It may be beneficial to have the system name be static text, and add a separate link that makes it clear what the exact destination is. Interested to hear your thoughts 🤔 

The sidebar navigation as viewed from the Assets page, where 'Site B' has been selected as the active site. A link to view 'Site B' in an new window has been added immediately after the system name.

brianjhanson:
Closing this for the moment as we'd like to think about it a bit more.

PR #12408, opened 2022-12-01
Scaffolding CLI commands

brandonkelly:
### Description

Adds commands for scaffolding various system components

#### Plugins/modules
- [x] `make plugin`
- [x] `make module`

#### Controllers
- [x] `make command`
- [x] `make controller`

#### System components
- [x] `make asset-bundle`
- [x] `make element-type`
- [x] `make element-action`
- [x] `make element-condition-rule`
- [x] `make element-exporter`
- [x] `make field-type`
- [x] `make filesystem-type`
- [x] `make queue-job`
- [ ] `make service`
- [ ] `make utility`
- [ ] `make validator`
- [x] `make widget-type`

#### GraphQL components
- [ ] `make gql-argument`
- [ ] `make gql-directive`
- [ ] `make gql-interface`
- [ ] `make gql-mutation`
- [ ] `make gql-query`
- [ ] `make gql-resolver`
- [ ] `make gql-type`

#### Supporting classes
- [ ] `make error`
- [ ] `make event`
- [x] `make generator`
- [ ] `make migration`
- [ ] `make model`
- [ ] `make record`
- [ ] `make test`

### Related issues

- #8908


michaelrog:
Super excited for this! Some ideas:

### 1. Keep the CLI namespace tidy

The CLI namespace is getting pretty busy... What about specifying the class type via argument, rather than having a separate action for each?

```
./craft make plugin ...
./craft make module ...
```

### 2. Dev features in `require-dev`

It'd be cool to wrap this functionality in a separate self-bootstrapping module, and file it under `require-dev`, so we can opt out of bloating production installs with unneeded weight...

### 3. Customizable namespace

I frequently use domain-driven folder structure, rather than the folder-per-component-type convention... so it'd be nice to have an option for specifying a custom namespace if I want to depart from the default convention.

khalwat:
> It'd be cool to wrap this functionality in a separate self-bootstrapping module, and file it under require-dev, so we can opt out of bloating production installs with unneeded weight...

Yeah having this in a separate stand-alone CLI package that is `require-dev`'d by Craft would be ideal because:

* As mentioned, you don't have to install it on production
* The separate CLI command could then be globally installed, and run from anywhere, with or without a working Craft install
* When used in the context of a Craft install, the `craft make` command would just pass through to the CLI command
* The CLI command could be context-aware, in terms of where the current directory is, and namespace things based off of this for modules/plugins, or Craft itself
* Having to be in the root directory of a working Craft install to run the command would negate or make really inconvenient taking advantage of the context of the current directory when creating things
* You'd have to either do gross things like `../../../php craft make controller` when in a plugin vendor dir, or something like `craft make controller --context=nystudio107/seomatic` both of which are just not fun

I put a lot of thought into it when I was considering doing it myself, and I think there are a whole lot of advantages to having the CLI command not be coupled to a working Craft install.

brandonkelly:
> The CLI namespace is getting pretty busy... What about specifying the class type via argument, rather than having a separate action for each?

Yeah that’s reasonable – [done](https://github.com/craftcms/cms/pull/12408/commits/b2550b81dbe1f08987458736cc88646a8e06075e).

The main `help` command will now only show a single `make` command (`make/generate`):

```sh
- make                                               Generates the scaffolding for a new system component.
    make/generate (default)                          Generates the scaffolding for a new system component.
```

And the help summary for `make` will list out each of the available generators within its `type` param docs:

```sh
USAGE

craft make [type] [...options...]

- type: string
  The type of component to generate. Options include:
  - plugin: Creates a new Craft plugin.
  - ...
```

(The list is dynamic, so will include any plugin-supplied generators.)

Regarding where it lives, the original plan was to build it within a standalone `craftcms` CLI tool, and I definitely see some advantages with that approach – especially the ability to generate things from within a plugin’s source directory as @khalwat mentioned.

But ultimately decided against it for a couple reasons:

- From a maintainability perspective, it’s going to be a lot simpler and cleaner to do this directly in Craft, where we can load/parse/reflect existing classes to aid in generation.
- No idea how we’d make it extensible. For example, a form plugin might want to register a form element class generator. That plugin probably isn’t going to be installed within another plugin’s own source directory, so the tool wouldn’t know anything about it. (And I don’t want a bunch of third party plugin developers PRing their own custom generators to the `craftcms` command.)

> You'd have to either do gross things like `../../../php craft make controller` when in a plugin vendor dir, or something like `craft make controller --context=nystudio107/seomatic` both of which are just not fun

Planning on:

```sh
craft make controller --plugin=seomatic
```

Doesn’t seem so bad to me.

> It'd be cool to wrap this functionality in a separate self-bootstrapping module, and file it under `require-dev`, so we can opt out of bloating production installs with unneeded weight...

Worth considering, but let’s see how big this ends up being first. I don’t expect it will be too significant, in which case it probably wouldn’t be worth the added complexity.

> I frequently use domain-driven folder structure, rather than the folder-per-component-type convention... so it'd be nice to have an option for specifying a custom namespace if I want to depart from the default convention.

Yep the location of component classes will be customizable.

khalwat:
> Planning on:
> craft make controller --plugin=seomatic
> Doesn’t seem so bad to me.

What if I have several custom site modules (not an uncommon situation), how do I specify the namespace to make a new controller for one of them?

What if I want to make a new Controller for Craft itself, for a PR (or for your own internal usage), how would that work?

> From a maintainability perspective, it’s going to be a lot simpler and cleaner to do this directly in Craft, where we can load/parse/reflect existing classes to aid in generation

This makes sense. Is it all going to be done with `heredoc`s mixed with PHP code, or will there be some kind of templating system?

> No idea how we’d make it extensible. For example, a form plugin might want to register a form element class generator. That plugin probably isn’t going to be installed within another plugin’s own source directory, so the tool wouldn’t know anything about it. (And I don’t want a bunch of third party plugin developers PRing their own custom generators to the craftcms command.)

I think it makes sense of the domain of the code generator is Craft itself, to help people scaffold things out while extending Craft. I'm not sure many people are going to want to extend it themselves?

brandonkelly:
> What if I have several custom site modules (not an uncommon situation), how do I specify the namespace to make a new controller for one of them?

```sh
craft make controller --module=my-module-id
```

> What if I want to make a new Controller for Craft itself, for a PR (or for your own internal usage), how would that work?

```sh
craft make controller --app
```

> > From a maintainability perspective, it’s going to be a lot simpler and cleaner to do this directly in Craft, where we can load/parse/reflect existing classes to aid in generation
> 
> This makes sense. Is it all going to be done with `heredoc`s mixed with PHP code, or will there be some kind of templating system?

Planning on creating a generic Craft component type generator, which can be configured with a base class/interface plus a handful of methods that should be pulled in. But I haven’t actually started exploring the idea yet, so we’ll see how successful it is.

> > No idea how we’d make it extensible. For example, a form plugin might want to register a form element class generator. That plugin probably isn’t going to be installed within another plugin’s own source directory, so the tool wouldn’t know anything about it. (And I don’t want a bunch of third party plugin developers PRing their own custom generators to the craftcms command.)
> 
> I think it makes sense of the domain of the code generator is Craft itself, to help people scaffold things out while extending Craft. I'm not sure many people are going to want to extend it themselves?

Just looking at our own first party plugins, we’d probably want to register generators for:

- Commerce purchasables
- Apple News channels
- Feed Me importers

Wouldn’t surprise me to see some third party plugins get on board too.

khalwat:
FWIW, I've used this PHP code generator to great effect... it's pretty glorious:

https://github.com/nette/php-generator

brandonkelly:
Had a need to pull in [PHP Parser](https://github.com/nikic/PHP-Parser) which is ~2MB, so decided to go with @michaelrog’s idea and separate this out to a separate package – [craftcms/generator](https://github.com/craftcms/generator). Will continue plugging away from there!

brandonkelly:
For posterity, Craft Generator 1.0 has been released!

Announcement: https://craftcms.com/blog/craft-generator
Packagist: https://packagist.org/packages/craftcms/generator

PR #12411, opened 2022-12-01
Add visible labels on login screen and remove placeholder text

gcamacho079:
### Description
Adds persistent labels on the login screen


brandonkelly:
Can we keep the existing design for the most part, by overlaying the labels over the field, and animating them out of the way a bit on focus (like https://codepen.io/katrina-isabelle/pen/PQBqyq)?

gcamacho079:
> Can we keep the existing design for the most part, by overlaying the labels over the field, and animating them out of the way a bit on focus (like https://codepen.io/katrina-isabelle/pen/PQBqyq)?

Absolutely. I also think that would be great to do for v5 in places like the condition builder. I'll create a new branch with the requested changes for the login screen.

PR #12415, opened 2022-12-02
Convert element index actions menu into disclosure menu

gcamacho079:
### Related issues
Resolves DEV-382


brandonkelly (changes requested):
The action menu isn’t currently closing when an action is selected, which should be the case:

https://user-images.githubusercontent.com/47792/211841261-c0b5859f-43be-4e41-8934-7478287c7965.mp4

PR #12432, opened 2022-12-08
email/username uniqueness shouldn't factor inactive users

timkelty:
@brandonkelly Once I got into this, I thought it made sense to scale back a bit, especially if we want this for 4.4.

Currently, this PR:
- fixes unique validation for `email`, `username`
- Moves "Delete…" to the "Save" dropdown (like other elements)
- Moves the `#action-menubtn` gear menu next to the save button
  - we originally said to the right of save, but I thought it felt weird for the rightmost thing to not be save, so I put it to the left. Also felt more similar to extra buttons on entry detail.
- Returns validation errors to the template (eg, when attempting to activate with an occupied username)

What it _doesn't_ do:
- Deprecate `\craft\services\Users` methods/events and move those to `\craft\elements\User`
  - Provide a better method for changing User statuses with validation, without going through the record
- Post entire form for actions that save the user & return errors (eg activate user)
  - This is what we discussed, but thought it felt weird once it was in place. Returning errors and just requiring you to save changes and then attempt the action again felt fine.

## User queries that relied on email/username being unique
`\craft\services\Users::getUserByUsernameOrEmail` was the only one I found, which is now only querying credentialed users. This is arguably a breaking change, but is better than the alternative of that query returning a random result between 2 users with the same email/username. We should check with @craftcms/commerce to see if they might be affected…

CleanShot 2023-02-20 at 21 26 13@2x


src/services/Users.php
```diff
@@ -790,6 +791,20 @@ public function activateUser(User $user): bool
...
+        $user->verificationCode = null;
+        $user->verificationCodeIssuedDate = null;
+        $user->invalidLoginCount = null;
+        $user->lastInvalidLoginDate = null;
+        $user->lockoutDate = null;
+
+        if (!$user->validate()) {
+            return false;
+        }
+
```
timkelty, line 807: Validate up front, so we can bail if there is already an active user with this email/username.

tests/unit/elements/UserElementTest.php
```diff
@@ -81,6 +81,27 @@ public function testValidateUnverifiedEmail(): void
...
+        $user = new User([
+            'active' => true,
+            'email' => $this->inactiveUser->email,
+            'username' => $this->inactiveUser->username,
+            'unverifiedEmail' => $this->inactiveUser->unverifiedEmail,
+        ]);
+
+        $this->tester->saveElement($user);
+
+        self::assertFalse($user->hasErrors());
```
timkelty, line 95: User is able to be saved, even though an inactive user has the same email/username.

tests/unit/elements/UserElementTest.php
```diff
@@ -81,6 +81,27 @@ public function testValidateUnverifiedEmail(): void
...
+
+        Craft::$app->getUsers()->activateUser($this->inactiveUser);
+
+        self::assertTrue($this->inactiveUser->hasErrors('email'));
+        self::assertTrue($this->inactiveUser->hasErrors('username'));
+
```
timkelty, line 101: Inactive user can now not be activated because there is an active user with the same email/username.

src/controllers/UsersController.php
```diff
@@ -749,9 +750,17 @@ public function actionActivateUser(): Response
...
+                Craft::t('app', 'Successfully activated the user.'),
+                $userVariable,
+            );
         } else {
-            $this->setFailFlash(Craft::t('app', 'There was a problem activating the user.'));
+            return $this->asModelFailure(
+                $user,
+                Craft::t('app', 'There was a problem activating the user.'),
+                $userVariable,
+            );
```
timkelty, line 763: Return validation errors to the template

PR #12436, opened 2022-12-09
Feature/979 copy field values

i-just:
### Description
On the entry/category edit page (elements that support autosaving drafts): 
- implemented a new button that allows you to copy all translatable field values from the selected site;
- translatable field icon now triggers a HUD, which contains a form that allows you to copy just that field’s value from the selected site;
- copy action is applied to the top-level fields (e.g. entire matrix field)
- plugins need to implement `CopyableFieldInterface` to use this feature

TODO: 
- [x] switch the sites selector in the copy form (in the hud) to a disclosure menu

**IMPORTANT**: once this feature is released, an update to `craftcms/html-field` needs to be released; [PR 6](https://github.com/craftcms/html-field/pull/6)

### Related issues
[DEV-979](https://linear.app/craftcms/issue/DEV-979/copy-field-values-from-other-sites)


i-just:
@brandonkelly as per our chat yesterday, `4.5` is merged into this, conflicts resolved (not many) and tested that it still works like it did in December. Please LMK if you have any questions or if it still doesn’t want to work for you.

benjamindavid:
I’m sorry if these things have already been discussed or ruled out, but here’s some feedback on this:
    - Explain somewhere that only translatable fields will be copied.
    - If there’s no translatable fields, or if copying won’t change anything, the copy button could be disabled or invisible
    - Add a blue flash for fields that have been copied and currently visible
    - The top area of the page is already a bit crowded with buttons, maybe we could try to find a shorter label for the copy button. Longer term we could clean up that UI with a three dot menu or something like that?

image

benjamindavid:
Probably another dumb idea but let's throw it anyway:
image

i-just:
Thanks @benjamindavid!

> Explain somewhere that only translatable fields will be copied.

good point! it never occurred to me that it might not be obvious; I added the “Copy content from site” heading and “Only translatable field values will be copied if their content differs from the current.” instructions to the hud.

> If there’s no translatable fields, or if copying won’t change anything, the copy button could be disabled or invisible

There’s already a “Nothing to copy” message showing if you trigger the copy action (for a single field or the entire entry), and the translatable content on both sites is the same;
Disabling or hiding the button could be tricky as we don’t know which site the user wants to copy from. If they have many sites on page load, we’d have to compare the values for the current site against all other sites. At the moment, we only do this check on demand and for one site - the one user said they want to copy from. I’m a bit worried about performance implications, but happy to be told it’s worth it or that I’m overthinking this.  
Given the changes from your first point + the “Nothing to copy” message - could that be enough now?


> Add a blue flash for fields that have been copied and currently visible

I sent you a question about it because I’m not sure I get what you’re referring to.

> The top area of the page is already a bit crowded with buttons, maybe we could try to find a shorter label for the copy button. Longer term we could clean up that UI with a three dot menu or something like that?

I changed the button to read “Copy content” and made the changes described in point 1; LMK, if you think it’s not enough and would still like to move it to the sites/drafts/revisions menu
 
Screenshot 2023-07-20 at 12 17 56

benjamindavid:
That's perfect, thanks for the changes @i-just!

Regarding the blue marker, I think it's more for cases where the user would already have made changes to the entry, with fields showing blue markers, copying the content from another site wouldn't change anything to those fields that would remain blue. So the flash would mostly be there to put emphasis on the fact that it's the copy action that changed the fields. It's probably overkill for now, we can discuss that later if people struggle.

i-just:
Targeting the `5.0` branch now. 
I’m putting it back into draft to work on minor compatibility changes & to test.

brianjhanson:
Closing in favor of #14056

Issue #12457, opened 2022-12-14
[4.x]: Date field condition ':notempty:' throws error


nilsenpaul:
### Description

In a custom element source, with the condition of a date field set to 'empty' or 'not empty', an error gets thrown:

`craft\fields\conditions\DateFieldConditionRule::elementQueryParam(): Return value must be of type ?array, string returned`

### Steps to reproduce

1. Create a custom element source on an element with a date field
2. Set a condition 'not empty' or 'empty' on the date field
3. Try to load the element index. It doesn't work (for me)

`elementQueryParam` must return null or an array, but it gets its input from `BaseDateRangeConditionRule`, which returns a string if 'empty' or 'not empty' was requested.

### Craft CMS version

4.3.4

### PHP version

8.1

i-just:
Thanks for reporting! I created a PR to fix this.

brandonkelly:
This is fixed for the next release via #12461.

brandonkelly:
Craft 4.3.6 is out with the fix for this.

PR #12461, opened 2022-12-15
empty and not empty date condition fix

i-just:
### Description
Fix for date field condition `:notempty:` and `:empty:` throwing a return value error: `craft\fields\conditions\DateFieldConditionRule::elementQueryParam(): Return value must be of type ?array, string returned`.

### Related issues
#12457


mathieuser:
Just so you know: I'm getting this error as well when I use the before and after date condition. Tested it locally and live on different installs.

brandonkelly:
@mathieuser Yeah same bug.

brandonkelly:
@mathieuser Craft 4.3.6 is out with this.

Issue #12459, opened 2022-11-15
Issue adding in fields to products


davidcraigmyle:
### Description
Hello,

I'm testing out the Shopify plugin and I've tried to add some fields to the products.
I had one tab with a title working fine but when I added a field using the Neo plugin, a tab with a field for the SEO plugin, and a tab with a couple of summary fields.
When I saved the page all of the tabs duplicated and then I got an error because of the duplicates.

'SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '83-44' for key 'idx_wlcbhmxrqizbeglfuifneaurqdbsemupfwzj'
The SQL being executed was: INSERT INTO `fieldlayoutfields` (`layoutId`, `tabId`, `fieldId`, `required`, `sortOrder`, `dateCreated`, `dateUpdated`, `uid`) VALUES (83, 1780, 44, 0, 0, '2022-11-15 16:24:56', '2022-11-15 16:24:56', '809f2855-bba5-43a8-967c-3f2d8ddbdce5')'

After trying to remove all the fields, they duplicated again and now I can't save any fields or use the plugin at all

### Steps to reproduce

1. Add new tabs to products under Shopify -> Settings -> Products
2. Add fields including a Neo fields nad SEO field

### Additional info

- Craft version: 4.3.1
- PHP version: 8.0.8
- Database driver & version: MySQL 5.7.34
- Plugins & versions: Neo - 3.5.2, SEO - 4.0.3, Shopify - 3.0.0.1

![Screenshot 2022-11-15 at 16 20 00](https://user-images.githubusercontent.com/19872210/201973529-ac73245b-2d09-41ff-b5ca-235ed7e08994.png)
![Screenshot 2022-11-15 at 16 29 42](https://user-images.githubusercontent.com/19872210/201973753-d94f420f-df85-4f58-a548-94f37aef9ecd.png)

brandonkelly:
This ended up benig mostly a Craft CMS bug, which is now fixed for the next release (2947da192efe8a78bd617565df0239baba46f931).

It was exacerbated in Shopify because the product field layout wasn’t getting saved to the project config. That’s been fixed for the next Shopify release as well (craftcms/shopify#50).

To get the Craft CMS fix early, change your `craftcms/cms` requirement in composer.json to `"dev-develop#2947da192efe8a78bd617565df0239baba46f931 as 4.3.5"` and run `composer update`.

brandonkelly:
[...]

PR #12462, opened 2022-12-15
additional check for layout without elements

i-just:
### Description
Additional check so that we only check for layout element UUID conflicts if there are elements in the layout. In other words - don’t check empty layouts, as that causes issues. For example, when you add a new block to a matrix field and try to add a field to that block - the field won’t be saved. This also caused the unit `FieldLayoutTest` to fail.

### Related issues
#12459

### DIFF
--- a/src/models/FieldLayout.php
+++ b/src/models/FieldLayout.php
@@ -319,23 +319,25 @@ class FieldLayout extends Model
 
             // Ensure there aren't any layout element UUID conflicts
             $layoutElements = $tab->getElements();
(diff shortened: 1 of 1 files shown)


brandonkelly:
Thanks for fixing that test :) 

The broken test got me a little spooked that we could be introducing a breaking change with 2947da192efe8a78bd617565df0239baba46f931, so I removed the PHP changes. The JS changes in that commit are enough to fix the main issue identified in #12459.

i-just:
Fair enough 🙂

Issue #12455, opened 2022-12-14
[4.x]: Large asset thumbnails appear stretched


mmikkel:
### Description

Since Craft 4.3.0, large asset thumbnail containers stretch across the entire width of the content pane, which looks a bit broken:

![CleanShot 2022-12-14 at 18 46 17](https://user-images.githubusercontent.com/298510/207669088-8533ca88-2ec1-4b75-b0a3-812dd0cada19.png)

Issue does not occur on Craft 4.2.8:

![CleanShot 2022-12-14 at 18 45 24](https://user-images.githubusercontent.com/298510/207668931-610ddd6f-5767-4748-9f24-fc56cc84893a.png)

### Steps to reproduce

1. Add a single asset to an asset folder
2. Toggle "Large thumbnails" view on in the asset index

### Craft CMS version

4.3.5

### PHP version

8.1

### Operating system and version

macOS 12.6

### Installed plugins and versions

- None

i-just:
Hi, thanks for reaching out. Around v4.3, some improvements were made to show as much of the asset titles as possible (for accessibility reasons), and I believe that’s when the centre alignment was introduced. I created a PR to make them left-aligned again.

brandonkelly:
Craft 4.3.6 is out with a fix for this.

PR #12466, opened 2022-12-16
adjust large thumbnail alignment

i-just:
### Description
Left-align large thumbs on assets index page.

1  multiple images
2  two images
3  one image - short title
4  one image - long title

### Related issues
#12455


brandonkelly:
The line items were still way too large with this PR:

A screenshot of a single asset in large thumbnail view, showing that it spans the full width of the page

Added `max-width` to the line items to prevent them from getting too wide in the first place.

PR #12474, opened 2022-12-19
Strip elisions from search keywords

### LINKED ISSUE #12467: [4.x]: Indexed keywords of sentences/words with single quotes/apostrophes are not correct in some cases
### Description

I have noticed that the search indexation is not working correctly when the search query includes words with single quotes or apostrophes, such as "l'orange" or "l'espace" in French. These words should be indexed as "orange" and "espace", but they are being indexed as "lorange" and "lespace" instead.

Upon investigation, I believe the issue may be related to this line of code: https://github.com/craftcms/cms/blob/develop/src/helpers/Search.php#L87.

An easy fix would be to give the the possibility to define a custom $asciiPunctuation variable to allow for better handling of these types of characters in search queries. This would greatly improve the usability of the search feature for users who speak languages that make use of single quotes or apostrophes.

### Steps to reproduce

1. Create an entry with this sentence `Bienvenue de l'espace`

### Expected behavior

Keywords in the database should be `bienvenue dans espace`

### Actual behavior

Keywords in the database are `bienvenue dans lespace`

### Craft CMS version

4.x

### PHP version

_No response_

_No response_

### Database type and version

_No response_

### Image driver and version

_No response_

-

brianjhanson:
### Description
I learned some things today. The first of which, is that when you smash together some sounds, it's called an [elision](https://en.wikipedia.org/wiki/Elision). They're not very common in English, but they're pretty common in French (they also happen in Catalon, Irish and Italian). 

Elisions should be stripped out of search keywords because they change the word and make it harder to find what you're looking for. This fixes that by creating an array of elisions that should be stripped from search terms. I only added the common list in French at the moment, primarily because [that's what Elastic search does by default](https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-elision-tokenfilter.html), but we may want to expand this in the future. 

### Related issues
Fixes #12467

### DIFF
--- a/src/helpers/Search.php
+++ b/src/helpers/Search.php
@@ -72,7 +72,7 @@ class Search
     }
 
     /**
-     * Returns an array of [elisions](https://en.wikipedia.org/wiki/Elision) that will be stripped from search keywords.
+     * Returns an array of elisions to remove from search keywords.
      *
      * @return array
      */
--- a/src/helpers/Search.php
+++ b/src/helpers/Search.php
@@ -51,6 +51,9 @@ class Search
         if ($processCharMap) {
             $str = strtr($str, StringHelper::asciiCharMap(true, $language ?? Craft::$app->language));
 
+            $elisions = self::_getElisions();
+            $str = str_replace($elisions, '', $str);
+
             // Remove punctuation and diacritics
             $punctuation = self::_getPunctuation();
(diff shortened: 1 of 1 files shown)


brandonkelly:
Cherry-picked your commits to `v3` and merged from there.

PR #12506, opened 2023-01-06
Dump debug panel

brandonkelly:
### Description

Adds a new `{% dump %}` tag, which will dump a variable (or the current context if no variable is provided), and capture the dump output, making it available in a new “Dumps” debug toolbar panel.

The new “Dumps” debug toolbar panel, showing two dumped variables.


brandonkelly:
@lewisjenkins Yep! Like so:

```php
use craft\debug\DumpPanel;

DumpPanel::dump($var);
```

PR #12508, opened 2023-01-09
Start transition from ID → Console

bencroker:
This PR is a starting point for transitioning usages of `Craft ID` to `Craft Console` in the code. It replaces all instances of `id.craftcms.com` with `console.craftcms.com`, as well as the button label in the plugin store.

Screenshot 2023-01-09 at 09 59 42

Note that the button label and URL also need to be fixed at https://plugins.craftcms.com/!

I haven’t touched any of the many instances of classes, methods and variables with `Craft ID` in them, as that’s likely something that should be handled internally.


brandonkelly:
Thanks for the PR!

I had been holding off for Craft 3.8 and 4.4 to do this since I thought it would require translation changes, but just realized that the affected translation strings aren’t actually present in the UI anymore. (They were used by the OAuth connection for in-app purchases, which was dropped in 3.7.37 in favor of redirecting you to `id.craftcms.com`.)

So this is now resolved for the next Craft 3 and 4 releases.

PR #12533, opened 2023-01-13
Replace $query->count() with count($query->all())

### LINKED ISSUE #12526: [4.x]: ResaveController ignores --offset when displaying element count in CLI output
### Description

Total number of elements shown in CLI output does not change when using offset

Without `--offset`:

image

With `--offset`:

image

### Steps to reproduce

1. Run `./craft resave/entries` to get the correct count of total entries being saved
2. Run `./craft resave/entries --offset=10`

### Expected behavior

The second run should show 10 less entries than the first

### Actual behavior

Total entries is not updated… however the number of entries saved is correct

### Notes

I think this might be an issue with Yii `count` method. 

### Craft CMS version

4.3.6.1

### PHP version

8.1

OSX

### Database type and version

Postgres

### Image driver and version

_No response_

-

brianjhanson:
Yii's $query->count() [ignores things like limit and offset](https://github.com/yiisoft/yii2/issues/13846).

Fixes #12526

### DIFF
--- a/src/console/controllers/ResaveController.php
+++ b/src/console/controllers/ResaveController.php
@@ -458,7 +458,7 @@ class ResaveController extends Controller
         /** @var ElementQuery $query */
         /** @var ElementInterface $elementType */
         $elementType = $query->elementType;
-        $count = (int)$query->count();
+        $count = count($query->all());
 
         if ($count === 0) {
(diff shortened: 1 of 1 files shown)


leevigraham:
@bencroker I have a feeling i've seen `->count()` in Blitz as well.

bencroker:
I don't think so @leevigraham, but please correct me if I'm wrong.

brianjhanson:
After talking internally, this isn't the best way to accomplish this.

Issue #12536, opened 2023-01-15
[4.x]: Assets::getAllDescendantFolders returns the `$parentFolder` folder in result


leevigraham:
### Description

`Assets.php` calls `$this->getAllDescendantFolders($folder)` on line 330 and passes in the `$folder`.

<img width="1471" alt="Screenshot 2023-01-15 at 9 32 45 pm" src="https://user-images.githubusercontent.com/25124/212535798-23879367-aa3e-4034-9c68-097b71518af9.png">

`getAllDescendantFolders($folder)` creates a query using the same `path` as the parentFolder

<img width="1471" alt="Screenshot 2023-01-15 at 9 35 27 pm" src="https://user-images.githubusercontent.com/25124/212535940-073d7b4a-7cee-416a-9ca1-0d6cc062a8b5.png">

As a result the parent is returned in the descendants:

<img width="1683" alt="Screenshot 2023-01-15 at 9 41 13 pm" src="https://user-images.githubusercontent.com/25124/212536169-fa31dea2-3674-4a9a-b1f3-5f143cd40a95.png">

### Expected behavior

`$parentFolder` should not be included in the returned results.

### Actual behavior

`$parentFolder` is included in the returned results.

### Craft CMS version

4.3.6.1

### PHP version

8.1

### Operating system and version

13.1 (22C65)

### Database type and version

MySQL

leevigraham:
@brandonkelly If you're looking at this Folder `155401` has the same path (weird duplicate) which is why it's also returned in the results.

brandonkelly:
This seems to be currently working as intended, as everything within Craft that calls the method is expecting the parent folder to be included in the results.

It’s obviously a bad method name though. To make amends. I’ve just given it a new `$withParent` argument for the next Craft 3 and 4 releases, which is `true` by default, but if you pass `false` it will omit the parent folder from the results.

brandonkelly:
Craft 3.7.64 and 4.3.7 have been released with that change.

PR #12537, opened 2023-01-15
Only return true descendants, not the parent.

leevigraham:
Fixes https://github.com/craftcms/cms/issues/12536.

Adding an `_` to the `like` query ensures there's at least one character after the parent folder slash.

MYSQL `_` reference: https://dev.mysql.com/doc/refman/8.0/en/string-comparison-functions.html#operator_like
Postgres `_` reference: https://www.postgresql.org/docs/current/functions-matching.html#FUNCTIONS-LIKE

Screenshot 2023-01-15 at 9 49 18 pm

### DIFF
--- a/src/services/Assets.php
+++ b/src/services/Assets.php
@@ -489,7 +489,7 @@ class Assets extends Component
         $query = $this->_createFolderQuery()
             ->where([
                 'and',
-                ['like', 'path', $parentFolder->path . '%', false],
+                ['like', 'path', $parentFolder->path . '_%', false],
                 ['volumeId' => $parentFolder->volumeId],
                 ['not', ['parentId' => null]],
             ]);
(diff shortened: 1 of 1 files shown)


leevigraham:
I've just reverted this PR to a draft because this change may have unintentional consequences when moving folders:

Screenshot 2023-01-15 at 10 01 00 pm

leevigraham:
A safer way to get descendants could be traversing the tree using parentId and multiple queries or switching to a nested set (like structures). In my case I've got duplicate paths and `getAllDescendantFolders` deletes folders that aren't part of the `parentId` structure.

leevigraham:
Example recursive function:

```sql
WITH RECURSIVE descendants AS
(
SELECT vf.* FROM volumefolders vf WHERE vf.parentId IS NULL
UNION ALL
SELECT vf2.* FROM volumefolders vf2 JOIN descendants 
ON descendants.id = vf2.parentId
)
SELECT * FROM descendants ORDER BY path;
```

image

brandonkelly:
See https://github.com/craftcms/cms/issues/12536#issuecomment-1414425227

PR #12550, opened 2023-01-18
Add floating labels to login form

gcamacho079:
### Description
Adds a new `FloatingLabel` Garnish class and implements on login form

### Related issues
Resolves DEV-315


src/templates/_includes/forms/text.twig
```diff
@@ -51,6 +53,15 @@
     }, recursive=true) %}
 {%- endif %}
 
+{%- if floatingLabel and label %}
```
gcamacho079, line 56: @brandonkelly I made this pretty barebones just for the login template, as having any floating labels in other parts of the CP would require a lot more style updates.

I was wondering if it would make sense to have both label types (floating- and heading-style) in a single template file, and in the case of the floating style, pass the output to this template.

gcamacho079:
Since persistent labels have been implemented, this is no longer relevant! 👍🏼

Issue #5809, opened 2020-03-18
Degraded performance in CP Assets view for folders with many subfolders


monachilada:
### Description

The performance of the assets page in the CP when it is asked to open a folder with a lot of subfolders in it is resulting page crashes in Chrome, Firefox and Safari on a recent MacBook Pro. Estimate is about 15k sub folders in the Assets volume that can be seen in the screenshots below.

If the folder has never been opened in the particular browser of choice the page will load ok, but as soon as the attempt to expand the folder is made, this will result in the freeze. Subsequent page loads will then also freeze, as the browser appears to attempt to remember the user preference and have that folder opened, making the assets page essentially unusable 😫.

![Chrome](https://user-images.githubusercontent.com/613183/76983311-7aa86800-693d-11ea-9eb6-c860138822ad.png)
![Firefox](https://user-images.githubusercontent.com/613183/76983318-7da35880-693d-11ea-8778-5f9e1157b604.png)

### Steps to reproduce

1. Have a lot of sub folders in a volume or folder
2. Attempt to expand the folder to show its contents

### Additional info

- Craft version: 3.4.10.1 (Pro)
- PHP version: 7.3.11
- OS version: Linux 4.19.0-0.bpo.6-amd64
- Database driver & version: MySQL 5.5.5
- Image driver & version: Imagick 3.4.3 (ImageMagick 7.0.8-58)
- Plugins & versions: 
    - Control Panel Nav 3.0.8.1
    - CP Field Inspect 1.1.3
    - CraftQL dev-master
    - Do It Yourself widget 2.0.1
    - Embedded Assets 2.2.1
    - Entry Instructions 1.0.7
    - Feed Me 4.2.1.1
    - Mailjet 1.2.0
    - MatrixMate 1.2.5
    - Preparse Field v1.1.0
    - Queue Manager 1.2.0
    - Redactor 2.6.1
    - Relabel 1.3.4
    - SEO 3.2.48
    - Servd Assets and Helpers 1.2.8
    - Tags 1.0.6
    - Typed link field 2.0.0-beta.7
    - Webhooks 2.2.0

monachilada:
Also affects modal asset windows triggered from asset fields, making them also unusable.

brandonkelly:
This is a known issue. We’re considering a couple ways to solve it, such as #1629. I’ll leave this open and update it once we have a solution.

brandonkelly:
[...]

PR #12558, opened 2023-01-21
Volume folder nav improvements (Craft 3)

brandonkelly:
### Description

Revamps how volume subfolder navigation works within asset indexes, to improve performance and usability.

Instead of trying to display every recursive subfolder within the sidebar as nested sources (which had a tendency to crash the browser on sites with 1K+ subfolders), we now show subfolders inline within the element listing pane:

An Assets index page, showing four folders and two assets side-by-side in the element listing pane.

Subfolders are always displayed first, generally in ascending alphabetical order, or descending order if assets are being sorted by Title or Filename in descending order.

As you navigate into subfolders, the current path is shown in the new “source path” component, above the element listing.

An Assets index page, navigated to the “art → van-gogh” subfolder.

Clicking on a parent folder within the source path will navigate to it. (Command-clicking to open the folder in a new tab is also supported.)

The currently active subfolder in the source path has a disclosure menu, with “New subfolder”, “Rename folder”, “Move folder”, and “Delete folder” options.

Additionally, a new “Move…” bulk element action has been added for asset indexes, which supports moving subfolders alongside other assets. Selecting it will open a new folder selector modal where a new destination can be chosen.

The new volume folder selector modal.

### Related issues

- #5809
- #8422
- #9171
- #12446


michaelrog:
This is so nice. Chef kiss. 😙🤌

brandonkelly:
@sebastian-lenz Can you post a new issue with steps to reproduce?

brandonkelly:
No, as `Träger` and `traeger` could both co-exist as unique folder names.

Issue #12557, opened 2023-01-20
[4.x]: Backups are broken under mysql 5.7.41 and 8.0.32


nitriques:
### Description

We run many craft instances, on many servers.
They all are cpanel/whm servers, which are a quite common.
With the last update, it installed mysql 5.7.41 / 8.0.32 on most our the servers.
This version was released by Oracle on Jan 17th (3 days ago).
Since then, all running craft instances on those server can no longer run backups.

We tested on other servers that still have mysql 5.7.39 and it stills works.
We tested with craft 4.2.5.2 and 4.3.6.1.
Both works/fails depending on mysql's version (so it's not directly craft related).

The problem is only present when the `--single-transaction` is used. Removing it creates a backup.
Nothing seems related to this flag in the release notes:   
https://dev.mysql.com/doc/relnotes/mysql/5.7/en/news-5-7-41.html  
https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-32.html  

Since `--single-transaction` locks the database, I can understand why mysql would want to FLUSH TABLES before doing so. But having a user that has this privilege might be something pretty rare: Under cpanel, your can't even create a user with such privilege. And it's a global privilege and can not be enable only for specific db. 
Using the `--single-transaction` seems like a good practice tho so maybe craft could check for the privilege before including it ?

I am pretty sure that we are not the only ones with broken backups right now and even tho this is not a bug in craft, we think it should try to maximize the chance of producing a backup.

### Steps to reproduce

1. Run `./craft db/backup`

### Expected behavior

It creates a backup, with a warning that it may be inconsistent because the DB user does not have the RELOAD privilege (?)

### Actual behavior

It errors out with the message:

```
failed with exit code 2: mysqldump: Couldn't execute 'FLUSH TABLES': Access denied; you need (at least one of) the RELOAD or FLUSH_TABLES privilege(s) for this operation (1227)
```

### Craft CMS version

4.x

### PHP version

7.4-8.1

### Operating system and version

Centos

### Database type and version

mysql 5.7.41 and 8.0.32

### Image driver and version

gd

### Installed plugins and versions

- Does not mater...

nitriques:
Side note: It is also curious that `--single-transaction` is used to dump the schema but not for the data...
So it's kind of useless as of now no?

Maybe the whole operation could be done in a single `mysqldump` operation ?

Ref: https://github.com/craftcms/cms/blob/develop/src/db/mysql/Schema.php#L200-L212

[...]

PR #12560, opened 2023-01-23
Fix for #12557

angrybrad:
Works around a MySQL bug https://bugs.mysql.com/bug.php?id=109685


mrsimonbennett:
Hey @angrybrad I am not sure if this is the best fix

This has been affecting our customers, (some of which use SnapShooter to backup craft)

By removing the --single-tranaction flag you're changing the way the backup will run and could cause issues that users would not expect. I know it was not ideal the way MySQL did the bug fix, but it was to fix an issue with the --single-transaction backup

We have been advising users to most ideally [update there grant permissions ](https://snapshooter.com/docs/fixing-mysqldump-couldn-t-execute-flush-tables-access-denied/)


```
GRANT RELOAD,PROCESS ON *.* TO 'backups'@'%';
FLUSH PRIVILEGES;
```

I know its not ideal at all!

afairhurst:
@mrsimonbennett

I ran into this issue today (unable to backup database). Your solution to update grant permissions worked for me. Thank you!

mohamedhafez:
@mrsimonbennett is there any reason we can't just grant the minimum permissions necessary with `GRANT SHOW_TABLES ON *.* TO 'backups'@'%';`? will it just fail with other errors later on, or is there a possibility it will work with just `SHOW_TABLES` for some people?

mrsimonbennett:
> @mrsimonbennett is there any reason we can't just grant the minimum permissions necessary with `GRANT SHOW_TABLES ON *.* TO 'backups'@'%';`? will it just fail with other errors later on, or is there a possibility it will work with just `SHOW_TABLES` for some people?
> 
> 
> 
> 

We need to allow the MySQL user to call some commands that are now required, you can read more about it in the MySQL bug fix. But basically they need to be able to flush the database first now as part of a single transaction backup

darylknight:
For users who land here and are using Ploi, this is what worked for us:

```
mysql -u username -p
GRANT RELOAD,PROCESS ON *.* TO 'ploi'@'%';   < this is for the automated backups
GRANT RELOAD,PROCESS ON *.* TO 'database_user'@'127.0.0.1';   < this is for the database user if you pull down your databases using Craft Scripts
FLUSH PRIVILEGES;
```

angrybrad:
@mrsimonbennett agree it's all gross, but it's a lowest common denominator workaround. The alternative is that we start requiring a _server wide_ privilege for less privileged users to be able to complete a backup in Craft. Even if we wanted to do that, plenty of hosts won't allow it (like AWS RDS), even though it's probably less of an issue if you're hosting on DO and/or using a provisioning service.

Honestly, I'm hoping they revert the change in the next release as it's quite a bad one.

darylknight:
Does removing the single-transaction flag mean that if someone saves an entry during the backup, the backup file will be corrupted? Because it might have already backed up one table that relies on another that has been changed?

jeromecoupe:
Was bitten by this one #13014

angrybrad:
@esbiorn I don’t think detecting the `mysqldump` version vs. the MySQL version addresses the underlying issue, though. The underlying problem is that the `mysqldump` version is out of sync with the database version it’s running against. Or at the very least, using an older version of `mysqldump` to back up a newer version of MySQL can be problematic.

`mysqldump`, in general, seems to be pretty backward compatible. I’ve never seen a problem running a newer version of `mysqldump` against an older database (not even ever had to use the [compatible](https://dev.mysql.com/doc/refman/8.0/en/mysqldump.html#option_mysqldump_compatible) flag).

PR #12564, opened 2023-01-23
Volume folder nav improvements (Craft 4)

brandonkelly:
Craft 4 port of #12558.


gcamacho079 (changes requested):
- [x] Focus styles should be customized to outline the entire breadcrumb link/button
- [ ] Because both the text content and `title` provide an accessible name, the `aria-label` can be removed from breadcrumb buttons with inner text
- [ ] Because the breadcrumb buttons trigger an on-page action, they should be given `role="button"` and should be activated via Spacebar as well; we have the new `handleActivatingKeypress` function to handle this
- [ ] A different visual treatment should be used for the last element to identify it as the active subfolder - perhaps using bold text or a different background color, like in the Source/Volume navigation
- [ ] The folder image doesn’t meet 3:1 against background, and has an even lower contrast on hover or select. This is essential as it provides info that isn't apparent elsewhere, so it should remain visible in all states. (Alternately, adding a new column for "Kind" like in Mac Finder could solve this issue, and also the next...)
- [ ] I'm not 100% sure having folders in the assets table is the right move, but I may be overthinking this. If anything, changing the `th` for the "Asset" column to "Name" may clarify that the table can contain either folders _or_ assets. As mentioned previously, having a new column for "Kind" in the default table view could also make this more clear.
- [ ] Focus management:
    - [ ] When activating a Volume path navigation button, focus should remain on the button when the new content loads
    - [ ] If navigating to a new folder from inside the `.elements` container, keyboard focus should move to the first focusable element in the view
    - [ ] When moving a folder to another folder, keyboard focus should be returned to the disclosure trigger
- [ ] Folder names should not have overflow hidden and should wrap onto new lines instead so that full folder names are always visible
    - Possible bug: when a folder is created and the breadcrumbs are updated via JS, instead of applying ellipses the folder names are shown in full and the nav scrolls off the page horizontally
- [ ] I think we should update the page title to accurately reflect the context: “Media subfolder - Blog volume - Assets - [Site Name]”
- [ ] Since there won’t be nesting in the Sources nav anymore, we can remove the markup for nested lists and just do headings with lists afterward

brandonkelly:
> I'm not 100% sure having folders in the assets table is the right move, but I may be overthinking this. If anything, changing the `th` for the "Asset" column to "Name" may clarify that the table can contain either folders _or_ assets. As mentioned previously, having a new column for "Kind" in the default table view could also make this more clear.

I was debating whether I should change the `<th>` to “Item”, and “asset(s)” to “item(s)” in the pagination text in the footer, but decided it’s probably not necessary since “asset” is already a bit of a generic term. (As opposed to “file” which would have been clearly wrong.)

That said, I’ve updated folders’ titles’ `aria-label` to `folder-name (Folder)`, which should help clear it up, right?

> I think we should update the page title to accurately reflect the context: “Media subfolder - Blog volume - Assets - [Site Name]”

We don’t currently even update it to show the selected source label. So while we could make this change, it feels like something we should consider for all element index pages, not just assets, as a separate PR.

> Since there won’t be nesting in the Sources nav anymore, we can remove the markup for nested lists and just do headings with lists afterward

There still technically could be nested sources, if another element type is choosing to define them, or if a developer adds them via a plugin/module. So if the current markup isn’t making it _worse_ for single-level source lists, my vote would be to leave it alone.

brandonkelly:
> Folder names should not have overflow hidden and should wrap onto new lines instead so that full folder names are always visible 

There’s no way to achieve this without messing up the chevron styling. I’m setting `title` and `aria-label` on them so the full folder label is still viewable for both sighted and screen reader users. If that’s in violation, could we just start completely hiding the entire label of higher-level subfolders until there’s enough space to display the full name of the selected subfolder?

Source path with the first subfolder name completely replaced with an ellipsis, to make room for the remaining folder names

brandonkelly:
> If navigating to a new folder from inside the `.elements` container, keyboard focus should move to the first focusable element in the view

First focusable DOM element (the “select all” checkbox), or first focusable content element? I read it as the latter, but then realized there’s code in place within Craft 4 to select the first focusable DOM element after the view is updated on page change.

https://github.com/craftcms/cms/blob/b3eb1d19a957939f0a78c2680fb29bbfa909f862/src/web/assets/cp/src/js/BaseElementIndex.js#L914-L917

gcamacho079:
> I was debating whether I should change the `<th>` to “Item”, and “asset(s)” to “item(s)” in the pagination text in the footer, but decided it’s probably not necessary since “asset” is already a bit of a generic term. (As opposed to “file” which would have been clearly wrong.)
> 
> That said, I’ve updated folders’ titles’ `aria-label` to `folder-name (Folder)`, which should help clear it up, right?

Yes, we can even do without the parentheses: `folder-name Folder`. With the text included in the link, the folder graphic can be marked as decorative.

> We don’t currently even update it to show the selected source label. So while we could make this change, it feels like something we should consider for all element index pages, not just assets, as a separate PR.

Agreed, it can definitely be its own PR and update all element index pages. 👍🏼 Will add this to Linear.

> There still technically could be nested sources, if another element type is choosing to define them, or if a developer adds them via a plugin/module. So if the current markup isn’t making it _worse_ for single-level source lists, my vote would be to leave it alone.

Alright, that's good to know! I do have an alternate markup suggestion that may make it easier to navigate, but I'll suggest that in a separate PR.

> If that’s in violation, could we just start completely hiding the entire label of higher-level subfolders until there’s enough space to display the full name of the selected subfolder?

This would be my suggestion as both screen reader and keyboard users would have access to the same level of information as sighted mouse users; it's also an expected navigation pattern to truncate parts of navigation/pagination with ellipses. In that case, the `...` list item would be static text without any added functionality.

> First focusable DOM element (the “select all” checkbox), or first focusable content element? I read it as the latter, but then realized there’s already code in place within Craft 4 to select the first focusable DOM element after the view is updated on page change.

I'm refactoring that code snippet in another PR to not move focus when the page changes, but when a user clicks into a folder it should probably move to the first focusable element in `tbody`.

gcamacho079 (changes requested):
Just a few more things in addition to the comments I left above:

- [x] For the active breadcrumb button, `aria-current="true"` should be switched to `aria-current="page"`
- [x] Inside the `.elements` container, folder links should be given `role="button"`
- [x] There's currently another PR open (#12213) that darkens selected items dramatically. In most cases, the blank table cells in other columns provide additional context that makes it clear that certain elements are folders. However, in cases where only the "Asset" column is visible, users lose that contextual information, and the folder image would also become difficult to see. Adding a border around the folder shape may help with this.
- [ ] Right now, most actions in the disclosure menu aren't working for me. I'm getting a success notification, but the loading spinner is persistent. Refreshing the page results in the correct view.
- [ ] For all actions in the disclosure menu, focus should return to the trigger. It's currently being dropped, but that may be because of the previous issue.
- [ ] Focus is a little difficult to see for the current folder button and the disclosure trigger. There's a `two-color-focus-ring` mixin that may help by providing a focus ring that works over both dark and light backgrounds.

brandonkelly:
> In that case, the `...` list item would be static text without any added functionality.

That could be problematic if even the direct parent folder isn’t visible. So maybe we need to make it a disclosure menu instead, like the main page breadcrumbs?

gcamacho079:
> So maybe we need to make it a disclosure menu instead, like the main page breadcrumbs?

Ahh yes, the disclosure menu would be a good solution for when that happens

brandonkelly:
@gcamacho079 Alright, overflow menu has been added!

The overflow menu for a source path

gcamacho079:
The overflow menu is beautiful! 💯 Just a few things:

1. Dependent on #12213 being merged, the new white outline on the folder graphic should be sufficient to distinguish it in the hover/selected state. **No action is necessary, just making a note for my brain.** 👍🏼 
2. When the root folder is collapsed inside the disclosure menu, the `span[data-icon="home"]` should be set to `aria-hidden` so the accessible name is consistent between collapsed and full versions of the breadcrumb nav
3. I'm still having some visibility issues with the focus outlines. Right now, the solid inner box shadow (#5ED0FA) for the `button[aria-current="page"]` and the disclosure trigger have poor contrast with all adjacent colors (dark and light gray button backgrounds), while the slightly transparent outer shadow also has the same issue. Having all the button outlines in this section two-toned (like the global skip link) would ensure contrast with all surrounding colors.
We could also achieve good focus visibility by having the text inside the button be what receives a focus outline (see third image)
A folder button in the breadcrumb section has a dark background color and light text. Its focus outline is very light blue, which contrasts well with the adjacent disclosure button background color, but not with any other adjacent colors
Global skip link with a two-color (white and black) focus outline

This following image is an example from [Understanding SC 1.4.11 Non-Text Contrast](https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast.html) documentation, but there are also alternative solutions in the [Focus Appearance documents](https://www.w3.org/WAI/WCAG22/Understanding/focus-appearance)

Screenshot 2023-01-31 at 3 08 42 PM

brandonkelly:
Alright I’ve made those additional changes! Went with your last suggestion, to have the text within the button be what gets the focus ring:

A source path with the current step’s text focussed, showing that its label gets a focus ring rather than the outer button shape.

gcamacho079:
@brandonkelly one last thing, I just checked on Firefox with Zoom Text Only, and noticed the following:

The focused navigation component has a light blue outline over a light gray page background when text is zoomed to 200%

- The root button is getting cut off due to the fixed `width: 13px` CSS. I recommend using a `rem` value here.
- The light focus outline on the text overlaps the light page background instead of the button background.

I’m unsure if refactoring `--ui-control-height` to use `rem` instead of `px` would be considered a breaking change, but I recommend doing that for at least this component, if not globally.

brandonkelly:
@gcamacho079 Fixed now!

Issue #12539, opened 2023-01-16
[4.x]: When removing money entry from entry type data stays saved as string


michtio:
### Description

When a money field has existed on multiple entry types, but is later removed from an entry type. The field value stays in the database content table. This value is now handled as string, which throws an error applying the `|money` filter.

### Steps to reproduce

1. Create 2 entry types on a section
2. Create a money field and add it to both entry types
3. Enter data in the entry in the money field on both entry types
4. Remove the money field from 1 entry type
5. The template will now error out if we try to parse an overview of all the entries in said section

### Expected behavior

When a money field is removed from an entry type, also remove all associated data of the money field in the database, so we can generate an overview without throwing any error. Or keep the data in the database but don't parse the field as it's disassociated with the field layout.

### Actual behavior

When adding `entry.moneyField` to loop over all entries (without checking the entry types, e.g. for an overview) the entry types that used to have a money field are now throwing the following error:

```
craft\web\twig\Extension::moneyFilter(): Argument #1 ($money) must be of type ?Money\Money, string given...
```

### Craft CMS version

4.3.6.1

### PHP version

8.0

### Additional info

Trying to find a workaround is a bit bulky, using the resave command for bulk editing isn't a solution since the field doesn't exist anymore. The only solution is to add/create the field again to the entry type. Save everything with `:empty:` deploy, do this on staging / production, to remove the field again and deploy again. This is quite cumbersome.

```
docker exec -it hdac-php-1 su-exec www-data php craft \
                                resave/entries --set semesterlyPrice --to :empty: --type==team --section=classes
Resaving 9 entries ...
    - [1/9] Resaving Chaos (3216) ... error: Setting unknown property: craft\elements\Entry::semesterlyPrice
    - [2/9] Resaving Team Infinity (3191) ... error: Setting unknown property: craft\elements\Entry::semesterlyPrice
    - [3/9] Resaving Team Felicity (3205) ... error: Setting unknown property: craft\elements\Entry::semesterlyPrice
    - [4/9] Resaving Team Diva Diversity (3180) ... error: Setting unknown property: craft\elements\Entry::semesterlyPrice
    - [5/9] Resaving Anthro-pos (3167) ... error: Setting unknown property: craft\elements\Entry::semesterlyPrice
    - [6/9] Resaving Team United (3156) ... error: Setting unknown property: craft\elements\Entry::semesterlyPrice
[...]

PR #12578, opened 2023-01-25
Bugfix/12539 fields removed from layout null vals

i-just:
### Description
If you add a custom field to a layout, populate that field with a value and then remove that field from the layout, if you try to get that value via a template, it will still be returned even though the field is no longer part of the layout. In terms of more complex fields (e.g. Money field for Craft 4, the value will be returned as string/number depending on your PHP version, and not as a Money field).

This PR sets custom field values for fields that are not part of the current layout as null.
The same applies to eager-loaded elements - only get them if they’re part of the layout (or if there’s no layout).

This can be merged into Craft 4 too.

### Related issues
#12539


michtio:
Thank you @i-just , this seems like the perfect solution! 🙌🏻

brandonkelly:
Craft 3.7.64 and 4.3.7 have been released with the fix for this.

Issue #11753, opened 2022-08-09
[3.x]: Database backup after composer update not working with SSL 


carolineboeykens:
### Description
Our server requires an SSL connection on our database. We added the following code to the config/db.php

` if(getenv('ENVIRONMENT') != 'dev') {
     $params['attributes']= [
         PDO::MYSQL_ATTR_SSL_CA => '/etc/ssl/certs/ca-bundle.crt'
     ];
 }`

Now the database backup on our production server won't work:

`mysqldump: Got error: 3159: “Connections using insecure transport are prohibited while --require_secure_transport=ON.” when trying to connect`

### Steps to reproduce

1. Enable 'backupOnUpdate' => true in general.php
2. Add PDO::MYSQL_ATTR_SSL_CA to db.php
3. ./craft update --> so that the database backup will be triggered after the updates

### Expected behavior
Does the database backup not use the parameters from db.php? 

### Craft CMS version

3.7.40

### PHP version

7.4

### Operating system and version

CentOS Linux

### Database type and version

MySQL 8.0.26

### Image driver and version

Imagick 3.7.0 (ImageMagick 6.9.12-55)

### Installed plugins and versions

-

angrybrad:
Was this working before an update and broke after or has it never worked for you? If the prior, do you recall the version you were on before?

carolineboeykens:
It was working before we had to add the SSL variable in the db.php. We recently moved our database to an environment where this is required

erickertzvfc:
Any updates on this? I also migrated to an env that requires SSL and DB backups no longer work using mysqldump.

`failed with exit code 2: mysqldump: Got error: 9002: "SSL connection is required. Please specify SSL options and retry."
 when trying to connect {"trace":["#0 [internal function]: \\controllers\\UtilitiesController->actionDbBackupPerformAction()",`

brandonkelly:
@carolineboeykens Does that help?

carolineboeykens:
@brandonkelly ,
This is not working for us because we're not using Alpine or Azure.

We got this error message:

[...]

Issue #10351, opened 2022-01-11
Error "Backing up the database" with SSL config


nickdunn:
### Description

When I run a project config update, Craft backs up the database. This has been working fine, but I recently updated to 3.7.28 (from 3.7.23) and when I deploy I now see:

> 2022-01-11T16:23:30.529761721Z Backing up the database ... error: The shell command "mysqldump --defaults-extra-file="/tmp/amvmktvlfncx.cnf" --add-drop-table --comments --create-options --dump-date --no-autocommit --routines --default-character-set=utf8 --set-charset --triggers --no-tablespaces --single-transaction --no-data --result-file="/var/www/craft/storage/backups/uat--2022-01-11-162330--v3.7.28.sql" craftcms && mysqldump --defaults-extra-file="/tmp/amvmktvlfncx.cnf" --add-drop-table --comments --create-options --dump-date --no-autocommit --routines --default-character-set=utf8 --set-charset --triggers --no-tablespaces --no-create-info --ignore-table=craftcms.assetindexdata --ignore-table=craftcms.assettransformindex --ignore-table=craftcms.sessions --ignore-table=craftcms.templatecaches --ignore-table=craftcms.templatecachequeries --ignore-table=craftcms.templatecacheelements --ignore-table=craftcms.cache --ignore-table=craftcms.templatecachecriteria craftcms >> "/var/www/craft/storage/backups/uat--2022-01-11-162330--v3.7.28.sql"" failed with exit code 2: mysqldump: Got error: 9002: "SSL connection is required. Please specify SSL options and retry." when trying to connect

I'm hosted in Azure and we have SSL enabled on the database:

```
return [
    'driver' => 'mysql',
    'server' => App::env('DB_SERVER'),
    'port' => App::env('DB_PORT'),
    'database' => App::env('DB_DATABASE'),
    'user' => App::env('DB_USER'),
    'password' => App::env('DB_PASSWORD'),
    'password' => App::env('DB_PASSWORD'),
    'tablePrefix' => '',
    'attributes' => App::env('DB_SSL') === '1' ? [
        // https://docs.microsoft.com/en-us/azure/mysql/howto-configure-ssl
        PDO::MYSQL_ATTR_SSL_CA => '/var/www/craft/config/BaltimoreCyberTrustRoot.crt.pem',
        // https://cloud.google.com/sql/docs/mysql/samples/cloud-sql-mysql-pdo-timeout
        PDO::ATTR_TIMEOUT => 240,
[...]

PR #12596, opened 2023-01-30
Add support for certificates in mysqldump

okolvik-avento:
### Description

As it is right now, craft does not support setting the ssl attributes in the configuration it uses for mysqldump. Add support for this instead of having to use the really cumbersome method of overriding the backup command setting.

### Related issues
#10351
#11753

### DIFF
--- a/src/db/mysql/Schema.php
+++ b/src/db/mysql/Schema.php
@@ -395,13 +395,13 @@ SQL;
         }
         
         // Certificates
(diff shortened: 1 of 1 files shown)


okolvik-avento:
@angrybrad Fixed the code style error (space before parentheses)

brandonkelly:
Thanks for the PR @okolvik-avento! We cherry-picked your fix over to Craft 3 via #12612, and have merged it into `develop` from there.

PR #12601, opened 2023-02-01
Only localized element types should get the site condition rule

lukeholder:
When using a condition builder with a non-localised element type, the "Site" condition rule should not be available for use.

This PR unsets the site rule from the condition if the element type is not localized


brandonkelly:
Craft 4.3.10 is out with this change.

Issue #12585, opened 2023-01-27
[4.x]: Asset index after emptying folder via FTP


mark-chief:
### Description
Back in Craft 3, when you deleted an Asset volume folder on the server, you could go to utilities and update the Asset Index and it would update the volume in the CP. 

But in Craft 4, after you delete all assets from a volume via FTP and then go to Utlities > Asset Index > Update asset index, it says there's nothing to index and the volume still shows all the assets as broken images as they obviously dont exist.

### Craft CMS version

4.3.6.1

### PHP version

8.1

### Installed plugins and versions

-

i-just:
Hi, thanks for getting in touch. This is an expected behaviour in Craft 4. It will only occur if there's absolutely nothing (no files and no folders) under the base path of the filesystem used by the volume you're indexing. That being said, we are going to make the message clearer in 4.4.

mark-chief:
Thanks for checking. So even though all the empty subfolders dont exist on the server you still have to manually delete all of them  via the craft CP? Seems counterintuitive, unless I have missed something completely obvious?

i-just:
Ok, I understand what you're after a bit better now. 

Yes, that's correct - at the moment, you'd have to delete them manually. Though it's not so much that the Asset Index doesn't pick it up; it's that it's not getting that far if there's absolutely nothing under your FS base path. 

You can "bypass" that if you have even one empty folder under your FS base path location.

We'll talk about this behaviour internally, and I'll update this issue once we've done that.

mark-chief:
Thank you

i-just:
We’ve discussed this, and I created a PR to change the current behaviour so that a list of missing files and folders is shown even if there’s nothing in the FS root folder.

brandonkelly:
This is resolved for Craft 4.4 via #12604 🎉

PR #12604, opened 2023-02-01
Feature/1078 asset indexing improvements

i-just:
### Description
Previously, when all files and folders were removed from the FS root folder, asset indexing was bailing before it started. This PR changes this behaviour, so indexing will still go ahead even if the FS root folder is completely empty.

It also introduces the `listEmptyFolders` option, which, if turned on, will list folders that exist but are empty or could be empty because all files in them are listed as missing and allows you to choose to delete such folders.

The review screen now divides into two lists (folders and files) to clarify the messaging in cases where `listEmptyFolders` is `true`.

Screenshot 2023-02-01 at 12 09 27

Screenshot 2023-02-01 at 12 10 09

### Related issues
#12585


brandonkelly:
Damn @i-just, this is great!

Issue #12602, opened 2023-02-01
[4.3.6.1]: Tab doesn't show,  when the tab name in chinese


ivesho:
### Description
Tab doesn't show,  when the tab name in chinese.
![image](https://user-images.githubusercontent.com/48641341/216006655-b9d095a0-ae9e-45af-b5d6-dbf3571153ca.png)
![image](https://user-images.githubusercontent.com/48641341/216006883-2e069eb0-8e43-4fef-b310-d8bf30e9cb00.png)

### Steps to reproduce

1.

### Craft CMS version

Craft Pro 4.3.6.1

### PHP version

8.1.8

### Installed plugins and versions

-

brandonkelly:
Thanks for reporting! This is fixed for the next release.

brandonkelly:
Craft 4.3.7 has been released with that fix.

PR #12606, opened 2023-02-01
let StringHelper::toAscii use language detection

i-just:
### Description
Field layout tabs don’t show when the tab name is in a language that doesn’t easily convert to ASCII. This affects languages like Mandarin, Hebrew, Korean, Japanese and potentially others.

This PR adds language detection via [patrickschur/language-detection](https://github.com/patrickschur/language-detection) to the `StringHelper::toAscii` (only) if the current approach returns an empty string.

### Related issues
#12602

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -48,6 +48,7 @@
     "mikehaertl/php-shellcommand": "^1.6.3",
     "moneyphp/money": "^4.0",
     "monolog/monolog": "^2.3",
+    "patrickschur/language-detection": "^5.2",
     "pixelandtonic/imagine": "~1.3.3.1",
     "samdark/yii2-psr-log-target": "^1.1",
     "seld/cli-prompt": "^1.0.4",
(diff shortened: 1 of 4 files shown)


brandonkelly:
Pushed up a [simpler fix](https://github.com/craftcms/cms/commit/2253f1eddf27adc2cc11849ced7c40f2bc11502c) – just switched to running the tab name through `md5()`, which will produce a consistent, ID-friendly result regardless of whether the tab name consists of ASCII-friendly characters.

i-just:
Cool! I got hung up on keeping the anchor links human-friendly.

brandonkelly:
Ah shoot, I didn’t think about that 🤦🏻 

Just reverted that back a bit, so `md5()` is only used as a fallback if `toAscii()` gives us an empty string. (db1ec08e0588ce4426637cc3a5e35c845a844be4)

PR #12618, opened 2023-02-03
`aria-labelledby` fixes

brandonkelly:
### Description

- The `_includes/forms/checkbox.twig`, `select.twig`, and `text.twig` templates will no longer output an `aria-labelledby` attribute when `labelledBy` is set, if `inputAttributes.aria.label` is also set.
- `craft\helpers\Cp::fieldHtml()` will no longer define `$config['labelledBy']` if the field doesn’t have a label to begin with.
- `craft\helpers\Cp::dateTimeFieldHtml()` now sets `$config['fieldset'] = true`. 

### Related issues

- Resolves #11961
- DEV-964

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Release Notes for Craft CMS 4
 
+## Unreleased
+
+- The `_includes/forms/checkbox.twig`, `select.twig`, and `text.twig` templates no longer add an `aria-labelledby` attribute to the input if an `aria-label` attribute is also specified.
+- `craft\helpers\Cp::dateTimeFieldHtml()` now returns inputs contained within a fieldset.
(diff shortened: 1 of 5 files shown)


gcamacho079 (changes requested):
Almost there! The datetime fields in the `.meta` container should also use a `fieldset` ([see change in old PR](https://github.com/craftcms/cms/pull/11961/files#diff-37afb4fd8af7be838f4108b66bb129e42f89f0cf1f10324c10d62ad816ad696bR1702)).

brandonkelly:
@gcamacho079 Just made that change centrally from `Cp::dateTimeFieldHtml()`. (The `dateTimeField()` macro in `_includes/forms.twig` is already doing this.)

gcamacho079 (approved):

PR #12625, opened 2023-02-05
Use `typeId` from `$this->_attributes` for drafts

### LINKED ISSUE #12592: [4.x]: Relational fields are cleared out when changing entry types
### Description

If the entry type is changed for an entry, relations made in a relational field that _doesn't exist in the new entry type's field layout_, are removed.

This is inconsistent with how Craft retains data for all other field types (plaintext, select fields, etc) when changing an entry's entry type. It's also not how this used to work (I'm not able to reproduce the behavior on Craft 3.7.63.1).

**I'm unsure if the current behavior is an intentional change in Craft 4, or a bug?**

_One interesting detail: From looking at the `relations` table, it doesn't look like the relations are deleted when the entry type is first changed to the type **without** the relational field; it's only when the entry type is changed _back_ to the type **with** the relational field, that the relations are actually deleted from the database. Don't know if that's relevant or not, but seemed a bit odd to me._

### Steps to reproduce

1. Create a section with two entry types, and add a relational field (assets, entries, Matrix ) to one of them.
2. Create an entry and add an element to the relational field. Save the entry.
3. Change the entry type to the one without the relational field. Save the entry.
5. Change the entry type _back_ to the type with the relational field. Notice that that the relational field is now empty.

### Expected behavior

Existing relations made in fields that are no longer in an element's field layout after changing entry types should be retained, similar to how Craft retains data for all other field types in the same situation.

### Actual behavior

Relations are deleted.

### Craft CMS version

4.3.6.1

### PHP version

8.1.12

macOS 12.6

### Database type and version

10.9.3-MariaDB

### Image driver and version

_No response_

None

brianjhanson:
Safer fix than #12592. Make sure to use the `typeId` from the request instead of from the element when creating a new draft.

Fixes #12592

### DIFF
--- a/src/controllers/ElementsController.php
+++ b/src/controllers/ElementsController.php
@@ -208,9 +208,9 @@ class ElementsController extends Controller
 
         // Redirect to its edit page
         $editUrl = $element->getCpEditUrl() ?? UrlHelper::actionUrl('elements/edit', [
-                'draftId' => $element->draftId,
-                'siteId' => $element->siteId,
-            ]);
+            'draftId' => $element->draftId,
+            'siteId' => $element->siteId,
+        ]);
 
         $response = $this->_asSuccess(Craft::t('app', '{type} created.', [
             'type' => Craft::t('app', 'Draft'),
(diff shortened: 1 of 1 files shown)


brianjhanson:
Closed for the moment because there aren't any good, obvious solutions, and the bug is probably better fixed by doing a better job of removing data when the entry type is changed.

mmikkel:
> Closed for the moment because there aren't any good, obvious solutions, and the bug is probably better fixed by doing a better job of removing data when the entry type is changed.

😢 Thanks for the effort though @brianjhanson! Hopefully a viable fix presents itself at some point - especially considering entry types are due for an even bigger role in Craft 5, w/ decoupled types and Inception fields 🙏

PR #12635, opened 2023-02-06
Accounts for non-array fieldId in MatrixBlockQuery

markhuot:
During a Craft 3.7 -> Craft 4.3 upgrade I ended up with a `$this->fieldId=69` which is not a `foreach`-able array. The fix was to just add the fieldId as a tag directly, but I'm not sure if that affects other parts of the code?

The big question: is this a result of a bad upgrade on my part or an actual bug? The type hints seem to imply that an `int` is a valid value for `->fieldId` but I'm not sure if they're just out of date?

### DIFF
--- a/src/elements/db/MatrixBlockQuery.php
+++ b/src/elements/db/MatrixBlockQuery.php
@@ -676,8 +676,13 @@ class MatrixBlockQuery extends ElementQuery
             }
         } else {
             if ($this->fieldId) {
-                foreach ($this->fieldId as $fieldId) {
-                    $tags[] = "field:$fieldId";
+                if (is_array($this->fieldId)) {
+                    foreach ($this->fieldId as $fieldId) {
+                        $tags[] = "field:$fieldId";
+                    }
+                }
+                else {
+                    $tags[] = "field:{$this->fieldId}";
                 }
             }
             if ($this->primaryOwnerId) {


brandonkelly:
`fieldId` should be getting normalized to an array or `false` via [`_normalizeFieldId()`](vhttps://github.com/craftcms/cms/blob/750fe591748cf79e06ab08f6c82cb80611d53c30/src/elements/db/MatrixBlockQuery.php#L593) (called at the beginning of `beforePrepare()`. Are you doing something to prevent that from getting called?

markhuot:
No, I get this when clicking "New entry" on the `/admin/entries` screen of a 4.3.6.1 install.

Here's the error I get when I click the button which triggers a POST to `index.php?p=admin%2Factions%2Fentries%2Fcreate&v=1675795476250`

Screenshot 2023-02-07 at 1 44 55 PM

brandonkelly:
Can you post the full stack trace?

markhuot:
I can do you one better. I can recreate this on a vanilla install of Craft 4.3.8.2 by creating a single section with a single matrix block with a single plain text field with a handle of `cacheTags`. Doesn't matter what you name the section, the block type, etc… if there's a matrix field with a handle of `cacheTags` it throws Craft for a loop.

image

The bug is that when you click to create a new entry in the CP it calls `->toArray()` on that new entry (after generating it). The `->toArray()` method then calls `->resolveFields()` on the entry (via [Yii's Arrayable trait](https://github.com/yiisoft/yii2/blob/95b0e6010e3960be4e206f9c89c2a4ae5092b86b/framework/base/ArrayableTrait.php#L125)) and tries to serialize all the fields of the entry. Eventually it reaches the matrix field which kicks off a recursive `->toArray()` call on the matrix block query. The matrix block query then calls its own `->resolveFields()` which sees the `cacheTags` handle and tries to serialize the `cacheTags` field. Unfortunately `MatrixBlockQuery` has its own `getCacheTags` getter which gets called _instead_ of our field because [Yii prioritizes getters](https://github.com/yiisoft/yii2/blob/95b0e6010e3960be4e206f9c89c2a4ae5092b86b/framework/base/Component.php#L137) over behaviors (which is where the custom field behavior _should_ have kicked in).

I'm not sure of the best fix for this. You could blacklist the `cacheTags` handle but that seems heavy handed. Ideally the custom fields would be queried directly via the behavior instead of the magic `__get`?

brandonkelly:
Thanks! I fixed this within `ElementQuery::fields()` so `toArray()` will explicitly pull custom field values from the behavior, rather than giving the magic methods a chance to interfere.

Craft 3.7.66 and 4.3.9 are both out with the fix.

(Technically the root bug existed in Craft 3, even though it wasn’t causing any problems there.)

markhuot:
🎉 this is _fantastic_ thanks! Will get it checked out and tested later today and confirm everything is good.

markhuot:
This is working great in our dev environment! Thank you.

brandonkelly:
Great to hear!

Issue #12631, opened 2023-02-06
[4.x]: Cannot eager load entry fields on matrix blocks


ivanBereznev:
### Description
There is an entry type named `Product` which has an `productAttributes` field. This field contains Matrix Blocks of several types where each type has its own entry field that references entries.
When I want to query products along with all Matrix Blocks with their relations I do the following:
```
$products = (new EntryQuery(Entry::class))
  ->type('product')
  ->with(['productAttributes.typeHandleA:attributeA', 'productAttributes.typeHandleB:attributeB'])
  ->all();
```
Then I do data processing without generating too many DB queries while iterating over products and their attributes.

This worked perfectly with Craft 4.3.1 but does not work with Craft 4.3.7, although data and data model has not been changed in any ways

### Steps to reproduce

1. Have a Matrix field on an entry type and some Entry fields to these block types.
2. Create an Entry of aforementioned type, add some Matrix Blocks and entries related to them.
3. Try to load Matrix blocks and their relations eagerly like in the description above.
4. Access eager loaded fields of Matrix blocks via magic getters or via `getEagerLoadedElements` method.

### Expected behavior
Related entries are loaded eagerly and available, magic getters and `getEagerLoadedElements` return collections of elements.

### Actual behavior
Entries related to matrix blocks are not loaded, magic getters return `ElementQuery` instances. `getEagerLoadedElements` returns blank result.

### Craft CMS version

4.3.7

### PHP version

8.0.22

### Operating system and version

Debian GNU/Linux 11

### Database type and version

PostgreSQL 13.6

### Installed plugins and versions

-

sebastian-lenz:
Actually eager loading of fields on matrix fields is completely broken, please see my comment here:
https://github.com/craftcms/cms/issues/12539#issuecomment-1418959299

i-just:
Hi, thanks for reporting! A new PR to fix this issue has been created.

brandonkelly:
This wasn’t actually fixed in 4.3.8 as intended, but 4.3.8.1 is out now with the proper fix.

brandonkelly:
Note that if you updated to 4.3.8 already, you will be affected by [this bug](https://github.com/craftcms/cms/issues/12647), which was also fixed in 4.3.8.1. If so you will need to run `composer update` to fix.

arentsen:
[...]

PR #12639, opened 2023-02-07
fix eager loading nested fields

i-just:
### Description
PR #12578 caused nested fields not to be eager-loadable. This PR fixes it.

### Related issues
#12631 
#12539

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Release Notes for Craft CMS 3.x
 
+## Unreleased
+
+- Fixed a bug where Matrix blocks weren’t getting eager-loaded. ([#12631](https://github.com/craftcms/cms/issues/12631))
+
 ## 3.7.64.1 - 2023-02-05
 
 - Fixed a bug where `craft\behaviors\CustomFieldBehavior::$owner` was getting nullified by `craft\base\Element::fieldByHandle()`. ([#12624](https://github.com/craftcms/cms/issues/12624))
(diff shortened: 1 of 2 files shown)


myleshyson:
Not sure whether to open up a new issue or not, but this still breaks eagerloading for custom element eagerloading definitions like `with(['children'])`, `with(['books']`, etc. I think you may need to add a check to see if the handle your comparing to an element's field layout is event a field in the first place.

```php
if ($checkIfInLayout) {
                    $filteredElements = array_values(
                        array_filter($filteredElements, function($filteredElement) use ($plan) {
                            $fieldLayout = $filteredElement->getFieldLayout();
                            $fieldHandle = preg_replace('/(\w+)([\.|:]\S+)/', '$1', $plan->handle, 1);

                            if (!Craft::$app->fields->doesFieldWithHandleExist($fieldHandle)) {
                                return $fieldHandle;
                            }

                            return $fieldLayout?->getFieldByHandle($fieldHandle) !== null;
                        })
                    );

                    // if we have nothing left in the $filteredElements, move on to the next iteration
                    if (empty($filteredElements)) {
                        continue;
                    }
                }
```

After adding 

```php
 if (!Craft::$app->fields->doesFieldWithHandleExist($fieldHandle)) {
                                return $fieldHandle;
                            }
```

Eagerloading starts working again like normal.

brandonkelly:
Craft 3.7.65.1 and 4.3.8.1 have been released with a fix for eager-loading attributes like `children`, via #12646.

Issue #12648, opened 2023-02-08
[4.x]: Eager loading broken


vandres:
### Description
I updated to Craft 4.3.8 and 4.3.8.1 afterwards. Since then eager loading stopped working, if the relation is empty or not set in the field layout.

We use Craft Commerce and have products, which can have allergy information. But not every product type has that information and the relation is optional.

```
{% do craft.app.elements.eagerLoadElements(
    className(product),
    [product],
    [
        'allergy',
    ]
) %}
```

Stack trace
```
2023-02-08 13:37:00 [web.INFO] [yii\db\Connection::open] Opening DB connection: mysql:host=db2;dbname=craft;port=3306 {"memory":11241416} 
2023-02-08 13:37:00 [web.INFO] [nystudio107\codeeditor\CodeEditor::bootstrap] CodeEditor module bootstrapped {"memory":15566160} 
2023-02-08 13:37:00 [web.INFO] [wbrowar\adminbar\AdminBar::init] Admin Bar plugin loaded {"memory":20292056} 
2023-02-08 13:37:00 [web.INFO] [bymayo\commercewidgets\CommerceWidgets::init] Commerce Widgets plugin loaded {"memory":23090016} 
2023-02-08 13:37:00 [web.INFO] [mmikkel\cpfieldinspect\CpFieldInspect::init] CP Field Inspect plugin loaded {"memory":23138168} 
2023-02-08 13:37:00 [web.INFO] [nystudio107\retour\Retour::init] Retour plugin loaded {"memory":24656744} 
2023-02-08 13:37:00 [web.INFO] [nystudio107\vite\Vite::init] Vite plugin loaded {"memory":24956544} 
2023-02-08 13:37:00 [web.INFO] [yii\web\Session::open] Session started {"memory":25188688} 
2023-02-08 13:37:00 [web.ERROR] [yii\base\ErrorException:2] yii\base\ErrorException: Undefined array key 0 in /var/www/vendor/craftcms/cms/src/fields/BaseRelationField.php:726
Stack trace:
#0 /var/www/vendor/craftcms/cms/src/web/ErrorHandler.php(79): yii\base\ErrorHandler->handleError(2, 'Undefined array...', '/var/www/vendor...', 726)
#1 /var/www/vendor/craftcms/cms/src/fields/BaseRelationField.php(726): craft\web\ErrorHandler->handleError(2, 'Undefined array...', '/var/www/vendor...', 726)
[...]

PR #12649, opened 2023-02-08
eager loading map - bail if sourceElements is empty

i-just:
### Description
Don’t attempt to get an eager loading map if the $sourceElements array is empty.

### Related issues
#12648


brandonkelly:
Fixed this at the source of the bug, in `craft\base\Element::eagerLoadingMap()`. (52abf8eca5f67053e301aefc39082ae9a14be6d0)

PR #12658, opened 2023-02-09
More explicit size ordering

### LINKED ISSUE #12652: [4.x]: craft asset query order by size not working
### What happened?

for some reason I cannot query assets ordered by size, it's just as it wouldn't consider it actually, not ordered at all, both in frontend query and also in the backend for the volumes
every other order works, by name, by date, works, but not by size
anyone an idea what this could be, don't find any hint online, please help
also I looked in the database, there the size looks correct and there I also can sort the table by size, so I don't know why craft can't

### Craft CMS version

4.3.x

### PHP version

8.1

_No response_

### Database type and version

_No response_

### Image driver and version

_No response_

-

brianjhanson:
### Description
Explicitly orders by `asset.size` just in case users have a field with a handle of `size`. Reserved handles don't help us here because the `columnMap` used by `orderBy` uses all columns on the `content` table, so any entry type with a `fieldHandle` of `size` will cause issues. 

### Related issues
Fixes #12652

### DIFF
--- a/src/elements/Asset.php
+++ b/src/elements/Asset.php
@@ -402,7 +402,11 @@ class Asset extends Element
         return [
             'title' => Craft::t('app', 'Title'),
             'filename' => Craft::t('app', 'Filename'),
-            'size' => Craft::t('app', 'File Size'),
+            'size' => [
+                'label' => Craft::t('app', 'File Size'),
+                'orderBy' => 'assets.size',
+                'defaultDir' => 'desc',
+            ],
             [
                 'label' => Craft::t('app', 'File Modification Date'),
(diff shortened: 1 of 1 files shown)


brandonkelly:
This feels a little too lighthanded. It solves this one guy’s problem, but the same could still happen from any of the other asset-specific sortable attributes that don’t specify the table name in `orderBy` (or any other element type, for that matter).

So I just beefed up element queries’ automatic column name mapping for 4.4 via 74211f8637db0e5fa661d351df98c915c7b94058. Going forward, all element extension table columns will be included automatically from `joinElementTable()` (unless the column name is already included in the mapping).

PR #12660, opened 2023-02-09
Add an edit button to elements and remove double-click-to-open from element index and element select components

gcamacho079:
### Description
See title

### Related issues
Resolves DEV-463, DEV-1084


mmikkel:
I understand the need for an Edit button, but does the double-click have to be removed? Can’t we have both? 🙂

brandonkelly:
Agree it makes sense to keep double-click support – especially if we’re considering a disclosure menu instead of an edit button in some cases (assets).

PR #12684, opened 2023-02-16
Set a default element type property for all element condition builders

lukeholder:
### Description

This PR adds a default $elementType value to each element condition builder. This is useful, as passing it into the constructor is no longer required.

### Related PR

https://github.com/craftcms/cms/pull/12601


brandonkelly:
Feels like the time to do this was before 4.0 shipped, when we could have overridden the extended conditions’ constructors to remove the `$elementType` argument altogether.

But either way this doesn’t really provide any DX improvement, since you already don’t have to specify an element type if you instantiate the conditions via `MyElementType::createCondition(0)`.

Issue #2911, opened 2018-05-22
FR: Consider globals/global sets refactor


timkelty:
Wanted to get my 2 cents in here, hoping that this might get considered with some of the other big-picture refactoring efforts (https://github.com/craftcms/cms/issues/1524)

I find globals to be quite cumbersome, particularly the levels of forced taxonomy/hierarchy it imposes, just to expose simple data.

For example, out of the gate you're forced to make a global set, which means you will likely end up with some sort of meaningless catch-all like "General". Beyond that, you also need to create a tab grouping, which is probably going to be another meaningless label like "Content" or something.

I wouldn't be as opposed to this forced hierarchy if it were merely presentational, and you could change it and your content evolved. However, globals are called explicitly by their global set handle, so and reorganization is going to require a code refactoring.

As Craft 3 has embraced multi-site, one possible direction could be to treat Sites as elements, with the ability to add fields/field layouts, instead of "globals" being a totally separate concept (discussed here: https://github.com/craftcms/cms/issues/2790)

Related globals/global set issues:
- https://github.com/craftcms/cms/issues/2682
- https://github.com/craftcms/cms/issues/916
- https://github.com/craftcms/cms/issues/2790

khalwat:
Having sites be elements with their own field layouts sounds pretty awesome. I'm not sure what impact that has under the hood, though. Because elements inherently can be localized based on... sites.

So it might be a bit of a weird infinite loop unless sites were then special-cased in some manner. Which then defeats the purpose of them being elements.

brandonkelly:
What if you still want to store actual global content that should be consistent across all sites?

timkelty:
> What if you still want to store actual global content that should be consistent across all sites?

I imagine you'd _not_ mark the field as "translatable", same as would for an entry field (or even globals, as they exist now).

brandonkelly:
Sorry I assumed you meant that each site should get its own field layouts as well (as globals do), but maybe that’s not the case.

Making sites elements, with a Site field layout, has been suggested before, but one issue with that is that sites are an admin-only thing. So we’d need to add some new, separate place to manage their field content. At which point, I’m not sure how it’s any different than the Globals other than the name.

brandonkelly:
[...]

PR #12689, opened 2023-02-17
Category/tag/global set entrification

brandonkelly:
### Description

Adds new entrification commands:

- `entrify/categories`
- `entrify/tags`
- `entrify/global-set`

Plus some supplemental section maangement commands:

- `sections/create`
- `sections/delete`

Here’s what the workflow for converting categories/tags/global sets to sections and entries will look like:

1. Run the `entrify/*` command(s) locally.

   ```sh
   php craft entrify/categories myCategoryGroupHandle
   ```

   The command will walk you through creating a new section based on the category/tag group or global set, convert the elements to entries in that section, and optionally also delete the old category/tag group or global set, and convert Categories/Tags fields over to Entries fields.

2. Update your templates, custom plugin/module code, and GraphQL queries accordingly.
3. Deploy the changes to another environment.
4. Run the same command immediately after deployment (after applying project config changes). This time, the new sections will already be in place, and the category/tag groups and global sets will be soft-deleted. So the command will only need to convert the actual category/tag/global set elements to entries.

> **Warning**
> **Do not** hand-delete your old category/tag groups or global sets via garbage collection before step 4. Doing so would also remove the category/tag/global set elements, at which point the content would be lost forever.

### Related issues

- #9781
- #8748
- #10560
- #8104
- #2911


vnali:
@brandonkelly this is great! 
I tried to test it but I received this error after choosing 'yes' on the 'Delete the “xyz” category group?' step on the local environment.
```
Exception 'yii\base\Exception' with message 'Unable to write new project config files'
in /var/www/html/craft4live/vendor/craftcms/cms/src/services/ProjectConfig.php:1608
Caused by: PHP Warning 'yii\base\ErrorException' with message 'rmdir(/var/www/html/craft4live/config/project/categoryGroups): Directory not empty'

```
i double-checked permission on the config/project folder and it is writable. 
And just for clarification, before deleting category group step, I can see the categoryGroups folder, and after choosing yes and this error, the categoryGroups folder is not in the project folder. although I can see a hidden folder like ..xyz containing category group .yaml file

brandonkelly:
@vnali Weird… it’s gotta be some sort of permission thing. PHP is probably being executed as a different user for CLI requests than it is for web requests. Can you try switching to whatever user account web requests are running as first, via

```sh
sudo su - username
```

vnali:
@brandonkelly  thanks! I wasn't sure how to switch to another user on my environment but changing permission on the config folder to 777 worked, so definitely it was a permission issue.

FrDH:
I had some issues with converting a global set, all went well local, but after pushing to staging I ran the command and got this error:

```
[me]$ php craft entrify/global-set meldingen --section=instellingen
 → Converting “Meldingen” … ✕
   Error: Changes to the project config are not possible while in read-only mode.
```

On staging:
- the single was created, but remained empty
- the global set was not removed, but emptied

brandonkelly:
@FrDH Did you run `craft project-config/apply` first? If the new section has already been created via that command (or `craft up`), then it should work as intended.

jevans1808:
I just want some clarification on this entrification process. I've run the commands locally, and taken a record of the commands to be run on the server. I've committed my changes to my repo. Once I have deployed, do I run the commands immediately, or do I apply yaml changes, and then run the commands? If I've deleted the categories groups locally, whilst following the prompts thrown by entrification (e.g. "Delete the 'Blog' category group? Yes") and committed these changes to the yaml file, will this cause any issues or conflicts? Or would the categories themselves still exist to be converted to entries, despite the deletion of the group?

Thanks

brandonkelly:
@jevans1808 You’ll run `craft up` like you normally would first, which will apply the project config changes (including deleting the category group), and then run the `entrify` command. Even though the category group was deleted, your categories will still exist in the database (soft-deleted), so the command is still able to convert them to entries.

jevans1808:
> @jevans1808 You’ll run `craft up` like you normally would first, which will apply the project config changes (including deleting the category group), and then run the `entrify` command. Even though the category group was deleted, your categories will still exist in the database (soft-deleted), so the command is still able to convert them to entries.

Awesome, thank you for the clarification!

PR #12694, opened 2023-02-17
Make non-global table row type names unique

brianjhanson:
If multiple table fields outside of the global context share a handle the type resolution process will end up smushing them all together into a single `{handle}_TableRow` graphql type.

This avoids that issue by making sure each table field within a nested context will output a unique name.

I'd really like the name generated here to be `{matrixHandle}_{handle}_TableRow` but as far as I can tell the matrix handle isn't easily accessible from the table field. This seems like the kind of problem [generators](https://craft-docs.netlify.app/docs/4.x/extend/graphql.html#generators) are intended to solve, but those only apply to elements and table fields are more of a quasi element field.


brandonkelly:
Wouldn’t this be a breaking change for existing queries?

brianjhanson:
Yeah, it should probably target 5 just to be safe. Currently both with and without the `... on` will work for table fields, so I'm not sure how many users are typing them, but it's probably not 0. 

### No explicit type
```graphql
{
  entries(section: "navigation") {
    id
    canonicalId
    title
    ... on navigation_default_Entry {
      navigation {
        ... on navigation_products_BlockType {
          buttons {
            title
            url
          }
        }
      }
    }
  }
}
```

### Explicit type (new syntax)
```graphql
{
  entries(section: "navigation") {
    id
    canonicalId
    title
    ... on navigation_default_Entry {
      navigation {
        ... on navigation_products_BlockType {
          buttons {
            ... on buttons_122_TableRow {
              title
              url
            }
          }
        }
      }
    }
  }
}
```

brianjhanson:
Turns out this isn't viable because IDs can change based on environments.

PR #12698, opened 2023-02-18
Preload Single entry variables

brandonkelly:
Preloads Single entries referenced by front-end templates by their handle.

For example, if you have a Single section with the handle `footer` and somewhere in your template you have:

```twig
{{ footer.copyrightNotice }}
```

then `footer` will be automatically available, even if the template didn’t explicitly fetch it.

Single entries will be queried in batches for each template. So if a template references 3 Single handles that hadn’t been preloaded by prior templates yet, those 3 entries would be fetched together in a single entry query, and memorized in case any future templates reference the same entries later on in the request.

### DIFF
--- a/CHANGELOG-WIP.md
+++ b/CHANGELOG-WIP.md
@@ -54,6 +54,7 @@
 - Added the `editable` and `savable` asset query params. ([#12266](https://github.com/craftcms/cms/pull/12266))
 - Added the `savable` entry query param. ([#12266](https://github.com/craftcms/cms/pull/12266))
 - The `editable` entry query param can now be set to `false` to only show entries that _can’t_ be viewed by the current user. ([#12266](https://github.com/craftcms/cms/pull/12266))
+- Front-end Twig templates now automatically preload any Single section entries that are referenced by their handle within the template. ([#12698](https://github.com/craftcms/cms/pull/12698))
 - Added the `{% dump %}` tag, which dumps variables into a new “Dumps” Debug Toolbar panel. ([#12506](https://github.com/craftcms/cms/pull/12506))
 - Added the `base64_decode` and `base64_encode` Twig filters. ([#12696](https://github.com/craftcms/cms/discussions/12696))
 - The `dump()` Twig function now utilizes `Craft::dump()`, and no longer requires Dev Mode to be active. ([#12486](https://github.com/craftcms/cms/pull/12486), [#12479](https://github.com/craftcms/cms/discussions/12479))
@@ -124,6 +125,7 @@
 - Added `craft\services\Elements::deleteElementsForSite()`.
 - Added `craft\services\Elements::EVENT_AFTER_DELETE_FOR_SITE`. ([#12354](https://github.com/craftcms/cms/issues/12354))
 - Added `craft\services\Elements::EVENT_BEFORE_DELETE_FOR_SITE`. ([#12354](https://github.com/craftcms/cms/issues/12354))
+- Added `craft\services\Entries::getSingleEntriesByHandle()`. ([#12698](https://github.com/craftcms/cms/pull/12698))
 - Added `craft\services\Fields::getFieldsByType()`. ([#12381](https://github.com/craftcms/cms/discussions/12381))
 - Added `craft\services\Path::getImageTransformsPath()`.
 - Added `craft\services\Search::normalizeSearchQuery()`.
--- a/src/helpers/Template.php
+++ b/src/helpers/Template.php
@@ -339,4 +339,18 @@ class Template
         // Template check copied from twig_var_dump()
         return array_filter($context, fn($value) => !$value instanceof TwigTemplate && !$value instanceof TemplateWrapper);
     }
+
+    /**
+     * Preloads Single section entries into the given context.
+     *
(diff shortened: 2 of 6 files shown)


khalwat:
Does it potentially make sense to namespace these, e.g.: 

```twig
     {{ singles.footer.copyrightNotice }}
```

I get that the intent is to allow things that were Global Sets to "just work" like they did previously... I just worry about namespace collisions (and yes, I see you are trying to avoid them in your PR, see below).

Probably I'm being overly paranoid, but I spent a good bit of time recently debugging an SEOmatic issue someone was having recently, and it ended up being that they created a Global Set named `SEOmatic` (which had a handle of `seomatic`), which was then stomping on the `seomatic.` variable that the plugin injects into Templates.

I worry that if every single on a site (on some sites, this could be a whole lot of them), it could result in a whole lot of additional variables added to the Twig context, which (depending on what they are named) could stomp on existing Craft variables, or plugin variables, or the like.

I do see that you're doing an `array_diff` that would preserve any existing variables, presumably in an attempt to mitigate this, but then you'd have some of these preloaded singles not being present and mysteriously not working.

Another option to consider is instead of just loading all Singles into the context as queries, in the Section settings, you could let people choose which Singles should be loaded... and you could have an opt-in setting on the Singles themselves which say "always load me", which would be set by default for Global Sets converted over to Singles.

This would allow templates to continue working as they did when they were Globals, but not pollute the namespace with _every_ single.

Bonus points for an "eager load" checkbox which would eager load the query.

Tangent: Some way to specify what is eager loaded for these queries, and other route variables like `entry.` that are automatically injected is something I've always wanted.

mmikkel:
> Another option to consider is instead of just loading all Singles into the context as queries, in the Section settings, you could let people choose which Singles should be loaded... and you could have an opt-in setting on the Singles themselves which say "always load me", which would be set by default for Global Sets converted over to Singles.

+1000%. Not only is this a good idea and would be a useful feature in its own right going forward, but I also share @khalwat's concerns here - especially considering that this PR is slated for a minor release.

A per-single "Preload" setting, which is disabled by default for existing and new singles (excepting entryfied globals), would be much appreciated.

khalwat:
There's also the possibility of things mysteriously working, too.

If you had a variable in Twig you were using like `computedStuff` and you also had a Single named `Computed Stuff` it would be pre-populated with an element query, and maybe somewhere you check whether it's empty or not and boom, now it's not empty so you skip initialization.

Just a contrived scenario, but I feel like there may end up being a number of similar scenarios that could happen if the context is pre-populated with a whole bunch of single handles.

mmikkel:
I'm a big fan of globals and typically create a bunch of them in every project. 

More than once, I've even created singles that are basically stand-ins for global sets - sometimes because I specifically _didn't_ want them injected in every template.

I've also never been particularly concerned about my singles' handles, because it wasn't possible to predict that they would some day all be injected into every single template.

But with this PR, if I have a single with the handle `contactInfo`, and an unrelated template partial that also happens to look for a variable called `contactInfo`, that - by design - may or may not have existed in the partial's context previously, and so the partial has a null coalescing operator to guard against undefined errors... that partial is now going to break, in all cases where the `contactInfo` variable hasn't been declared somewhere further up the Twig include stack, because now the `contactInfo` variable is the single entry, and not whatever object the partial was originally designed to deal with:  

```twig
{% set contactInfo = contactInfo ?? null %}

{% if contactInfo %}
   ...
{% endif %}
```

I don't believe the above is an uncommon pattern.

Singles are potentially even going to be injected into templates that have been included specifically _without_ the current Twig context (i.e. `with_context=false`). That's the way it works currently for global sets, so it has to work that way for entryfied global sets as well – but it makes it even harder to guard against problems like the above, when it'll happen even for single entries that were never intended to be global in the first place.

A "Preload" (or "Make global") section setting for singles (like @khalwat suggests) would be the perfect solve, in my opinion:

- Makes the new behavior completely opt-in; i.e. nothing changes unless the developer wants it to
- Makes singles even more flexible, vs. the deprecated global sets (i.e. the use case of having a "global set" that is *not* preloaded and automatically injected, is now solved)

aelvan:
Everything @khalwat and @mmikkel said. This seems like a massive can of worms.

brandonkelly:
Good points. Reverted this change for now. We’ll consider alternative approaches in another release.

brandonkelly:
4.4.0-beta.4 is out with that change.

brandonkelly:
@khalwat @mmikkel @aelvan Just PR’d an alternate approach, which I think does a good job of addressing your concerns while still keeping the DX improvement attempted here: #12787

It’s pretty similar to this one – templates still track which Single handles are referenced as variables, and ensure that each of those Singles have been preloaded & memorized at the beginning of the template rendering. The main difference is that the singles no longer enter the `_context` whatsoever (unless explicitly defined by the template, of course). They are only ever pulled in as a fallback value, via a new `craft\helpers\Template::fallback()` method, right inline where the `$_context` array is being referenced.

Would love to hear your thoughts on this approach! I personally think it’s a pretty good compromise.

khalwat:
I have to say, that's an incredibly clever way to handle it @brandonkelly

PR #12699, opened 2023-02-18
Use DI for cookie config

### LINKED ISSUE #12531: [4.x]: User photos volume not recognized as valid
### Description
With a valid volume (containing an uploaded image) set in user settings, I'm still getting "Please set a valid volume for storing the user photos in user settings page first." message under my user account.

Screen-Shot-2023-01-12-at-1 40 31-PM
Screen Shot 2023-01-12 at 1 40 11 PM
Screen-Shot-2023-01-12-at-1 40 45-PM

This is currently a Servd assets volume, but I have also tested with a local assets volume with the same results.

### Steps to reproduce

1. Create an assets filesystem/volume
2. Assign the volume to user photo location in user settings

### Expected behavior
User photo can be chosen/uploaded in user preferences

### Actual behavior
User preferences panel displays "Please set a valid volume for storing the user photos in user settings page first."

### Craft CMS version

Craft Pro 4.3.6.1

### PHP version

8.1.13

Linux 5.15.68-0-virt

### Database type and version

MySQL 10.4.27

### Image driver and version

Imagick 3.7.0 (ImageMagick 6.9.11-60)

    "craftcms/cms": "4.3.6.1",
    "craftcms/redactor": "^3.0",
    "doublesecretagency/craft-cpcss": "^2.6",
    "mmikkel/retcon": "^2.6",
    "nystudio107/craft-imageoptimize": "^4.0",
    "nystudio107/craft-minify": "4.0.0-beta.2",
    "nystudio107/craft-retour": "^4.1",
    "nystudio107/craft-seomatic": "4.0.19",
    "servd/craft-asset-storage": "^3.2",
    "spicyweb/craft-neo": "^3.5",
    "verbb/cloner": "^2.0",
    "verbb/field-manager": "^3.0",
    "vlucas/phpdotenv": "^5.4.0"

timkelty:
Instead of requiring the passing around of a common cookie config, we should use Yii's DI container to load our cookie config anywhere `\yii\web\Cookie` is instantiated.

Related: https://github.com/craftcms/commerce/pull/3091

### DIFF
--- a/bootstrap/bootstrap.php
+++ b/bootstrap/bootstrap.php
@@ -251,6 +251,8 @@ $config = ArrayHelper::merge(
 /** @var \craft\web\Application|craft\console\Application $app */
 $app = Craft::createObject($config);
 
+Craft::bootstrapContainer();
+
 // If there was a max_input_vars error, kill the request before we start processing it with incomplete data
 if ($lastError && strpos($lastError['message'], 'max_input_vars') !== false) {
     throw new ErrorException($lastError['message']);
--- a/src/Craft.php
+++ b/src/Craft.php
@@ -192,7 +192,7 @@ class Craft extends Yii
     public static function bootstrapContainer(): void
     {
         self::$container->setDefinitions([
-            \yii\web\Cookie::class => \Craft::cookieConfig()
+            \yii\web\Cookie::class => \Craft::cookieConfig(),
         ]);
     }
 
--- a/src/Craft.php
+++ b/src/Craft.php
@@ -173,7 +173,9 @@ class Craft extends Yii
                     $request = static::$app->getRequest();
                 }
 
-                $generalConfig->useSecureCookies = $request->getIsSecureConnection();
+                if (!$request->isConsoleRequest) {
+                    $generalConfig->useSecureCookies = $request->getIsSecureConnection();
+                }
             }
 
             self::$_baseCookieConfig = [
@@ -185,6 +187,13 @@ class Craft extends Yii
         }
 
         return array_merge(self::$_baseCookieConfig, $config);
+    }
+
+    public static function bootstrapContainer(): void
+    {
+        self::$container->setDefinitions([
+            \yii\web\Cookie::class => \Craft::cookieConfig()
+        ]);
     }
 
     /**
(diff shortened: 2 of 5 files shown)


bootstrap/bootstrap.php
```diff
@@ -251,6 +251,8 @@
 /** @var \craft\web\Application|craft\console\Application $app */
 $app = Craft::createObject($config);
 
+Craft::bootstrapContainer();
+
```
timkelty, line 255: This needs to be _after_ we initialize `$app`, as the config we are using for DI may reference Craft's config component.

src/Craft.php
```diff
@@ -173,7 +173,9 @@ public static function cookieConfig(array $config = [], ?Request $request = null
                     $request = static::$app->getRequest();
                 }
 
-                $generalConfig->useSecureCookies = $request->getIsSecureConnection();
+                if (!$request->isConsoleRequest) {
+                    $generalConfig->useSecureCookies = $request->getIsSecureConnection();
+                }
             }
```
timkelty, line 179: Required check since the this method will now be called during bootstrap. Alternatively we could only set the DI definition for `Cookie` on web requests, but I preferred just adding this check.

src/Craft.php
```diff
@@ -187,6 +189,13 @@ public static function cookieConfig(array $config = [], ?Request $request = null
         return array_merge(self::$_baseCookieConfig, $config);
     }
 
+    public static function bootstrapContainer(): void
+    {
+        self::$container->setDefinitions([
+            \yii\web\Cookie::class => \Craft::cookieConfig()
+        ]);
+    }
+
```
timkelty, line 198: So it can be shared between app and test bootstraps.

tests/CraftTest.php
```diff
@@ -61,4 +63,14 @@ public function testParseEnvReturnsFalse(): void
...
+
+    public function testDependencyInjection(): void
+    {
+        $cookie = Craft::createObject(Cookie::class);
+        $cookieConfig = Craft::cookieConfig();
+
+        foreach ($cookieConfig as $property => $value) {
+            $this->assertEquals($cookie->$property, $value);
+        }
+    }
```
timkelty, line 75: I have no idea where this test should go…so I put it here for now.

PR #12719, opened 2023-02-22
Add `string` @var type to `orderBy`

bencroker:
This PR adds a `string` @var type to the `ElementQuery::orderBy` PHPDoc to keep PHPStan happy, since its parent specifies `array|null`.
https://github.com/yiisoft/yii2/blob/f388ca71b08b89e940f1ffbe4afa19ae9d5e115f/framework/db/QueryTrait.php#L39-L47

```php
if (empty($elementQuery->orderBy) || !is_array($elementQuery->orderBy)) {
    return false;
}
```

Screenshot 2023-02-21 at 19 24 32

### DIFF
--- a/src/elements/db/ElementQuery.php
+++ b/src/elements/db/ElementQuery.php
@@ -354,6 +354,7 @@ class ElementQuery extends Query implements ElementQueryInterface
 
     /**
      * @inheritdoc
+     * @var array|string|null
      * @used-by orderBy()
      * @used-by addOrderBy()
      */


brandonkelly:
This is causing an issue with our own PHPStan checks: https://github.com/craftcms/cms/pull/12719/files#diff-f94272123a3a550345c7f794061fd3664d8cff428758179dad3cfe3cc89eaef2R361

bencroker:
Yeah I see why that might be. I'll just add a `@phpstan-ignore-next-line` to my code. Thanks for looking at this!

PR #12722, opened 2023-02-22
Structure view mode

brandonkelly:
### Description

Introduces a dedicated element index view mode for structured element sources, rather than including “Structure” as one of the Table mode’s sort options.

The view mode listbox for a Structure section, showing two options – a structured table view (selected) and a normal table view.

When a column is activated to be sorted by, the view mode listbox will switch to the table view, indicating that the results will no longer be structured.

Temporary view state changes like searching will not immediately update the view mode listbox, as the structure view will resume once the search query is cleared. (If the sorting attribute is changed mid-search, that will in fact visually switch the view mode, though.)

### Related issues

- #12718


mmikkel:
This change is a good call 👏 

Small bug: Sometimes the new structure/table toggle changes position in the element index toolbar when switching back and forth between structure and non-structure sources. It can be reliably reproduced by doing a fresh page reload on a structure source, then switching to a channel source and finally back to the structure source again.

https://user-images.githubusercontent.com/298510/220755597-8a33ab10-9cd4-465e-aeb4-c193bbe0af6e.mp4

brandonkelly:
Thanks! I was able to reproduce that on the Assets index page on 4.3 as well, so just fixed it for the next 4.3 and 4.4 releases. (c7c9cac9dad2d7c2b4c23f2e09daf76d01b03a2f)

PR #12724, opened 2023-02-23
Updates asset alt text to use getThumbAlt

gcamacho079:
### Description
...and updates `getThumbAlt` for instances where the file type icons are used instead of the asset itself. This fixes instances where generic file type icons used the asset Alternative Text.

In addition, it removes the use of the asset title from the alt attribute as it’s considered a [generic/irrelevant string under ATAG](https://www.w3.org/TR/2015/NOTE-IMPLEMENTING-ATAG20-20150924/#sc_b232).

### Related issues

Also resolves DEV-588


src/web/assets/cp/src/js/ElementThumbLoader.js
```diff
@@ -129,7 +129,7 @@ Craft.ElementThumbLoader.Worker = Garnish.Base.extend({
     var $img = $('<img/>', {
       sizes: $container.attr('data-sizes'),
       srcset: $container.attr('data-srcset'),
-      alt: $container.attr('data-alt') || '',
+      alt: $container.attr('data-alt') || null,
```
brandonkelly, line 132: @gcamacho079 Isn’t it important for `alt` attributes to explicitly be set to `""` if there’s no alt text?

gcamacho079: > @gcamacho079 Isn’t it important for `alt` attributes to explicitly be set to `""` if there’s no alt text?

The null attribute (`alt=""`) explicitly conveys to screen reader users that [the image is decorative and can safely be ignored](https://www.w3.org/WAI/WCAG21/Techniques/html/H67). In the context of the control panel, the image thumbnail (or its alternative) should be considered essential since it provides important information about the asset that isn't communicated by just the asset title.

This could also provide users with another way of finding missing alt text inside the control panel (other than using the condition builder to surface images without alt text). An image with `alt=""` won't be flagged as an accessibility issue using an automated testing tool; it requires a manual test to verify whether or not the image is actually decorative. But an image lacking the alt attribute will be flagged on an automated accessibility test.

brandonkelly: Yeah but setting it to `null` is preventing an `alt` attribute from being added at all:

```html
<img sizes="34px" srcset="/assets/site/_34x32_crop_center-center_none_ns/skateboard.jpg?v=1423589079 34w, /assets/site/_68x65_crop_center-center_none_ns/skateboard.jpg?v=1423589079 68w">
```

gcamacho079: @brandonkelly that’s what I was hoping for. Right now, preventing it from being added to the `img` tag seems to be the best solution. Here are the three ways of handling the missing Alternative Text and their drawbacks/solutions:

1. **Removing the alt attribute from the image entirely.** Of course, this will introduce an accessibility violation, but this seems to mesh with the ATAG concepts of not attempting to repair missing text alternatives and truthfully reflecting content accessibility problems (as noted in the [Part A Conformance Applicability Notes](https://www.w3.org/TR/IMPLEMENTING-ATAG20/#partA-tool-accessible-notes)). 
In those notes, it's stated that "...where an [authoring tool user interface accessibility problem](https://www.w3.org/TR/IMPLEMENTING-ATAG20/#def-Authoring-Interface-Accessibility-Problem) is caused directly by the content being edited (e.g. if an image in the content lacks a text alternative), then this would not be considered a deficiency in the accessibility of the [authoring tool user interface](https://www.w3.org/TR/IMPLEMENTING-ATAG20/#def-Authoring-Tool-Interface)"
2. **Marking the image as decorative using the null alt (`alt=""`)**. Because the image provides essential information about the asset, this choice seems misleading. It fixes the accessibility error but, in the process, conveys the false message to screen reader users that the contents of the image are unimportant.
3. **Using a generic alternative, like the asset title.** This goes against the [ATAG Success Criterion B.2.3.2 Automating Repair of Text Alternatives](https://www.w3.org/TR/2015/REC-ATAG20-20150924/#sc_b232). While I think this SC is more related to the "repair" of the alt text in front-end templates ([like what we fixed in this PR](https://github.com/craftcms/cms/pull/12854)), we should probably consider this SC in the context of the control panel as well; using generic strings doesn’t increase usability for screen reader users, it just makes the content redundant.

brandonkelly: 🤦🏻 Sorry, I completely misread your last comment. Got it now!

PR #12732, opened 2023-02-24
Wrap migration folder create in condition

timkelty:
@brandonkelly after some more investigation, it isn't quite this simple, as other `migrate` commands will still fail if we sidestep here. Notes ↓


src/console/controllers/MigrateController.php
```diff
@@ -203,8 +204,10 @@ public function beforeAction($action): bool
                 }
             }
 
-            $this->migrationPath = $this->getMigrator()->migrationPath;
-            FileHelper::createDirectory($this->migrationPath);
+            if (!App::isEphemeral()) {
+                $this->migrationPath = $this->getMigrator()->migrationPath;
```
brandonkelly, line 208: Seems like this should still live outside the condition?

src/console/controllers/MigrateController.php
```diff
@@ -202,7 +202,6 @@ public function beforeAction($action): bool
             }
 
             $this->migrationPath = $this->getMigrator()->migrationPath;
-            FileHelper::createDirectory($this->migrationPath);
```
timkelty, line 205: I'm pretty sure this is superfluous, as later we call `parent::beforeAction`, which already tries to create this dir.

src/console/controllers/MigrateController.php
```diff
@@ -211,8 +210,12 @@ public function beforeAction($action): bool
...
 
-        if (!$this->traitBeforeAction($action)) {
-            return false;
+        try {
+            if (!$this->traitBeforeAction($action)) {
+                return false;
+            }
+        } catch (InvalidConfigException $e) {
+            return true;
         }
```
timkelty, line 219: `$this->traitBeforeAction` calls `parent::beforeAction`, which throws `InvalidConfigException` if it can't create the dir.

Technically it will throw `yii\base\Exeption` if this is the `create` action, but I don't think we care. If you can't create the dir, you obviously wont be able to create a migration file.

brandonkelly: If we are catching the wrong exception, why `try/catch` it in the first place?

timkelty: Not the wrong exception, just some exception silliness here from Yii:

When `migrations` can't be created:

- `migrate/create` will throw `yii\base\Exception`: we're not bothering catching this, because we can't create a migration on a RO fs.
- `migration/*` (anything besides `create`) will throw `InvalidConfigException`, which we are catching

Issue #12728, opened 2023-02-23
[4.4.0-beta.2]: Dropdown Fields overlap if not set to full width 


wsydney76:
### Description

We have three dropdown fields set to 25% width.

They overlap now:

![image](https://user-images.githubusercontent.com/60010597/220953670-62d26f11-2ef9-415c-8916-5286952db80b.png)

EDIT:

Dropdowns do not work in Live Preview, only text input possible.

### Steps to reproduce

See above.

### Expected behavior

Do not overlap.

### Actual behavior

See image

### Craft CMS version

4.4.0-beta-2

### PHP version

8.1.13

### Operating system and version

Linux 5.15.79.1-microsoft-standard-WSL2

### Database type and version

MySQL 10.4.27

### Image driver and version

Imagick 3.7.0 (ImageMagick 6.9.11-60)

### Installed plugins and versions

-

i-just:
Hi, thanks for reporting! I created a PR to address both issues.

brandonkelly:
Fixed now for the next release, via #12734!

brandonkelly:
4.4.0-beta.3 is out with that fix.

PR #12734, opened 2023-02-24
Bugfix/12728 selectize dropdown amends

i-just:
### Description
When you have multiple dropdowns in a row (e.g. 3 dropdowns, each set to occupy 25% width), if your browser window is from 1536px to roughly above 2100px wide, the dropdowns will overlap. Fix: set `max-width` instead of `width`.

Now that the dropdown fields are selectized, they stopped working in Preview. 
For dropdown fields, we’re setting the `dropdownParent: 'body'`, and it’s what’s breaking it in the Preview in 4.4.
Setting it to `null` (or removing that setting, as `null` is the default option) seems to work fine in and outside the Preview. 
I don’t know the reason behind explicitly using `'body'`, so I’m not sure if this is a good idea?

For context: multi select uses the default `dropdownParent` (as in `null`) and works in and out of the Preview.

### Related issues
#12728


brandonkelly:
`dropdownParent: 'body'` is needed to ensure the menu doesn’t get cut off when placed in a container with `overflow: hidden`. So brought those back, and fixed the menu z-index so it’s visible within Live Preview.

Issue #12749, opened 2023-02-27
[4.4.0-beta.3]: a volume get indexed by 'all' option, even if that volume is not listed by EVENT_LIST_VOLUMES


vnali:
### Description
because 'all' option is always available in volumes list in the asset index utility page,  by selecting that option, all volumes are getting indexed, even if a volume is not passed by EVENT_LIST_VOLUMES

i-just:
Hi, thanks for reporting! I created a PR to address this.

mmikkel:
Thanks a lot @i-just! 🙌

brandonkelly:
4.4.0-beta.4 is out with the fix for this, via #12762.

PR #12762, opened 2023-03-01
asset indexing by "all" to account for EVENT_LIST_VOLUMES

i-just:
### Description
When using Utilities > Asset Indexes and choosing to index by “All”, get all listed volume ids instead of relying on `*` being passed to account for volumes that could have been excluded by the new `EVENT_LIST_VOLUMES` event.

### Related issues
#12749


vnali:
@i-just Thanks! Maybe also Craft should check passed volumes on actionStartIndexing action if they are in $event->volumes list and the user hasn't manipulated input values?

brandonkelly:
@vnali Can you post a new discussion and explain the use case for what you need?

PR #12787, opened 2023-03-04
Feature/preload singles

brandonkelly:
### Description

Ensures any Single entries are preloaded & memoized at the beginning of each template, if their handles are referenced as variables anywhere in the template.

Then, wherever their handles are referenced, the Single entries will be swapped in as fallback values, _if and only if_ the variable is otherwise undefined at that point in the template.

For example, if you have a Single section with the handle `footer` and somewhere in your template you have:

```twig
{{ footer.copyrightNotice }}
```

then `footer` will be automatically available, even if the template didn’t explicitly fetch it.

However if the template did this:

```twig
{% set footer = null %}
{{ footer.copyrightNotice }}
```

you’d get a Twig error, because `footer` is `null` and does not include a `copyrightNotice` property.

Additionally, the Single entries will never touch the variable context. So `{{ dump(_context) }}` would not include the Single entry handles unless they had been explicitly defined by the template.

### Related issues

- #12698


bencroker:
Will `craft.app.elements.eagerLoadElements()` continue to work with this?

mmikkel:
From testing a little bit, it looks like the `is defined` test is still potentially problematic:

```twig
{% if contactInfo is defined %}
    ...
{% endif %}
```

In Craft 4.3, the above conditional will evaluate to `false`, unless there's a variable `contactInfo` in the template's context. In Craft 4.4, it will evaluate to `true` even if there _isn't_ a `contactInfo` variable in the context, but the site happens to be a single with the handle `contactInfo`.

I assume this is an oversight in the new PR and that this potential BC can be mitigated.

To be completely honest though, I still feel like this feature just isn't a good idea overall, and especially not for a minor release. It feels like too big a change and I'm not confident that all bases are truly covered in terms of it not introducing (potentially silent) errors on a lot of sites.

Curious: What's the reason for not wanting to go with a "Preload" (or "Make global") section setting that was proposed in [#SADSD](https://github.com/craftcms/cms/pull/12698#issuecomment-1452025096)? I appreciate the DX effort here, but from my POV this seems like a prime candidate for an opt-in approach; especially considering the fact that many sites won't have any use for the feature in the first place as entryfying globals is going to be a manual effort for the time being.

brandonkelly:
@bencroker Yep, although there wouldn’t be any point in calling that for a single; you could just set `with` params when fetching individual Matrix/relational fields off of the single, for the same performance gain.

brandonkelly:
> From testing a little bit, it looks like the `is defined` test is still potentially problematic:
> 
> 
> 
> ```twig
> 
> {% if contactInfo is defined %}
> 
>     ...
> 
> {% endif %}
> 
> ```
> 
> 
> 
> In Craft 4.3, the above conditional will evaluate to `false` if there isn't a variable `contactInfo` defined in the template's context. But with this PR, it will evaluate to `true` even if there _isn't_ a `contactInfo` variable in the context, but there happens to be a single with that same handle.
> 
> 
> 
> I assume this is an oversight in the new PR and that this potential BC can be mitigated.

That was intentional (`is defined` checks and normal variable references were each implemented separately), but I could be persuaded to remove the fallback for `is defined` checks.

> To be completely honest though, I still feel like this feature just isn't a good idea overall, and especially not for a minor release. It feels like too big a change and I'm not confident that all bases are truly covered in terms of it not introducing (potentially silent) errors on a lot of sites.
> 
> 
> 
> Curious: What's the reason for not wanting to go with a "Preload" (or "Make global") section setting that was proposed in [#12698](https://github.com/craftcms/cms/pull/12698#issuecomment-1452025096)? I appreciate the DX effort here, but from my POV this seems like a prime candidate for an opt-in approach; especially considering the fact that many sites won't have any use for the feature in the first place as entryfying globals is going to be a manual effort for the time being.

If it were opt-in, I imagine it would lead to support because developers won’t understand why some singles are immediately available to their templates, but others aren't.

I like this because it significantly improves the DX for working with singles, which is currently super awkward to me, in that you have to go through the whole `craft.entries()` thing just to get a single entry. This will drastically simplify templates while improving performance at the same time (by querying all referenced singles in one query and memorizing them for other templates). So to me, it seems like something we should be making as accessible as possible so long as we're not breaking templates (which I believe we’ve accomplished here).

mmikkel:
> That was intentional (is defined checks and normal variable references were each implemented separately), but I could be persuaded to remove the fallback for is defined checks.

It's potentially going to break existing templates either way, isn't it? Which way you go with it should probably be the option that carries the least risk of things breaking. Which I think this FR generally does a very good job with, but it's probably impossible to make it air-tight, hence why I think this isn't an appropriate change for a minor release, unless it's made opt-in.

> If it were opt-in, I imagine it would lead to support because developers won’t understand why some singles are immediately available to their templates, but others aren't.

Maybe. Personally, I don't think that a "Preload" or "Make available globally" setting would be that much harder for developers to grok than, for example, the URI setting. It might even be helpful in putting emphasis on the feature, and making developers aware of how it works.  

> I like this because it significantly improves the DX for working with singles, which is currently super awkward to me, in that you have to go through the whole craft.entries() thing just to get a single entry. This will drastically simplify templates while improving performance at the same time

I appreciate that, but singles are created for all sorts of purposes. 

For example, some singles that have URIs and serve as either listing pages (e.g. "News", "Projects", "Employees") or special one-off pages ("Contact", "Privacy Policy") are rarely needed outside of their own entry template. Being able to pull these entries in from anywhere via their handles isn't really going to be that helpful or simplify anything, especially not for existing sites built without this feature in mind. 

For _some_ singles, the preloading will definitely be a boon, and perhaps I'm overreacting here. I just think that, even if the concepts of singles and globals are merging, effectively starting to treat _all_ singles as globals isn't going to be the best approach. It's great that this FR does a good job of avoiding variable conflicts, but personally I still struggle with the idea of having an entry potentially pulled in just because I happen to type something like `news`, `projects` or `employees` into a template (instead of the `Variable "news" does not exist` exception I might've otherwise had, and expected). In my opinion, it's going to add some extra cognitive overhead for - in many cases - little benefit.

brandonkelly:
Discussed this internally and decided to toggle this behavior with a new `preloadSingles` config setting. It’s disabled by default in Craft, so there’s zero chance it will break existing projects, but I’m going to have it enabled by default in the `craftcms/craft` starter when 4.4 ships.

Given that you’ll be opting into these changes, I left the `is defined` check alone (so it will return `true` for single handles). Feels right considering the variable _is_ defined, even if not in the `context`.

brandonkelly:
4.4.0-beta.5 is out with these changes!

gglnx:
If I enable `preloadSingles` I get Twig compile errors in combination with `??`:

```
ParseError: syntax error, unexpected token "," in storage/runtime/compiled_templates/3d/3d4c0246f66fb939a8dbf6908b6652f0.php:159
Stack trace:
#0 vendor/twig/twig/src/Environment.php(350): Twig\Cache\FilesystemCache->load('/Users/dennis/P...')
#1 vendor/twig/twig/src/Template.php(322): Twig\Environment->loadTemplate('__TwigTemplate_...', '_layouts/main.t...', NULL)
#2 storage/runtime/compiled_templates/52/526229511e201f6b0fb6264b84be52c0.php(41): Twig\Template->loadTemplate('_layouts/main.t...', '_index.twig', 1)
#3 vendor/twig/twig/src/Template.php(394): __TwigTemplate_eaf119847b72ab5787a16abece6af570->doDisplay(Array, Array)
#4 vendor/twig/twig/src/Template.php(367): Twig\Template->displayWithErrorHandling(Array, Array)
#5 vendor/twig/twig/src/Template.php(379): Twig\Template->display(Array)
#6 vendor/twig/twig/src/TemplateWrapper.php(40): Twig\Template->render(Array, Array)
#7 vendor/twig/twig/src/Environment.php(277): Twig\TemplateWrapper->render(Array)
#8 vendor/craftcms/cms/src/web/View.php(451): Twig\Environment->render('_index.twig', Array)
#9 vendor/craftcms/cms/src/web/View.php(504): craft\web\View->renderTemplate('_index.twig', Array)
#10 vendor/craftcms/cms/src/web/TemplateResponseFormatter.php(56): craft\web\View->renderPageTemplate('_index.twig', Array, 'site')
#11 vendor/yiisoft/yii2/web/Response.php(1098): craft\web\TemplateResponseFormatter->format(Object(craft\web\Response))
#12 vendor/craftcms/cms/src/web/Response.php(286): yii\web\Response->prepare()
#13 vendor/yiisoft/yii2/web/Response.php(339): craft\web\Response->prepare()
#14 vendor/yiisoft/yii2/base/Application.php(390): yii\web\Response->send()
#15 web/index.php(14): yii\base\Application->run()
#16 {main}
```

brandonkelly:
@gglnx Doh, thanks for reporting! Could have sworn I tested that 🤦🏻  Fixed now in 4.4.0-beta.6. You will need to clear your compiled template caches after updating, either from Utilities → Caches, or by running

```sh
php craft clear-caches/compiled-templates
```

PR #12798, opened 2023-03-06
Fixes #12797

leevigraham:


leevigraham:
Ok maybe not as simple as I thought…

```
$result = UrlHelper::encodeParams('http://example.test?foo[0]=bar&foo[1]=baz#hash');
```

returns: `http://example.test?foo%5B0%5D=bar&foo%5B1%5D=baz#hash` as expected

Issue #12695, opened 2023-02-17
[3.7.59]: Additional fields for assets not saved when temporary asset location in Local Temp Folder and permanent location in AWS Bucket until the whole entry is saved


mnlmaier:
### Description
We have defined several additional fields for assets, including a text field for image credits. Assets are stored in an S3 Bucket. The issue occurs when uploading images directly via the "Upload a file" button while using a temporary asset folder which is stored locally.
![image](https://user-images.githubusercontent.com/15020024/219698785-49d25fdb-7a9e-4d6f-9bc6-d52f99cec69d.png)
![image](https://user-images.githubusercontent.com/15020024/219699772-e50e1bb5-2812-472c-83e7-7bac47030198.png)

After the image is uploaded, double click the image **without saving the ** entry first to edit it. Note that the image will still be stored in the temporary location, which is expected.
![image](https://user-images.githubusercontent.com/15020024/219700570-b2983a89-5537-4f06-a689-3c8e083826fb.png)

Note and fill the custom text field:
![image](https://user-images.githubusercontent.com/15020024/219700833-49e76530-7915-4218-85a7-269024dd8ced.png)

Save the asset using the button on the bottom right, re open-the asset again (without saving the entry first) and note that the custom field will be empty.
![image](https://user-images.githubusercontent.com/15020024/219701312-a5e527c9-7774-4aa8-bffe-5e12be3578be.png)

### Steps to reproduce

1. Assets should be stored in S3 Bucket
2. Temporary Asset location should be set to Local
3. Open a page and add a module which uses the Asset Upload
4. Make sure assets use custom fields besides `title`
5. Upload an asset using the direct upload functionality
6. Wait for the asset to be uploaded successfully
8. **Do not save the entry**, this is a crucial step!
9. Double click on the asset and note that it is still stored in the temporary asset folder
10. Insert value into custom field
11. Save the asset, which causes the modal to close.
12. Double-Click the asset again.
13. Custom Text Field will be empty

**This behavior does not occur when the entry is saved before editing the asset.** Once the Save Event has moved the asset into the AWS bucket, everything works as expected.

### Expected behavior
Additional asset fields will be stored reliably no matter of the temporary asset folder's location

### Actual behavior
Additional asset fields are only stored reliably after the entry has been saved at least once, seemingly after the asset has been moved to its permanent destination.

Note that there are no errors on the console (so no JS issue) and the POST request when saving the asset seems to be firing correctly.

[...]

PR #12805, opened 2023-03-06
adjust field layout if asset uploaded to temp volume

i-just:
### Description
Ensure we don’t show custom fields if Temp Uploads Location isn’t set to a volume.

### Related issues
#12695


src/elements/Asset.php
```diff
@@ -1140,11 +1140,12 @@ public function getTitleTranslationKey(): string
      */
     public function getFieldLayout()
     {
-        if (($fieldLayout = parent::getFieldLayout()) !== null) {
```
brandonkelly, line 1143: I take it that `$fieldLayout` was getting set here? Maybe because something was calling `setFieldLayout()` earlier?

i-just: `setFieldLayout()` is not actually called here. The layout is determined based on the `defaultFieldLayoutId` body param. 

The order of events is as follows:
- `Asset->getEditorHtml()` starts with `fieldLayoutId = null`, which means it then uses the `defaultFieldLayoutId` body param (https://github.com/craftcms/cms/blob/v3/src/elements/Asset.php#L1821-L1828)
- then `Element->getEditorHtml()` checks if the element has a field layout, 
- `Element::hasFieldLayout()` calls `Asset::getFieldLayout()`  
- which calls `Element::getFieldLayout()`, which gets the layout by `fieldLayoutId`, which at this point is set to the `defaultFieldLayoutId` body param 
- and so the default layout is returned

brandonkelly: So maybe we just need to get rid of `defaultFieldLayoutId`? I’m guessing that was added before we had the modern element editor.

Can you test removing that, and then make sure that the proper field layout is still getting shown for freshly-uploaded assets, when they are able to be uploaded directly to a real (non-temp) volume?

i-just: Yup, you’re right. `defaultFieldLayoutId` is no longer needed. I have:
- an “Uploads” volume on Local FS with the field layout containing a “Number” field
- an “S3” volume on S3 FA with the field layout containing a “Plain Text” field
- Temporary uploads location set to a temp folder (so no field layout)
- autosaving drafts is disabled


https://user-images.githubusercontent.com/4500340/223683291-d8b1959e-0b05-4238-82ca-265419a91ab2.mov

Also tested with autosaving drafts enabled, in which case the file gets sent to the default uploads location (S3 in my case), and when you double-click it, it shows the S3 field layout.

brandonkelly:
Fixed this by checking for the `defaultFieldLayoutId` value via `Asset::getFieldLayout()` rather than `getEditorHtml()`.

Issue #12795, opened 2023-03-06
[4.3.10]: sidebarTemplate() method - sidebar template is injected on the right of content instead on the left


piotrpog:
### Description

When i use `$this->asCpScreen()->sidebarTemplate()` method in the controller, sidebar template is injected on the right of content instead of the left side, like when i use `{% block sidebar %}` with regular control panel templates that are not rendered using `asCpScreen`.

### Craft CMS version

4.3.10 

### Installed plugins and versions

-

i-just:
Hi, thanks for reporting! I created a PR to fix this.

brandonkelly:
This is actually working as expected. `sidebar` in the CP screen context means the right-and sidebar, whether within a slideout modal, or a full edit page (as those two sidebars should generally be consistent).

We should probably introduce new `leftSidebar` functionality though.

piotrpog:
@brandonkelly 
Would be great if you added that. I wanted to switch my plugin control panel pages to `asCpScreen` so they work as standalone pages and also in slideouts, but right now i cannot use sidebar properly with this method.

piotrpog:
@brandonkelly by the way - maybe there could be option for using `asCpScreen ` for pages that are not a form that can be submitted? Because unless i omitted something, every page rendered this method has its content injected into form.

piotrpog:
@brandonkelly thanks for adding this. althought i hoped it would get to 4.4.

brandonkelly:
It required a little internal discussion, and we didn’t have a chance to do that until post-4.4.

brandonkelly:
Craft 4.5.0 is out with this.

PR #12812, opened 2023-03-06
Bugfix/12795 cp screen response sidebar

i-just:
### Description
When adding a sidebar via `$this->asCpScreen()`, it was injected into the `details-container` and not `sidebar-container`.

This PR ensures sidebar content is injected as an actual sidebar via the `CpScreenResponseFormatter` and adds `details` param, `details()` and `detailsTemplate()` methods to the `CpScreenResponseBehavior` which are then injected into the `details-container` via the formatter.

### Related issues
#12795


brandonkelly:
See https://github.com/craftcms/cms/issues/12795#issuecomment-1457649186

Issue #12815, opened 2023-03-07
[4.x]: craft\fields\Matrix: Method craft\elements\ElementCollection::clearCachedResult does not exist


barnslig:
### Description

Saving an entry with a matrix field sometimes fails due to the following line of code: https://github.com/craftcms/cms/blob/e41aaf5aa54b7393795e3cb168a7ff1b0b320823/src/fields/Matrix.php#L1078

Apparently, `$value` can either be a `MatrixBlockQuery` or a `Collection`, as documented in line 1073:

https://github.com/craftcms/cms/blob/e41aaf5aa54b7393795e3cb168a7ff1b0b320823/src/fields/Matrix.php#L1073-L1074

However, the `Collection` type does not have a method `clearCachedResults`, resulting in an error when saving the entry.

The error seems to be fixed by moving the `$value->clearCachedResults();` call into the `$value instanceof MatrixBlockQuery`condition.

### Steps to reproduce

I am not sure in which case a Matrix is represented as Collection during saving.

### Expected behavior

The entry can be saved.

### Actual behavior

Saving an entry with a matrix field fails with the following exception:

```
BadMethodCallException: Method craft\elements\ElementCollection::clearCachedResult does not exist.
#16 vendor/illuminate/macroable/Traits/Macroable.php(113): Illuminate\Support\Collection::__call
#15 vendor/craftcms/cms/src/fields/Matrix.php(1078): craft\fields\Matrix::afterElementPropagate
#14 vendor/craftcms/cms/src/base/Element.php(4761): craft\base\Element::afterPropagate
#13 vendor/craftcms/cms/src/elements/Entry.php(2003): craft\elements\Entry::afterPropagate
#12 vendor/craftcms/cms/src/services/Elements.php(1612): craft\services\Elements::duplicateElement
#11 vendor/craftcms/cms/src/services/Elements.php(1151): craft\services\Elements::updateCanonicalElement
#10 vendor/craftcms/cms/src/services/Drafts.php(304): craft\services\Drafts::applyDraft
#9 vendor/craftcms/cms/src/controllers/ElementsController.php(1322): craft\controllers\ElementsController::actionApplyDraft
#8 [internal](0): call_user_func_array
#7 vendor/yiisoft/yii2/base/InlineAction.php(57): yii\base\InlineAction::runWithParams
#6 vendor/yiisoft/yii2/base/Controller.php(178): yii\base\Controller::runAction
[...]

PR #12816, opened 2023-03-07
additional safeguards for matrix field and cached results

i-just:
### Description
When trying to set/clear cached results for a matrix field, make sure we’re working with a query and not a collection.

### Related issues
#12815


brandonkelly:
Went with a more focused fix for now (ca7aa001aca9887f2bf0a91e0ad01ff12b896cad).

PR #12817, opened 2023-03-07
Add missing `@since` comment

bencroker:
Adds a missing `@since` comment to the `Html::svg()` method.

(this tripped me up in https://github.com/putyourlightson/craft-blitz/issues/480)


brandonkelly:
Why thank you!

PR #12819, opened 2023-03-07
Check passed volume ids on actionStartIndexing() if they are in $event->volumes list

vnali:
### Description
EVENT_LIST_VOLUMES allows plugin to restrict volumes on asset index utility page. but currently, on back-end, Craft doesn't check if passed volumes are valid -not manipulated by the end user- and available on EVENT_LIST_VOLUMES
this PR only proposes what i meant by checking passed volumes in the back-end.


vnali:
@i-just Let me know if this PR needs more detailed description. thanks!

brandonkelly:
Craft 4.4.6 is out now with that change 🎉

PR #12825, opened 2023-03-08
Darkens text and outline color for generic file type image

gcamacho079:
### Related issues
Resolves DEV-419


brandonkelly:
@gcamacho079 Can this be closed now thanks to #13262?

Issue #12871, opened 2023-03-12
[4.4.1]: After changing slugWordSeparator to "/", entries urls in admin panel are broken


bkrsmanovic:
### Description
Imagine you have structure and your url for that structure is `{slug}`. Typical url for my structure, if my slug is: contact-bojan would be something like:
https://myawesomecraftsite.com/contact-bojan.

Now if I want to add for this specific page slug like this:
https://myawesomecraftsite.com/contact/contact-bojan, I needed to change `slugWordSeparator` to "/", and then it allowed me to save. Otherwise, it always replaces "/" with "-" by default.

It gets saved, you open the page it works. But on backend in Entries, you cannot access anymore to that page because admin url of that page is broken.
Example:
This is proper edit URL: `/admin/entries/publicPages/1654-contact-bojan`

But in admin panel, when I press on my entry, url is something like this:
`/admin/entries/publicPages/1654-contact/contact-bojan` - and ofc opens up [HTTP 404 – Not Found].

If you press CTRL ALT SAVE instead of normal save, it instantly throws error message.

I think its happening because that slugWordSeparator should affect only frontend, but for some reason it affects backend as well.

### Steps to reproduce

1. Change `slugWordSeparator` to "/" in General options
2. Create structure and set for URL `{slug}`
3. Add in entry slug `word/another-word`
4. Save it
5. Try to open that page from admin panel - it doesnt work. Even CTRL ALT SAVE gives you an error.
6. You can manually figure out edit url - which is same as before the save with new slug.

### Expected behavior
Backend should properly keep old edit URL and not add according to slug in admin panel / or it should work when slug changes in backend as well.

### Actual behavior
Entries URLs are not correct once these steps are done.

### Craft CMS version

4.4.1

### PHP version

8.1

### Operating system and version

Ubuntu 20

### Database type and version

MYSQL

### Installed plugins and versions

-

brandonkelly:
Made the CP edit page URL generation defensive against `/`s in the slug, for the next Craft 3 and 4 releases.

brandonkelly:
Craft 3.8.2 and 4.4.2 are out with that fix.

PR #12874, opened 2023-03-13
disallow forwards slashes in slugs

i-just:
### Description
Disallows forward slash in slugs when saving an element.

I’m torn between this approach and a more drastic version where we check if the `slugWordSeparator` is set to `/` on getting the general config (similar to how we check the `securityKey`). 

On the one hand, the approach in this PR should provide better backward compatibility on the off-chance that someone was using `/` as a slug word separator. It would still allow it in, e.g. autogenerated entry URI format on section creation. 
On the other hand, throwing an exception on getting the general config is a more robust way of seriously disallowing `/`.

@brandonkelly, when you have a second, could you please LMK your thoughts?

### Related issues
#12871

### DIFF
--- a/src/translations/en/app.php
+++ b/src/translations/en/app.php
@@ -1735,6 +1735,7 @@ return [
     'week' => 'week',
     'weeks' => 'weeks',
     'yesterday' => 'yesterday',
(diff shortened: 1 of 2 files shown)


brandonkelly:
Just added a little protection to swap `/`s with `-`s in CP edit URLs for now.

PR #12876, opened 2023-03-13
Fixed test suite type compatibility

boboldehampsink:
Fixes

```
Fatal error: Declaration of craft\test\Craft::_afterSuite() must be compatible with Codeception\Module\Yii2::_afterSuite(): void in /app/user/vendor/craftcms/cms/src/test/Craft.php on line 143
```

### DIFF
--- a/src/test/Craft.php
+++ b/src/test/Craft.php
@@ -140,7 +140,7 @@ class Craft extends Yii2
     /**
      * @inheritdoc
      */
-    public function _afterSuite()
+    public function _afterSuite(): void
     {
         parent::_afterSuite();
 


boboldehampsink:
This is fixed in Craft 4. Note that in this file of Craft 3 the _beforeSuite already has a void return type as well.

brandonkelly:
Thanks!

boboldehampsink:
@brandonkelly this was not included in 3.8.2?

brandonkelly:
Doh… missed that the PR target was `3.8`. That branch is no longer active. Just merged into `v3` for the next release.

Issue #12884, opened 2023-03-14
[4.4]: Custom field validation stopped working on front end


pdaleramirez:
### Description

Adding a required custom field to an Element stopped validating when saving the field on a front-end form after upgrading to CMS version 4.4

### Steps to reproduce

1. Create a plain text custom field with a handle of "description".
2. Add the created field to the User Fields field layout and make it required.
3. Create a front-end [Registration Form](https://craftcms.com/knowledge-base/front-end-user-accounts#registration-form)
4. Add the custom field as a twig input
```
        {{ user ? _self.errorList(user.getErrors('description')) }}
        <label for="description">Description</label>
        {{ input('text', 'fields[description]', null, {
            width: 11,
            class: [
                'w-48',
                'border border-gray-300 hover:border-gray-500 px-4 py-2 leading-tight rounded',
            ],
        }) }}
```
5. Hit the Register button.

### Expected behavior
It should not save the user and display the validation error of "Description cannot be blank."

### Actual behavior
It saves the user and no validation error text is displayed.

### Craft CMS version

4.4

### PHP version

8

### Installed plugins and versions

-

i-just:
Hi Dale, 
I’m unable to replicate this. If the Settings > Users > Settings > Validate custom fields on public registration is checked and the custom field is required, it gets validated on registration. Are you able to replicate it with this option turned on?

That being said, I can replicate the related Commerce issue #3109 on Craft 4.4 but not on Craft 4.3.9 (using the same version of Commerce in both cases, 4.2.4 Pro), and I see what causes it. I'll update this issue again once the fix is ready.

i-just:
PR submitted.

pdaleramirez:
> Hi Dale, I’m unable to replicate this. If the Settings > Users > Settings > Validate custom fields on public registration is checked and the custom field is required, it gets validated on registration. Are you able to replicate it with this option turned on?
> 
> That being said, I can replicate the related Commerce issue #3109 on Craft 4.4 but not on Craft 4.3.9 (using the same version of Commerce in both cases, 4.2.4 Pro), and I see what causes it. I'll update this issue again once the fix is ready.

Thank you for checking; it is my fault my project config did not make the custom field back to  "required" when changing databases.

i-just:
Cool, thanks for the update Dale!

Issue #3109, opened 2018-07-18
Sorting eager loaded elements only works with "order" not "orderBy"


kristiansp:
### Description
It seems that ordering eager loaded elements only works with the deprecated `order`, not the new `orderBy`. Thus the following code returns the wanted result, sorting the `shows` in ascending order by `showTime`:

```
{% set currentMovies = craft.entries({
                         section: 'movies',
                         type:    'movie',
                         orderBy:   'startingDate asc',
                         limit:   null,
                         with: [
                           ['shows', { order: 'showTime asc' }],
                           'poster',
                           'stills'
                         ]
                       }).all()
%}
```
It will then throw a deprecation warning with the following origin:
`/Applications/MAMP/htdocs/craft3/vendor/craftcms/cms/src/elements/db/EntryQuery.php:216`

Whereas the following code does not sort the shows:

```
{% set currentMovies = craft.entries({
                         section: 'movies',
                         type:    'movie',
                         orderBy:   'startingDate asc',
                         limit:   null,
                         with: [
                           ['shows', { orderBy: 'showTime asc' }],
                           'poster',
                           'stills'
                         ]
                       }).all()
%}
```

### Steps to reproduce

1. Make element query with eager loaded elements
2. Sort the eager loaded elements by a custom field

### Additional info

- Craft version: Craft Pro 3.0.15
- PHP version: 7.1.12
- Database driver & version: MySQL 5.6.38
- Plugins & versions: N/A

brandonkelly:
Thanks! Fixed that for the next release.

PR #12893, opened 2023-03-15
don't skip custom field validation if field was passed in without 'field:' prefix

i-just:
### Description
Don’t skip custom field validation if the fields were passed without the `field:` prefix. (Introduced in [this commit](https://github.com/craftcms/cms/commit/081660a4792262054f32b663465d914ea3037147)).

Alternatively, I can submit a PR for Commerce to prefix custom field attributes with `field:`, which will also work but just for the Commerce plugin.

### Related issues
[#3109](https://github.com/craftcms/commerce/issues/3109)
#12884

### DIFF
--- a/src/base/Element.php
+++ b/src/base/Element.php
@@ -2393,7 +2393,11 @@ abstract class Element extends Component implements ElementInterface
                 $field = $layoutElement->getField();
                 $attribute = "field:$field->handle";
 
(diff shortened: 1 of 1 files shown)


i-just:
Doh... This solution doesn’t respect Commerce’s `validateCartCustomFieldsOnSubmission`. Closing and submitting a PR to Commerce instead.

Issue #12878, opened 2023-03-13
[4.4]: Optimizing image tasks throws error: "fclose(): supplied resource is not a valid stream resource"


RobinWithouck:
### What happened?

The Optimizing image tasks (from the ImageOptimize plugin) throws an error: "fclose(): supplied resource is not a valid stream resource".
Our file system is of type "Google Cloud Storage".
We think it started after the update to Craft 4.4. We changed the filesystem to local to test and this works.

We allready filled an bug report on the [ImageOptimize plugin](https://github.com/nystudio107/craft-imageoptimize/issues/371).

### Craft CMS version

4.4.1

### PHP version

8.1

### Installed plugins and versions

- ImageOptimize: 4.0.4
- Google Cloud Storage: 2.0.0

tomfischerNL:
Same issue here. Also with Google Cloud Storage. When updating from 4.3.11 to 4.4.0 or 4.4.1/4.4.2, this occurs.

i-just:
Thanks both. Just wanted to say that we are aware of this and are working on a solution. I'll update again once I have more info.

i-just:
Thanks for your patience. A PR to fix this has been submitted.

brandonkelly:
4.4.3 is out now with the fix for this!

PR #12908, opened 2023-03-16
only fclose the stream if it's a valid resource

i-just:
### Description
When Google Cloud Storage is done with the `$stream`, it’s no longer recognised as a valid resource. It comes back with `type=Unknown` and then causes `fclose()` to trigger an error: `TypeError: fclose(): supplied resource is not a valid stream resource`. 

Wrapping `fclose()` in an `is_resource()` check makes it behave as expected. Also tested against Local FS and Amazon S3. In both cases, `$stream` is still a valid resource which makes `fclose()` work as expected.

### Related issues
#12878


DynamiteGoesBoom:
FWOW This is happening with Azure Blob Storage as well.

PR #12930, opened 2023-03-19
Added default value as placeholder for number and money fields

bertoost:
### Description

I think it's good to show end users the default value as placeholder.

image

### Related issues

Not aware of


bertoost:
Hmm that sounds not really like a default value... Shouldn't it be default when it's NULL or empty? at any time?

Issue #12949, opened 2023-03-22
[3.x]: Duplicated entries/categories/products in relation fields


pierrestoffe:
### Description

We noticed that the relation fields weren't working as they used to. When I had content in a relation field (Entries, Categories, Products,…) the elements which were already listed in the field are listed a second time. Here is a video showing the issue.

https://user-images.githubusercontent.com/1706605/226870516-f92265a6-aa41-46e6-aa74-cb5df7f9e8d8.mp4

This issue is only experienced when populating an empty field. It doesn't occur when editing a field which has already been saved with content and still holds content. It happens when creating Entries and Commerce Products and has been tested on multiple browsers and multiple projects running on the same version.

### Expected behavior

1. Create a new entry
2. Edit a relation field with a limit >= 3 (maybe >=2 too, not tested)
3. Select one element (element A)
4. Select a second element (element B)
5. Notice that the relation field is now element A - element B
6. Select a third element (element C)
7. Notice that the relation field is now element A - element B - element C

### Actual behavior

1. Create a new entry
2. Edit a relation field with a limit >= 3 (maybe >=2 too, not tested)
3. Select one element (element A)
4. Select a second element (element B)
5. Notice that the relation field is now element A - element A - element B
6. Select a third element (element C)
7. Notice that the relation field is now element A - element A - element B - element A - element B - element C

### Craft CMS version

3.8.5

### PHP version

8.1.13

### Installed plugins and versions

- craftcms/aws-s3 1.3.2
- craftcms/cms 3.8.5
- craftcms/commerce 3.4.20.1
- craftcms/postmark 2.1.0
- craftcms/redactor 2.10.12
- nystudio107/craft-seomatic 3.4.51
- verbb/super-table 2.7.4
- vlucas/phpdotenv 5.4.0

i-just:
Hi, thanks for reporting and for the detailed steps. I submitted a PR for it.

brandonkelly:
Fixed for the next release!

PR #12955, opened 2023-03-22
potential fix for 12949

i-just:
### Description
When adding elements to some relational fields (e.g. Entries, Users, but not Assets, Categories), if you:
- add one element, 
- close the modal, 
- open the modal again and add another one from the same source
- the element that you selected in the first step will show twice in the field
 
(Data is then saved correctly (unique related elements).)

It was working correctly in `3.7.68`, and it is working in `4.4.5`.

There might be a better fix for this, but I can’t figure it out. 

The `this.elementSelect` from removed `if` statement seems to always be `undefined` in Craft 3 and Craft 4, but the described issue only occurs in Craft 3.

### Related issues
#12949


brandonkelly:
I’m pretty sure that condition was needed; it was just checking the wrong object.

i-just:
Seems so obvious now that I see what you've done!

PR #12962, opened 2023-03-22
Convert more actions menu to disclosure menu

gcamacho079:


src/web/assets/cp/src/js/Craft.js
```diff
@@ -2581,14 +2581,16 @@ $.extend($.fn, {
...
+    return this.on('activate', function (ev) {
       let $btn = $(ev.currentTarget);
       let params = $btn.data('params') || {};
       if ($btn.data('param')) {
         params[$btn.data('param')] = $btn.data('value');
       }
 
-      let $anchor = $btn.data('menu') ? $btn.data('menu').$anchor : $btn;
+      let $anchor = $btn.closest('.menu--disclosure').length
+        ? $btn.closest('.menu--disclosure').data('trigger').$trigger
```
brandonkelly, line 2592: It doesn’t seem like there would ever be a time when `$btn.closest('menu--disclosure')` would yield any results? `closest()` will return the closest ancestor of `$btn` that matches the selector, but the button is never nested within the disclosure menu.

gcamacho079: @brandonkelly this is true! I've updated that line. 👍🏼

brandonkelly: The new code is still assuming that `$btn` will be nested inside `.menu--disclosure` though:

https://github.com/craftcms/cms/blob/17cc70c50b031faf13321622e31bef59a39e894f/src/web/assets/cp/src/js/Craft.js#L2591

gcamacho079: I misunderstood your comment (and my code) the first time and made an irrelevant change. 🤦🏼 This is what I had before the latest commit:

```      
let $anchor = $btn.closest('.menu--disclosure').length
        ? $btn.closest('.menu--disclosure').data('trigger').$trigger
        : $btn
```
My reasoning for that was because inside that callback `$btn` refers to [any one of the secondary form submit buttons](https://github.com/craftcms/cms/blob/17cc70c50b031faf13321622e31bef59a39e894f/src/web/assets/cp/src/js/Craft.js#L2585), either:

1. “Create a draft” - which is _not inside_ the disclosure menu, or 
2. One of the buttons _inside_ the disclosure menu (“Save and continue editing,“ “Save and add another,“ etc.)

Let me know if I’m misunderstanding that particular bit of code.

brandonkelly: Ahhhh, ok! Should the original code be brought back then?

Issue #12956, opened 2023-03-22
[4.x]: Replace filter not working with an array of regular expressions


maximeDore:
### Description
Using an array as a parameter for the `replace` filter in Twig appears to be working if used with simple expressions but a regular expression doesn't seem to have any impact whatsoever. 

Here is an example of code that works in a Twig 3 template that's not in a Craft environment:
```twig
{% set str = 'Déjà Vu!' %}
{% set slug = str|replace({
   '/[^\\w\\s-]/u': '',
   '/\\s+/u': '-',
   '/-+/u': '-',
})|lower %}
```

However it would appear that the only way to make this work in Craft is by chaining multiple `replace` filters with 2 string parameters instead of an array like so:
```twig
{% set slug = item.contentTitle|replace('/[^\\w\\s-]/u', '')|replace('/\\s+/u', '-')|replace('/-+/u', '-')|lower %}
```

As for a little bit of context, I know I could use the `kebab` filter to end up with the same result. This is merely an example. My original goal was to create a slug with no accents or special characters, which I ended up doing with `str|kebab|ascii`. 

### Steps to reproduce

1. Set a string variable with a space character;
2. Set another variable with a `replace` filter chained to it;
3. Use an array as a parameter for this `replace` filter.

### Expected behavior
Return the modified string as per the regex used in the array parameter. In this case it should output `déjà-vu`.

### Actual behavior
Returns the string completely unchanged. In this case it outputs `déjà vu!`, the `lower` being the only filter that's applied correctly.

### Craft CMS version

4.3.5

### PHP version

8.0.8

### Operating system and version

macOS Ventura 13.2.1

### Installed plugins and versions

-

i-just:
Hi, thanks for reaching out.

> Here is an example of code that works in a Twig 3 template that's not in a Craft environment

Looking at the code Twig 3 code, their `replace` filter doesn’t actually support regular expressions: https://github.com/twigphp/Twig/blob/3.x/src/Extension/CoreExtension.php#L530-L537

Craft’s version of this filter adds support for passing in separate search and replace parameters and allows the search parameter to be a regular expression string.

That being said, I have submitted a PR to add the ability to have an array of regular expressions and an array of replacements. If it gets accepted, the way to go about your case would be to provide an array of regular expressions as the first parameter and an array (or string) of replacements, e.g.  `str|replace(['/[^\\w\\s-]/u', '/\\s+/u', '/-+/u'], ['', '-', '-'])` as a second one.

[...]

PR #12964, opened 2023-03-23
twig replace - add support for array of regex in search

i-just:
### Description
Adds support for providing the `$search` parameter as an array containing regular expressions. 

If `$search` and `$replace` are arrays of the same length, a one-to-one replacement will be performed. 
If `$replace` is shorter, its last element will be used for `$search` items that don’t have a corresponding `$replace` item.
If `$replace` is a string, it will be used for all items in the `$search` array.

With this PR, you wouldn’t have to chain regex replacements, so instead of this:
`str|replace('/[^\\w\\s-]/u', '')|replace('/\\s+/u', '-')|replace('/-+/u', '-')`

you could do this:
`str|replace(['/[^\\w\\s-]/u', '/\\s+/u', '/-+/u'], ['', '-', '-'])`

A mixture of regex strings and “regular” strings is allowed in the `$search` array.

### Related issues
#12956

### DIFF
--- a/src/web/twig/Extension.php
+++ b/src/web/twig/Extension.php
@@ -855,6 +855,38 @@ class Extension extends AbstractExtension implements GlobalsInterface
             return strtr($str, $search);
         }
 
+        // if search is an array, and replace is provided
+        if (is_array($search) && $replace !== null) {
+            // go through each search item
+            foreach ($search as $i => $searchItem) {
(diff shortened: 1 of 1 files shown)


brandonkelly:
Closing in favor of 3090f198e5bf56ce9c6f750829632b16c910035d, added for Craft 4.5.

Issue #12960, opened 2023-03-22
[4.x]: Tab display when autosaving


robin-gauthier:
### Description
When a draft is autosaving an entry, if you try to change to another content tabs in the entry the content stay from the older tabs or both are display at the same type. Moving back to another tab and reopening the broken one will fix the display.

### Steps to reproduce

1. Edit an entry with at least 2 tabs of fields
2. Type in a field to start the autosave
3. While the loader is display change to another tabs
4. 

### Expected behavior
When should be able to switch tabs with the good fields on display. (Or the tabs switching should be disable while the autosave isn't done)

### Actual behavior

The content should be the new in the same tabs, or the other tab is active with the older content. (Sometimes both are displayed)

### Craft CMS version

4.4.5

### PHP version

8.2

### Operating system and version

macos Monterey 12.6.3

### Installed plugins and versions

-

i-just:
Hi, thanks for reporting. I was able to replicate this and submitted a PR for it.

brandonkelly:
Fixed for the next release via #12968.

brandonkelly:
Craft 4.4.6 is out now with that fix 🎉

PR #12968, opened 2023-03-23
stay on the selected tab after autosaving

i-just:
### Description
Issue: If you edit a field, and while, e.g. entry is autosaving, you switch to a different tab; once autosaving is complete, you will see the content from the tab you switched to under the “old” tab (the one that triggered autosaving).
If the tab you switched to is conditionally hidden after editing, you would see the first tab’s content under the “old” tab.

I added checks to ensure the tab you switched to (new tab) shows properly (content and tab indicator). Both full page and slideout were affected by this.

### Related issues
#12960


src/web/assets/cp/src/js/ElementEditor.js
```diff
@@ -1464,16 +1469,35 @@ Craft.ElementEditor = Garnish.Base.extend(
 
             // Make the first tab visible if no others are
             if (!$allTabContainers.filter(':not(.hidden)').length) {
-              $allTabContainers.first().removeClass('hidden');
+              if (this.isFullPage) {
+                Craft.cp.selectTab($allTabContainers.first());
+              }
```
brandonkelly, line 1474: This change is preventing any content from showing when the originally-selected tab isn’t present after the autosave (e.g. when switching entry types, or when a tab condition comes into effect.) `$allTabContainers` is for the content  container elements, not the tabs themselves, so not something you would pass into `selectTab()`.

PR #12990, opened 2023-03-27
Globally-scoped Twig variables

brandonkelly:
### Description

Adds the ability to define custom “global variables” in twig templates, by introducing a new `global` keyword to `set` tags:

```twig
{% set global foo = 'foo' %}

{% set global foo, bar = 'foo', 'bar' %}

{% set global foo %}
foo
{% endset %}
```

When `global` is present, the variable values will be registered as fallback variables via `craft\helpers\Template::setFallback()`, and pulled into templates whenever referenced by name, if a local variable doesn’t already exist with the same name in the current context.

### Related issues

- #12951

### DIFF
--- a/src/helpers/Template.php
+++ b/src/helpers/Template.php
@@ -16,6 +16,8 @@ use craft\web\View;
 use Twig\Environment;
 use Twig\Error\RuntimeError;
 use Twig\Markup;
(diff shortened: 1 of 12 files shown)


brandonkelly:
Closed in favor of #13050

PR #13000, opened 2023-03-28
Two-factor authentication

i-just:
### Description
2FA and WebAuthn support for Craft 4.
2FA methods are: Google Authenticator (default) and Email Code (code sent via email).

- You can enable/enforce MFA on a per-user and per-user group basis
Screenshot 2023-03-28 at 16 10 10

- If you enable MFA for, e.g. all admins, the next time they log in, they will be asked to provide the TOTP. They can use ‘Email Code’ to have the code emailed to them, or (by default) they will be provided with a setup screen for Google Authenticator
- You can only set up a Security Key for WebAuthn once you’re logged in. WebAuthn is grouped under MFA though the way I (and Andris) chose to use it is - you don’t need a password to login via WebAuthn, so it looks like a single-step login.
Screenshot 2023-03-28 at 16 10 28

- Google Authenticator and Security Key for WebAuthn can only be set up for the currently logged in user (as in you can’t set it up on someone’s behalf)
- If MFA is turned on and your session expires, and you see the modal when visiting the control panel, you have to MFA in again.
- Email content for “Email Code” is content managed like other system emails (e.g. forgot password, account activation etc)
- Developers can register new MFA types/manipulate existing ones via new `EVENT_REGISTER_MFA_TYPES`

#### Other things to note:
- Because we plan to release it for 4.5, the code in `UsersController` is not as “clean” as it could be. It could be cleaned up for Craft 5.
- Google Authenticator and Email Code are enforced and work via the front-end login too, though changes to the login form are needed. I have a rough FE login template which can be used to update the KB article.
- Apparently, a known Firefox bug prevents WebAuthn from fully working: https://bugzilla.mozilla.org/show_bug.cgi?id=1479500 and https://bugzilla.mozilla.org/show_bug.cgi?id=1448408. They are marked as resolved, but still, I can’t make it work in Firefox and Andris’ solution also fails in the same way. Support doesn’t look great for FF either: https://developer.mozilla.org/en-US/docs/Web/API/Web_Authentication_API#browser_compatibility

### Related issues
- DEV-13
- #8453


brandonkelly (changes requested):
Off to a great start!

There’s a few high-level changes I’d like to see:

- [ ] Always show a Password field on the login form, and have the security key just be another form of 2FA, rather than an alternative for entering a password. Hiding the password field initially makes the login flow worse for the vast majority of users who haven’t set up a security key. (And this is how GitHub does it.)
- [ ] Remove the “Enable MFA” checkbox for user accounts (and the ability to force an individual user to have a 2FA method).
  - `User::$requireMfa` and `users.requireMfa` can be renamed to `$has2fa`/`has2fa` if it’s still useful to keep track of whether a user has enabled any 2FA methods.
- [x] Change any other `MFA` references to `2FA` for consistency.
- [ ] Drop the email option. Seems like that’s generally not considered a secure 2FA method.
- [ ] Add the ability to assign/view account recovery codes, possibly using the [pragmarx/recovery](https://packagist.org/packages/pragmarx/recovery) library.

i-just:
@brandonkelly - here's an update on what's been done since we last talked about it:

> Single page with email/password form, as well as passkey and SSO buttons (but no passkey button on Firefox). 

Done. Instead of detecting Firefox and just disabling the passkey option, I added a second check in there to see if the browser supports "platform authenticators" and if it doesn't, there's a popup saying just that and you can proceed or cancel; that should be more graceful, should work in any browser that has this problem (I'm only aware of Firefox on Mac having this problem, but that doesn't mean there's not more out there). In theory, you should still be able to use passkey with Firefox on a Mac if you have a "roaming authenticator" (external device like yubikey or google titan)
Screenshot 2023-05-12 at 12 07 27

> If you press “Sign In with a Passkey”, the browser immediately takes over. The passkey contains the user ID (email) + auth. No need to enter your email address manually or anything.

done;

> Very (maybe too) basic passkey management in user settings. Just says “You have X passkeys enabled”, a link to delete them all, and a “Create a Passkey” button. (Listing all passkeys + renaming/deleting them individually like Google does seems preferable.)

I kept what I had before; list of all passkeys, you can delete them individually, you can name them upon creation but not rename them;

> Awkwardly, if you already have a passkey for the account/device and then press “Create a Passkey”, you will go through the full passkey setup process, and then afterward, you get an error saying “A passkey was already set up”. (The browser doesn’t seem to retain the passkey, at least.)

we don't show a false positive message at the end; you go through the whole process and at the end it says "Try a different device. You already registered this device. You don't have to register it again."
Screenshot 2023-05-12 at 11 49 53

> If you have multiple accounts, you are able to create a passkey for each of them. In which case, the passkey browser UI will let you choose which account to use.

done
Screenshot 2023-05-12 at 12 39 04


> No cookies are used to track passkeys whatsoever. Passkey login continues to work exactly the same after clearing cookies.

yup

> ideally the elevated session modal could be a little smarter, since we know whether or not the current user has any passkeys. So in that case it should just be a simple Password input, and if they have passkeys, also a passkey button

done

- we use passwordless and usernameless option for login
- we use just passwordless option for elevating session and in the AuthManager

brandonkelly:
> I added a second check in there to see if the browser supports "platform authenticators" and if it doesn't, there's a popup saying just that and you can proceed or cancel; that should be more graceful, should work in any browser that has this problem (I'm only aware of Firefox on Mac having this problem, but that doesn't mean there's not more out there). In theory, you should still be able to use passkey with Firefox on a Mac if you have a "roaming authenticator" (external device like yubikey or google titan)

I think we should just not show the button at all. Most people don’t have an external authenticator, so for the most part we’re just adding noise to the page. (And Google/Best Buy both made the same decision, so it doesn’t seem like there’s going to be any expectation for WebAuthn support from Firefox users.)

> I kept what I had before; list of all passkeys, you can delete them individually, you can name them upon creation but not rename them;

That’s fine, but ideally we should at least give the passkey a name automatically (e.g. “Chrome on Mac”, “Chrome on Mac 2”, etc.) rather than requiring them to think of something.

Besides that, there seems to be a bug where once you’ve added a passkey, normal password sign-in is effected: the next time you sign in, you get the “set up a 2FA method” screen, as if you belong to a user group that requires 2FA. Normal login shouldn’t be affected by whether the account has a passkey.

i-just:
@brandonkelly all items in your last comment have been actioned

i-just:
@brandonkelly I discussed error messaging placement for this with @gcamacho079. Technically, current solution doesn't violate WCAG, but it could be better in terms of usability. I'm going to make some adjustments that Lupe suggested.

i-just:
@brandonkelly, I made the accessibility changes, @gcamacho079 is happy with them.

Issue #13006, opened 2023-03-29
[4.x]: Asset-selection modal opens on default upload folder, even when the user has no access to this folder


FlorisDerks:
### Description

We have a multisite setup within Craft, where each site has it's own asset-filesystem. Some of our asset-selection fields have a default folder in order to redirect editors to the correct location to upload their files. This used to work somewhat in older versions (would only work on one site, others would load the root-folder of their own first authorized filesystem), but since v4.4.x it loads the folder of the site mentioned in the field-settings, even when the user does not even have access to that filesystem.

### Steps to reproduce

1. Create 2 filesystems, create a folder with files in one of them
2. Create a user which only has access to the filesystem without the folder
3. Create a field which uses the folder as the default path & add to a section
4. Log in as the user without folder-access
5. Try to select a file, you will see that it opens the folder in the unauthorized filesystem, and when you try to upload it throws an error (which is expected)

### Expected behavior

When a user has no access to a filesystem, the root of the first authorized filesystem should be loaded (as it was in v4.3.x). Even better would be to check if the folder exists in the first authorized filesystem and use that instead.

### Actual behavior

The folder of an unauthorized filesystem is shown, which should've been hidden.

### Craft CMS version

4.4.5

### PHP version

8.2

### Installed plugins and versions

-

FlorisDerks:
Additional findings: when the first authorized root-folder is correctly displayed (due to not having a default folder), new uploads are still trying to be added to the default folder. Only when you navigate to a different folder & back & then try to upload a file, it's uploaded to the correct location.

i-just:
Hi, thanks for reporting! I raised a PR to address this.

qrazi:
Just another +1 on this issue.

I see - I think :sweat_smile: - that the PR is targeting Craft 3.8? In which case; we have the issue in Craft 3.8.5. I can confirm that when I apply the changes from the PR - I guess specifically commit d6236c0b8127eba941e03160431288fe205f4bb2 - to our local project, the upload works as expected again.

Basically we can reproduce this locally, and can also confirm that for us the PR fixes the issue :sweat_smile:

brandonkelly:
This is fixed for the next Craft 3 and 4 releases via #13010 🎉

brandonkelly:
Craft 3.8.6 and 4.4.6 are out with that fix 🎉

PR #13010, opened 2023-03-29
check if user has access to the default asset location

i-just:
### Description
When the assets field has a default assets location set, check if the user has permission to `viewVolume` before setting the `defaultSource` and `defaultSourcePath` template variables.

Please let me know if you need me to raise a separate PR for Craft 4.

### Related issues
#13006

### DIFF
--- a/src/fields/Assets.php
+++ b/src/fields/Assets.php
@@ -709,13 +709,17 @@ class Assets extends BaseRelationField
         if ($this->useSingleFolder) {
             $variables['showSourcePath'] = false;
         } else {
-            $uploadFolder = $this->_uploadFolder($element, false);
-            if ($uploadFolder->volumeId) {
-                $folders = $this->_folderWithAncestors($uploadFolder);
-                $variables['defaultSource'] = $this->_sourceKeyByFolder($folders[0]);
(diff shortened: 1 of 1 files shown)


brandonkelly:
This seems kindof at odds with these lines, which explicitly authorizes the user to access the upload folder, regardless of permissions:

https://github.com/craftcms/cms/blob/bff6946e76f99d058d20dd0b48ea5d84764928f8/src/fields/Assets.php#L638-L639

Maybe that’s just not working as intended currently?

i-just:
I tracked it down to enforcing `defaultSourcePath` for Element Index.

In 3.7, it worked as expected. The `defaultSource` was still set to what’s defined in the field’s settings, but it was then “ignored” in favour of showing the volume user has permission to view. If a user didn’t have permission to view any volumes, they could see the field in the entry layout but not modify it.

In 3.8, the `defaultSource` and `defaultSourcePath` are enforced even if a user cannot view the volume those two reference. It then caused the criteria for the element query to contain a `folderId` for an unpermitted volume. PHP side of things works roughly the same, but because of this: https://github.com/craftcms/cms/blob/v3/src/web/assets/cp/src/js/BaseElementIndex.js#L306, the defaults were enforced regardless of permissions.

I reverted the previous change and committed a new fix.

PR #13012, opened 2023-03-30
if newly selected tab is now hidden, switch to the first one

i-just:
### Description
One more adjustment for switching tabs while autosaving. 
If the tab you switch to while autosaving is no longer visible after the autosave (e.g. conditionally hidden), switch to the first visible tab.

### Related issues
#12960


brandonkelly:
Works great!

Issue #12986, opened 2023-03-27
[4.x]: Plugins with valid licences are still listed as 'Active Trials' in the plugin store


MoritzLost:
### Description

Not sure if this issue belongs here or in a different repository, please let me know if this is the wrong place to report this.

I have just published a new website and bought licences for Craft CMS and the two commercial plugins used on the site. However, the plugin store still shows the two plugins as 'Active Trials':

![Screenshot of the plugin store showing the plugins _Navigation_ and _Retour_ as active trials](https://user-images.githubusercontent.com/10146880/227960238-425bfa86-18c2-4ac9-80c9-6e493754ee32.png)
![Screenshot of the _Retour_ plugin in the plugin store showing _Installed as a trial_](https://user-images.githubusercontent.com/10146880/227960244-540052fa-e1e4-4541-8b98-0098c82ba4ae.png)

I've cleared all caches, logged out and back in again and restarted the server, and I'm not getting any warnings regarding invalid licences. But the store still shows those plugins as active trials. Could this be an issue with any server-side caching of the Craft licensing API? Not sure if this is just temporary – I've also noticed the warnings about invalid/missing licences are pretty persistent after buying the licence, but those usually go away after clearing caches and logging out and back in.

We're using environment variables for all licence keys, in case that is relevant to this issue.

On a related note, I noticed that the link in the lower left of the plugin store that is supposed to say `Craft Console` says `Craft Consolebuild` instead:

![Link saying Craft Consolebuild instead of Craft Console](https://user-images.githubusercontent.com/10146880/227961894-15119641-bbfb-4665-a809-e95e84ced11a.png)

### Steps to reproduce

1. Buy a licence for a commercial plugin and insert the licence key as the environment variable used for the plugin licence.
2. Check if the plugin is still listed under _Active Trials_ in the plugin store.

### Expected behavior

After buying a licence, the store should clearly indicate that we have valid licences for those plugins and not display them as trials any more.

### Actual behavior
 
The store still displays the plugins as Active Trials.

### Craft CMS version

4.4.5

### PHP version

8.2

lindseydiloreto:
I came here to post about "Craft Consolebuild". I'm not sure it's related to your main issue, so I opened up a separate thread for it...

- ["Craft Consolebuild" appearing at bottom of CP Plugin Store](https://github.com/craftcms/cms/issues/13004)

brandonkelly:
This has been fixed for the next Craft 3 and 4 releases, via #13023.

[...]

PR #13023, opened 2023-03-31
Get correct cache key with plugin- prefix

timkelty:
### Description
We were retrieving items from the `licenseInfo` cache by plugin handle, but they were stored with a `plugin-` prefix.
This seems to fix https://github.com/craftcms/cms/issues/12986, but I'm not 💯 sure it is in the right place, as @brandonkelly did the recent updates to licensing info.

### Related issues
https://github.com/craftcms/cms/issues/12986


timkelty:
@brandonkelly confirmed this was happening in 3.x as well, so rebased to `v3`.

Issue #12954, opened 2023-03-22
[4.4.5]: Using "Field Limit" after creating new "Plain Text Field" -> Database Exception


mfell:
### Description
Got a Database Exception when using "Field Limit" of a just created and saved new "Plain Text Field".

```
Database Exception – [yii\db\Exception](https://www.yiiframework.com/doc-2.0/yii-db-exception.html)
SQLSTATE[42000]: Syntax error or access violation: 1118 Row size too large. The maximum row size for the used table type, not counting BLOBs, is 65535. This includes storage overhead, check the manual. You have to change some columns to TEXT or BLOBs
The SQL being executed was: ALTER TABLE `content` ADD `field_assetDownloadName_hdmffqxr` varchar(200)
Error Info: Array
(
    [0] => 42000
    [1] => 1118
    [2] => Row size too large. The maximum row size for the used table type, not counting BLOBs, is 65535. This includes storage overhead, check the manual. You have to change some columns to TEXT or BLOBs
)
↵
Caused by: PDOException
SQLSTATE[42000]: Syntax error or access violation: 1118 Row size too large. The maximum row size for the used table type, not counting BLOBs, is 65535. This includes storage overhead, check the manual. You have to change some columns to TEXT or BLOBs
in /var/www/html/vendor/yiisoft/yii2/db/Command.php at line 1302
```

### Steps to reproduce

1. Create a new "Plain Text Field"
2. Safe 
3. Set the option "Field Limit" (in my case 50)

### Expected behavior

no Database Exception

### Actual behavior

- Database Exception
- Reload after Database Exception, see pdf

[Database Exception – yii_db_Exception.pdf](https://github.com/craftcms/cms/files/11040945/Database.Exception.yii_db_Exception.pdf)

Can send you config.json and DB, if needed.

### Craft CMS version

Craft Pro 4.4.5

### PHP version

8.1.13

### Operating system and version

Linux 5.15.49-linuxkit (DDEV)

### Database type and version

MySQL 5.7.38

### Image driver and version

Imagick 3.7.0 (ImageMagick 6.9.11-60)

### Installed plugins and versions

- "carlcs/craft-redactorcustomstyles": "4.0.3",
- "craftcms/cms": "4.4.4",
- "craftcms/redactor": "3.0.4",
- "doublesecretagency/craft-cpcss": "2.6.0",
- "nystudio107/craft-seomatic": "4.0.22",
- "nystudio107/craft-templatecomments": "^4.0",
[...]

Issue #7750, opened 2021-03-31
Database error - content table maximum rows exceeded


jakepm:
### Description

Receiving the following error when attempting to create a new field.

Seems the content table is full?

I'm seeing 5891 rows in total. The install has a lot of fields, most of which are translatable across 3 (multi) sites. 

Database Exception – yii\db\Exception
SQLSTATE[42000]: Syntax error or access violation: 1118 Row size too large. The maximum row size for the used table type, not counting BLOBs, is 8126. This includes storage overhead, check the manual. You have to change some columns to TEXT or BLOBs
The SQL being executed was: ALTER TABLE `content` ADD `field_landingPageButtonALink` varchar(255)
Error Info: Array
(
    [0] => 42000
    [1] => 1118
    [2] => Row size too large. The maximum row size for the used table type, not counting BLOBs, is 8126. This includes storage overhead, check the manual. You have to change some columns to TEXT or BLOBs
)
↵
Caused by: PDOException
SQLSTATE[42000]: Syntax error or access violation: 1118 Row size too large. The maximum row size for the used table type, not counting BLOBs, is 8126. This includes storage overhead, check the manual. You have to change some columns to TEXT or BLOBs
in /{path-removed}/vendor/yiisoft/yii2/db/Command.php at line 1299

### Steps to reproduce

1. Create new field
2. Save (submit event gives DB error)

### Additional info

- Craft version: 3.6.11.2
- PHP version: 7.4.12
- Database driver & version: MySQL 5.7.32
- Plugins & versions:

jakepm:
I can see this is to do with the row size of the content table so not specifically related to the field I'm trying to create. Updated the title and description.

jakepm:
Perhaps more concerning is that I just deleted about 10 translatable fields and 2 corresponding single entries, yet I still have 5891 rows (same amount as before) in the content table. Does the content table not get pruned under these circumstances?

brandonkelly:
We are working on a solution to this for Craft 4 (#1524). In the meantime, if you email a database backup and Composer files over to support@craftcms.com we may be able to offer some suggestions on how to work around the issue.

jakepm:
@brandonkelly - I resolved this by deleting some fields that were redundant. 

However, this same issue is now preventing me from doing a 'reapply project config' on the staging site after deploy, as the auto sync was not successful. We changed a lot of config stuff and I need to deploy this before we can move on to a rebuild, which will include halving the number of fields we currently use.

[...]

Issue #7221, opened 2020-12-03
Can no longer create new fields. Maximum table row size reached error warning. 


riusi88:
We just ran into an issue on one of our large Craft sites, and can no longer create text content fields (with the exception of "number" fields). The following error gets logged. 

`SQLSTATE[42000]: Syntax error or access violation: 1118 Row size too large. The maximum row size for the used table type, not counting BLOBs, is 8126. This includes storage overhead, check the manual. You have to change some columns to TEXT or BLOBs`

Our temporary solution was to modify a mysql config, and set `innodb_strict_mode=0`, but we're concerned that this won't be a stable fix. We'd appreciate any guidance to resolve this problem any other way, and can provide additional info/db's if requested. 

### Additional info

- Craft version: 3.4.30
- PHP version: 7.1.33
- Database driver & version: MySQL 5.5.5

angrybrad:
There are other MySQL config options to help, too: https://github.com/craftcms/cms/issues/2009#issuecomment-333394434

Ultimately, this is a MySQL limitation that you're running into. We hope to lower the chance that people actually hit it with the planned element content refactoring: https://github.com/craftcms/cms/issues/1524

brandonkelly:
Row size limits are no longer an issue in Craft 5.

Issue #4308, opened 2019-05-24
table hits row size limit


rolfkokkeler:
### Description
table with to columns; label and url.
Adding up to 20 rows gives database exception:

> SQLSTATE[42000]: Syntax error or access violation: 1118 Row size too large (> 8126). Changing some columns to TEXT or BLOB or using ROW_FORMAT=DYNAMIC or ROW_FORMAT=COMPRESSED may help. In current row format, BLOB prefix of 768 bytes is stored inline. 

### Steps to reproduce

just try an recreate a table with this content:
[...]

PR #13025, opened 2023-04-02
Field column size improvements

brandonkelly:
### Description

Makes the following improvements for custom field `content` table columns:

- When adding or renaming a column, if a column with the same name already exists, a new table will be created (`content_<columnName>_bak_<timestamp>` to store the existing column’s non-null values, along with the rows’ `id`, `elementId`, and `siteId` values, and then the original column will be deleted from the `content` table. (Previously, the column was just renamed to `<columnName>_old`, adding to the table’s row size.)
- When changing a custom field’s column type, if a SQL error occurs due to the existing data being too long, the existing column data will be backed up in a new table, and then any existing values that would overflow the new column size will be set to `null`, then the `alter table` command will be attempted once again. (Previously, the column was just renamed to `<columnName>_old` and a new column was created, effectively wiping out the field’s existing values until manually reconstructed.)
- Dropdown and Radio Buttons fields now use smart column lengths, based on their longest option value’s length. (Note this doesn’t come into effect until the fields are re-saved.)

#### Example

If you have a Dropdown field with `yes`, `no`, and `maybe` options, its column type will be `varchar(5)` (instead of `varchar(255)` as before).

If its `maybe` option is removed later on, its column type will be changed to `varchar(3)`. If any entries exist that had `maybe` selected, a new `content_<columnName>_bak_<timestamp>` table will be created, which will contain _all_ of the non-null Dropdown field values, and all `maybe` options will be set to `null`.

### Related issues

- #2009
- #4308
- #7221
- #7750
- #12954

### DIFF
--- a/CHANGELOG-WIP.md
+++ b/CHANGELOG-WIP.md
@@ -21,3 +21,6 @@
 ### System
 - All generated URL param characters are now properly encoded. ([#12796](https://github.com/craftcms/cms/issues/12796))
 - `migrate` commands besides `migrate/create` no longer create the migration directory if it doesn’t exist yet. ([#12732](https://github.com/craftcms/cms/pull/12732))
+- Dropdown and Radio Buttons fields now use smart `content` table column lengths, based on their longest option value’s length. ([#13025](https://github.com/craftcms/cms/pull/13025), [#12954](https://github.com/craftcms/cms/issues/12954))
+- When `content` table columns are resized, if any existing values are too long, all column data is now backed up into a new table, and the overflowing values are set to `null`. ([#13025](https://github.com/craftcms/cms/pull/13025))
+- When `content` table columns are renamed, if an existing column with the same name already exists, the original column data is now backed up into a new table and then deleted from the `content` table. ([#13025](https://github.com/craftcms/cms/pull/13025))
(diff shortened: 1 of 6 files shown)


Mosnar:
@brandonkelly Is there any risk of the automatic column resizing on option fields causing problems for people using the `DefineInputOptionsEvent` event? I'm not using it myself anywhere, but I've been keeping it in my pocket since it seems like it'd be really useful.

To be more specific, my thought is that since the resizing occurs when the field is saved, additional options defined via that event might be longer than what's supported by the column size.

brandonkelly:
@Mosnar Good point! Yes, that could be an issue. And I don’t think we can safely call the event when determining the column length, because event handlers could return different options depending on the current entry, etc.

I’ve just added a new “Column Type” setting for Dropdown and Radio Buttons fields, which has two options – “Automatic” and “varchar”. The setting will be set to “varchar” for existing fields, and “Automatic” for new fields. So you will need to opt into the smart column sizing for existing fields (if you know it’s safe to do so), and you can opt out of it for new ones.

brandonkelly:
@lineke Craft 3 is now feature-locked. Bug/security fixes only now.

lineke:
@brandonkelly We are now upgrading to craft 4 in order to solve the issue. Any idea when 4.5 will be released?

Issue #12916, opened 2023-03-16
Plugin Store not handling 403 API responses for organization carts


brandonkelly:
### Description

If you create a cart in the control panel and press **Checkout**, then choose to make it an organization cart within Console, it is no longer possible to make any changes to the cart within the control panel, as API requests will come back with `403 Login Required` responses, which are not being handled properly.

(This affects Craft 3 and 4.)

### Steps to reproduce

1. Add something to the in-app Plugin Store cart
2. Press **Checkout**, and choose to make it an organization cart
3. Go back to the in-app Plugin Store
4. Add another item to the cart, or modify an existing item’s Expiry Date

### Expected behavior

Either the cart modification should work, or the Plugin Store should handle the 403 API response

### Actual behavior

Endless spinner, and an error logged out to the browser console

### Craft CMS version

3.8 and 4.4

timkelty:
@brandonkelly:
- #13038 fixes the uncaught exception and will now display a flash error.
- Craftnet will now return an error response when an IAPS request tries to get a cart that has been already claimed by an org. We were already doing this for carts that were pending approval, so it made sense just to open that up to any org carts. The existing logic clears the stored cart and creates a new one, so any page loads after the cart is claimed by an org will have a new cart.

We talked about including a message like "reload to reset cart" or something…but that could be problematic – the error flash just happens in the request `catch`, so we can't really just assume what is happening. It could just be a network blip or craftnet momentarily down.

If we want to something further, I'm thinking a better option would be to add a "clear cart" button somewhere, that would just discard the stored local `orderNumber` and create a new one.

PR #13038, opened 2023-04-04
Fix uncaught errors from org carts

timkelty:
### Description
This fixes the uncaught error when updating the expiry in the cart.

### Related issues
https://github.com/craftcms/cms/issues/12916


brandonkelly:
Thanks!!

PR #13050, opened 2023-04-04
Feature/dev 1101 set new global variables from twig

brandonkelly:
### Description

Adds a new `_globals` global Twig variable for front-end templates, set to an empty collection. All templates have access to it, and can set/retrieve values on it, providing a way for templates to store custom values in a global scope.

A new `set()` macro has also been added to collections, which can work as an alias of `Collection::put()` (key + value arguments), and it also accepts an array of key/value pairs.

```twig
{# setting values: #}
{% do _globals.set('foo', 'bar') %}
{% do _globals.set({
  foo: 'bar',
  baz: 'qux',
}) %}

{# getting values: #}
{{ _globals.foo }}
{{ _globals.get('foo') }}
{% dump_globals.all() %}
```

### Related issues

- #12951

### DIFF
--- a/CHANGELOG-WIP.md
+++ b/CHANGELOG-WIP.md
@@ -13,6 +13,7 @@
 ### Administration
 
 ### Development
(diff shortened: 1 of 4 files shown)


brandonkelly:
@romainpoirier It’s hard to imagine it could have a negative performance impact. In theory the more you’re taking advantage of it, the less values have to be fetched multiple times, so you’d more likely see a performance improvement.

Issue #13053, opened 2023-04-05
[4.4.5]: Edit User Slide out panel is blank when accessing from Users page


seandelaney:
### Description

Edit User slide out panel is blank when attempting to edit a user from Users page.

### Steps to reproduce

1. Log into CP
2. Click Users in the sidebar.
3. Check a user in the list (to enable/toggle options in the secondary menu button) (The button with the cog wheel icon centred between pagination and export button).
4. Click edit user. 
5. Note the slide out panel which opens is blank.

### Expected behavior

Slide-out panel shows user fields. If no user fields, show a message instead of blank panel.

### Actual behavior

Slide-out panel is blank/empty

### Craft CMS version

4.4.5

### PHP version

8.1.2

### Installed plugins and versions

-

i-just:
Hi, thanks for reaching out. 

Could you please clarify if it’s completely blank or something like this:
<img width="921" alt="Screenshot 2023-04-05 at 12 40 12" src="https://user-images.githubusercontent.com/4500340/230069991-0e53c5be-6f95-40db-8d03-5ec11d631ccc.png">

seandelaney:
@i-just Completely blank.

![SCR-20230405-lxdl-2](https://user-images.githubusercontent.com/580403/230078045-402f1563-5c2e-4d28-97e2-11bf860aa2fa.png)

i-just:
Right, so your screenshot differs from mine because your screen is narrower, and the sidebar is hidden by default. In the top-right corner, you have a sidebar icon that will show the sidebar and make it look similar to the one I posted.

This is the expected behaviour, but I have submitted a PR with a slight improvement.

brandonkelly:
This has been resolved for Craft 4.5 via #13056. Going forward, elements that don’t have field layouts will have their sidebar content shown full-screen within CP slideouts.

brandonkelly:
Craft 4.5.0 is out with that change. Thanks for bringing it up!

PR #13054, opened 2023-04-05
empty sidebar content message and show sidebar

i-just:
### Description
If the element you want to edit opens in a sidebar and that element has no visible fields, show a message that says so.

Additionally, if there are no visible fields and the screen width makes the sidebar hidden when the slideout opens up - open it.

Screenshot 2023-04-05 at 14 16 49

Screenshot 2023-04-05 at 14 17 16 

This only affects Craft 4.

### Related issues
#13053


brandonkelly:
Took this in a different direction (#13056).

Issue #13060, opened 2023-04-06
[4.x]: Error sending activation email for user when there are conditional tabs with required fields


jalendport:
### Description

I've got a site where there is a tab (with required fields) in the User Fields settings that are conditional based on the user group. When creating a new user, those tabs are not shown dynamically once the user group is selected, but only once the user is saved/created - which is fine, except if the "Send an activation email now" checkbox is selected, I get an error: `Unable to set verification code on user: X Field cannot be blank., Y Field cannot be blank.`. This seems to be a loophole, because I can't access those fields until after the user is saved/created. It seems those fields should not be validated until they've had a chance to actually be edited...

### Steps to reproduce

1. Create a User Group ("X Group")
2. Add some required user fields to a tab in the User Fields Field Layout
3. Make the tab conditional with a "User Condition" where "User Group" is one of "X Group"
4. Create a new user in the control panel - fill in the Username, Full Name, Email, and select "X Group" under Permissions
5. Select "Send an activation email now" and the click the "Save" button
6. Enjoy the error screen 🙃 

### Craft CMS version

Craft Pro 4.4.3

### PHP version

8.1.16

i-just:
Hi, thanks for reporting! I submitted a PR to adjust this.

PR #13068, opened 2023-04-11
Bugfix/13060 user activation email send validation

i-just:
### Description
We validate a user on the initial save and then again before sending the activation email. On initial save, the conditionally required fields are not yet considered a part of a layout, but on activation email send, they’re considered visible. This causes an exception to be thrown, and an activation email isn’t sent.

Because of this, when sending the activation email when “send an activation email now” was checked, we shouldn’t validate against a `live` scenario because that will validate the conditionally required fields, which cannot be filled out at this point.
We should, however, use the `live` scenario if a user was saved and the cog > "Send activation email" option is used.

### Related issues
#13060


brandonkelly:
This ended up being caused by #12432 so I had Tim fix it via #13093. And I pulled in the checkbox fix in 0d561600c60a2bb2b142a84e27ee4a342cd1a1c0.

Issue #13061, opened 2023-04-06
[4.x]: Sending activation email that fails to send causes user permissions to get wiped


jalendport:
### Description

If for whatever reason activation emails won't send successfully (e.g. a misconfigured Email setting), any selected checkboxes on the Permissions tab are cleared, which can cause issues if the user is then saved without reloading the page or reselecting the desired permissions.

In my case I'm actually using Yii DI containers to add a simple conditional to the `sendActivationEmail` method on the Users service:

```php
public function sendActivationEmail(User $user): bool
{
	$url = $this->getActivationUrl($user);

	// Prevent activation email from being sent if user conditions are not met
	if ($user->isInGroup('clients') && !$user->accountManager->one() && !$user->tier->one())
	{
		return false;
	}

	return Craft::$app->getMailer()
		->composeFromKey('account_activation', ['link' => Template::raw($url)])
		->setTo($user)
		->send();
}
```

So that's why the activation emails are erroring out for me, but the same should be true in any case where the Mailer failed to successfully send, with perhaps one of the biggest potential reason being misconfigured email settings.

### Steps to reproduce

1. Add a new or "Pending" user to a group or customize individual permissions, and make sure it's saved.
2. Prevent the activation email from being sent successfully, either by returning `false` from the `sendActivationEmail` method or by intentionally misconfiguring the email settings.
3. Click "Send activation email" from the sidebar, and you should get the "Couldn’t send activation email. Check your email settings." flash message.
4. Inspect the Permissions tab and notice all the permissions are unchecked.

### Craft CMS version

Craft Pro 4.4.3

### PHP version

8.1.16

i-just:
Hi, thanks for reporting and great replication steps. I raised a PR to address this.

brandonkelly:
Craft 4.4.7 is out now with a fix for this!

jalendport:
Thanks for the quick fix! 🙌

PR #13071, opened 2023-04-11
added non-json response to actionSendActivationEmail

i-just:
### Description

> If for whatever reason activation emails won't send successfully (e.g. a misconfigured Email setting), any selected checkboxes on the Permissions tab are cleared, which can cause issues if the user is then saved without reloading the page or reselecting the desired permissions.

Return `asSuccess`/`asFailure` if the request accepts json; otherwise, redirect back to prevent the above from happening.

### Related issues
#13061

### DIFF
--- a/src/controllers/UsersController.php
+++ b/src/controllers/UsersController.php
@@ -1713,9 +1713,17 @@ JS,
             );
         }
 
-        return $emailSent ?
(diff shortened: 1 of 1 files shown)


brandonkelly:
Source of the bug was that the templates were just checking if it was a post request, rather than specifically a post request to the `users/save-user` action.

Issue #13072, opened 2023-04-11
[4.x]: Default Upload Location combined with Sources allows for impossible combinations that look like a bug.


MattWilcox:
### Description

I at first thought this was an Asset upload issue (video attached of why I thought that something was broken and "putting assets in the wrong Source").

In actuality I later spotted the problem, and it's this:

<img width="499" alt="Screenshot 2023-04-11 at 16 07 13" src="https://user-images.githubusercontent.com/814039/231207118-dc246c0f-257d-4fc9-be35-0ed47fe3fc7a.png">

Despite the Source being set to only allow "Files", the "Default Upload Location" isn't in any way related to that choice, so unless you know to also adjust Default Upload Location (which itself defaults to the first source defined), the Asset field will happily let you configure it to upload new files into a location that the field itself is not configured to read from.

This results in fields which appear broken, by allowing uploads into a Source that the same field is not able to display or select from.

### Expected behavior

Only Sources that are enabled for an Asset field should be available to set as a Default Upload Location for that field. 
Sources that are not valid for that field should be impossible to set as an upload target.

### "Buggy" experience

https://www.youtube.com/watch?v=hIc2JDAx1jE

### Craft CMS version

4.4.6.1

### Installed plugins and versions

-

brandonkelly:
Craft 3.8.8 and 4.4.8 are out with a couple fixes for this!

- Directly uploading assets to Assets fields no longer respect the Default Upload Location, if the selected volume isn’t one of the allowed volumes. (#13079)
- Disallowed volumes are no longer selectable within the Default Upload Location setting in the first place, when editing Assets field settings. (#13088)

PR #13077, opened 2023-04-12
don't use default upload location if it's not in sources

i-just:
### Description
For the Assets field, if the default upload location is set to a volume which is not selected under Sources, set the default upload location to the first selected volume.

### Related issues
#13072


i-just:
Closing this one in favour of #13079. PR 13079 makes it work in the same way as prior to `3.8`/`4.4`.

PR #13080, opened 2023-04-13
Description for $propagateTo, use FG_RED on error

vnali:
### Description
This PR describes the $propagateTo option and also modifies the 'Invalid site handle:' error color in red to be consistent with other console errors.

Also when wrong site handle error is passed, after the 'Invalid site handle:' error, there is an exception `Exception 'TypeError' with message 'craft\console\Controller::runAction(): Return value must be of type int, null returned` which i don't know if this exception needs a fix or this is expected behavior. 

Thanks!


brandonkelly:
Thanks!

Issue #13082, opened 2023-04-13
[4.x]: When Entrifying Categories, reference tags to them no longer function


dlindberg:
### Description

When Entrifying categories reference tags to those pages (in this instance in Redactor fields) stop working after Entrification.

### Steps to reproduce

1. use `entrify/categories` on a category
2. Links to those categories using reference tags, eg: `{category:19305:url}`, stop working

### Expected behavior

Links to continue to work

### Actual behavior

Links break

### Notes

Changing the reference from `{category:19305:url}` to `{entry:19305:url}`, of course, fixes a specific instance of the problem, but in a site with a substantial amount of content fixing all of this manually is not practical. It may make sense for tag and category reference tags to also check entries for Entrified sites or on a setting, I'm not sure of the practicality of accurately updating all reference tags used in a site.

### Craft CMS version

4.4 +

### PHP version

8.1.x

### Database type and version

mysql

### Installed plugins and versions

- Redactor

brandonkelly:
This has been fixed for the next release. Going forward, category/tag/global set reference tags will automatically be mapped to entries, if there aren’t any category groups/tag groups/global sets defined. (If there aren’t any category groups, we know for sure that a `category:x` reference tag wouldn’t get resolved to a category. And most likely that will be because the category groups were converted to sections.)

I’ve also updated `craftcms/html-field` (the base class for CKEditor and Redactor field types) to automatically update category/tag/global set reference tags to entry reference tags, as fields are resaved, including via the `craft resave/entries` command. (https://github.com/craftcms/html-field/commit/7ecccab102b4e86abb5526411e96594f430f2bb5)

brandonkelly:
Craft 4.4.8 has been released with that change.

PR #13084, opened 2023-04-14
adjustment for reference tags after entrification

i-just:
### Description
After entrifying, e.g. categories, reference tags (e.g. `{category:15:url}`) no longer work. This PR makes a special provision for Categories, Tags and GlobalSets so that they can be found as Entries if no matches weren’t found under their initial type.

### Related issues
#13082

### DIFF
--- a/src/services/Elements.php
+++ b/src/services/Elements.php
@@ -2437,17 +2437,12 @@ class Elements extends Component
                 foreach ($tokensByType as $refType => $tokensByName) {
                     // Get the elements, indexed by their ref value
                     $refNames = array_keys($tokensByName);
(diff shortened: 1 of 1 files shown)


brandonkelly:
Fixed via https://github.com/craftcms/cms/commit/2dd1c6ab7de2c0e05684130f039ec3eda6b18991 (and https://github.com/craftcms/html-field/commit/7ecccab102b4e86abb5526411e96594f430f2bb5) instead.

Admittedly it’s not quite as effective as this PR (it won’t help if you’ve only converted some category groups but not all, for example), but it avoids extra DB queries, and should be good enough for the vast majority of sites which will be going all-in on entrification.

Issue #13083, opened 2023-04-13
[3.x]: Argument 1 passed to craft\helpers\Html::_namespaceInputs() must be of the type string, null given


birdkiwi:
### Description

When filling the field with a large amount of data CraftCMS returns **TypeError: Argument 1 passed to craft\helpers\Html::_namespaceInputs() must be of the type string, null given, called in /app/vendor/craftcms/cms/src/helpers/Html.php on line 601**

With the help of Xdebug I found out that the problem is `$html` variable stored as `null`. Null value returns because of `PREG_BACKTRACK_LIMIT_ERROR` inside preg_replace_callback(). **pcre.backtrack_limit** is "1000000" by default.

https://github.com/craftcms/cms/blob/develop/src/helpers/Html.php#L854
```
$markers = [];
$html = preg_replace_callback('/(<textarea\b[^>]*>)(.*?)(<\/textarea>)/is', function(array $matches) use (&$markers) {
    $marker = '{marker:' . StringHelper::randomString() . '}';
    $markers[$marker] = $matches[2];
    return $matches[1] . $marker . $matches[3];
}, $html);
return $markers;
```

I know that storing such a large amount of text is not a good idea, but I think that such an error should be handled in Craft to avoid customer problems.

I haven't tested it with Craft 4 yet, but looking through source I see the same issue as well.

### Steps to reproduce

1. Create a custom textarea field up to 16mb size limit
2. Attach it to some section and fill this input in the entry. There should be more then 1 million chars in the field
3. Try open / refresh entry in CP again

### Expected behavior
CP should not throw an Exception but show an error or notification maybe? I still don't know the best way to to avoid this behavior.

### Actual behavior
Internal Server error

### Craft CMS version

3.8.5

### PHP version

7.2.34

### Operating system and version

Linux 5.15.90.1-microsoft-standard-WSL2

### Database type and version

MySQL 5.7.40

### Image driver and version

Imagick 3.4.4 (ImageMagick 7.0.10-48)

### Installed plugins and versions

Blitz | 3.12.8
Blitz Recommendations  | 1.3.0
Blitz | 3.12.8
Blitz Recommendations | 1.3.0
CP Field Inspect | 1.4.4
Groupie | 1.0.8
Incognito Field | 1.3.0
Minifier | 3.0.0
Password Policy | 1.2.0
Position Fieldtype | 1.0.17
Protected Links | 0.0.4
Redactor | 2.10.12
Sentry Logger | 1.3.1
[...]

PR #13085, opened 2023-04-14
escape textareas adjustment for v.long content

i-just:
### Description
Adjustment for `Html::_escapeTextareas()` to handle `$html` that’s long enough to trigger `pcre.backtrack_limit` in `preg_replace_callback()`.

### Related issues
#13083


src/helpers/Html.php
```diff
@@ -760,11 +760,27 @@ function(array $match) use ($namespace, $ids): string {
...
+            /* @phpstan-var $data string */
+            $html = $data;
+        } else {
+            preg_match('/(.*?)(<textarea\b[^>]*>)/is', $html, $matches);
+            $start = $matches[0];
+            preg_match('/(<\/textarea>.*)/is', $html, $matches);
+            $end = $matches[0];
+            $marker = '{marker:' . StringHelper::randomString() . '}';
+            $markers[$marker] = str_replace([$start, $end], '', $html);
+            $html = $start . $marker . $end;
```
brandonkelly, line 781: This looks like it’s assuming there is only 1 `<textarea>` tag. Think you’d need to keep doing this in a loop until `preg_match` returns `0`.

And if this method is more performant than `preg_replace_callback()`, we should probably just go straight to using it, rather than just as a fallback if `$data` is` null`.

i-just: I adjusted it for multiple `<textarea>` tags and dropped the `preg_replace_callback()` entirely.

Tested against "more then 1 million chars" as metioned in the issue and also against 16MB of text in a field and this version is still able to process it all.

i-just:
I adjusted it for multiple `<textarea>` tags and dropped the `preg_replace_callback()` entirely.

Tested against "more then 1 million chars" as metioned in the issue and also against 16MB of text in a field and this version is still able to process it all.

Issue #13087, opened 2023-04-17
[3.x]: EVENT_AFTER_APPLY_DRAFT data in DB is not consistent


eduardoalonsoalbella:
### Description

In EVENT_BEFORE_APPLY_DRAFT if we try to get a relationship value ex:

$entry->fieldName->one() we get correctly our custom relationship

But in EVENT_AFTER_APPLY_DRAFT if we try to get that relationship

$entry->fieldName->one() we get null

### Steps to reproduce

1. (explained above)

### Expected behavior

The code should always return the related object

### Actual behavior

In one event the data is returned as expected, in the other one it doesn't work

### Craft CMS version

latest 3.8.7

### PHP version

8.1

### Installed plugins and versions

- latest versions of neo blocks, supertables

eduardoalonsoalbella:
```sql
SELECT `elements`.`id`,
       `elements`.`fieldLayoutId`,
       `elements`.`uid`,
       `elements`.`enabled`,
       `elements`.`archived`,
       `elements`.`dateCreated`,
       `elements`.`dateUpdated`,
       `elements_sites`.`id`      AS `siteSettingsId`,
       `elements_sites`.`slug`,
       `elements_sites`.`siteId`,
       `elements_sites`.`uri`,
       `elements_sites`.`enabled` AS `enabledForSite`,
       `elements`.`canonicalId`,
       `elements`.`dateLastMerged`,
       `jira_issues`.`key`
FROM (SELECT `elements`.`id` AS `elementsId`, `elements_sites`.`id` AS `elementsSitesId`
      FROM `craft_elements` `elements`
               INNER JOIN `craft_jira_issues` `jira_issues` ON `jira_issues`.`id` = `elements`.`id`
               INNER JOIN `craft_elements_sites` `elements_sites` ON `elements_sites`.`elementId` = `elements`.`id`
               INNER JOIN `craft_relations` `relations` ON (`relations`.`targetId` = `elements`.`id`) AND
                                                           ((`relations`.`sourceId` = 375) AND (`relations`.`fieldId` = 121)) AND
                                                           ((`relations`.`sourceSiteId` IS NULL) OR
                                                            (`relations`.`sourceSiteId` = 14))
      WHERE (`elements_sites`.`siteId` = 17)
        AND (`elements`.`archived` = FALSE)
        AND (`elements`.`dateDeleted` IS NULL)
        AND (`elements`.`draftId` IS NULL)
        AND (`elements`.`revisionId` IS NULL)
[...]

PR #13091, opened 2023-04-18
update query joins to not reference the draft

i-just:
### Description
When applying a draft, make sure related queries reference the new canonical element and not the deleted draft.

### Related issues
#13087


brandonkelly:
Fixed via 44df25a6a4a273df42b61a9183e3282fc2b46364

PR #13093, opened 2023-04-18
Add SCENARIO_ACTIVATION

### LINKED ISSUE #13060: [4.x]: Error sending activation email for user when there are conditional tabs with required fields

timkelty:
### Description
https://github.com/craftcms/cms/pull/12432 added element validation when activating users to enforce uniqueness for emails/usernames. To avoid errors from other attributes, this PR adds `User::SCENARIO_ACTIVATION`, which is used when directly activating or generating an activation code for a user.

### Related issues

- Fixes #13060


src/templates/users/_edit.twig
```diff
@@ -89,7 +89,7 @@
                     name: 'email',
                     value: user.email,
                     type: 'email',
-                    errors: user.getErrors('email')|merge(user.getErrors('unverifiedEmail')),
```
timkelty, line 92: We already copy/clear `unverifiedEmail` errors to `email` [in the controller](https://github.com/craftcms/cms/blob/develop/src/controllers/UsersController.php#L1444-L1448), so we shouldn't need to render both here.

If we do, we can end up rendering the same error twice, in the case of an invalid email.

Issue #6338, opened 2020-07-07
'Double-instantiating a menu button on an element' garnish messages in console.log


Friksel:
### Description
Not sure what's triggering this, but obviously something inside 'garnish.js' doesn't quite like something as it's throwing 'Double-instantiating a menu button on an element' in the console log a couple of times.

Just having a simple entry with matrix field on it. No redactor field. No plugin installed. Fresh Craft install.

Not sure when exactly this happens, but this issue is around for quite some time now, but I didn't post it as I had several plugins loaded and don't have the time to debug this. And now I see it happening even on a fresh Craft install without a plugin installed. Looks like it's related to the Matrix block.

### Steps to reproduce

1. Make a matrix field
2. Add a new entry and edit the matrix field inside Entries --> Save

### Additional info

- Craft version: 3.4.27
- PHP version: 7.1.16
- Database driver & version: MariaDb 10.4.10
- Plugins & versions: none

brandonkelly:
I have seen that as well from time to time. Generally perfectly safe to ignore though.

Friksel:
@brandonkelly just for the record: I don't see this from time to time, but regularly. It might be minor, but as a front-end developer it gives, together with other issues I have when running Craft on development environment (many errormessages in the cp which seem to appear if a user is clicking something in the CP when something is being saved or something), and issues I've seen like something isn't initiated, but called, the feeling the front-end of the CP isn't really stable.
Which I am surprised about, as Craft in general feels so well thought through and is so fun to use and I love it.

If I would have more information on when these issues especially appear I tell you.

brandonkelly:
@Friksel Are you seeing any actual unexpected behavior? Or just concerned about the various console logs/warnings/errors?

It’s expected you’d get an error logged if you leave a page mid-Ajax request.

Friksel:
@brandonkelly I see actual unexpected behaviour AND am concerned about the various console logs/warnings/errors.

Can't really point my finger on it when exactly this happens to create an issue about it helping you guys to look into it (if I have a little more time I dive deeper into this and do some specific testing), but it looks like it's especially happening on a little slower environments, like the dev-environment I'm working on. Although not all issues are. Sometimes it just feels random.
For the record: the system is not running out of max_timing/memory or whatever.

If I know more I let you know as I'd like to solve this as much as you do as this doesn't feel nice when using a CMS.
[...]

PR #13179, opened 2023-05-10
different approach to double-instantiating issue

i-just:
### Description
Different approach to fixing “Double-instantiating a menu button on an element” for matrix fields.

Having class `menubtn` on the “Add a block” button for the Matrix field causes the button to be double-instantiated. I removed that class which means when we call `new Garnish.MenuBtn` for that button it’s being instantiated for the first time.

### Related issues
#13177 and #6338


brandonkelly:
Fixed via 2d35352752db5dc7a8de574b9005e24dc7e3ab13. Calling `.menubtn()` to just ensure that the MenuBtn has been instantiated (it will do nothing if it’s already done), and then attaching an event handler to it, rather than using `onOptionSelect`.

Issue #13181, opened 2023-05-10
[4..4.10]: Matrix blocks not propagated when entry is translated/enabled into another website since


yingban:
### Description

When a entry is enabled to another website, the matrix blocks are not propagated, even if the propagation method is set to 'Save blocks to other sites in the same site group".

### Steps to reproduce

1. Create a matrix field with different block types.
2. Assign it to an entry
3. Create blocks inside entry
4. Enable entries into another website

### Expected behavior

The matrix blocks should be copied and 1:1 between the two different versions.

### Actual behavior

Matrix field is empty and the blocks are not propagated.

### Craft CMS version

4.4.10.1

### PHP version

8.1

i-just:
Hi, thanks for raising a separate issue! I created a PR for this.

weotch:
@i-just, I tried out [your PR](https://github.com/craftcms/cms/pull/13185) and it didn't fix the issue for me.  

Here's a video showing how I experience this:

https://github.com/craftcms/cms/assets/77567/791d318a-46ed-495b-bef7-b042635261c8

One thing I've noticed is that setting the `'autosaveDrafts' => false,` config option fixes this.  Here's the same steps with that setting enabled:
 
https://github.com/craftcms/cms/assets/77567/ff6aeef4-13e7-475f-afed-b4cabadbe79f

I also don't experience this when `'autosaveDrafts' => true` if I save the entry after adding the new site super quickly, like before the autosave completes.

i-just:
@weotch, is your “Blocks” field a matrix or a neo? What is that field’s propagation method set to?

weotch:
It is Matrix (though we're using [Spoon](https://plugins.craftcms.com/spoon?craft4) with it).  

Propagation is set to "Save blocks to other sites with the same language"

brandonkelly:
This is fixed now for the next release. We should be getting a hotfix out later today, but if you want to test before then, you can change your `craftcms/cms` requirement in composer.json to `dev-develop#c575444a6e7243e8fb7655c68b8ce44569c74f99 as 4.4.10.1` and run `composer update`.

brandonkelly:
Craft 4.4.11 is out now with the fix.

yingban:
@brandonkelly thanks for fixing

weotch:
4.4.11 did _not_ fix [my issue](https://github.com/craftcms/cms/issues/13181#issuecomment-1544975639).  @yingban did it work for you?

[...]

Issue #13155, opened 2023-05-03
[4.x]: First save of a multisite entry with Matrix blocks orphans the blocks from the primary site


aaronbushnell:
### What happened?

I'm seeing an issue on a Craft site where a handful of Matrix blocks are being orphaned in the database and lost on my primary site when saving the "Home" single in my newly created site.

Here are the steps I'm performing:

### 1. Add a new Site

![Screenshot 2023-05-03 at 9 49 34 AM](https://user-images.githubusercontent.com/315202/235935578-167e4dc7-6333-4e7d-b8c4-157521f43554.png)

### 2. Change my Matrix field's "Propagation Method" so blocks are only on the site they were created in

![Screenshot 2023-05-03 at 9 53 05 AM](https://user-images.githubusercontent.com/315202/235936436-afc716e0-cf0c-4ab4-b09c-24462f59d3f5.png)

### 3. Enable the Home single for the Spanish site

![Screenshot 2023-05-03 at 9 56 33 AM](https://user-images.githubusercontent.com/315202/235937362-6ef2e573-eec6-4f96-bc70-793d8bb331dc.png)

### 4. Visit the Spanish version of the Home single and then save it

**Note:** It appears that the primary site's blocks are in this view already. I _thought_ this would normally be empty because of step 2, but I could be mistaken.

![Screenshot 2023-05-03 at 9 59 37 AM](https://user-images.githubusercontent.com/315202/235938215-242a9a92-3650-454f-856b-58d6d7d7140d.png)

### 5. The primary site Home single has had all but one block removed

![Screenshot 2023-05-03 at 10 00 57 AM](https://user-images.githubusercontent.com/315202/235938732-6f09c478-f775-418d-a618-a4bf71a31667.png)

## Database note
I've attached a screenshot of the blocks with the `ownerId` of `4` (the Home single) before I saved the Home single and after (step 4). You can see that a number of these drop off.

![Screenshot 2023-05-03 at 10 05 13 AM](https://user-images.githubusercontent.com/315202/235940546-d6de336e-ad87-4e0d-8949-2a89d29172ac.png)

I was able to find one of the blocks that had dropped off to see where it went and here's what I found. It looks like that block was assigned a new `ownerId` and it points to a draft entry.

![Screenshot 2023-05-03 at 10 11 04 AM](https://user-images.githubusercontent.com/315202/235941582-9ef863d8-881d-4095-a05c-7a8339ca871d.png)

Let me know if you need me to send anything!

[...]

PR #13185, opened 2023-05-11
matrix blocks propagation when owner added for new site

i-just:
### Description
Following a fix for #13155, matrix blocks were not getting propagated when the owner element was enabled for a new site.

Please note: I can’t actually replicate 13155 to test if this still works as expected. I believe it does, but I couldn’t fully test it.

### Related issues
#13181


brandonkelly:
Resolved via c575444a6e7243e8fb7655c68b8ce44569c74f99

PR #13249, opened 2023-05-24
Update composer.json

leevigraham:
Bump composer to `^2.3.0`

The only possible negative side affects are the updated requirements on the other symfony packages:

image

### Related issues

https://github.com/craftcms/cms/discussions/13248


brandonkelly:
There are compatibility issues with Composer 2.3+, which we need to sort through.

PR #13250, opened 2023-05-25
layout container abstraction

marcusgaius:
### Description
While working with element structures recently, I have noticed how it's not possible to use `craft\base\ElementInterface` to access the structure containing the field layout programmatically. Instead, it would either have to be mapped or called explicitly based on the type of the Element.

This PR is an attempt to make the retrieval of such structures a bit more uniform.

### Related issues
Implementing new feature, no related issues.


brandonkelly:
What is the point of this? `LayoutContainerInterface` doesn’t define any methods, so even if there’s a consistent way of getting the element’s “container”, you would still need to ceck what type of object is returned before you knew how to interact with it:

```php
$container = $element->getLayoutContainer();
if ($container instanceof Section) {
    // ...
}
```

At which point, you could have just as easily checked the element itself.

```php
if ($element instanceof Entry) {
    $section = $element->getSection();
    // ...
}
```

MarcusGaius:
Sections don't contain field layouts.

The idea was mainly to add some uniformity to the retrieval of the structures containing the field layouts, add some distinguishment between those who have them and those who don't (I am not sure about the application of this aspect, possibly for UI), and possibly to include the method to the ElementInterface, as it's the common way to type hint the `$element`-type params.

`LayoutContainerInterface` was just a proposal for the collective name for such structures, and it's not required for the overall idea.

marcusgaius:
I may have went a bit overboard with the introduction of the interface, even from one example you provided, I would think that you or anyone else who is working more with the elements index could find better use cases for this idea, as the information about the volume/entryType/etc could be provided at the ElementInterface level, and not per-element.

For now, I would like to modify the PR to only include the abstraction for the retrieval of the model (and the return type to be set just to `craft\base\Model`).

If you think there could be any room for the models that house element field layouts, such as the example you provided. 
Personally, I am looking into building a verbose RESTful CRUD API within Craft, and a method like `craft\base\ElementInterface::getLayoutContainer(): ?Model` would allow for possibility of dynamically structuring requests and responses without needing to specify which element type it is, like you mentioned. But if you don't find it worth while, I guess I could also make such mapping for my own use, and reference it each time. I just feel the more abstract method is just 8 lines of code away if added at the source, in the element itself. :)

brandonkelly:
> Personally, I am looking into building a verbose RESTful CRUD API within Craft, and a method like `craft\base\ElementInterface::getLayoutContainer(): ?Model` would allow for possibility of dynamically structuring requests and responses without needing to specify which element type it is

Ah, that makes sense. So you could include the container’s attributes in a JSON response, and have the element type-specific logic live only on the receiving end. Thanks for the context.

Still, don’t like the idea of adding a new interface method that doesn’t actually have any purpose internally.

I do have a need for something similar in Craft 5, which should be able to solve your need here as well, so I’ve just backported that change for Craft 4.5 (41b42abec4ecbfad3cdde9b961aa86b2aa4e9ae1).

Going forward you will be able to get the “layout container” like so;

```php
$container = $element->getFieldLayout()?->owner;
```

If `$element` is an entry, `$container` will be its entry type, etc.

marcusgaius:
This is very much exactly what I was looking for. Thanks, Brandon! :)

brandonkelly:
Ended up needing to add a `FieldLayoutProviderInterface`, which the provider classes must implement in order to gain support for this (102ce31547b5dfab42aab0884f643a8546d5b637).

brandonkelly:
Craft 4.5.0 is out with that change.

PR #13251, opened 2023-05-25
Fix namespacing of pseudo-classes

bencroker:
This (tiny) PR fixes the namespacing of pseudo-classes within `<style>` tags by adding a `:` to the list of possible characters that can follow a class name. 

Before:

```php
// Outputs `<style>.ns-a, .b:hover</style>`
Html::namespaceHtml('<style>.a, .b:hover</style>', 'ns', true))
```

After:

```php
// Outputs `<style>.ns-a, .ns-b:hover</style>`
Html::namespaceHtml('<style>.a, .b:hover</style>', 'ns', true))
```

Discovered and reported via Discord by @maxstrebel.

### DIFF
--- a/src/helpers/Html.php
+++ b/src/helpers/Html.php
@@ -836,7 +836,7 @@ class Html extends \yii\helpers\Html
                         return $match[0];
                     }, $match[2]);
                 if ($withClasses) {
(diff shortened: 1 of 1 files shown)


brandonkelly:
Thanks!

brandonkelly:
Craft 4.4.14 is out with this fix.

PR #13255, opened 2023-05-26
load fallback asset icons as inline svgs

i-just:
### Description
Load fallback asset icons as inline SVGs.

@timkelty LMK if I have missed the point here!

### Related issues
[#242](https://github.com/craftcms/cloud/issues/242)

### DIFF
--- a/src/helpers/Assets.php
+++ b/src/helpers/Assets.php
@@ -905,7 +905,44 @@ class Assets
             return $path;
         }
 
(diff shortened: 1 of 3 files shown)


brandonkelly:
Allowing an SVG to be returned by `getThumbUrl()` felt weird. Implemented in a different way in #13262, with a new `getThumbSvg()` method.

Issue #13253, opened 2023-05-25
[4.4.13]: Thumbnails not loading when quickly switching between entry sources


thupsi:
### Description

Asset thumbnails not loading when quickly switching between entry sources

### Steps to reproduce

1. Have multiple sections with an asset field that is displayed as a thumb in the entry index.
2. Quickly switch between the corresponding sources/sections in the entry index. Do this multiple times and try to do it pretty fast, so there is not enough time to load all the thumbs before you switch.

### Expected behavior
All thumbnails should load.

### Actual behavior
Thumbnails stop appearing/loading. 
The following error appears in the console:

<img width="923" alt="image" src="https://github.com/craftcms/cms/assets/19857599/398c593e-91cf-47c2-9070-33d5cb9c6090">

If the user refreshes the page the thumbs reappear.

### Craft CMS version

4.4.13

### Installed plugins and versions

-

i-just:
Hi, thanks for reaching out. Is it only happening when the transforms are being generated for the first time?

thupsi:
At first I thought so, but no. It happens even if transforms have been generated.

thupsi:
The issue seems to be a bit more widespread. It also sometimes happens in asset select modals without the editor frantically switching sources. Sorry for being vague, haven't had time to isolate the exact conditions, but there have been similar reports in Discord.

brandonkelly:
Craft 4.4.14 is out with a fix for this. Thanks again for reporting!

PR #13263, opened 2023-05-30
free up thumb loader workers if all active on load

i-just:
### Description
If your element sources are set to show assets, and you quickly switch between sources before they all load, no element thumbs will load until you reload the page.

This is because all workers have their `active` property set to `true` and can’t do any processing. When you refresh the page, `init` is called again, which creates new workers that can load thumbs again.

I have added a bit of logic to the `src/web/assets/cp/src/js/ElementThumbLoader.js` `load` method to check if all workers are “busy” on load and if they are, free them up.

### Related issues
#13253


brandonkelly:
Probably overkill, but went about this in a different way in a80aec46b6155024bae55d00c22d121ad835a57f.

Issue #13270, opened 2023-06-02
[4.x]: datefmt_create: No such time zone: '+01:00': U_ILLEGAL_ARGUMENT_ERROR


kamejosh:
### Description
After upgrading from Craft `4.4.10_1` I can no longer edit a matrixBlock type. presumably because some field triggers this error. It is not happening with all matrixBlocks, so i will provide the config.
I tried creating a field for all the different types available in my project but outside of this matrixBlock nothing triggered the error

### Steps to reproduce

1. Install Craft 4.4.13
2. Go To Settings
3. Go to Fields
4. Try opening a Matrixblock
5. Error appears

### Expected behavior
Editing matrixBlock is possible without Server Error

### Actual behavior
Editing certain matricBlocks trigger a Server Error

### matrixblock Config
```
columnSuffix: null
contentColumnType: string
fieldGroup: 9c75cf65-133a-4757-8c1f-c03ae2d4ce6e # Common
handle: contentBlocks
instructions: null
name: Seiteninhalte
searchable: false
settings:
  contentTable: '{{%matrixcontent_contentblocks}}'
  maxBlocks: null
  minBlocks: null
  propagationKeyFormat: null
  propagationMethod: all
translationKeyFormat: null
translationMethod: site
type: craft\fields\Matrix

```

### matrixBlockTypes config files
[matrixBlockTypes.zip](https://github.com/craftcms/cms/files/11634018/matrixBlockTypes.zip)

### Craft CMS version

4.4.13

### PHP version

8.2.6

### Operating system and version

macOS 13.4 and Linux Ubuntu

### Database type and version

Mysql 8.0.29

### Installed plugins and versions

    "craftcms/contact-form": "3.0.1",
    "craftcms/contact-form-honeypot": "2.0.0",
    "craftcms/redactor": "3.0.4",
    "ether/seo": "4.1.2",
    "misterbk/mix": "^1.6",
    "shornuk/craft-plausible": "2.0.0",
    "verbb/field-manager": "3.0.6",
    "verbb/super-table": "3.0.9",

kamejosh:
Just performed an additional test, the first version the error is happening in ist Craft `4.4.12` so it was probably introduced in this version. `4.4.11` is running without issue.

i-just:
Hi, thanks for reporting! This affects the default date and time values in an editable table. I raised a PR for it.

brandonkelly:
[...]

PR #13277, opened 2023-06-05
different handling of Date and Time for tables

i-just:
### Description
Fixes a problem where after changes from https://github.com/craftcms/cms/commit/e1714a84ec8a3db3c1a158e21bf749c4ba2816aa, default date and time fields in a table field could no longer be displayed when trying to edit the field, throwing `datefmt_create: No such time zone: '+01:00': U_ILLEGAL_ARGUMENT_ERROR`

The `IntlDateFormatter` doesn’t seem to like the `timezone_type=1` (e.g. `+01.00`).

### Related issues
#13270

### DIFF
--- a/src/fields/Table.php
+++ b/src/fields/Table.php
@@ -136,7 +136,7 @@ class Table extends Field
                 if (in_array($col['type'], ['date', 'time'], true)) {
                     foreach ($config['defaults'] as &$row) {
                         if (isset($row[$colId])) {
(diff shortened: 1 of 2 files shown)


brandonkelly:
Restored the `date` and `time` input templates to pre-e1714a84ec8a3db3c1a158e21bf749c4ba2816aa behavior, and now exposing `|date` and `|time` filters’ `timezone` args via a new `timeZone` property that can be passed to the template instead. (e1714a84ec8a3db3c1a158e21bf749c4ba2816aa)

PR #13287, opened 2023-06-06
Feature/dev 1082 modals selecting elements beyond limits

i-just:
### Description
Prevents selecting more elements than the “max relations” or “branch limit” (depending on which limit is set for the field).

If a limit is set: 
- `selectAll` will select all elements from the first one up to the allowed limit
- `selectRange` will select from the first selected element to the one that doesn’t go above the limit
- `selectItem` won’t allow selection if a limit has been reached

If “max relations” is set, we count the elements.
If “branch limit” is set, we count only top-level (level 1) elements.

If “max relations” is set and the limit has been reached, the “Add a {type}” button is hidden.
If “branch limit” is set and the limit has been reached, the “Add a {type}” button is not hidden because, technically, you can still select additional elements if they’re the descendants of the already selected element(s) - see question 2.

Questions:
1. For `selectAll`, alternatively, we could disallow this option if there’s a limit and that limit is greater than the count of all the items?
2. Based on the GH issues related to dev-1082, this will be a confusing experience for some. Maybe we should add some messaging when the modal is open or visualise/highlight what can still be selected?

### Related issues
[DEV-1082](https://linear.app/craftcms/issue/DEV-1082/dont-allow-selecting-entriescategories-beyond-the-branch-limit-setting)


timkelty:
> Based on the GH issues related to [dev-1082](https://linear.app/craftcms/issue/DEV-1082), this will be a confusing experience for some. Maybe we should add some messaging when the modal is open or visualise/highlight what can still be selected?

Maybe something as simple as:
CleanShot 2023-12-15 at 17 08 03@2x

For `selectAll` and `selectRange`, I feel like it might get awkward to start interfering there…What if we just don't do anything there, but disable the select button and show some error text:
CleanShot 2023-12-15 at 17 09 52@2x

PR #13288, opened 2023-06-06
Update markup on account page to use tabpanel roles and label panels

gcamacho079:


gcamacho079:
The account page no longer uses tabs - closing and marking Linear task as cancelled.

Issue #13296, opened 2023-06-08
[4.x]: Field conditions, not all fields visible


lennonpuype:
### Description
When you have multiple fields with the same name (I know this is not the way to go 😉  ) You have an issue with the field conditions that when there are with the same name, only 1 field with that name is visible.

It is just a bug I came across, but nothing urgent. I just didn't saw why it would work, but then eventually did 😉 

### Steps to reproduce

1. Have a field you want to hide for example when a toggle is checked
2. Make fields with the same name, but different handles. For example "Toggle" with handle "toggle" and "Toggle" with handle "toggle2"
3. Add this field to a structure or neo block (plugin field) or any other thing to add this to
4. Set the condition of the field you want to hide and you see that only 1 toggle (the one with handle "toggle") will be visible to use as condition, but I want to use "Toggle" with handle "toggle2"

### Expected behavior
I want both fields with the same name to be checked in the condition. So not title related like it is now, but handle related.

![Screenshot 2023-06-08 at 12 57 29](https://github.com/craftcms/cms/assets/30866096/fc83a22c-6738-43b3-b0e4-f4c1dccf90e6)

### Actual behavior
Only 1 field of the same name is now visible

![Screenshot 2023-06-08 at 12 58 21](https://github.com/craftcms/cms/assets/30866096/5f52f81b-5f00-4e86-9b27-c7500fbda9b8)

### Craft CMS version

4.3.10

### PHP version

8.0

### Installed plugins and versions

-

i-just:
Hi, thanks for reporting and for the clear description! I raised a PR to adjust this.

brandonkelly:
Craft 4.6.0 was just released with a fix for this. Now custom fields’ condition rule options will start showing the field handle as well, for admins with the “Show field handles in edit forms” preference enabled. So those option labels will no longer be duplicates of each other.

(Non-admin users will still have no way of selecting duplicate condition rules though. If that’s a problem, you will need to give your fields unique names.)

PR #13300, opened 2023-06-09
adjustment for fields with the same label in conditions

i-just:
### Description
Condition builders that show custom fields were not showing all the fields if the fields had the same label. E.g. if you have a custom field named “Toggle” with a handle of `toggle` and another custom field named “Toggle” with a handle of `toggle2`, only one would show in the condition builder.

I changed how we compile a list of options for the condition builder so that the field handle is used as a key for field condition rules. 
I also adjusted how the custom field options are displayed in the menu - they will now show with the field handles (like they do in the field layout builder).

Screenshot 2023-06-09 at 10 09 23

Once a field is selected, it shows as it used to (just with the Label), and if you expand the menu, you can see the handle for it.
Screenshot 2023-06-09 at 10 10 19

### Related issues
#13296


i-just:
@brandonkelly I just merged `develop` into this PR and re-tested - all good. Could you please have a look for the next release (the issue was raised again)?

brandonkelly:
Handles aren’t meant to be author-facing, so I don’t think we should be surfacing them within element index filters.

Maybe it’s OK for conditions that get stored in the project config (`$this->condition->forProjectConfig`), which should only be viewable by developers, such as within custom element source settings and conditional field settings. Otherwise, I think the current behavior is fine. (If someone has an issue with it, they can rename their fields to be unique.)

i-just:
@brandonkelly, I changed when we show the handle. It’ll show only if you have the `showFieldHandles` preference turned on. I think that has a nice consistency to it. LMK if you’re okay with that or if you’d like to show it only if the condition is for project config.

brandonkelly:
I still think it would be better to only do it for conditions that will be save to the project config. Otherwise it could be a little confusing why some field types are listed sometimes, but not others.

i-just:
It is adjusted so that the handle will only show if `forProjectConfig` is `true` and only if the `showFieldHandles` preference is turned on.

brandonkelly:
On second thought, it makes sense to just show field handles everywhere, for users with the “Show field handles in edit forms” preference enabled, so reverted that last change you made.

Also changed it up a bit so there’s now generic support for condition rules to supply “hints” (similar to Selectize menus), rather than `BaseCondition` having explicit support for `FieldConditionRuleInterface`.

PR #13302, opened 2023-06-09
Feature/dev 1098 ability to split up full name

i-just:
### Description
Adds the ability to overwrite parsed User’s First and Last names. The same mechanism applies in the control panel and via front-end submission.

- you create a new user or edit an existing one and provide only Full Name - we parse First and Last names from it (like we do now)
- you create a new user or edit an existing one and provide only First and Last names - we parse Full Name from them (like we do now)

We show the Full Name field when creating a new user and for existing users when we determine that the first and last names parsed from the full name are the same as the values in the individual first and last name columns. 

If we detect that they’re different, the Full Name field is hidden, and we show First Name and Last Name fields. 

A new “Edit user” lightswitch is added to the edit user page in the Control Panel to allow you to switch to First Name and Last Name fields if you want to use those.

In other words, we only show one set of fields - either Full Name or First Name and Last Name.

User with a full name: Mary Jane van Smith (first name: Mary, last name: van Smith)
Screenshot 2023-06-15 at 16 16 52

User with first name: Mary Jane, last name: van Smith (full name: Mary Jane van Smith)
Screenshot 2023-06-15 at 16 17 07

### Related issues
[DEV-1098](https://linear.app/craftcms/issue/DEV-1098/ability-to-split-up-full-name-fields)


brandonkelly:
I spent some time trying to improve the UI on this and just not very happy with it, so made it a config setting instead, for Craft 4.6 (#14097).

Issue #13282, opened 2023-06-05
[4.x]: After updating to Craft 4, `/admin/utilities/upgrade` is a dead link


lindseydiloreto:
### Description

Very frequently, just before updating to Craft 4, I am looking at the "Craft 4 Upgrade" screen...

```
https://mywebsite.com/admin/utilities/upgrade
```

**BEFORE UPGRADING TO CRAFT 4**

<img width="555" alt="Before upgrading to Craft 4" src="https://github.com/craftcms/cms/assets/5309692/94dd74d7-eeba-403e-9fb3-45f86d407cba" style="margin-bottom:20px">

### The Problem

After upgrading to Craft 4, that page becomes a _broken URL!_

**AFTER UPGRADING TO CRAFT 4**

<img width="495" alt="After upgrading to Craft 4" src="https://github.com/craftcms/cms/assets/5309692/1a7dbe02-d8e8-4746-973e-ec73b53312e6">

### Steps to reproduce

1. In Craft 3, go to the "Craft 4 Upgrade" screen.
2. Upgrade everything to Craft 4.
3. Refresh the "Craft 4 Upgrade" page.
4. See a broken page.

### Expected behavior

It should (at the very least) redirect to the "Updates" page (which would likely show that everything is now up-to-date).

### Actual behavior

A broken page.

### Craft CMS version

3 --> 4

### PHP version

n/a

### Operating system and version

n/a

### Database type and version

n/a

### Image driver and version

n/a

### Installed plugins and versions

n/a

brandonkelly:
Just released Craft 4.4.14 with a fix for this.

lindseydiloreto:
Awesome, thanks @i-just & @brandonkelly! 🎉

PR #13309, opened 2023-06-13
send utilities/upgrade requests to system report

i-just:
### Description
After Upgrading to Craft 4, the `utilities/upgrade` link will redirect to Utilities > System Report.

### Related issues
#13282


brandonkelly:
Implemented a simpler fix via 26bbabaf98edd601044a24a3f683e3f1748fec8c.

Issue #13350, opened 2023-06-23
[4.x]: URI autofilled when creating new SINGLE section


remcoov:
### Steps to reproduce

1. Create a new section
2. Fill with a name (and thus handle)
3. Change section to a SINGLE
4. The URI is empty (you can even do an extra backspace or delete to be sure) - which is what I want
5. When saving the SINGLE the URI gets filled anyway

### Craft CMS version

4.4.14

### Installed plugins and versions

-

i-just:
Hi, thanks for getting in touch. I have raised a PR for this. 

For now, if you edit the single section again, you will be able to remove the URI.

brandonkelly:
Craft 4.4.15 is out with a fix for this. Thanks for reporting!

PR #13355, opened 2023-06-26
new section - update uri (format) even if field hidden

i-just:
### Description
Improves the UX of new section creation by updating the URI/URI format field based on the name, even if the URI/URI format field is hidden. This is achieved by adding the `updateWhenHidden` property to the `BaseInputGenerator` settings. It’s `false` by default resulting in the same behaviour as currently, but if set to `true`, even hidden inputs will be populated with the generated value.

Previously, when you went to create a new section, and you filled out the name, the handle and URI got populated with the generated value. However, if you then changed the section type from channel or structure to single, the URI remained blank until you went on to save the section, at which point the URI got populated based on the section name and saved. You then had to edit the single section again and remove the URI if that’s what you wanted in the first place.

A similar logic applied the other way around too. If you went to create a new section, changed the type to single, and then filled out the name, the handle and URI were populated with the generated value. If you then changed the section type to channel or structure, the URI was blank and only populated on save.

Now, fields for the URI format for the new section get updated with the generated value even if the field is hidden, which means that the user sees the actual URI/URI format that will be saved and can adjust it without it being updated and overwritten on save.

### Related issues
#13350


brandonkelly:
Nice one!

Issue #13345, opened 2023-06-21
[4.x]: Addresses have no UK County (administrativeArea) option


bymayo:
### Description

Since using the new address options in Craft 4, we've noticed that when you select United Kingdom as a country, that County field just vanishes, and theres no `administrativeArea` for UK.

This is strange as we definitely have Counties 😂 Lancashire, Manchester, Yorkshire, London etc. 

It's particularly problematic when using addreses with Commerce, as we've had to add our own County field, but then this causes confusion with people processing orders as theres sometimes 2 county fields in the address, one mainly being blank unless it's a UK address.

### Steps to reproduce

1. Go to Addresses
2. Add/edit address
3. Change address country to United Kingdom

### Expected behavior

County / administrativeArea field should appear

### Actual behavior

County / administrativeArea field doesn't appear

### Craft CMS version

4.4.11

### PHP version

8.1

### Installed plugins and versions

- Craft Commerce 4.2.11

i-just:
Hi, thanks for reaching out. It sounds like a duplicate of: #13101. There's more info there which should help clarify things!

brandonkelly:
Craft 4.5.0 is out with a new event that makes it possible to customize address subdivisions (#13361).

Issue #13101, opened 2023-04-19
[4.x]: UK counties no longer available


chrismou:
### Description

After upgrading to Craft Commerce 4, and spending far too long debugging to try to find out what Ive missed, it seems that the new Administrative Area concept doesn't include British Counties.

The issue is mostly around shipping in Commerce. The UK incorporates England, Scotland Wales and Northern Ireland, and while we're effectively considered a single country in Craft, in reality Northern Ireland is on the other side of the Irish Sea, and the Scottish Highlands are well outside of most shipping companies scope for, say, one day delivery.

So up to now we've based shipping rules on counties. But since upgrading, its no longer possible. This massively complicates things as, using DPD as an example, this is what they consider the Scottish Highlands

I tried 

### Steps to reproduce

1.

### Craft CMS version

4.x

### PHP version

8.1

### Installed plugins and versions

-

i-just:
Hi, thanks for reaching out. It sounds like you might want to have a look at: https://github.com/craftcms/cms/discussions/11788. 

The library ([commerceguys/addressing](https://github.com/commerceguys/addressing)) doesn’t come with UK subdivisions (more details in that discussion), but the solution mentioned there should allow the input and saving of plain text values.

You might also find this helpful: https://github.com/craftcms/cms/issues/12644.

chrismou:
@i-just Thanks, that actually is really useful. I've already switched the shipping rules to be postcode based, but I'd not had a chance to look into the counties not saving so hopefully this will solve it for me :+1:

bymayo:
@i-just Just jumping in on this as I posted a duplicate (Sorry!) But I don't see the solution Brandon posted actually allowing to display the Counties for UK, only display a field (Which i'm assuming as it's untested, just displays an input field, not a select)

I'm also a little confused as to why this can't be added to the core? As I assume a lot of developers will be using Craft and Craft Commerce in the UK, and to neglect having UK Counties in the system and needing to add them manually with a module seems a bit crazy to me.

i-just:
Yes, if you use those events, the `administrativeArea` field will be a text input in the Control Panel, but you can render it as a dropdown in the front end. Additionally, from `4.5`, you’ll be able to override the default address formatter and display the `administrativeArea` as part of the card view (https://github.com/craftcms/cms/pull/13242).

[...]

PR #13242, opened 2023-05-24
Allow override of default address formatter

lukeholder:
### Description

Allows replacement of the default formatter that is used by the Addresses service.

Custom Formatter class:

```php
<?php

namespace modules;

use CommerceGuys\Addressing\AddressInterface;

class MyFormatter extends \CommerceGuys\Addressing\Formatter\DefaultFormatter
{
    public function format(AddressInterface $address, array $options = []): string
    {
        return parent::format($address, $options) . $address->phone;
    }
}
```

Inject into app config:

`app.php`

```php
'components' => [
    'addresses' => [
        'class' => \craft\services\Addresses::class,
        'formatter' => new \modules\MyFormatter(
            new \CommerceGuys\Addressing\AddressFormat\AddressFormatRepository(),
            new \CommerceGuys\Addressing\Country\CountryRepository(),
            new \CommerceGuys\Addressing\Subdivision\SubdivisionRepository()
        )
    ],
],
```

### Related issues and discussions

Craft:

https://github.com/craftcms/cms/discussions/12615

Commerce issues:

https://github.com/craftcms/commerce/issues/3152
https://github.com/craftcms/commerce/issues/3082

PR #13361, opened 2023-06-27
Feature/address subdivisions event

i-just:
### Description
This PR adds:
- a mechanism to supplement the `commerceguys/addressing` subdivision data, 
- event to allow developers to manipulate the subdivision data, 
- GB counties data.

Calls to `Craft::$app->getAddresses()->getSubdivisionRepository()->getList()` are now going through `commerceguys/addressing`, and our method. Afterwards a `EVENT_DEFINE_ADDRESS_SUBDIVISIONS` event is triggered to allow developers to manipulate the data even further.

Taking `GB` as an example, the counties data is now shipped with Craft, however it's still up to the developer if they want to show the `AdministativeArea` field in the Control Panel. If yes, you still need to use the `EVENT_DEFINE_USED_FIELDS` event, and you can change the label of the field via `EVENT_DEFINE_FIELD_LABEL`.

To manipulate the subdivisions data, you can now use the `EVENT_DEFINE_ADDRESS_SUBDIVISIONS`, like so:

```php
Event::on(
    Addresses::class,
    Addresses::EVENT_DEFINE_ADDRESS_SUBDIVISIONS,
    function(DefineAddressSubdivisionsEvent $event) {
        if (count($event->parents) == 1) {
	    if ($event->parents[0] === 'GB') {
	        // adds a blank option to the beginning of the counties list
	        $event->subdivisions = ['' => 'Please select'] + $event->subdivisions;
	    } else if ($event->parents[0] === 'PL') {
	        // adds a list of voivodeship when selected country is Poland
	        $event->subdivisions = [
	                'Małopolskie' => 'Małopolskie',
			'Podkarpackie' => 'Podkarpackie',
	                'Ślaskie' => 'Ślaskie',
	                'Wielkopolskie' => 'Wielkopolskie',
	                ...
	            ];
	        }
	    }
        }
    }
);
```

The data added via the new `addresses/SubdivisionRepository` and via the `EVENT_DEFINE_ADDRESS_SUBDIVISIONS` is then used consistently throughout the Control Panel.

e.g. when adding Address for a User
Screenshot 2023-06-27 at 14 00 47

when using Administrative Area Address Condition (e.g. in a Shipping Zone in Commerce):
Screenshot 2023-06-27 at 13 36 27
Screenshot 2023-06-27 at 13 37 22

### Related issues
#13101
https://github.com/craftcms/cms/discussions/11788
#13345


samhibberd:
Just testing this feature out in advance of the 4.5 release, got the 'County' included by adding `administrativeArea` via the `EVENT_DEFINE_USED_FIELDS` event, but the label identifies the field as a 'Province', which looks to be the default / fallback `administrativeAreaType`. 

It looks like the address format repository for the GB countryCode, returns `null` for `$format->administrativeAreaType()`.

New to the Craft 4 addressing logic, and understand that I can tweak via the `EVENT_DEFINE_FIELD_LABEL` event, but not sure if I have missed anything or any required steps?

Also not sure if I have misunderstood, but grabbing the format and then requesting the used fields, skips any craft events, is that by design or should they be called:

```php
// When called via the service the craft events are called:
Craft::$app->getAddresses()->getUsedFields('GB');

// If accessing them the format then they are not?
$format = Craft::$app->getAddresses()->getAddressFormatRepository()->get('GB');
$format->getUsedFields();
$format->getRequiredFields();
$format->getPostalCodePattern();
```

samhibberd:
Hi @i-just, thanks for the detail, all makes sense! So always going to be best running through the Craft service when building out our front end forms.

>It's also worth noting that in 4.5 you'll be able to create your own formatter thanks to this PR: https://github.com/craftcms/cms/pull/13242

Already got this up and running, really useful!

4.5 getting close?

Thanks!

i-just:
Not a problem 🙂 

4.5 is not too far away, but we don't have a firm date.

samhibberd:
Hey @i-just, not sure if i'm again missing some obvious stuff here, but struggling to get my head around this PR, the main issue we have is that we are migrating over 130k+ addresses of which most are UK addresses and have a County (assuming maps to an administrativeArea) and an Address Line 3 (which is now a custom field). 

This PR allows us to identify the `administrativeArea` as a used field / and we can give it a custom label of County, but when using a custom formatter (so that these changes can reflect in the CP) none of the values / data (for administrativeArea or addressLine3) are available when building out the address format string, because the formatters use the `getUsedFields()` methods directly on the formats provided by the third party library.

Is it a case that we need to completely rewrite custom formatter logic to just build output an address string in the same format that craft requires) but by using our own logic and values that are set on the $address for uk addresses only.

```php
public function format(AddressInterface $address, array $options = []): string
{
    if($address->getCountryCode() == 'GB')
    {
        return $address->addressLine1 . $address->administrativeArea;
    }
    return parent::format($address, $options);
}
```
It feels like we absolutely need this PR, as County is always asked for in the UK, but the complexities to get it setup are a creating a bit of a barrier, over here anyway.

Also not sure if these are bugs or further configurations, but when editing in the cp:
1) The County is shown above locality, it should be below?
2) If you change the value set on the now shown County field, it wipes out any value set in the locality (Town/City) input on the first change, although other values are retained.

https://github.com/craftcms/cms/assets/1846063/2cbc6ec7-bad3-4522-9fe0-24b277613f79

i-just:
> This PR allows us to identify the administrativeArea as a used field / and we can give it a custom label of County

To be clear, you were able to define used fields, define field labels and define used subdivision fields since 4.3. This PR adds `EVENT_DEFINE_ADDRESS_SUBDIVISIONS` which lets you manupulate the subdivision data and also adds that data for GB.

> but when using a custom formatter (so that these changes can reflect in the CP) none of the values / data (for administrativeArea or addressLine3) are available when building out the address format string, because the formatters use the getUsedFields() methods directly on the formats provided by the third party library.

I just tested the format you're after for GB - first line of address and county (administrative area) and that shows as expected. Screenshot below. Maybe I'm missing what you're after here?
Screenshot 2023-07-26 at 20 04 49


RE point 1:
The formatter affects how address shows on the card (e.g. on the list of addresses in user's arccount, like in the screenshot above). It doesn't affect the order of fields in the Control Panel. (And you can control the order of fields on the front-end yourself.)

RE point 2:
That happens for other countries too (e.g. USA). I'll have a look into it - thanks for reporting!

samhibberd:
Thanks @i-just, I think perhaps a little bit more reading up required, felt like it was getting rather complex to cover what i thought should be pretty simple.

> I just tested the format you're after for GB - first line of address and county (administrative area) and that shows as expected. Screenshot below. Maybe I'm missing what you're after here?

Sorry, my code wasn't clear, I was just providing a basic example of how I might format the address directly not using the logic / methods from the default formatter. I did a bit more reading up and ended up being able to use the `getValues()` and `buildView()` methods from the default formatter to get the formatter to behave correctly. Great because I could also include the custom field Address Line 3 in the address output for all addresses (if it has data), and output in the same format as the default formatter:

```php
<?php

namespace modules\findarace\formatters;

use craft\helpers\Html;
use CommerceGuys\Addressing\AddressInterface;
use CommerceGuys\Addressing\AddressFormat\AddressFormat;
use CommerceGuys\Addressing\Locale;

class AddressFormatter extends \CommerceGuys\Addressing\Formatter\DefaultFormatter
{
    protected $defaultOptions = [
        'locale' => 'en',
        'html' => true,
        'html_tag' => 'p',
        'html_attributes' => [],
    ];

    public function format(AddressInterface $address, array $options = []): string
    {
        $this->validateOptions($options);
        $options = array_replace($this->defaultOptions, $options);
        $countryCode = $address->getCountryCode();
        $addressFormat = $this->addressFormatRepository->get($countryCode);

        // Add the country to the bottom or the top of the format string,
        // depending on whether the format is minor-to-major or major-to-minor.
        if (Locale::matchCandidates($addressFormat->getLocale(), $address->getLocale())) {
            $formatString = '%country' . "\n" . $addressFormat->getLocalFormat();
        } else {
            $formatString = $addressFormat->getFormat() . "\n" . '%country';
        }

        // Include addressLine3 for al addresses
        $formatString = str_replace("%locality\n", "%locality\n%administrativeArea\n", $formatString);

        // Include administrativeArea for GB addresses only
        if($countryCode == 'GB') {
            $formatString = str_replace("%addressLine2\n", "%addressLine2\n%addressLine3\n", $formatString);
        }

        $view = $this->buildView($address, $addressFormat, $options);
        $view = $this->renderView($view);

        // Insert the rendered elements into the format string.
        $replacements = [];
        foreach ($view as $key => $element) {
            $replacements['%' . $key] = $element;
        }
        $output = strtr($formatString, $replacements);
        $output = $this->cleanupOutput($output);

        if (!empty($options['html'])) {
            $output = nl2br($output, false);
            // Add the HTML wrapper element.
            $output = Html::tag($options['html_tag'], $output, $options['html_attributes']);
        }

        return $output;
    }

    protected function getValues(AddressInterface $address, AddressFormat $addressFormat): array
    {
        $values = parent::getValues($address, $addressFormat);
        $values['addressLine3'] = $address->addressLine3 ?: null;
        if($address->getCountryCode() == 'GB') {
            $values['administrativeArea'] = $address->administrativeArea ?: null;
        }
        return $values;
    }

    protected function buildView(AddressInterface $address, AddressFormat $addressFormat, array $options): array
    {
        $view = parent::buildView($address, $addressFormat, $options);
        $view['addressLine3'] = [
            'html' => $options['html'],
            'html_tag' => 'span',
            'html_attributes' => ['class' => 'address-line3'],
            'value' => $address->addressLine3 ?: null,
        ];
        if($address->getCountryCode() == 'GB') {
            $view['administrativeArea'] = [
                'html' => $options['html'],
                'html_tag' => 'span',
                'html_attributes' => ['class' => 'administrative-area'],
                'value' => $address->administrativeArea ?: null,
            ];
        }
        return $view;
    }
}
```




1) Ok understood, so currently no way to control the order of fields in the cp.
2) Thanks.

i-just:
Hi @samhibberd, a quick update to clarify that the `locality` (City/Town) getting cleared if the `administrativeArea` changes is the intended behaviour. But you are also right, there are some issues there, for which I raised this: https://github.com/craftcms/cms/issues/13486. 

Here's how the locality/administrativeArea funtionality works on `4.4.16.1`:

https://github.com/craftcms/cms/assets/4500340/34d38cd5-d3ad-43ce-9f2e-1d7a029761d2

Issue #13360, opened 2023-06-26
[4.x]: \craft\log\ContextProcessor not filtering variables correctly


myleshyson:
### Description
Looking at the comments for `yii\log\Target`, and you should be able to specify not only root level global keys to log but nested keys as well.

```php
    // yii\log\Target line 17

    /**
     * @var array list of the PHP predefined variables that should be logged in a message.
     * Note that a variable must be accessible via `$GLOBALS`. Otherwise it won't be logged.
     *
     * Defaults to `['_GET', '_POST', '_FILES', '_COOKIE', '_SESSION', '_SERVER']`.
     *
     * Since version 2.0.9 additional syntax can be used:
     * Each element could be specified as one of the following:
     *
     * - `var` - `var` will be logged.
     * - `var.key` - only `var[key]` key will be logged.
     * - `!var.key` - `var[key]` key will be excluded.
     *
     * Note that if you need $_SESSION to logged regardless if session was used you have to open it right at
     * the start of your request.
     *
     * @see \yii\helpers\ArrayHelper::filter()
     */
    public $logVars = [
        '_GET',
        '_POST',
        '_FILES',
        '_COOKIE',
        '_SESSION',
        '_SERVER',
    ];
```

However `craft\log\ContextProcessor::filterVars` is only filtering by root level keys using `in_array`, making nested key filtering impossible.

Line 96 in `ContextProcessor` can be updated to
 ```php
$filtered = Collection::make(ArrayHelper::filter($GLOBALS, $vars));
``` 
to allow proper variable filtering. I'm a little rusty on Laravel collections so I couldn't find a clean way to get the same functionality as `ArrayHelper::filter`. 

This is important for us in our production environment. We want to filter to a few server variables so that our log files are a lot easier to parse. The option right now is either all context or no context in logs. 

### Steps to reproduce

1. In app.php, add something like 
```php
'log' => function () {
            return Craft::createObject([
                'class' => Dispatcher::class,
                'monologTargetConfig' => [
                  'logVars' =>  ['_SERVER.REQUEST_URI']
                ],
            ]);
        },
```
2. Visit a page on the site with logging enabled.
### Expected behavior
Expect only to see `_SERVER.REQUEST_URI` information in the log context.

### Actual behavior
Nothing appears at all for context.

### Craft CMS version

4.4.14

### PHP version

8.2

### Installed plugins and versions

-

i-just:
[...]

PR #13362, opened 2023-06-27
filter Globals array using Yii

i-just:
### Description
When `logVars` is provided for the log component, filter `$GLOBALS` array using Yii’s `filter` method, which can deal with negation (e.g. `!var.key`) and nested key patterns (e.g. `var.key`).

All the hard work done by @myleshyson.

### Related issues
#13360

### DIFF
--- a/src/log/ContextProcessor.php
+++ b/src/log/ContextProcessor.php
@@ -8,6 +8,7 @@
 namespace craft\log;
 
 use Craft;
+use craft\helpers\ArrayHelper;
 use Illuminate\Support\Collection;
 use Monolog\Processor\ProcessorInterface;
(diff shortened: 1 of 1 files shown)


timkelty (approved):
Nice work @myleshyson! I wasn't even aware you could use dot notation or `!` negation.

> I'm a little rusty on Laravel collections so I couldn't find a clean way to get the same functionality as `ArrayHelper::filter`.

Collections support dot notation, but no sense bending over backward to support the `!` when we can just use `ArrayHelper::filter`.

brandonkelly:
Craft 4.4.15 is out with this change. Thanks for the PR!

Issue #1340, opened 2017-02-03
500 error for images preview in CP on Windows - GD


fcaron:
### Description
After uploading images in Craft CP, I see them in my folder web/images, but the thumbnails in the CP are not working and when I edit the image, I don't see it either. When I check the url Craft try to load, it says that there is a 500 Internal Server Error:
http://craftcms3.dev.localtest.me/admin/resources/resized/8/30?d=1486129361
The image “http://craftcms3.dev.localtest.me/admin/resources/resized/8/30?d=1486129361” cannot be displayed because it contains errors.

I'm on PHP 7 on IIS Windows 10 with the image driver GD.

### Additional info

- Craft version: 3.0 beta 2
- PHP version: 7.14
- Database driver & version: MySQL
- Plugins & versions: No plugins

Thank you for your great CMS :)

andris-sevcenko:
Can you shoot an email to support@craftcms.com and include the contents of your craft/storage/logs folder?

fcaron:
I see that it works when I remove "set_time_limit" in "disable_functions" from php.ini.
But it would probably be a good idea to catch the error or alert the user in the CP to enable set_time_limit.

Thanks

andris-sevcenko:
Alerting the user in the CP would be of little use if the user was a content editor with no idea what "php.ini" is at all. This is a great middle ground, as you get the error logged in the log files and you also get a nice red HTTP 500 error in your browser console to alert you that something definitely is wrong.

PR #13391, opened 2023-07-06
added entrify event

i-just:
### Description
This PR adds `EntrifyEvent`, fired at the end of each entrify command (`entrify/categories`, `entrify/global-set`, `entrify/tags`).

(After this is released, a Feed Me PR (work in progress) can be released too.)

### Related issues
[#1340](https://github.com/craftcms/feed-me/issues/1340)


brandonkelly (changes requested):
I think it would be better to have separate `EVENT_ENTRIFY_CATEGORIES`, `EVENT_ENTRIFY_TAGS`, and `EVENT_ENTRIFY_GLOBAL_SET` events, each with their own dedicated event classes, with typed properties for all the associated data (`CategoryGroup $categoryGroup`, `Section $section`, `EntryType $entryType`), etc. And then there’s no need for `$elementType`.

PR #13409, opened 2023-07-10
Add `targetElementRecord` to before/after move element in structure function and events

engram-design:
I have a use case where my element needs validation rules when moving it within a structure. See https://github.com/verbb/navigation/issues/359 where I want to enforce user-defined rules on how many nodes for a particular level there can be.

Fortunately, I can use the `Element::beforeMoveInStructure()` function (or the `Element::EVENT_BEFORE_MOVE_IN_STRUCTURE` event if I wanted) to trigger validation with `element->validate()`. However, one glaring issue is that I cannot get the to-be-moved element information from the function or event, because it's not passed to it.

For example, if I want validation rules to check if something can be placed at `level` 2 if something is currently at level 1, there's no way for me to know what the new level will be, until after the fact. I only know about the `level` of it currently.

This information can be grabbed from [`$targetElementRecord`](https://github.com/craftcms/cms/blob/75089d5f7a15396f2601ac92f031e4f13d1375e1/src/services/Structures.php#L478).

Happy to revise this, or let me know if I've gone about this the wrong way.

### DIFF
--- a/src/base/Element.php
+++ b/src/base/Element.php
@@ -66,6 +66,7 @@ use craft\helpers\UrlHelper;
 use craft\i18n\Formatter;
 use craft\models\FieldLayout;
 use craft\models\Site;
+use craft\records\StructureElement;
 use craft\validators\DateTimeValidator;
 use craft\validators\ElementUriValidator;
 use craft\validators\SiteIdValidator;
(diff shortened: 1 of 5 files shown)


brandonkelly:
A breaking ElementInterface method signature change like this would need to wait until 5.0.

Adding additional properties to the `MoveElementEvent` class would be OK though (used by the `EVENT_BEFORE_MOVE_ELEMENT` and `EVENT_AFTER_MOVE_ELEMENT` events on the `Structures` service, where all the target element info is already known). So I’ve just added `$action`, `$targetElementId`, and `getTargetElement()` to that, for Craft 4.5 (#13429).

Also decided to deprecate `Element::beforeMoveInStructure()` / `afterMoveInStructure()` and the events they trigger, since they’re a little inconsistent with the service events (they get called/triggered for moves _and_ inserts), and now the service events are more capable, so we should be steering people toward those.

brandonkelly:
Craft 4.5.0 is out with those new events.

PR #13415, opened 2023-07-11
Fluent/Static Config Fixes

AugustMiller:
### Description

@olivierbon [noticed that a setting or two](https://github.com/craftcms/docs/pull/513) still used the "static" syntax, and issued a docs PR to correct—but it'll have to be handled in `cms` to prevent the changes getting squashed on the next build.

This only affects docblocks; no actual code updates.

### Related issues

- https://github.com/craftcms/docs/pull/513


brandonkelly:
Thanks!

Issue #13435, opened 2023-07-17
[4.x]: Date Range Conditionals don't include the 'To:' date


tommysvr:
### Description
When using Conditional Date Range filters, ‘From:’ is `>=` but ‘To:’ is `<` so a check for entries until today doesn’t include entries created today.

### Steps to reproduce

1. Create a new entry
2. Set a Date Created / Post Date filter with a Range from any date until today's date

### Expected behavior
The newly created entry will show in the filtered results.

### Actual behavior
The new entry will not show (entries matching the 'To:' date will not show)

![screenshot_2023_07_17 _12 11](https://github.com/craftcms/cms/assets/867566/bb47278a-ca18-4dc9-9b3a-0909d64ff7e6)

### Craft CMS version

4.4.15

### Installed plugins and versions

-

MatthewBUK:
Hi @tommysvr ,

Just adding those missing items.

**PHP version**
8.1.5

Operating system and version
Ubuntu 18.04.6 LTS / Linux 4.15.0-163-generic

**Database type and version**
MySQL 5.7.36

**Image driver and version**
Imagick 3.5.1 (ImageMagick 6.9.7-4)

**Installed plugins and versions**
Article Editor | v2.0.1
Code Field | 4.0.9
Colour Swatches | 4.3.0
Control Panel CSS | 2.6.0
Control Panel JS | 2.6.0
Craftagram | 2.0.2
Icon Picker | 2.0.13
Navigation | 2.0.19
Neo | 3.8.2
Redactor | 3.0.4
SEOmatic | 4.0.28

brandonkelly:
Craft 4.6.0 is out now, with a change to date range condition rules, so that they are now inclusive of their end dates.

PR #13438, opened 2023-07-17
date range - add inclusive endDate option

i-just:
### Description
Adds the ‘Including “To” Date’ checkbox to the Date Range condition (for backwards compatibility). When checked, the “To” date  (endDate) is included in the query. The checkbox is unchecked by default.

Screenshot 2023-07-17 at 12 48 00

### Related issues
#13435


brandonkelly:
The added setting seems unnecessary. I think most people would expect that the “To” date would be inclusive, and that the current behavior is wrong.

i-just:
@brandonkelly, sure, a bit, but I added it for backwards compatibility. My thinking was that people might be relying on the current “wrong” behaviour, and without the added setting, e.g. custom sources would start pulling in a broader set of elements. If that’s unnecessary/too cautious, I’m happy to drop it. Or, we could go ahead with this and then drop the setting for 4.6 or 5.0?

brandonkelly:
Yeah let’s just make the change for 4.6 and let it be a sight change in behavior.

i-just:
Extra setting dropped.

Issue #13451, opened 2023-07-18
[4.x]: Creating a draft with unsaved changes overwrite all simple content to other sites


smcyr:
### Description

When you change some content in an entry, it triggers an auto-save with this message "Showing your unsaved changes.". If you create a draft after that, all the non-relational content (textarea, redactor, etc.) from the current site will be copied to the other sites. So, for example, the entry will have all the text fields in English in the French draft except for the matrix or relational fields.

If you create a new draft without doing changes before in the entry, the content isn't copied over the other sites and everything is working as expected.

### Steps to reproduce

1. Do some changes in a text field in an entry in the English site.
2. Click on "Create a draft"
3. Go over the draft in the other site (French for exemple)
4. Now all the text fields are now in English instead of French

### Expected behavior

If you create a draft with unsaved changes, it shouldn't overwrite content in the other sites/languages.

### Actual behavior

If you create a draft with unsaved changes, it overwrites content in the other sites/languages.

### Craft CMS version

4.4.16

### PHP version

8.0.29

### Operating system and version

Linux 4.14.318-240.529.amzn2.aarch64

### Database type and version

MariaDB 10.6.10

### Image driver and version

Imagick 3.7.0 (ImageMagick 7.1.0-50)

smcyr:
No, it's not inside a Matrix field. The problem arises for field with "translated for each language" and also "translated for each site". We were also able to reproduce the problem on Craft websites including a Craft website with only one field and no plugin.

smcyr:
Here is a video of the bug in action:
[screen-capture.webm](https://github.com/craftcms/cms/assets/664920/3f101e7a-2494-4967-b829-7e7019ceb96c)

i-just:
Thanks for all the info. I can confirm I can replicate this, and I see when this issue was introduced.

brandonkelly:
Craft 4.4.17 is out with a fix for this. Thanks again for reporting!

PR #13457, opened 2023-07-20
Track if we're in the process of dropping provisional draft

i-just:
### Description
The issue was introduced in `4.4.12` via https://github.com/craftcms/cms/commit/9cd0a532c0fb387b69f0945c8b143e6b7ead58ba.

I added new `droppingProvisionalDraft` property for elements to track whether we’re in the process of dropping a provisional draft so that `Element::getLocalized()` can build a query based on that.

### Related issues
#13451


i-just:
Closing in favour of #13497

Issue #13456, opened 2023-07-20
[4.x]: Craft not deleting nested fields properly


mikesnoeren:
### Description
When deleting a Matrix (or SuperTable) field, the inner fields are not deleted properly and leave traces of themselve in the project file and database.

I have tested this using a clean installation on the latest version of Craft.

### Steps to reproduce

1. Create a matrix field (or supertable field) with a plain text field.
2. Collect the text fields UID from the project file. (`project.yaml` -> `meta` -> `__names__`)
3. Delete the field matrix field in Craft.
4. Search the project file and the database for the UID.
5. Find traces of the text field in the project files and in the database.

### Expected behavior
All fields have to be deleted properly

### Extra test results when deleting fields
| Type | Removed in project.yaml | Removed in DB |
|--------|--------|--------|
| Plain Text | Yes | Yes |
| Matrix | Yes | Yes |
| Matrix Block | Yes | Yes |
| Matrix Block Field (Plain Text) | No | No |
| SuperTable | Yes | Yes |
| SuperTable Field (Plain Text) | No | No |

### Craft CMS version

4.4.14

### PHP version

8.0

### Operating system and version

DDEV project type: `craftcms`

### Database type and version

Mysql 5.7

### Installed plugins and versions

- Super Table: 3.0.9

mikesnoeren:
@i-just, thanks for the fix, is there an easy way for us to clean up projects that were affected by this bug?

brandonkelly:
@mikesnoeren Looks like @i-just is working on a new command to simplify this (#13503), but in the meantime you can do the following:

1. Delete the entire `meta.__names__` array from `config/project/project.yaml`
2. Run `php craft project-config/apply` (to pull in the manual change)
3. Run `php craft project-config/rebuild` (to rebuild everything, including the `meta.__names__` array)

brandonkelly:
Craft 4.4.17 is out with a fix that ensures nested deleted component names are removed from the `meta.__names__` map immediately, as well as an improvement to the `project-config/rebuild` command to ensure the whole `meta.__names__` map is rebuilt.

PR #13462, opened 2023-07-21
PC - removing nested field names from meta > __names__

i-just:
### Description
When deleting a matrix field, ensure that the nested fields are removed from the `project.yaml` -> `meta` -> `__names__`.

When deleting a nested field from a matrix block, remove that field from the `project.yaml` -> `meta` -> `__names__` too.

### Related issues
#13456


brandonkelly:
Managed to fix this purely in `ProjectConfigData` (026c5d5cc3a62706761a5104b8f06fc7b2b3aad4).

PR #13463, opened 2023-07-21
Fix BaseDateRangeConditionRule not handling negative intervals correctly.

myleshyson:
### Description
This fixes a bug in the `DateTimeHelper::toDateInterval` method, where if passed a negative value, returns a positive interval. This caused an issue in `DateRange::dateIntervalByTimePeriod`, and subsequently the `BaseDateRangeConditionRule`. 

I noticed this when trying to filter elements on the element index page by date period (e.g. After 1 day ago).

I went ahead and added a check for negative numbers in `DateTimeHelper::toDateInterval` and updated the `DateTimeHelperTest` to make sure it works. `DateCreatedConditionRuleTest` also passes now thanks to this fix.

Ran these tests using PHP version 8.2.4.

### DIFF
--- a/src/helpers/DateTimeHelper.php
+++ b/src/helpers/DateTimeHelper.php
@@ -654,7 +654,8 @@ class DateTimeHelper
         if (is_numeric($value)) {
             // Use DateTime::diff() so the years/months/days/hours/minutes values are all populated correctly
             $now = static::now(new DateTimeZone('UTC'));
-            $then = (clone $now)->modify("+$value seconds");
+            $operator = $value < 0 ? '-' : '+';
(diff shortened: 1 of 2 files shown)


myleshyson:
I also noticed some other tests failing when addressing this, like "MoneyTest" and "LanguageValidatorTest". I honestly didn't dig too deep into them. Maybe it's php 8.2 specific or I was missing something that testsuite needed.

brandonkelly:
Thank you!

> I also noticed some other tests failing when addressing this, like "MoneyTest" and "LanguageValidatorTest".

Those are passing for me locally with this PR, so maybe a fluke.

PR #13475, opened 2023-07-24
Update toggletip element roles

gcamacho079:
### Description
Updated roles so that toggletip uses button role instead of `li`. The button label has also been updated to use a disclosure message rather than the contents of the toggletip. Upon testing, live region messages have been nixed as they don’t communicate a status update and may cause the interface to be too chatty.

A couple of outstanding questions I’m hoping to get your opinion on @brandonkelly
- The Tooltip.js uses a HUD, which isn’t ideal for the message container. In addition to having `role="status"`, the message container will be empty and present in the accessibility tree on page load (not `display: none` when closed), and the message container will be emptied and then re-injected when the button is toggled open. In that case, should the message container be something other than a HUD?
- Upon testing, the hover-to-open could have implications for screen magnifier users. Because a screen magnifier user will need to move their cursor to center the message on screen, the message will continue to close unless they click to keep it open; however, this requires two clicks, as the first click would close the toggletip. Should we have the first click toggle it open if the tip is initially opened by hover?


src/web/assets/cp/src/js/ElementEditor.js
```diff
@@ -2023,14 +2023,17 @@ Craft.ElementEditor = Garnish.Base.extend(
...
                     const activity = data.activity[i];
-                    const $li = $('<li/>', {
-                      tabindex: 0,
-                      'aria-label': activity.message,
-                    }).appendTo($ul);
+                    const $li = $('<li/>').appendTo($ul);
+                    const $button = $('<button/>', {
+                      type: 'button',
+                      class: 'activity-container__trigger',
+                      'aria-label': `${activity.userName} active, more info`,
```
brandonkelly, line 2030: Couldn’t this be the same message as within the HUD? Then there’s no need to activate it.

gcamacho079: I think that might be confusing from a screen reader perspective, as a button signals that there's some sort of interactivity available. Having it function as a disclosure (with expanded/collapsed states, and with a label that signals that more info is available) solves that by having the screen reader experience be identical to the visual experience.

brandonkelly:
Gotcha. Most of the HUD code is just for displaying/positioning it properly, so I think it makes sense to use that. We can tweak other behaviors to match the toggletip recommendations though. So to make sure we are on the same page, you are saying:

- The container should have `role="status"`
- The inner contents should be cleared out when the tooltip is closed
- The tooltip shouldn’t get `display: none` when closed

Is that right? And if so, what should we do instead of `display: none`?

gcamacho079:
> The container should have role="status"

Yes, if by container you mean the HUD/tooltip container. 👍🏼 All looks good with those bulletpoints.

> And if so, what should we do instead of display: none?

The `visually-hidden` class could be used as an alternative. Or if those styles are problematic, something like setting opacity to zero, or positioning it outside the viewport could work as well.

brandonkelly:
Alright, made those changes. Let me know how it looks. Also there’s this question, ICYMI: https://github.com/craftcms/cms/pull/13475#discussion_r1283227777

gcamacho079:
> Alright, made those changes. Let me know how it looks. Also there’s this question, ICYMI: [#13475 (comment)](https://github.com/craftcms/cms/pull/13475#discussion_r1283227777)

Looks and sounds great! I made a quick update to add a region role to the container.

I only noticed a couple things:

- [ ] The tooltip should only automatically open on hover, not on focus.
- [ ] If the toggle is focused and the activity center is removed from the DOM, the tooltip stays visible. Ideally it would also be destroyed/deleted.

brandonkelly:
Alright, those are addressed!

PR #13499, opened 2023-08-02
Persistent/default filters

brandonkelly:
### Description

Element indexes now store the serialized filter condition in the query string, so the current results can be shared with others, and filters will persist when navigating back to an element index via browser history.

This also adds support for native sources to define a `defaultFilter` key, which can be set to an element condition instance or config, which defines the default filter condition that should be applied when the source is first selected (per page load). The user will still have the ability to modify or remove it after that.

```php
use craft\base\Event;
use craft\elements\conditions\entries\TypeConditionRule;
use craft\elements\Entry;
use craft\events\RegisterElementSourcesEvent;
use craft\helpers\ArrayHelper;
use craft\models\EntryType;

Event::on(Entry::class, Entry::EVENT_REGISTER_SOURCES, function(RegisterElementSourcesEvent $event) {
    $newsSection = Craft::$app->getSections()->getSectionByHandle('news');
    /** @var EntryType $articleType */
    $articleType = ArrayHelper::firstWhere(
        $newsSection->getEntryTypes(),
        fn(EntryType $type) => $type->handle === 'article'
    );
    $condition = Entry::createCondition();
    $condition->setConditionRules([
        new TypeConditionRule([
            'values' => [$articleType->uid],
        ]),
    ]);

    $event->sources[] = [
        'key' => 'articles',
        'label' => 'Articles',
        'criteria' => ['section' => 'news'],
        'defaultFilter' => $condition,
    ];
});
```

### Related issues

- #11215
- #12590
- #13464


myleshyson:
The default filter doesn't seem to persist when selecting a new source. It's only applied on the first page load for the initial source that's selected. 

https://github.com/craftcms/cms/assets/11747187/6bacc2b6-2cf3-4f13-8264-974fddb20984

Also noticed that I had to use class array syntax when defining `defaultFilter` on `nested` sources, otherwise the elementIndex twig template would error out.

brandonkelly:
@myleshyson Thanks for testing! Just fixed both of those (37ec49484832b05ab95a23db97091e2979a96388).

PR #13503, opened 2023-08-03
utility command to clean orphaned project config meta names

i-just:
### Description
A console command utility to clean orphaned project config meta names.

### Related issues
#13456


brandonkelly:
With just a small change, I was able to get the `project-config/rebuild` command to rebuild the `meta.__names__` array from scratch, so going with that approach instead, as it feels like that would be the first thing most people would try anyway.

i-just:
Makes sense! I got hung up on a wrong idea 🤦‍♀️

mikesnoeren:
> With just a small change, I was able to get the `project-config/rebuild` command to rebuild the `meta.__names__` array from scratch, so going with that approach instead, as it feels like that would be the first thing most people would try anyway.

Hi @brandonkelly, great to hear! When do you think this will be released? Would be great if we could apply that before launching some sites.

PR #13515, opened 2023-08-07
Return target as container config

timkelty:


src/base/ApplicationTrait.php
```diff
@@ -1224,6 +1225,17 @@ public function getLocale(): Locale
...
+    /**
+     * Returns the log dispatcher component.
+     *
+     * @return Dispatcher the log dispatcher application component.
+     */
+    public function getLog(): Dispatcher
+    {
+        /** @noinspection PhpIncompatibleReturnTypeInspection */
+        return $this->get('log');
+    }
```
brandonkelly, line 1237: This is identical to `yii\base\Application::getLog()` so could be removed.

src/log/Dispatcher.php
```diff
@@ -46,7 +46,7 @@ public function init(): void
     }
 
     /**
-     * @return MonologTarget[]
+     * @return Array[]
```
brandonkelly, line 49: should be `array<MonologTarget|array>`

brandonkelly (changes requested):

src/base/ApplicationTrait.php
timkelty: This returns our `Dispatcher`, not Yii's.

brandonkelly: Oh sorry, I missed the updated return type.

src/log/Dispatcher.php
timkelty: Right you are. Fixed!

Issue #13133, opened 2023-04-27
[4.x]: Error when saving Dropdown field without Default option selected


gateszies:
### What happened?
Entries with empty dropdown fields are not able to be saved because of 'Invalid' error.

### Description
We have several Dropdown fields that do not have a Default option selected in the field settings, thus creating many entries with 'undefined' or null dropdown data. Upon updating from Craft 4.3.6.1 to 4.4.7.1, our authors noticed that when saving an entry with one of these 'empty' dropdown fields, it would now produce an error that prevented saving the entry.

I would assume that if defining a dropdown option is required for an entry to be saved, you would also be required to select a Default option when creating the field itself. However, you are able to create and define Dropdown fields with no Default option checked. Once that field is used on an entry it must be defined by the author for the entry to be saved.

**Error**
<img width="1408" alt="Screen Shot 2023-04-27 at 4 04 59 PM" src="https://user-images.githubusercontent.com/66807975/234991005-b3ed8de3-f703-42e9-a91c-d2cd836f6912.png">

**Field Settings**
<img width="1664" alt="Screen Shot 2023-04-27 at 4 05 53 PM" src="https://user-images.githubusercontent.com/66807975/234991279-e6c729bf-c29c-478f-b52b-f5c7beb94817.png">

### Steps to reproduce

1. Create a new dropdown field. Do not check a Default option.
2. Add your dropdown field to an entry type (existing or new)
3. Create/navigate to an existing entry type
4. Do not select a dropdown option as the field is not required
5. Click Create Entry or Save Entry
6. Empty dropdown produces an 'Invalid' error and entry cannot be saved

### Expected behavior
I would expect either Craft to accept an 'empty' dropdown on an entry when no Default option is defined on the field settings. Or, if the dropdown needs to be defined, there should be a similar error when creating a new Dropdown field prompting the user to check a Default option. Otherwise, the latter scenario essentially makes any dropdown field without a Default defined in the field settings a required field for the author.

### Actual behavior
Any field without a Default option defined in the field settings needs to be defined by an author. Dropdowns cannot be 'empty'. No indication of this when creating fields on the back-end.

### Craft CMS version

4.4.7.1

### Installed plugins and versions

-

i-just:
Hi, thanks for reaching out.

[...]

Issue #13034, opened 2023-04-03
[4.x]: Dropdown field appears blank even with a default set


mjniland1:
### Description
The default dropdown property is ignored when adding a new dropdown field to an entry. This isn't a huge issue for one-off dropdowns but if you have a matrix/neo field layout, you have to then set all of the defaults manually on the entry. 

### Steps to reproduce

1. Create a new dropdown field
2. Add a few sample dropdown options to this field
3. Set one of the options as the default
4. Add the dropdown field to an entry

### Expected behavior

The new dropdown field on the entry should have the default applied therefore not requiring a user to make a selection. 

### Actual behavior

The new dropdown field appears blank and throws a "Field is Invalid" error when a user attempts to save. 

### Craft CMS version

Craft Pro 4.4.5

### PHP version

8.0.8

### Installed plugins and versions

-
Batch Actions	1.3.0
Contact Form	3.0.1
Contact Form Extensions	4.2.2
Control Panel CSS	2.6.0
CP Field Inspect	1.4.4
Embedded Assets	3.1.3
Empty Coalesce	4.0.0
Neo	3.7.5
Quick Field	2.0.3
Redactor	3.0.4
Table Maker	4.0.4

brandonkelly:
This is working as expected. The default value will only come into play for _new_ entries, where we definitely know that the existing `null` value was not intentional.

You can bulk-set a specific value on your existing entries using the `resave/entries` command. See [this knowledge base article](https://craftcms.com/knowledge-base/bulk-resaving-elements#resaving-with-specific-field-values) for full details on how to use it.

linusschwab:
@brandonkelly In my opinion, it makes sense to add an empty option to new dropdown fields with no default value on existing entries to "force" the user to make a decision and select a value. But for new fields with a default value set, this is really annoying. 

Are you sure that this is intended behaviour? If I understand correctly, the idea in #12235 was to only add this behaviour for new fields on existing entries with no default value set.

Of course I could use the resave command or write a simple content migration every time a new dropdown field with a default value is added. But this is just unnecessary work in my opinion and was working as expected before the 4.4 update. I don't mind that existing entries have a null value before they are resaved again - just that it throws an error when not explicitly choosing a value even though there is a default value set.

brandonkelly:
@linusschwab The default option has never come into play for existing elements; only brand new ones.

linusschwab:
[...]

Issue #12548, opened 2023-01-18
[4.x]: Default value of Dropdown fields are not saved when using "resave/entries" CLI option


mikesnoeren:
### Description
When using the "resave/entries" CLI option to resave entries in Craft CMS, entries that have a dropdown field with a default value set and have never been saved after the field was added do not have the default value saved to the database. This is likely due to the CLI not taking default values into account, unlike the web UI which automatically selects the default value because a form is used to save the entry.

It is possible that this issue may also occur for other fields where a default value is set, but this has not been confirmed as this behavior has only been observed with dropdown fields by me personally.

### Steps to reproduce

1. Create a new dropdown field in Craft CMS with a default value set.
2. Add the field to a section with existing entries.
3. Use the "resave/entries" CLI option to resave all entries.
4. Check the value of the new dropdown field for the entries within said section, they won't be filled with the defaults.
5. Navigate to an entry and click save without making changes, the default value will now be filled.

### Expected behavior
The new dropdown field should have the default value saved to the database.

### Actual behavior
The new dropdown field does not have the default value saved to the database.

### Craft CMS version

4.3.5

### PHP version

8.0.26

### Operating system and version

Linux 5.15.82-0-virt

### Database type and version

MySQL 5.7.38

### Image driver and version

Imagick 3.7.0 (ImageMagick 6.9.11-60)

### Installed plugins and versions

The version of any installed plugins does not seem to be relevant to this issue, and due to the lack of an easy way to copy and paste this information from the CMS, it will not be included in this bug report

brandonkelly:
Craft 4.5.0 is out with an improvement to Dropdown field behavior: now if a Dropdown field’s value is invalid, and the field contains an explicit Default option, the default option will be auto-selected, and the field will be marked as changed. (#13540)

Issue #12262, opened 2022-11-04
[3.7.59]: Editing Dropdown field labels or values with an existing default causes a blank option to display in the list.


KettyLezama:
### Description
Editing Dropdown field labels or values with an existing default causes a blank option to appear as the first option in the list in an entry that already exists. 

This only happens on existing components; if I create a brand new component with this field, the blank value does not show. This doesn't cause problems when a field is optional, but as soon as the field is mandatory, authors are forced to select a value and resave. With so many entries already using this field, you can see how tedious this can be.

Is there a way to avoid this from happening?

### Steps to Reproduce
1. Let's say I had this as my original Dropdown field, with a default option set:
<img width="1012" alt="Old Dropdown Field" src="https://user-images.githubusercontent.com/3981398/200016075-ceac87d0-7704-4716-afdc-eb91e94b8766.png">

This is how the author sees the field, and this is exactly what I want.
<img width="284" alt="Old Dropdown" src="https://user-images.githubusercontent.com/3981398/200016046-180fdf15-0174-42ac-a429-6be11d13662a.png">

2. But then I need to update the value of the dropdown options, as such:
<img width="1004" alt="New Dropdown Field" src="https://user-images.githubusercontent.com/3981398/200016322-97be8870-fc40-49be-8be4-e9cf60ef9792.png">

This is what the author now sees on existing entry pages that are using this field:
<img width="272" alt="New Dropdown" src="https://user-images.githubusercontent.com/3981398/200016664-eb2a6e25-cd11-4ed2-8ba8-533a472d6cdd.png">

### Expected Behavior
I would expect the dropdown options to display the same way they were originally -- without a blank option.

### Actual Behavior
The first option displayed (and selected) is now blank (null).

### Craft CMS version

3.7.59

### PHP version

7.4.30

### Operating system and version

macOS Monterey 12.6

### Database type and version

MySQL 5.7

### Installed plugins and versions

- "ext-zip": "^1.15",
- "born05/craft-twofactorauthentication": "2.6.1",
- "composer/composer": "2.2.15",
- "craftcms/aws-s3": "1.3.0",
- "craftcms/cms": "3.7.55.3",
- "craftcms/element-api": "2.8.6.1",
- "craftcms/feed-me": "4.5.3",
- "craftcms/plugin-installer": "1.5.6",
- "craftcms/redactor": "2.10.10",
- "ether/tags": "1.0.9",
- "internetztube/craft-element-relations": "1.3.10",
[...]

PR #13540, opened 2023-08-10
Improve invalid Dropdown value behavior

brandonkelly:
### Description

Improves the behavior of Dropdown fields when their value is invalid.

As of 4.4, invalid Dropdown fields which didn’t have a blank option would get a blank option prepended, and that would be selected by default in the UI (see https://github.com/craftcms/cms/discussions/12235#discussioncomment-4059192). This would lead to a validation error on save (because a blank value wasn’t one of the allowed options), forcing the author to make a valid selection.

That confused authors because they were getting validation errors for fields they hadn’t even edited.

With this PR, invalid Dropdown fields which have a default option will now automatically select that instead, and the field will immediately be marked at changed, to indicate that the selected option is not equal to the field’s actual current value, until it’s saved.

A Dropdown field that is marked as changed

### Related issues

- #12262
- #12548
- #13034
- #13133

### DIFF
--- a/src/fields/BaseOptionsField.php
+++ b/src/fields/BaseOptionsField.php
@@ -337,12 +337,25 @@ abstract class BaseOptionsField extends Field implements PreviewableFieldInterfa
             $selectedValues[] = $val;
         }
 
+        $selectedBlankOption = false;
         $options = [];
         $optionValues = [];
         $optionLabels = [];
         foreach ($this->options() as $option) {
             if (!isset($option['optgroup'])) {
-                $selected = in_array($option['value'], $selectedValues, true);
+                // special case for blank options, when $value is null
+                if ($value === null && $option['value'] === '') {
+                    if (!$selectedBlankOption) {
+                        $selectedValues[] = '';
+                        $selectedBlankOption = true;
+                        $selected = true;
+                    } else {
+                        $selected = false;
+                    }
+                } else {
+                    $selected = in_array($option['value'], $selectedValues, true);
+                }
+
                 $options[] = new OptionData($option['label'], $option['value'], $selected, true);
                 $optionValues[] = (string)$option['value'];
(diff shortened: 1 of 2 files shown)


northform:
I've a multisite that was recently updated to 4.5.3 and there are certain select fields which have values assigned to them but they are hitting validation errors.

image

The errors seem to take you in an infinite loop between the sites, so editing Site A shows a validation error for Site B, then when you go to resave Site B, it shoes a validation error for Site C etc.

brandonkelly:
@northform Could you send your Composer files and a database backup over to support@craftcms.com so we can reproduce and look into it from there?

elivz:
I'm running into the same issue as @northform. I am having a hard time putting my finger on exactly what is going on, but I _think_ it has to do with a dropdown field being invalid in two different sites. Craft seems to only "fix" the current site to the default option, but having that invalid value in the other site nevertheless prevents the current site from being saved, so there is no way to update either one of them.

brandonkelly:
@elivz Cross site validation behavior is improving in the next release (see https://github.com/craftcms/cms/discussions/13675#discussioncomment-6982305). Maybe it will solve that for you?

elivz:
Sounds like it should, in theory anyway. Right now I have some pages that are stuck in a state where they are impossible to update short of fiddling with the raw data in the DB, so anything will be an improvement!

brandonkelly:
Any chance you have time to test in the next few minutes as I’m preparing the release? If so and it turns out to still be a problem for you, I could try to improve it further.

If you can, then change your `craftcms/cms` requirement to `dev-develop as 4.5.5`, and run `composer update`.

elivz:
Works now for me! On an entry where I was previously stuck in that validation loop between two sites, I am now able to save each site separately, which updates the dropdown fields to their default values. No errors or other issues. Thanks!

Issue #12290, opened 2022-11-10
[4.x]: 


i-just:
### Description

When editing content in a slideout, if there are validation errors, not all of them are shown under their respective fields and if there are multiple tabs in the element you're editing, tabs with fields containing errors don't have the usual error indication.

Errors are not shown at all for fields inside a matrix block.
Tabs are not indicated as having fields with errors.

### Steps to reproduce

1. Create e.g. entry with a required "entries" or "categories" field (element 1).
2. Create entry or category with at least 2 tabs and some fields in each of them. One of those fields should be a matrix field with at least one block type and in the block type there should be at least one required field. All fields for this element should be required (element 2).
3. Edit element 1 and add to it element 2 via the relations field. Double click on the selected entry/category to open the slideout. Make sure all required fields are empty and attempt to save. 

### Expected behavior
All errors are displayed under their respective fields.
Tabs have indication of having errors in them.

### Actual behavior
There's no indication of which tabs contain errors. 
For the matrix field, if it's required and no blocks are added, the error will show, but if you add a block with a required field and leave that field empty, error won't show for it - your element won't save but there's no indication why.

### Craft CMS version

4.3.1

### PHP version

8.1

mmikkel:
Any chance this will be addressed (i.e. fixed or improved) in a 3.x/4.x release?

I'm actually avoiding adding Matrix sub field restrictions (required fields, max/min limits etc) as much as possible ATM, as clients are very confused when a sub fields fails validation and the only feedback is the "Couldn't save entry" error toast.

brandonkelly:
@mmikkel Yeah we’re looking to improve it in 4.4.

PR #13572, opened 2023-08-18
Bugfix/12290 slideout error indicators

i-just:
### Description
Adds error indication across tabs in a slideout.

### Related issues
#12290


brandonkelly:
This isn’t working reliably for me for Matrix fields with multiple block types, because the `idRegex` is only matching blocks with the same type as the error, and assuming that all blocks have the same type.

Even if we could find a way to fix that, it would be pretty specific for Matrix and wouldn’t help nested Matrix/Super Table/Neo combos.

A better approach would be to change the error list key format to include specific block UUIDs rather than a numeric index (e.g. `articleBody.e72958fc-1db6-43f7-ae06-b42a559d48ea.heading` rather than `articleBody[12].heading`), and set fields’ expected error keys on a `data-error-key` attribute, so the JS doesn’t have to do any guesswork. (Added that to the v5 list: DEV-1177)

Issue #13574, opened 2023-08-18
[4.5]: `Craft.cp.notification` that contains buttons prevents shortcuts from working


i-just:
### Description
In the Control Panel, when `Craft.cp.notification` that contains a `button` (or `input`) is triggered, it prevents form shortcuts from working. 

Caused by: https://github.com/craftcms/cms/blob/develop/src/web/assets/cp/src/js/CP.js#L1536-L1541

### Steps to reproduce

1. On the `4.5-beta.2` branch, open the entry edit page and attempt to save it and trigger a validation error. 
2. “This {type} has been updated.” notice should show with a “Reload” button.
3. Try to use `Ctrl/Cmd+S` to save and continue editing.

### Expected behavior
Once the notice has been dismissed, the shortcut should trigger the usual action. In the case of `Ctrl/Cmd+S`, it should trigger save and continue editing.

### Actual behavior
Once the notice has been dismissed, the shortcut triggers the browser’s native Save As dialogue.
<img width="1727" alt="Screenshot 2023-08-18 at 12 35 30" src="https://github.com/craftcms/cms/assets/4500340/50775ca8-389a-4ac2-bbc8-cb7a2ecd9dd5">

### Craft CMS version

4.5-beta.2

### PHP version

8

### Operating system and version

Mac OS

### Database type and version

MySQL

### Image driver and version

n/a

### Installed plugins and versions

n/a

brandonkelly:
The notification is supposed to add a new UI layer when present, which would temporarily block any shortcuts registered to the previous UI layer. Without this PR, don’t the shortcuts resume once the notification is closed?

i-just:
No, shortcuts don’t resume once the notification is closed (and no matter how you close it). They only resume if you reload the page. A notification has to contain a `button` (or `input`) tag for this behaviour to occur.

https://github.com/craftcms/cms/assets/4500340/61e160f5-4ece-4d19-882a-48d1577309f7

arifje:
Same issue here; shortcuts aren't enabled after the notification is gone. Only on a page refresh.

brandonkelly:
@arifje Fixed in 4.5.1 – see https://github.com/craftcms/cms/issues/13591#issuecomment-1690574203.

PR #13575, opened 2023-08-18
re-initialise shortcuts after dismissing notification

i-just:
### Description
Re-initialise shortcuts after dismissing a notification. 

(If the notification contained `input` or `button` tags, the original shortcut registrations were lost.)

### Related issues
#13574


brandonkelly:
Notifications should have been removing their UI layer on close, to resume any keyboard notifications registered to the previous layer. Fixed that.

Issue #13570, opened 2023-08-18
[4.4.17]: An error occurs when trying to export subfolders of assets


rhlt:
### Description

Craft CMS provides an "Export" button on all element pages in the CP; including Assets. This allows exporting the field data of all elements in the page. If one or more elements are selected, only those elements are exported. However, this does not work when a subfolder is selected. (Subfolders only exist for Assets.)

### Steps to reproduce

1. Open the "Assets" page in the CP;
2. Select at least one subfolder by clicking the checkbox to the left of the folder name;
3. Click the "Export" button at the bottom right;
4. Click the red "Export" button in the pop up that appears (there is no difference in behavior for different Export Types or Limits);

### Expected behavior

A file is downloaded that contains all data for all selected elements (in this case, all assets in the selected subfolder).

### Actual behavior

A small pop up appears with the message **"An internal server error occured."**

Inspecting the network request sent by the CP reveals the following error:

```
An Error occurred while handling another error:
TypeError: array_keys(): Argument #1 ($array) must be of type array, string given in /var/www/html/vendor/craftcms/cms/src/web/CsvResponseFormatter.php:92
Stack trace:
#0 /var/www/html/vendor/craftcms/cms/src/web/CsvResponseFormatter.php(92): array_keys('Exception')
#1 /var/www/html/vendor/yiisoft/yii2/web/Response.php(1098): craft\web\CsvResponseFormatter->format(Object(craft\web\Response))
#2 /var/www/html/vendor/craftcms/cms/src/web/Response.php(286): yii\web\Response->prepare()
#3 /var/www/html/vendor/yiisoft/yii2/web/Response.php(339): craft\web\Response->prepare()
#4 /var/www/html/vendor/yiisoft/yii2/web/ErrorHandler.php(136): yii\web\Response->send()
#5 /var/www/html/vendor/craftcms/cms/src/web/ErrorHandler.php(192): yii\web\ErrorHandler->renderException(Object(yii\base\InvalidArgumentException))
#6 /var/www/html/vendor/yiisoft/yii2/base/ErrorHandler.php(152): craft\web\ErrorHandler->renderException(Object(yii\base\InvalidArgumentException))
#7 /var/www/html/vendor/craftcms/cms/src/web/ErrorHandler.php(66): yii\base\ErrorHandler->handleException(Object(yii\base\InvalidArgumentException))
#8 [internal function]: craft\web\ErrorHandler->handleException(Object(yii\base\InvalidArgumentException))
[...]

PR #13577, opened 2023-08-18
include assets from folders in asset export

i-just:
### Description
When exporting all assets, include assets from the subfolders too.

When exporting assets by selecting rows on the asset index page, if a folder is selected, include all the assets in the selected folder(s) and its subfolders.

### Related issues
#13570


brandonkelly:
Fixed this via 9610cc7d48bbbe815a099a7a16678a672aef4273 by omitting folders without IDs from `getSelectedElementIds()`. Anything beyond that is going to require a more major rearchitecture of element exporting (see https://github.com/craftcms/cms/issues/13570#issuecomment-1684880296).

PR #13589, opened 2023-08-22
Exclude phpsessions in backups

timkelty:


src/db/Connection.php
```diff
@@ -214,10 +214,11 @@ public function getIgnoredBackupTables(): array
     {
         return [
             Table::ASSETINDEXDATA,
+            Table::CACHE,
             Table::IMAGETRANSFORMINDEX,
             Table::RESOURCEPATHS,
+            Table::PHPSESSIONS,
```
timkelty, line 220: Exclude `phpsessions` table if using `DbSession`

src/db/Connection.php
```diff
@@ -214,10 +214,11 @@ public function getIgnoredBackupTables(): array
     {
         return [
             Table::ASSETINDEXDATA,
+            Table::CACHE,
```
timkelty, line 217: Replaces string reference with class const

PR #13608, opened 2023-08-25
Store updated user from register-user event

trevor-cuffe:
### Description
In actionEditUser, the RegisterUserActionsEvent supposedly gives plugins a chance to modify any of the event properties, including the $user object. However, the $user stored in the event is never referenced after the event is triggered. This update stores the $event->user object back to the $user variable

### Related discussions
[#13607](https://github.com/craftcms/cms/discussions/13607)

### DIFF
--- a/src/controllers/UsersController.php
+++ b/src/controllers/UsersController.php
@@ -1008,6 +1008,8 @@ class UsersController extends Controller
         ]);
         $this->trigger(self::EVENT_REGISTER_USER_ACTIONS, $event);
 
+        $user = $event->user;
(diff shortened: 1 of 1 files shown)


brandonkelly:
That event is just for building out the list of actions that the user should have. It’s not intended that event handlers could actually change the user being edited in the process though. 

What’s the use case for this?

trevor-cuffe:
Got it. I was interested in pre-setting certain field values when creating new users, and this seemed like the best place to be able to it.

PR #13609, opened 2023-08-25
\craft\console\Controller::do improvements

timkelty:
- Fix missing closing paren


timkelty:
@brandonkelly this is now just the typo fix.

The exception behavior you can mostly control on your own, and I wanted to do some further improvement (support depth, add error/success messages), so its feeling like those would be more appropriate in their own class.

Issue #13632, opened 2023-08-31
[4.x]: Unable to use Asset element select field in Vue (inline `<style>` elements)


engram-design:
### What happened?

This is similar to https://github.com/craftcms/cms/issues/13597 but Asset icons have the same issue where inline `<style>` tags are being used, which means they throw a fatal error [in Vue](https://github.com/verbb/vizy/issues/250).

In Vizy, Hyper or Icon Picker which uses Vue to render Craft-rendered fields, you'll get the following fatal error. For example, we pick a PDF asset in an Assets field.

<img width="286" alt="image" src="https://github.com/craftcms/cms/assets/1221575/9c1f62c8-518d-4e6c-9d43-7c8f00b08b78">

<img width="946" alt="image" src="https://github.com/craftcms/cms/assets/1221575/6db2b567-d851-4cd7-bb48-747640d9e051">

This is because including this is invalid.

```html
<style type="text/css">
.bkcnixcaoe-st0{fill:hsl(212, 50%, 93%);}
.bkcnixcaoe-st1{fill:#FFFFFF;}
.bkcnixcaoe-st2{fill:hsl(210, 10%, 53%);}
</style>
```

I imagine there might be other SVGs that this is happening for? Passing these through SVGO is an easy (and automated) way to inline styles for SVGs.

### Craft CMS version

4.5.3

### Installed plugins and versions

-

engram-design:
Also, here's a PR! https://github.com/craftcms/cms/pull/13633

brandonkelly:
Craft 4.5.4 is out with that fix. Thanks again!

PR #13633, opened 2023-08-31
Change icons to remove `<style>` elements and use inline `style` attributes to get around Vue inline style warnings

engram-design:
Fixes https://github.com/craftcms/cms/issues/13632


brandonkelly:
Thanks!

PR #13648, opened 2023-09-06
Check if key 'value' isset

MarcelWeidum:
### Description
When I save an entry with the Money field added later I get the following error:
`Undefined array key \"value\" at /pathtoproject/vendor/craftcms/cms/src/fields/Money.php:182`

This is solved by checking is the key exists.


brandonkelly:
Thanks for submitting that! Merged in for the next release.

brandonkelly:
Craft 4.5.4 is out with that fix.

Issue #13626, opened 2023-08-31
[4.x]: Photos are not restored after the revision is restored


iZerus:
### Description

I created a record, added photos to it. Then I deleted the photos. But when I decided to restore the previous version (where there were photos), the photos never recovered. 
P.S. When deleting photos, photos remain in assets

### Steps to reproduce

1. Create a section with the Assets field (with versioning control)
2. Create an entry in this section and attach photos to it
3. Save it
4. Delete photos from entry and save
5. Try to restore the version with photos

### Expected behavior

Photos must be present in the entry

### Actual behavior

Photos are missing

### Craft CMS version

 4.5.3

### PHP version

8.1

### Installed plugins and versions

- Redactor 3.0.4
- Site Switcher 2.3.0

i-just:
Hi, thanks for reporting! I raised a PR to adjust this.

brandonkelly:
Craft 4.5.4 is out with that fix. Thanks again!

olivierbon:
There might be a part two to this. It also happens with Matrix Blocks.


https://github.com/craftcms/cms/assets/146041/079987fb-49eb-4def-8244-0a26b2b30f32

brandonkelly:
Thanks @olivierbon! Fixed for the next release via #13740.

brandonkelly:
Craft 4.5.6 is out now with that fix.

PR #13652, opened 2023-09-06
don't preset dirty attrs and fields to empty array

i-just:
### Description
When updating the canonical element from a given derivative, don’t set the `dirtyAttributes` and `dirtyFields` to an empty array. 

Setting it triggers `setDirtyFields()`, which sets `_allDirty` to `false`, which causes issues when reverting from a revision.

### Related issues
#13626

### DIFF
--- a/src/services/Elements.php
+++ b/src/services/Elements.php
@@ -1238,8 +1238,6 @@ class Elements extends Component
             'revisionId' => null,
             'isProvisionalDraft' => false,
             'updatingFromDerivative' => true,
-            'dirtyAttributes' => [],
-            'dirtyFields' => [],
         ];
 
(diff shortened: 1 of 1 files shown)


brandonkelly:
It’s important that those values get set, even if to empty arrays, so `duplicateElement()` can retain the correct dirty fields here:

https://github.com/craftcms/cms/blob/71151312946cbd810799c0843bea7320f43de601/src/services/Elements.php#L1558-L1565

Fixed the bug in 71151312946cbd810799c0843bea7320f43de601 by explicitly setting all custom fields to dirty, just in this one case where the revision doesn’t really know all of the fields that have changed since then, so it’s best to play it safe.

PR #13657, opened 2023-09-07
Fix FS URL validation

timkelty:
Makes URL actually required when `hasUrls` is true.


src/base/Fs.php
```diff
@@ -54,6 +54,7 @@ public function attributeLabels(): array
         return [
             'handle' => Craft::t('app', 'Handle'),
             'name' => Craft::t('app', 'Name'),
+            'url' => Craft::t('app', 'URL'),
```
timkelty, line 57: Give it a label so errors refer to it as "URL"

Issue #13627, opened 2023-08-31
[4.5.3]: typeId appearing under Entry Type label


RyanRoberts:
### What happened?

When `Show field handles in edit forms` is enabled a copy text button with `typeId` appears below the Entry Type label. Maybe it's intentional, but doesn't feel intentional.

![Google Chrome - 2023-08-30--11-47-20@2x](https://github.com/craftcms/cms/assets/108874/62deb27e-b955-463a-a887-d795f264279b)

### Steps to reproduce

1. Enable Show field handles in edit forms in user profile
2. Make sure you have more than one entry type so the Entry Type label appears in the sidebar

### Expected behavior

Either not to appear at all, or not to break label layout.

### Craft CMS version

4.5.3

### PHP version

8.2

i-just:
Hi, thanks for reporting! I raised a PR to adjust this.

brandonkelly:
Craft 4.5.4 is out with that fix. Thanks again!

PR #13659, opened 2023-09-07
hide field handle for entryType

i-just:
### Description
PR https://github.com/craftcms/cms/pull/13138 added missing validation messages to the meta fields. In the process, it was necessary to add the `attribute` param to the Entry Type select field to trigger the containing `div` to have the `data-attribute="typeId"` attribute, which is needed for the validation errors to show in a slideout.

Adding `attribute`, however, makes the field handle show for the Entry Type field when the user has “Show field handles in edit forms” enabled, and this wasn’t the case before (nor should it be).

This PR removes the `attribute` from the Entry Type select field and replaces it with `fieldAttributes[data][attribute]`, which will give the field container the correct attribute for the error messages to show in a slideout, without making the field handle show.

### Related issues
#13627

### DIFF
--- a/src/elements/Entry.php
+++ b/src/elements/Entry.php
@@ -1655,8 +1655,14 @@ EOD;
                     'value' => $this->getTypeId(),
                     'options' => $entryTypeOptions,
                     'disabled' => $static,
-                    'attribute' => 'typeId',
                     'errors' => $this->getErrors('typeId'),
+                    // we can't simply use 'attribute' as that will make the field handle show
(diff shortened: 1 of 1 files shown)


brandonkelly:
Fixed this in a different way via 565eaa28b18e869b3ec3c37ec7da5afd83aeaae5 – I’d rather have `CustomField` have to do a little extra work explicitly stating that its “attribute” should be shown on edit forms, rather than make every other field layout field have to identify its attribute in a more complicated way.

PR #13666, opened 2023-09-07
Allow env config for attributes

timkelty:
### Description
Detects if you pass `\craft\helpers\App::envConfig` a Model, and if so, uses the the model's attributes instead of just getting public properties.

This allows config overrides to work if you have an attribute with a getter/setter.

Another option could be to just allow you to pass an array of properties to override, so when you called `\craft\helpers\App::envConfig`, you could just pass if your model's attributes.

One potential issue here might be if you have a read-only attribute (no setter), and you try to set the corresponding env for it, it will throw.


brandonkelly:
> This allows config overrides to work if you have an attribute with a getter/setter.

`Typecast` would need to be updated to support that as well.

brandonkelly:
Closing

PR #13683, opened 2023-09-11
Matching parenthesis in craft\console\Controller::do

AugustMiller:
Noticed this was missing when doing some debugging output!


AugustMiller:
Nevermind! @timkelty found this in https://github.com/craftcms/cms/pull/13609

PR #13690, opened 2023-09-13
Keep branches up to date with GHA

timkelty:
Has proved useful elsewhere.

### DIFF
--- a/.github/workflows/update-branches.yml
+++ b/.github/workflows/update-branches.yml
@@ -1,0 +1,40 @@
+name: Update branches
+
+on:
+  push:
+    branches: develop
+
+permissions:
+  contents: write
+
+env:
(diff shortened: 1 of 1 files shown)


.github/workflows/update-branches.yml
```diff
@@ -0,0 +1,40 @@
...
+env:
+  base_ref: develop
+
+jobs:
+  update-branches:
+    strategy:
+      matrix:
+        ref:
+          - '4.5'
+          - '5.0'
```
timkelty, line 19: Anything else we want to keep in sync with `develop`?

brandonkelly:
99% of merges include resolving conflicts, either in the changelog or `config/app.php` or built resource files, so I don’t see this being helpful.

Issue #13710, opened 2023-09-18
[4.x]: Some checkboxes are pre-checked after updating from 4.4.17 to 4.5.5


Mixxey:
### Description

After updating from Craft 4.4.17 to 4.5.5 several checkboxes used to control styling options and sometimes other functions are checked across several entries which were not checked before the update and are not meant to be checked in their specific contexts.

This happens independent of the entry and field types they are attached to. I found the checked checkboxes in constructions as complex as "a matrix field inside of a neo block inside of a neo block" but also in checkboxes on entry-level so it doesn't seem to have something to do with neo blocks or the overall complexity.

I also noticed that all affected checkbox-type fields didn't have the "Die Feldwerte als Suchschlüsselwörter verwenden"-checkbox checked (the one just above the dropdown where you set the field type), but checking those before the update didn't prevent the error.

Sadly, I wasn't able to find out more on my own, but maybe you can help? Many thanks in advance for looking into this mystery :) 

### Steps to reproduce

1. Update from Craft 4.4.17 to 4.5.5

### Expected behavior
No additional checkboxes should be checked that aren't meant to be checked.

### Actual behavior
Checkboxes are checked without any too obvious pattern: Sometimes all occurrences within an entry are affected, sometimes only specific ones; some entries don't seem to be affected at all. 

### Craft CMS version

Craft Pro 4.5.5

### PHP version

8.0.30

### Operating system and version

Linux 6.2.0-32-generic

### Database type and version

MySQL 5.7.43

### Image driver and version

Imagick 3.7.0 (ImageMagick 6.9.10-23)

### Installed plugins and versions

- Asset Usage 3.2.0
- Autologin 3.0.0
- Batch Actions 1.3.0
- Blitz 4.5.4
- CKEditor 3.4.0
- Cloudflare 2.0.1
- Contact Form 3.0.1
- Control Panel CSS 2.6.0
- CP Clear Cache 1.3.0
- CP Field Inspect 1.4.4
- Craft JWT Auth dev-develop-craft-4
- Dumper 3.0.1
- Elements Panel 2.0.0
- Embedded Assets 3.1.8
- Field Manager 3.0.7
- Google Cloud Storage 2.0.0
- ImageOptimize 4.0.5
- Instagram Feed 2.1.0
- Logs 4.0.0
- Minify 4.0.0-beta.2
- Navigation 2.0.21
- Neo 3.8.6
- Redactor 3.0.4 (no longer in use but not yet uninstalled)
[...]

PR #13727, opened 2023-09-21
limit the use of new blank option & null handling

i-just:
### Description
Only use the new blank option and `null` value handling for dropdown fields.

Using the blank options and `null` value handling for all fields that extend `BaseOptionsField` causes the following issue with the `Checkboxes` field. If you have a checkbox field with only one blank option, you add it to an existing entry and then edit that entry - the checkbox shows as selected even though it shouldn’t. What’s more, if you were to save that entry without making any changes, the checkbox would continue to show as checked, and its value would continue to be `null` in the database.

### Related issues
#13710


src/fields/BaseOptionsField.php
```diff
@@ -390,6 +378,20 @@ public function normalizeValue(mixed $value, ?ElementInterface $element = null):
...
+     * @param array $option
+     * @param mixed $value
+     * @param array $selectedValues
+     * @param bool $selectedBlankOption
+     * @return array
+     */
+    protected function isOptionSelected(array $option, mixed $value, array $selectedValues, bool $selectedBlankOption): array
+    {
+        return [in_array($option['value'], $selectedValues, true), $selectedValues, $selectedBlankOption];
+    }
```
brandonkelly, line 393: Given the method name, I think it would make more sense to have it return `bool`, and have `$selectedValues` and `$selectedBlankOption` passed by reference, in case they need to be modified.

```suggestion
    protected function isOptionSelected(array $option, mixed $value, array &$selectedValues, bool &$selectedBlankOption): array
    {
        return in_array($option['value'], $selectedValues, true);
    }
```

i-just: Makes sense! Adjusted.

PR #13730, opened 2023-09-21
Norwegian bokmål - fixes typo, changes some translations to be consistent with others, adds two missing translations

kbergha:
### Description

- Fixed typo on line 1950, `slal` makes no sense, and is supposed to be `skal` (should).
- Changed `kan` (can) and `må` (must) to `skal` (should) to be consistent in the related translations on line 1948 + 1949.
- Added the translation for two newly added strings that were missing.
- Changed `kataloger` (catalogues) to `mapper` (folders / directories) in an existing translation to be consistent with the two new ones.

### Related issues

None.


brandonkelly:
Thank you! We use CrowdIn to manage translations, so I just applied your changes over there (https://translate.craftcms.com). Pulled in for the next release!

Issue #13708, opened 2023-09-18
[4.5.5]: Links in error summaries don't work if the error is in a collapsed Matrix Block


MoritzLost:
### Description

The new error summary (great addition BTW!) at the top of element edit pages includes anchor links that allow the editor to jump directly to the field that contains the error. However, if the field that contains the error is inside a collapsed Matrix Block, the anchor links don't work.

### Steps to reproduce

1. In an entry with a matrix field, add a matrix block and leave a required field empty.
2. Collapse the matrix block.
3. Save the entry.
4. Click the link to the field in the error summary.

### Expected behavior

The link should always jump to the field, even if it's inside a collapsed matrix block. Ideally, it should open the matrix block if it's collapsed before jumping to the field.

In addition, we're injecting a small piece of JS so that all Matrix Blocks start out collapsed to improve usability if there are lots of Matrix Blocks:

```php
$view->registerJs(<<<'JS'
    $('.matrixblock').each(function() {
        $(this).addClass('collapsed');
        Craft.MatrixInput.rememberCollapsedBlockId($(this).data('id'));
    });
JS);
```

Ideally, any solution to the above problem should even work if matrix blocks are collapsed on page load using this snippet.

### Actual behavior

If the matrix block is collapsed, the anchor links don't work, and there's no indication what went wrong.

### Craft CMS version

4.4

### Installed plugins and versions

-

i-just:
Hi, thanks for reporting! I raised a PR to adjust this.

brandonkelly:
Fixed for the next release via #13741.

brandonkelly:
Craft 4.5.6 is out now with that fix.

PR #13741, opened 2023-09-25
Bugfix/13708 error summary links collapsed blocks

i-just:
### Description
Improved anchoring from the error summary item to the field.
If the field we’re anchoring to is in a collapsed block, expand that block.

### Related issues
#13708


brandonkelly:
This seems to only halfway fix the bug – the block will expand when the link is clicked on, but the browser won’t actually scroll down to it, even for subsequent clicks.

i-just:
I was finally able to reliably replicate it not scrolling down and pushed an update.

Issue #13760, opened 2023-09-29
[4.x]: Money field breaks Save and Continue on users


CharlieGentle:
### What happened?

When there is a Money field visible in a user profile, it breaks the Save and Continue functionality. Clicking on the arrow next to save just saves instantly and takes you back to the index. Cmd + S also saves but takes you back to the index.

### Steps to reproduce

1. Create a money field
2. Add the money field to a user profile.
3. Go to the user and try to Save and Continue.

### Expected behavior
- Clicking on the save dropdown should show the dropdown
- Cmd + S should Save and Continue

### Actual behavior
- Clicking on the arrow next to save just saves instantly and takes you back to the index.
- Cmd + S also saves but takes you back to the index.

### Craft CMS version

4.5.5

### PHP version

8.0.28

### Operating system and version

Linux 5.10.104-linuxkit

### Database type and version

MariaDB 10.4.28

CharlieGentle:
The CMD + S issue is happening on one site which has more complex fields with Money fields inside a Super Table. On a test site with just a money field, CMD + S is working ok but the save dropdown is still broken.

i-just:
Hi, thanks for reporting and for all the details! I can see what's happening, and I'm working on a solution.

brandonkelly:
Craft 4.5.7 is out now with a fix for this. Thanks again for reporting!

PR #13800, opened 2023-10-09
inputmask yiiAsset data-form clash

i-just:
### Description
The Money field includes `MoneyAsset`, which depends on `yii\widgets\MaskedInputAsset`, which depends on `yii\web\YiiAsset`, which includes `@yii/assets` => `yii.js` where [the following line](https://github.com/yiisoft/yii2/blob/master/framework/assets/yii.js#L265) clashes with what we use on the User edit page for the additional form actions.

Renaming our attribute from `data-form` to `data-formsubmit` prevents this clash.

### Related issues
#13760


brandonkelly:
That change would probably break other plugins which are also taking advantage of `data-form`. Fixed by adding a new `InputmaskAsset` that just pulls in the `inputmask` library without the `YiiAsset` baggage.

Issue #13771, opened 2023-10-03
[5.0.0-alpha]: Cannot save matrix entries that have URL enabled


mattbloomfield:
### Description

When trying to save an entry that has a matrix field and the field has URLs enabled, you get an error saying `slug cannot be blank`

### Steps to reproduce

1. Create a matrix field
2. Add a path for the URL
3. Add the matrix field to an entry type
4. Add the entry type to a section
5. Create a new entry in that section
6. Add a block to the matrix
7. Save

### Expected behavior

It would save. I am curious how I would know the slugs for the matrix blocks after save. 

### Actual behavior

It gives the error: 

<img width="810" alt="image" src="https://github.com/craftcms/cms/assets/37582550/61b22b58-7c65-492b-bd82-a3715234135f">

### Craft CMS version

Craft Solo 5.0.0-alpha

### PHP version

8.2.10

### Operating system and version

MacOS Sonoma 14.0

### Database type and version

ddev standard

### Image driver and version

ddev standard

### Installed plugins and versions

- NONE

i-just:
Hi, thanks for reporting! I raised a PR for this.

brandonkelly:
This is fixed now! Thanks again.

PR #13824, opened 2023-10-13
preset slug for matrix entries with URI format

i-just:
### Description
Preset slug for matrix entries when a Matrix field has a URI format.

### Related issues
#13771


brandonkelly:
Just tweaked the default `slug` required validation rule, to only apply when the element’s URI format actually contains `slug`. (Previously it was required for any element that had a URI format at all, regardless of what it contained.)

PR #13825, opened 2023-10-13
[5.x] Fix an error when querying tags

engram-design:
There's no content table any more, so this orderBy param will fail.


brandonkelly:
Thanks! The column got moved to the `elements_sites` table, so just updated the property to point to that instead. (Though `TagQuery` is a dead man walking at this point anyway…)

Issue #13854, opened 2023-10-19
[4.x]: Entries with `showStatusField: false` can be disabled in bulk operations


realjoshharrison:
### Description

I have a Single with `showStatusField` set to false, which operates as my site globals. It cannot be disabled from the Entries screen when only this entry is selected, as the button is disabled as expected:

![Screenshot 2023-10-19 at 14 37 46](https://github.com/craftcms/cms/assets/3930056/6b73a80f-5c96-4de7-b86f-7685b983ca08)

However, if I select another entry which has an enabled status field, the button becomes available and I can disable my site globals as part of the bulk operation:

![Screenshot 2023-10-19 at 14 37 53](https://github.com/craftcms/cms/assets/3930056/16fe3bc5-7a2c-45eb-95a6-123fc0a9cb13)

![Screenshot 2023-10-19 at 14 39 05](https://github.com/craftcms/cms/assets/3930056/7af206b4-66f9-4d12-ab9a-ca72a1122dea)

### Steps to reproduce

1. Add two sections: one with it's entry type status field hidden, and another with the status field visible.
2. Create two entries from these and select them both in the Entries admin screen, then disable them both.

### Expected behavior

The entry with `showStatusField: false` is unaffected by the bulk operation. Any others are processed as normal.

### Actual behavior

All selected entries are disabled indiscriminately. 

### Craft CMS version

4.5.6.1

### PHP version

8.2.0

### Operating system and version

MAMP Pro 6.8.1, Darwin 22.6.0, MacOS 13.5.2

### Database type and version

MySQL 5.7.39

### Image driver and version

Imagick 3.7.0 (ImageMagick 6.9.6-2)

### Installed plugins and versions

- CKEditor 3.6.0
- CP Field Inspect 1.4.4

i-just:
Hi, thanks for reporting! I raised a PR for this.

brandonkelly:
Craft 4.5.9 is out with a fix for this. Thanks again!

datpaitoon17:
> ### What happened?
> 
> ### Description
> 
> I have a Single with `showStatusField` set to false, which operates as my site globals. It cannot be disabled from the Entries screen when only this entry is selected, as the button is disabled as expected:
> 
> ![Screenshot 2023-10-19 at 14 37 46](https://github.com/craftcms/cms/assets/3930056/6b73a80f-5c96-4de7-b86f-7685b983ca08)
> 
[...]

PR #13856, opened 2023-10-20
Bugfix/13854 set status bulk action

i-just:
### Description
Don’t change the element’s status via bulk action if its type’s `showStatusField` is `false`.

### Related issues
#13854

### DIFF
--- a/src/base/Element.php
+++ b/src/base/Element.php
@@ -4944,7 +4944,7 @@ JS,
      * @return bool
      * @since 4.5.0
      */
-    protected function showStatusField(): bool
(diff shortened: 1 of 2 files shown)


brandonkelly:
This would be a breaking change for any element types that are overriding `showStatusField()` and keeping its visibility set to `protected`. So we should make this a 5.0 change instead.

brandonkelly:
Just remembered this is already supposed to be handled via the SetStatus action trigger, but there was a JS bug where only the first selected element was being respected.

Issue #13850, opened 2023-10-18
[4.x] Disabled/Static Dropdown fields are back, but illegible


amphibian:
### Description

Dropdown fields on completed orders are [no longer hidden](https://github.com/craftcms/cms/issues/13753), but their disabled state makes them almost invisible, and thus not very useful for admins looking to see which options might have been selected via these fields.

<img width="397" alt="Screenshot 2023-10-18 at 4 20 06 PM" src="https://github.com/craftcms/commerce/assets/53333/caa6d401-c13a-40c3-ae1a-9819f4c39bcd">

### Craft CMS version

4.5.7

### Craft Commerce version

4.3.1

i-just:
Hi, thanks for reporting! I raised a PR for this.

brandonkelly:
This is resolved in Craft 4.5.8, via #13857 :tada:

PR #13857, opened 2023-10-20
disabled fields look

i-just:
### Description
Ensures all field types that ship with Craft are visible and consistent when rendered statically (and therefore are not editable).

For example, before this PR, some fields were rendered with the `.disabled` class, which made them partially transparent (to a various degree), and some were fully visible (like the editable fields are).

Removes the “clear” button functionality from the disabled date & time fields.

Ensures disabled fields have `cursor: not-allowed` to differentiate them from the enabled/editable fields.

Before:
one of each - before

After:
one of each - after

(Related PR for CKEditor: https://github.com/craftcms/ckeditor/pull/142)

### Related issues
#13850


brandonkelly:
Wayyyy nicer. Thanks @i-just!

PR #13879, opened 2023-10-26
Fixes potential FOUT with icon font

kennethormandy:
### Description

CleanShot 2023-10-24 at 22 13 04@2x

This PR prevents the flash of unstyled text you can potentially see with the icon font, by:

- Adding a `.woff2`, with is ~40% smaller filesize than the current `.woff` file
- Adding `font-display: block`, which will stop the browser from using a fallback font when the icon font hasn’t loaded yet—we never want a fallback font in this scenario


kennethormandy:
Here’s what I did to generate the WOFF2 file.

Clone [the google/woff2 repo](https://github.com/google/woff2#build--run), and use the standard build instructions:

```sh
git clone --recursive https://github.com/google/woff2.git
cd woff2
make clean all
cd woff2
```

…and then I used the existing `.ttf` file as the source file:

```sh
./woff2_compress ./path/to/Craft.ttf

Processing path/to/Craft.ttf => path/to/Craft.woff2
Compressed 20813 to 11063.
```

brandonkelly:
Looks like the [ttf2woff2](https://www.npmjs.com/package/ttf2woff2) npm package could be used to make this part of the `cp` asset’s build script. So we could just continue dumping the IcoMoon-generated fonts in `src/web/assets/cp/src/fonts/` without having to worry about it.

brianjhanson:
@brandonkelly added a custom `Ttf2Woff2Plugin` which will run `ttf2woff` on build. Any `.ttf` files in the `src/fonts` directory will be converted to `woff2`

timkelty:
Wondering if it would be feasible to remove _any_ loading time by inlining these as svg instead.
@i-just did something similar for the default asset icons recently…

brandonkelly:
I’m working toward that for v5. Lot of work though. Maybe we can have a goal to complete it in v6.

PR #13880, opened 2023-10-27
`db/drop-table-prefix` command

angrybrad:
This can be us to change the table prefix for an existing Craft install.

Tested on MySQL and Postgres. Tested going from none to some, some to none, and some to some.


brandonkelly:
Changed it to `drop-table-prefix` since that’s the only direction we actually want to encourage. And also because having the logic rely on the `tablePrefix` setting feels sketchy (maybe they cleared that out first). Now it just detects the current prefix by looking at the existing table names, and there’s a `$prefix` argument that can be explicitly passed if it’s ambiguous.

PR #13883, opened 2023-10-27
(fix) Prevent crash when the Fs has no settings

nitriques:
This fix is required for Fs that do not have any settings. Without it, the filesystem page crashes with an undefined "settings" key error message.

Please tell me if I did something wrong. If not, this fix is required.


brandonkelly:
Thank you!

PR #13898, opened 2023-11-02
added `onCanceledInlineEditing()` method

i-just:
### Description
When you’re viewing an embedded element index in a table view, and you trigger inline editing, the `New {type}` button gets disabled. If you then click `Cancel`, the view gets updated to the regular table view, but the `New {type}` button stays disabled.

This PR adds an `onCanceledInlineEditing()` method to the `BaseElementIndex` that gets triggered in `TableElementIndexView` when the `Cancel` button is activated and for `EmbddedElementIndex` ensures the `New {type}` button gets re-enabled.

### Related issues
[CMS-1210](https://linear.app/craftcms/issue/CMS-1210/nested-elements-new-element-button-stays-disabled-after-inline-editing)


brandonkelly:
Fixed via beaa9145bf70882e6d4d5b90468fef22d74aaabf.

It was not intended that new entries couldn’t be added when inline editing them in the first place.

i-just:
> It was not intended that new entries couldn’t be added when inline editing them in the first place.

Ah, I assumed it was intended, hence my solution. Thanks for clarifying!

PR #12946, opened 2023-03-21
Convert context menu on element edit page to disclosure menu

gcamacho079:
### Description
@brandonkelly marking as a draft because I still need to make changes for this component in the JavaScript.

This is similar to the change we made on element index pages to break the **Sort Attribute** and **Sort Direction** into separate form controls, this PR separates Site and Revision/Draft into two separate form controls inside of a disclosure menu.

Screenshot 2023-03-21 at 2 36 02 PM

PR #13902, opened 2023-11-02
Fancy crumbs and element action menus

brandonkelly:
High level: this redesigns control panel breadcrumbs, and unifies element action menus across edit pages, slideouts, and selectable element chips & cards.

## Overview

- Breadcrumbs have been redesigned to accommodate complex crumb UIs, including menus and chips.

The breadcrumbs for an Edit Entry page. The “Locations” section crumb now includes a menu, and the current entry is marked by a chip that includes the entry’s status indicator. The crumbs are followed by a context menu with “Current” selected, and an action menu trigger.

- Entry and asset edit pages’ section and volume crumbs now include menus, making it possible to quickly navigate to another section or volume’s index source.

The breadcrumbs for an Entry Edit page, with its section crumb’s menu expanded, revealing options for each of the configured sections, and showing that “Locations” is selected.

- Multi-site installs now include a site selector as the initial breadcrumb, on pages that need it.

The breadcrumbs for an Edit Entry page on a multi-site install, showing that the first breadcrumb is for the selected site. Its menu is expanded to reveal “English”, “French”, and “German” options, with “English” selected.

- Page context menus have been moved into the global header beside the breadcrumbs.

The global header for an Entry Edit page, with its context menu expanded to reveal options for the current revision, active drafts, and recent revisions.

- Elements can now define action menus, which will be displayed beside the breadcrumbs and context menu. These should include any actions that are unrelated to saving the content on the page.

The global header for an Entry Edit page, with its action menu expanded to reveal “View in a new tab” and “Delete entry” actions.

- Element action menus also appear within element slideout headers, and selectable element chips & cards (with any destructive items omitted).

A selected asset within an Assets field, with its action menu expanded to reveal “Preview file”, “Download”, “View in a new tab”, “Edit asset”, “Replace file”, “Open in Image Editor”, and “Remove” actions.

## Technical Details

- Control panel screens can now include a site breadcrumb via `CpScreenResponseBehavior::site()`, and can make it a menu with additional selectable sites via `CpScreenResponseBehavior::selectableSites()`.
- Control panel screens can now specify an action menu via `CpScreenResponseBehavior::actionMenuItems()`.
- Control panel screens can now specify context menu items via `CpScreenResponseBehavior::contextMenuItems()`. `contextMenuHtml()` and `contextMenuTemplate()` have been removed in favor of that.
- Elements can define their action menu items via `safeActionMenuItems()` and `destructiveActionMenuItems()`.
- Element action menu items can be customized via `Element::EVENT_DEFINE_ACTION_MENU_ITEMS`.
- Elements can now define breadcrumbs via `crumbs()` (rather than `prepareEditScreen()`).
- Menus can be added to any breadcrumb via a `menu` key, set to an array of menu options.
- Disclosure menus can now easily be created throughout the control panel via `craft\helpers\Cp::disclosureMenu()`.

## Related

- This iterates on the changes in #13847, so user actions are now included in users’ action menus, rather than getting their own unique menu beside the “Status” meta value.
- Should resolve the need for #12946.
- Resolves ACC-7


Mosnar:
Some minor unsolicited feedback 😅 

> Entry and asset edit pages’ section and volume crumbs now include menus, making it possible to quickly navigate to another section or volume’s index source.

This UI seems like it may not scale well with more than ~8-10 sections - I think this will be exacerbated by entryfication and it won't be unusual to have 20+ sections. I wish I had a suggestion, but I can't think of a good alternative off the top of my head.

michaelrog:
🙌🏼  YES! I'm very excited for these little UI candies, unifying element action controls, and making them available in more places. Nice work yall.

I agree with @Mosnar: The Sections dropdown could quickly get unwieldy. I work with several sites with more than 24+ (if I'm including Categories, Tags, etc).

Also, maybe nice to (optionally?) hide the breadcrumb dropdown controls behind a hover?

brandonkelly:
> This UI seems like it may not scale well with more than ~8-10 sections

Custom selects (like those used within condition rules to determine the rule Type) have the ability to auto-focus on an option as you type its label, similar to native `<select>`s. I was just thinking today we should probably bring that feature to disclosure menus as well, which would help here.

The menu will become scrollable if it’s too tall to display in the viewport, so that combined with the type-to-focus shortcut, I imagine it will be pretty usable.

gcamacho079 (changes requested):
## Fancy crumbs
### Critical
- [x] For entry type and multi-site items, the menu button is nested inside the anchor; interactive elements should not be nested. This also causes some confusion on mobile, as the entire row (link and chevron) are inside the focus outline, so a user might think they are opening another disclosure and accidentally trigger the link to the new page.
- [x] Hamburger menu button focus outline is clipped and only visible on one side
- [x] When tabbing through the breadcrumbs, the focus outline on the current entry is invisible
- [x] The site selector control should have a [consistent name](https://www.w3.org/WAI/WCAG22/Understanding/consistent-identification) across all screens. I recommend using an icon button so that the accessible name of the button is consistent between entries and sites. For example: `<li><a>{{ site name }}</a> <button aria-label=”Change site”>{{ world icon }}</button></li>`
- [x] The name of the version switcher should be consistent across screens, for example “Change version”, instead of referring to the name of the version itself. (See recommendation for site switcher.)
### Serious
- [x] The `.menubtn` for switching between entry types or sites must be 24px large, or [leave sufficient space](https://www.w3.org/WAI/WCAG22/Understanding/target-size-minimum.html). (Note: this may be resolved when the button is no longer nested in the anchor.)
- [x] `#action-btn` and `#context-btn` use the disclosure button pattern, but rely on semantic `label` for an accessible description. Semantic label elements should not be with the disclosure pattern. For example: for the Action menu specifically, the accessible text should come from an `aria-label` or `visually-hidden` text inside the button itself.
- [x] On hover, the mobile hamburger menu button drops below 3:1 contrast
- [x] On hover, the context button text drops below 4.5:1 contrast ratio against the button background.
- [ ] When context menu is open, heading levels skip from `h2` to `h6`. Recommendation: Add a visually-hidden heading to the context menu (i.e. “Entry Context”), and make all the other headings inside the disclosure `h3`
- [ ] ~~There are some instances where disclosure menus overflow and cause a horizontal scroll. Unsure if this is related to this PR, or should be tracked as a separate issue.~~ (unable to recreate)

### Recommendations
- [ ] For instances where a checkbox is used to identify selected language/version, the “check” pseudo-element should be given the `role="img"` and accessible name of "Selected". Ideally, this should follow the name of the link.
- [ ] The first list in the Context menu has list markup, but I think it only ever has one item it it (i.e. Current). If this is true, the item probably shouldn’t be wrapped in a list.

## Element action menu
### Critical
- [ ] Focus outlines aren’t visible for element links

### Serious
- [ ] On hover, the icon in the disclosure trigger drops below 3:1 ratio against the background

theskyfloor:
This is beautiful, wanted this for years!

Issue #13910, opened 2023-08-25
[4.x]: Paid status falls on to 2 lines on smaller screens


jonleverrier:
### Description

My order view summary has around 11 columns worth of data. On smaller screens, the Paid Status column data falls onto 2 lines. Is there anyway to force this to always be on 1 line? On larger screens it displays fine.

<img width="129" alt="Screenshot 2023-08-25 at 09 37 43" src="https://github.com/craftcms/commerce/assets/2373940/7cd0f041-4393-457d-bcbc-030a824df529">

### Steps to reproduce

1.

### Craft CMS version

4.5.2

### Craft Commerce version

4.2.11

### Installed plugins and versions

-

i-just:
@jonleverrier thanks for reporting! PRs raised to adjust this.

brandonkelly:
@nfourtythree You can fix this by adding `class="nowrap"` to your status column `<td>`s.

PR #13915, opened 2023-11-08
ensure status wording and circle are in one line

i-just:
### Description
Ensure the status wording and circle (shape) are shown on one line in the element index table.

Compliments the commerce PR: https://github.com/craftcms/commerce/pull/3329.

Before:
Screenshot 2023-11-08 at 12 28 37

After:
Screenshot 2023-11-08 at 12 28 09

### Related issues
#13910


brandonkelly:
This would only work if the “Status” label wasn’t translated, and feels a little too targeted anyway. The `<td>` could gain `class="nowrap"`, or its contents could be wrapped by `<span class="nowrap">`, and get the same effect.

PR #13938, opened 2023-11-16
Enable browser autocomplete on user's own address fields

gcamacho079:
### Description
Enable autocomplete for most address fields for user’s own address (_not_ other user addresses), the exception being selectize fields. (@brandonkelly would this be a good opportunity to un-selectize locale/address fields entirely, or save that for a later release?)

This PR — specifically the `getInputPurpose` method — will also help achieve WCAG 2.1 AA compliance in Console, as we can use it to set input purposes on the billing address fields.

### Related issues
Resolves ACC-42


brandonkelly:
> This PR — specifically the `getInputPurpose` method — will also help achieve WCAG 2.1 AA compliance in Console, as we can use it to set input purposes on the billing address fields.

Feels a little sketchy to be relying on that method on front ends. Couldn’t the Console UI be setting its own `autocomplete` values?

gcamacho079:
> Feels a little sketchy to be relying on that method on front ends. Couldn’t the Console UI be setting its own `autocomplete` values?

I’ll need to refresh myself on why I went this route and see if that alternative makes more sense. Will get back to you on this shortly!

gcamacho079:
@brandonkelly moved that method to `Cp` and made the other methods private and static. What are your thoughts about un-selectizing the other fields (country, state, etc.) in the address forms?

brandonkelly:
I think we should keep Selectize there until we have a better alternative (like the disclosure menu with a search input + inline list, like we talked about). Until then, it does make working with very large selects a lot more manageable.

gcamacho079:
> I think we should keep Selectize there until we have a better alternative (like the disclosure menu with a search input + inline list, like we talked about). Until then, it does make working with very large selects a lot more manageable.

Ok! I’ve been working on a mockup/wireframe for this in Figma, trying to think through the a11y UX; I’ll aim to have something concrete by Monday.

src/helpers/Cp.php
```diff
@@ -1396,6 +1396,49 @@ public static function addressCardHtml(Address $address, array $config = []): st
...
+    }
+
+    /**
+     * Returns a randomized string for the autocomplete value to disable browser autocomplete
+     *
+     * @return string
+     */
+    private static function getBogusAutofillValue(): string
+    {
+        return 'disable-autofill-' . mt_rand();
```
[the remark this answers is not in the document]
gcamacho079: On some browsers, autocomplete="off"` doesn’t have an effect, so a hack is to set the value to a random string. It seems like with Chrome, we may have to _also_ randomize the `name` and `id` strings or set them all to `new-password` in order to ensure the autocomplete menu doesn't pull up ([source](https://adamsilver.io/blog/stopping-chrome-from-ignoring-autocomplete-off/)). 🤔

src/templates/_includes/forms/selectize.twig
```diff
@@ -104,6 +105,17 @@
 
     const $select = $(`#${id}`);
 
+    const allowAutocomplete = () => {
+        const $originalSelectInput = $($select[0]);
```
brandonkelly, line 109: @gcamacho079 What’s the point of this line, vs. just referring to `$select` directly in this function?

gcamacho079: @i-just, could you chime in on this? I’m not too familiar with this change.

i-just: I tried a few thing before ending up with this code and that's a silly left over, sorry! I pushed an update.

src/helpers/Cp.php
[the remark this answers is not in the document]
brandonkelly: Hm alright… yeah looks like Chrome is the main culprit here, and [as of Chrome 72](https://stackoverflow.com/a/55045439/1688568) it ignores random strings. Not really interested in playing a cat and mouse game with them on this… if Chrome doesn’t want to support `autocomplete="off"` like a normal browser, that’s their (dumb) choice.

Issue #13943, opened 2023-11-18
[4.x]:`{% namespace %}` Twig tag should allow `0`


engram-design:
### What happened?

Thinking that the `{% namespace %}` tag should accept `0` as a value, and it currently doesn't. Probably to do with a falsey check, but I can't figure out how that works, sorry!

```twig
{% set items = ['first', 'second'] %}

{% for item in items %}
    {% namespace loop.index0 %}
        {{ tag('input', {
            name: item,
        }) }}
    {% endnamespace %}
{% endfor %}

// Produces
<input name="first">
<input name="1[second]">

// Expected
<input name="0[first]">
<input name="1[second]">
```

In reality, this would be surrounded by `fields` and the field handle, but in this reduced example, you can't have multiple `{% namespace %}` tags, so in a real-world use-case you'd produce:

```
// Expected
<input name="fields[myField][0][first]">
<input name="fields[myField][1][second]">
```

My use-case here is that the value for a field is an array, and I'd like to keep things simple by using an array syntax, rather than `new1` or similar like Matrix does.

### Craft CMS version

4.5.9

### Installed plugins and versions

-

engram-design:
Oh, managed to track down where that is, keen to hear your opinion on https://github.com/craftcms/cms/pull/13944

brandonkelly:
Namespaces are used to prefix `id` HTML attributes, and `id` attributes [aren’t allowed](https://www.w3.org/TR/html4/types.html#type-name) to begin with a number. So namespaces shouldn’t be either.

I’ve just added a check to ensure namespaces follow `id` attribute rules for Craft 5.

You can work around that by prefixing your keys with a string, e.g. `value:0`, `value:1`, etc.

brandonkelly:
Nevermind, that turned out to be pretty heavy-handed, and I forgot that we’re already ensuring namespaces are properly formatted for `id` attributes via `craft\helpers\Html::id()`.

Your PR is now merged for the next Craft 4 release, and I’ve [adjusted](https://github.com/craftcms/cms/commit/aa18b0f444de9e8fd4b28093e8feb212e228bb03) `Html::id()` to trim out any leading digits for Craft 5. (Entirely-numeric IDs will end up getting normalized to a 10-digit random string.)

engram-design:
Cheers! I know its kinda a "dumb" PR, but might be handy in some instances...

brandonkelly:
4.5.13 is out with that change!

PR #13944, opened 2023-11-18
Add support for `{% namespace %}` Twig tag to allow `0` as a value

engram-design:
Refer to https://github.com/craftcms/cms/issues/13943


brandonkelly:
See https://github.com/craftcms/cms/issues/13943#issuecomment-1850751820.

PR #13949, opened 2023-11-20
Improve German translations in field rules

rostockahoi:
Both these instructions were grammatically incorrect and slightly confusing. I think with this proposed changes it should be pretty clear.


brandonkelly:
Thanks! Pulled these changes in via CrowdIn (#14021).

Issue #13951, opened 2023-11-20
[4.5.11]: resave/entries --to=:empty: does not clear the value in field type = entries 


adamseabrook:
### Description
Running the resave/entries command to clear a field type of entries does not clear the field.
1. Tested running the same command but clearing a plain text field and it worked fine.
2. Tested running the same command without the `--site` option and the `createdBy` field is still not cleared.

### Steps to reproduce
1. In a es_MX site with a site handle of MX create a field called `createdBy` with a field type of `Entries` min and max relations of 1, show the site menu ticked, and manage relations on a per-site basis ticked.
2. Create an entry and select an entry inside the `createdBy` field and save the entry.
3. run `php craft resave/entries --site mx --set createdBy --to=:empty:`

### Expected behavior
`createdBy` should be empty.

### Actual behavior
`createdBy` is not empty and still retains the old value.

### Craft CMS version

4.5.11

### PHP version

8.2.10

### Operating system and version

Linux 6.1.0-0.deb11.11-amd64

### Database type and version

MariaDB 10.11.5

### Image driver and version

Imagick 3.7.0 (ImageMagick 7.1.1-13)

### Installed plugins and versions

Anchors 3.3.0
Architect	4.0.1
Field Manager	3.0.8
Google Cloud Storage	2.0.0
Linkit	4.0.4.1
Minify	4.0.0-beta.2
Position Fieldtype	4.0.0
Postmark	3.0.0
Retour	4.1.14
Servd Assets and Helpers	3.5.4
Super Table	3.0.12

i-just:
Hi, thanks for reporting! I raised a PR for this.

brandonkelly:
Craft 4.6.0 is out now with that fix.

PR #13953, opened 2023-11-21
set to `:empty:` for relation fields

i-just:
### Description
When setting a relation field to `:empty:` via the resave command, the normalised `$to` value needs to return an empty string (not `null`). Otherwise, the updated (empty) value won’t be set.

### Related issues
#13951


brandonkelly:
All of the built-in field types’ `normalizeValue()` methods treat empty strings as “empty”, so I changed the `:empty:` value to `''` instead of `null`, which fixed this.

PR #13955, opened 2023-11-22
Include column type argument when preparing field content to be saved

johnnynotsolucky:
### Description

With a psql database, if a custom field has a `Schema::JSON` field type, Craft inserts/updates the field data as a string instead of valid json. With mysql, this works fine. 

The functionality to prevent any encoding already exists in `prepareValueForDb`, just the type argument was missing in `saveContent`. This PR adds that argument so that data is saved correctly for psql.

### Related issues

N/A


brandonkelly:
A similar change has already been made for Craft 4.6 via d02708c52b1f570b37cc31fc8c4b58329e9a9fcd :)

Issue #11405, opened 2022-06-08
[3.x]: Unexpected behavior when an Asset field is using a volume that has been set as the "Temp Uploads Location"


angrybrad:
### Steps to reproduce

1. On a fresh Craft 3.7.44 install, create a local Asset volume.
2. Go to Settings->Assets->Settings and select that volume as the "Temp Uploads Location"
3. Go to Assets Index page and create a subfolder. Upload an image to that subfolder.
4. Create an Assets field and add it to a section's field layout.
5. Create a new entry, click the Asset field, and select the uploaded asset from the subfolder.
6. Save the entry.
7. Notice, on the Assets Index page, that the asset now lives in the root of the volume, not the subfolder.

### Expected behavior

The asset should be in the subfolder.

### Actual behavior

The asset is at the root of the volume.

### Notes

This is likely a change in behavior caused by https://github.com/craftcms/cms/commit/fa6fd6351d0e8d3453667b072d729c0e9dfd3367 that was released in 3.7.39.

Ideally, no one would select a volume that was intended to serve public assets, which mitigates this problem.

Things like the Asset field's source settings should remove any volumes so they can't be selected if "Temp Uploads Location" is set to a volume.

### Craft CMS version

3.7.4.4

### PHP version

7.4

### Operating system and version

macOS

### Database type and version

MariaDB 10

### Image driver and version

imagick

### Installed plugins and versions

None

jamie-s-white:
This is still affecting Craft 4.5.9 and 3.9.5. 

Can confirm it doesn't affect 3.7.16.

I wish I had seen this two days ago, I was just about to file a bug report with these very details... Although I absolutely agree that "no one would select a volume that was intended to serve public assets, which mitigates this problem", evidently my predecesor did, and lay this trap waiting for me when I started upgrading our Craft CMS websites past 3.7.16.

This issue also manifests itself when clicking preview, for whatever that is worth.

brandonkelly:
Craft 4.7.0 is now tagged with those changes.

PR #13957, opened 2023-11-22
Feature/cms 989 temp upload location filesystem

i-just:
### Description
- removes Control Panel > Settings > Assets > Settings page along with the `AssetSettingsController`
- adds `tempUploadsFs` general config property and method
- temporary uploads can now be stored either in the local storage folder (like before) or in a designated filesystem;
- if the filesystem is used to store temp uploads, it cannot be reused to create a volume or to hold transforms

### For documentation
**for the load-balanced environments in particular:**
- It would be good to add some sort of explanation on what the Temp Uploads Location is used for. I couldn’t find any info in the docs on what this setting does - did I miss it? Most people just leave that as is (local storage folder) or set that to a dedicated volume, but there’s nothing in v3 and v4 stopping you from re-using the volume you selected for Temp Uploads Location in your Assets fields, and that can lead to issues like this: [#11405](https://github.com/craftcms/cms/issues/11405). It’s a lesser-known setting, and I’m not sure it’s explained anywhere in the docs. 
- Long story short, it was introduced for load-balanced environments, and that’s what it should really be used for. In other cases, it should be set to the default, which is the local storage folder (for v3 and v4). 
- Because of the issue [#11405](https://github.com/craftcms/cms/issues/11405) and because it’s beneficial for the Cloud, in v5, we’re removing that setting from UI completely, putting it into General Config and moving it from Volume to Fs.

**for the upgrade guide:**
Before the upgrade, you need to check your setting in v4 in Control Panel > Settings > Assets > Settings. If Temp Uploads Location was set to an actual volume, you need to create a Filesystem that will be used to store the temporary uploads and then add the handle of that filesystem to the `config/general.php`, to `->tempUploadsFs()`. please note that this filesystem won’t be available for volume creation.

### Related issues
[CMS-989](https://linear.app/craftcms/issue/CMS-989/temp-upload-location-should-be-a-filesystem-not-a-volume)


src/config/GeneralConfig.php
```diff
@@ -2795,6 +2795,23 @@ class GeneralConfig extends BaseConfig
...
+     * ```
+     * ```shell Environment Override
+     * CRAFT_TEMP_UPLOADS_FS=myTempFs
+     * ```
+     * :::
+     *
+     * @group Assets
+     * @since 5.0.0
+     */
+    public ?string $tempUploadsFs = null;
```
timkelty, line 2813: Should this be called `$tempAssetUploadsFs`, to clarify it is specifically for assets, and not all uploads? Or is that too verbose?

src/translations/en/app.php
```diff
@@ -1553,6 +1553,7 @@
     'This field was updated in the Current revision.' => 'This field was updated in the Current revision.',
     'This field’s target subfolder path is invalid: {path}' => 'This field’s target subfolder path is invalid: {path}',
     'This field’s values are used as search keywords.' => 'This field’s values are used as search keywords.',
+    'This filesystem is used as a Temporary Uploads filesystem. Please choose a different one for your volume.' => 'This filesystem is used as a Temporary Uploads filesystem. Please choose a different one for your volume.',
```
timkelty, line 1556: Maybe:

> This filesystem has been reserved for temporary asset uploads. Please choose a different one for your volume.

timkelty (approved):
Love it!

src/translations/en/app.php
i-just: That sounds better - changed!

src/config/GeneralConfig.php
i-just: Either one is fine with me. @brandonkelly - want to be a tiebreaker?

brandonkelly: Hmmm… there are several `*Upload*` config settings, none of which specify `asset`, but they also each could technically be respected by other upload functions if they existed.

Then there’s `craft\services\Path::getTempAssetUploadsPath()`, which does have `asset`.

So I guess for consistency with that, yes we should go with `tempAssetUploadFs`.

i-just: `tempAssetUploadFs` it is!

AugustMiller:
Edited PR description to use new `tempAssetUploadFs` config name, for posterity! 💞

PR #13960, opened 2023-11-22
Add money input field helpers

nfourtythree:
### Description
Create HTML helpers in the `Cp` helper class for creating money input fields. This keeps consistency with other fields in the system.


brandonkelly (changes requested):
This breaks Money fields. New values are no longer saved.

nfourtythree:
Thanks @brandonkelly, it appears a couple of attributes went AWOL in the move from render template to the CP helper method.

Have fixed that now.

Issue #13959, opened 2023-11-22
[4.x]: Markdown in UI elements: Quotes inside inline code are incorrectly escaped


MoritzLost:
### Description

Not sure if this is an issue in the Markdown engine or in Craft. I'm talking about UI Elements like the Tip in Field Layouts. Single or double quotes inside inline code (marked up with backticks) are converted to entities, but the entities are visible in plaintext to the end user.

### Steps to reproduce

1. Add a Tip to any Field Layout with the following content:

```
`"Double Quotes"` or `'Single Quotes'`
```

Then open the entry form, the text displays like this:

```
&quot;Test&quot; &#039;Test&#039;
```

![Screenshot showing the incorrectly escaped tip contents](https://github.com/craftcms/cms/assets/10146880/9a0b611b-80c5-4cba-b83a-b8245019ac3b)

### Expected behavior

The quotes should be displayed as quotes instead of entities, as shown in the first example above.

### Actual behavior

The quotes are output as HTML entity shortcodes to the page.

### Craft CMS version

4.5.11

### PHP version

8.2

brandonkelly:
Craft 4.5.13 is out with a fix for this. Thanks again!

PR #13962, opened 2023-11-23
don't encode quotes before passing to markdown

i-just:
### Description
Don’t encode single and double quotes in a string that’s passed to Markdown for further processing.

Before:
Screenshot 2023-11-23 at 13 33 17

After:
Screenshot 2023-11-23 at 13 33 05

### Related issues
#13959

### DIFF
--- a/src/fieldlayoutelements/Tip.php
+++ b/src/fieldlayoutelements/Tip.php
@@ -101,7 +101,7 @@ class Tip extends BaseUiElement
             $classes[] = 'dismissible';
         }
 
-        $tip = Markdown::process(Html::encode(Craft::t('site', $this->tip)));
+        $tip = Markdown::process(Html::encodeNoQuotes(Craft::t('site', $this->tip), false));
         $closeBtn = $this->dismissible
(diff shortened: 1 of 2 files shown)


brandonkelly:
There was a similar bug with HTML tags that were output within backticks:

```md
`<strong>test</strong>`
```

Culprit for both was that the Markdown parser’s `renderCode()` and `renderInlineCode()` methods were running the content through `htmlspecialchars()`, which typically makes sense because you wouldn’t ever want HTML tags within `<code>` tags. However it doesn’t make sense in cases where all of the content has been pre-encoded, because the content isn’t trusted and the _only_ acceptable HTML should be whatever `Markdown::process()` returns (like for Tips and Warnings).

Worked around it by adding a new `pre-encoded` Markdown “flavor” (parser), which overrides `renderCode()` and `renderInlineCode()` and has them output the content verbatim. And now using that for rendering Tips and Warnings, as well as in a couple of other places where we were HTML-encoding Markdown content before passing it to `Markdown::process()`.

Issue #13963, opened 2023-11-23
[4.x]: 'Type of craft\log\MonologTarget::$logger must be ?Psr\Log\LoggerInterface (as in class samdark\log\PsrTarget)'


janhenckens:
### Description
Today's update of `samdark/yii2-psr-log-target`(1.1.4) also upgrade psr/log to 3.0.0, which causes the following error:

````php
PHP Compile Error 'yii\base\ErrorException' with message 'Type of craft\log\MonologTarget::$logger must be ?Psr\Log\LoggerInterface (as in class samdark\log\PsrTarget)'

in /var/www/html/vendor/craftcms/cms/src/log/MonologTarget.php:29

Stack trace:
#0 [internal function]: yii\base\ErrorHandler->handleFatalError()
````

Downgrading back to samdark/yii2-psr-log-target:1.1.3 also downgrades psr/log and fixes it for now.

### Steps to reproduce

1. Run craft 4.5.3 or later and run composer update
2. Run `php craft`

### Craft CMS version

4.5.11

### PHP version

8.1

### Installed plugins and versions

-

**Comment by pbmills** [comment]
I'm getting the same issue. How do we resolve it?

**Comment by brandonkelly** [comment]
Thanks all! Craft 4.5.11.1 is out with a fix for this.

**Comment by mark-chief** [comment]
How do I fix this? I am unable to upgrade craft as this has already broken it. Thanks

**Comment by MoritzLost** [comment]
@mark-chief If you can't run `php craft update` because the incompatible `psr/log` version is already installed, you need to update to 4.5.11.1 manually using Composer. Your `composer.json` has a requirement for `craftcms/cms`, you need to manually edit that to require `4.5.11.1`:

```json
"craftcms/cms": "4.5.11.1",
```

Then run `composer update --with-all-dependencies`. This will upgrade Craft to the specified version, which pins the `psr/log` dependency, so it will also downgrade that library to the supported version. After that, you can run `php craft update` again.

Alternatively, if you're on an older version and can't update Craft for any reason, you can workaround this issue by explicitly requiring `psr/log` 1.x in your project:

```
composer require psr/log:^1.0.0
```

This will downgrade the package and fix the compatibility issue. Once you upgrade to Craft 4.5.11.1, you can remove the direct depency again.

PR #13965, opened 2023-11-23
Fix changes for with psr/log > 3.0.0

sfsmfc:
The patch fix the error with psr/log > ^3.0. 

See issue https://github.com/craftcms/cms/issues/13963.


brandonkelly:
Thanks for the PR! This change wouldn’t have worked with yii/log v2 though, which will still get pulled in if you have Codeception installed. Fixed it via 25143382a7489aee11dfc5cc51bf87aa7b9e561f by removing the property override altogether, since it wasn’t really needed in the first place.

PR #13978, opened 2023-11-30
per-site element search scores

i-just:
### Description
Calculate search scores per element and per site so that when an element is included in the results multiple times, it doesn’t get extra points.

Previously, the scoring array contained key-value pairs where the key was the `elementId` and the value was the calculated search score. This approach meant that if the search was performed across multiple sites, the element’s score was the sum of all its scores across all sites.

This PR adjusts the scoring array so that it’s still indexed by the `elementId`, but the value is an array where the key is the `siteId` and the value is the score for that element for that site.
A sum of scores for the element across all sites is still used for ordering, but each element’s score is unique per site.

Before:
Screenshot 2023-11-30 at 08 55 42

After:
Screenshot 2023-11-30 at 08 56 45

Before:
Screenshot 2023-11-30 at 08 56 13

After:
Screenshot 2023-11-30 at 08 56 57

### Related issues
[cms-1198](https://linear.app/craftcms/issue/CMS-1198/calculate-element-search-scores-per-site)

### DIFF
--- a/src/elements/db/ElementQuery.php
+++ b/src/elements/db/ElementQuery.php
@@ -1577,8 +1577,10 @@ class ElementQuery extends Query implements ElementQueryInterface
         // Should we set a search score on the elements?
         if (isset($this->_searchResults)) {
             foreach ($rows as &$row) {
-                if (isset($row['id'], $this->_searchResults[$row['id']])) {
-                    $row['searchScore'] = (int)round($this->_searchResults[$row['id']]);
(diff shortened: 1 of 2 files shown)


brandonkelly:
Resolved via 5d8ac94f0972e2662b52199922c0ae9c19d3a596. Scores are now indexed by both element ID and site ID, so we can sort the full list of scores as a single array.

PR #13979, opened 2023-11-30
Add Development mode to the System report

boudewijn-zicht:
It would be terrible to run in development mode on a production server, this makes it clear whether the configuration is correct.


brandonkelly:
Dev mode is already identified by the caution tape stripes at the bottom of the global sidebar.

Screenshot of the caution tape stripes shown at the bottom of the global sidebar.

Issue #13974, opened 2023-11-28
[4.x]: Entry editor jumps to top after draft notification is dismissed.


msimpson:
### Description

The entry editor will jump to the top after the draft creation notification is dismissed.

### Steps to reproduce

1. Edit an entry.
2. Press "Create a draft".
4. Scroll down.
5. Pages scrolls back up as the "Title" field is given focus.

### Expected behavior

For arbitrary fields to not be focused upon notification dismissal.

### Actual behavior

https://github.com/craftcms/cms/assets/730334/c7840f56-3479-4737-99b8-255b97bd8887

### Craft CMS version

Craft Pro 4.5.9

### PHP version

8.2.3

### Operating system and version

Linux 3.10.0-1160.83.1.el7.x86_64

### Database type and version

MariaDB 10.3.35

### Image driver and version

Imagick 3.7.0 (ImageMagick 7.1.1-15)

### Installed plugins and versions

- Calendar 4.0.9
- Forms 4.1.11
- Navigation 2.0.22
- Redactor 3.0.4
- Retour 4.1.13
- Super Table 3.0.12
- Tags 2.0.0

brandonkelly:
Fixed for the next release via #14011.

PR #13986, opened 2023-12-01
only manage focus if there's something actionable in the toast

i-just:
### Description
Only manage focus if there’s something truly actionable in the toast notification. 
If there’s a UI element in the toast, move the focus to the notification when it opens and move it back to the previously focused element when it closes.
For cases where there are no UI elements in the notification toast, just keep the layer that allows the toast to be dismissed via the ESC key.

### Related issues
#13974


gcamacho079 (approved):
@i-just tested with VO/Safari and keyboard for:

1. “Create a draft” (no actionable control in the notification)
2. Dashboard widget deletion (Undo button in notification)

This seems like a great middle-ground solution that prevents page scrolling in certain circumstances, but still manages focus when there’s an important action (like "Undo") inside the notification. 👌🏼 Love it!

brandonkelly:
Whoops, missed this, and made a similar fix (for #14011) via af94736e6a4848707966134cceda4504dbff332b.

PR #14001, opened 2023-12-07
Use Craft queue as proxy queue

boboldehampsink:
### Description

This way we can also use `Queue::EVENT_AFTER_EXEC_AND_RELEASE` on the proxy queue, for example.

This PR is not tested but ment as a discussion on why the proxy uses the yii base queue in the first place


brandonkelly:
The proxy queue could by any queue driver, e.g. SQS. The only driver you _wouldn’t_ likely want to go with is Craft’s `Queue` class. (If you want to use Craft’s queue as the main queue driver, there’s no need to run it by proxy.)

PR #14004, opened 2023-12-07
Bugfix/volume empty fs selector

i-just:
### Description
When adding/editing a volume:
- add a validation rule so that it’s not possible to save a volume without a filesystem selected
- instead of hiding unavailable filesystems, show them as disabled (and don’t allow them to be selected)
- add an empty “Select a filesystem” option so that when there are no available filesystems (none have been created or all are already in use), the “Create a new filesystem” option isn’t selected by default and doesn’t automatically trigger the slideout.

If we still prefer not to show the filesystems in use, I can revert that part and go with validation and improved empty option.

https://github.com/craftcms/cms/assets/4500340/2d36912f-b7bf-4f04-a004-3d612e314b57

### Related issues
from an internal discussion on Slack


brandonkelly:
Couple things:

- Not a big fan of the “Please select” label. Can we pull the same thing off with a blank option or non-breaking space? (Thought that was already the case?)
- `Volume::attributeLabels()` needs to start including `'fsHandle' => Craft::t('app', 'Asset Filesystem')`, so  validation errors match the visible field label instead of calling it `Fs Handle`. (Same goes for “Transform Filesystem” if it can also receive validation errors.)

i-just:
I changed the label to “Select a filesystem” after recording the video. We do a similar thing on Settings > Users > Settings page with “User Photo Location” (the empty value label is “Select a volume”), but yes, we can do the same with an empty option.
Screenshot 2023-12-11 at 08 41 54
Screenshot 2023-12-11 at 08 41 39


Noted! I adjusted the `Volume::attributeLabels()` to include `fsHandle`, `transformFsHandle` and `transformSubpath`.

Issue #14003, opened 2023-12-07
[4.x]: Ambiguous error message when uploading multiple files


MoritzLost:
### Description

When uploading multiple files at once through the assets index, individual uploads sometimes fail. Not sure why this happens, probably due to timeouts when the server is at maximum capacity due to handling a bunch of files. In any case, you only get this error message as a notification:

> Uploads failed.

![craft-upload-failed](https://github.com/craftcms/cms/assets/10146880/82162dcf-1c0c-42e4-bb6a-eb60e23ded6b)

Other than that, there's no indicator that something has gone wrong. If you are uploading a bunch of files in the background while doing something else, you might even miss this notification entirely (since they disappear after a couple of seconds by default, unless you've changed that in the accessibility settings).

The error message is very ambiguous. Has the entire upload failed, or only _some_ of the files? It's the latter, but the error message should indicate that. It also doesn't tell you _which_ upload has failed, so you have to manually compare your local file folder with the uploaded assets to see which one is missing.

### Steps to reproduce

1. Upload a lot of files at once (unfortunately, I don't have a reproducible way to cause an upload to fail).

### Expected behavior

The error message should be improved to indicate that _some_ of your files have failed to upload. It should also tell you _which_ file has failed to upload, so you can upload that specific file again.

### Actual behavior

The error message is vague and unclear.

### Craft CMS version

4.5.11.1

### PHP version

8.2

### Installed plugins and versions

-

i-just:
Thanks @MoritzLost. Specific error messages help with replication. Based on a quick look, we already have specific error messages that mention the filename if that data is made available to the uploading script, and the fallback is the generic “Upload failed.” message. I’ll have a closer look and keep you updated.

i-just:
Hi @MoritzLost, I had a closer look into it, and the generic “Upload failed.” message is the expected behaviour when there’s no error/reason passed up the chain. I’ve triggered all sorts of errors related to uploading and saving an asset, which all resulted in a descriptive error message being presented to the user. I could only trigger the generic one when a 502 Bad Gateway was thrown.

Given all that, if you were to open developer tools in your browser before you start uploading multiple assets, you can follow the progress in the Network tab. If it happens again, you can then see what actually caused it (and which files didn’t upload). If you spot anything that you think should be passable up the chain, let me know, and I’ll have another look, but at the moment, it does look to be working as expected.

I’m going to close this now, but feel free to reply if you find further info that could mean that this is a bug.

[...]

PR #14029, opened 2023-12-12
Fix error messages for custom uploaders

timkelty:
### Description

[This PR](https://github.com/craftcms/cms/pull/14019) broke things for custom uploaders (eg Cloud), as `data` is undefined in that context.

This is because we need to support the ancient [blueimp/jQuery-File-Upload](https://github.com/blueimp/jQuery-File-Upload) events, as well as standard JS events for alternate uploaders.

So, we need account for `data` being null, as well as `event` being an object with different props.

To try and clarify this a bit, I added a default `null` value for `data`.

@i-just I also used `data.files` rather than getting it from the FormData like you were, as that seemed more straightforward. If there is a reason not to use that let me know.

### Related issues
- https://github.com/craftcms/cms/issues/14003
- https://github.com/craftcms/cms/pull/14019


i-just (commented):
Thanks @timkelty!

I tested your solution, and it works like my previous one in terms of plain cms uploads. `data.files` seems more straightforward, and I can’t think of a reason not to use it.

I guess the same approach should be used in the 3 other places, too, so I pushed changes for the following files: 
- `elements/Asset.php` (lines: 1485-1558) => triggered when you’re editing an individual asset and use the “Replace file” button
- `web/assets/cp/AssetSelectInput.js` => triggered when you use direct upload in an Assets field (“Upload files” button)
- `web/assets/cp/ImageUpload.js` => triggered when you upload a photo to a user profile

Any chance you can test this against the cloud?

timkelty:
@i-just your additions look 💯, but i'll give it a spin in a cloud project to make sure.

timkelty:
@i-just tested on cloud and looks good!

Issue #13645, opened 2023-09-05
[4.x]: Delete For Site behavior for multi-site entries is inconsistent (hard-/soft-deletion)


michaelrog:
### What happened?

The _Delete For Site_ behavior appears to be inconsistent with respect to soft-/hard-deletion:
- If I delete an entry that is only enabled in one site, the entry is soft-deleted and appears in _Trashed_.
- If I delete a multi-site-enabled entry that is currently only active on one site, the entry is soft-deleted and appears in _Trashed_.
- **If I delete a multi-site-enabled entry that is currently active on multiple sites, the entry (for that site) is hard-deleted and cannot be restored.**

This behavior is confusing; It feels like a bug.

**As a content editor, I would *expect* that, if soft-deletes are enabled for entries, an entry I delete from a site should be recoverable from Trashed, no matter how many sites it's active in at the time of deletion.**

### Steps to reproduce

1. Accidentally Delete-For-Site an entry that is only enabled in one site. See that you can restore the entry from Trashed.
2. Accidentally Delete-For-Site an entry that is enabled in multiple sites. See that it's irrecoverably gone. Cry. 😢 

---

Craft CMS 4.5.3
PHP 8.1.20
MySQL 8.0

michaelrog:
(Apologies, this should have been tagged `craft4`. Although I suspect the same issue affects `craft3`, too.) 😬

i-just:
Hi, thanks for getting in touch! 

Deleting an entry for a site when that entry exists for other sites doesn’t go through the same delete mechanism as in other cases. When you delete a multi-site element for just one site, it still exists for the other sites. That’s why it behaves differently.

We'll discuss this one internally and I'll get back to you once I have an update.

brandonkelly:
Yeah this is currently working as intended, and it would take some schema changes to make this possible. Going to close, but feel free to open up a discussion requesting soft deletion for site-specific entry content.

PR #14032, opened 2023-12-13
Bulk element operation tracking + event

brandonkelly:
### Description

Adds the concept of bulk element operation tracking, giving plugins an opportunity to do things only after a bulk operation is fully completed, fetching all affected elements at once.

This could be useful for expensive actions like sitemap generation, static page generation, updating Algolia indexes, etc.

```php
use craft\elements\Entry;
use craft\events\BulkElementOpEvent;
use craft\services\Elements;
use yii\base\Event;

Event::on(Elements::class, Elements::EVENT_AFTER_BULK_OP, function(BulkElementOpEvent $event) {
    // Fetch all the entries that were affected
    $entries = Entry::find()
        ->inBulkOp($event->key)
        ->status(null)
        ->all();
    // ...
});
```

All built-in element (multi-)save and deletion actions will ensure that a bulk operation is active, even if only impacting a single element. So plugins could potentially stop listening to `Elements::EVENT_AFTER_SAVE_ELEMENT`/`Element::EVENT_AFTER_SAVE` altogether, and could go all-in on `EVENT_AFTER_BULK_OP`.

Plugins that are performing bulk operations on multiple elements can define their own bulk operations like so:

```php
$bulkKey = Craft::$app->elements->beginBulkOp();

// save/delete elements
// ...

Craft::$app->elements->endBulkOp($bulkKey);
```

There’s also a new `BaseBatchedElementJob` queue job class, which extends `BaseBatchedJob`, but ensures that all of the individual batches are wrapped in a single bulk element operation.

### Related issues

- #14000

### DIFF
--- a/CHANGELOG-WIP.md
+++ b/CHANGELOG-WIP.md
@@ -61,6 +61,7 @@
 - Entry type names and handles must now be unique globally, rather than just within a single section. Existing entry type names and handles will be renamed automatically where needed, to ensure uniqueness.
 - Assets, categories, entries, and tags now support eager-loading paths prefixed with a field layout provider’s handle (e.g. `myEntryType:myField`).
 - Element queries now have an `eagerly` param, which can be used to lazily eager-load the resulting elements for all peer elements, when `all()`, `collect()`, `one()`, `nth()`, or `count()` is called.
+- Element queries now have an `inBulkOp` param, which limits the results to elements which were involved in a bulk operation. ([#14032](https://github.com/craftcms/cms/pull/14032))
 - Address queries now have `addressLine1`, `addressLine2`, `administrativeArea`, `countryCode`, `dependentLocality`, `firstName`, `fullName`, `lastName`, `locality`, `organizationTaxId`, `organization`, `postalCode`, and `sortingCode` params.
 - Entry queries now have `field`, `fieldId`, `primaryOwner`, `primaryOwnerId`, `owner`, `ownerId`, `allowOwnerDrafts`, and `allowOwnerRevisions` params.
 - Entries’ GraphQL type names are now formatted as `<entryTypeHandle>_Entry`, and are no longer prefixed with their section’s handle. (That goes for Matrix-nested entries as well.)
@@ -171,6 +172,7 @@
 - Added `craft\enums\PropagationMethod`.
 - Added `craft\enums\TimePeriod`.
 - Added `craft\events\BulkElementsEvent`.
+- Added `craft\events\BulkOpEvent`. ([#14032](https://github.com/craftcms/cms/pull/14032))
 - Added `craft\events\DefineEntryTypesForFieldEvent`.
 - Added `craft\events\DefineFieldHtmlEvent::$inline`.
 - Added `craft\fieldlayoutelements\BaseField::$includeInCards`.
@@ -226,6 +228,11 @@
 - Added `craft\models\Section::getCpEditUrl()`.
 - Added `craft\models\Volume::getSubpath()`.
 - Added `craft\models\Volume::setSubpath()`.
+- Added `craft\queue\BaseBatchedElementJob`. ([#14032](https://github.com/craftcms/cms/pull/14032))
+- Added `craft\queue\BaseBatchedJob::after()`.
+- Added `craft\queue\BaseBatchedJob::afterBatch()`.
+- Added `craft\queue\BaseBatchedJob::before()`.
+- Added `craft\queue\BaseBatchedJob::beforeBatch()`.
 - Added `craft\services\Auth`.
 - Added `craft\services\Entries::refreshEntryTypes()`.
 - Added `craft\services\Fields::$fieldContext`, which replaces `craft\services\Content::$fieldContext`.
--- a/src/config/app.php
+++ b/src/config/app.php
@@ -4,7 +4,7 @@ return [
     'id' => 'CraftCMS',
     'name' => 'Craft CMS',
     'version' => '5.0.0-alpha',
-    'schemaVersion' => '5.0.0.10',
+    'schemaVersion' => '5.0.0.11',
     'minVersionRequired' => '4.4.0',
     'basePath' => dirname(__DIR__), // Defines the @app alias
     'runtimePath' => '@storage/runtime', // Defines the @runtime alias
--- a/src/db/Table.php
+++ b/src/db/Table.php
@@ -44,6 +44,8 @@ abstract class Table
     public const DRAFTS = '{{%drafts}}';
     /** @since 4.5.0 */
     public const ELEMENTACTIVITY = '{{%elementactivity}}';
+    /** @since 5.0.0 */
+    public const ELEMENTBULKOPS = '{{%elementbulkops}}';
     public const ELEMENTS = '{{%elements}}';
     /** @since 5.0.0 */
     public const ELEMENTS_OWNERS = '{{%elements_owners}}';
--- a/src/db/Table.php
+++ b/src/db/Table.php
@@ -44,9 +44,9 @@ abstract class Table
     public const DRAFTS = '{{%drafts}}';
     /** @since 4.5.0 */
     public const ELEMENTACTIVITY = '{{%elementactivity}}';
-    /** @since 5.0.0 */
-    public const ELEMENTBULKOPS = '{{%elementbulkops}}';
     public const ELEMENTS = '{{%elements}}';
+    /** @since 5.0.0 */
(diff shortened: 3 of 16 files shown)


khalwat:
I had some objections, but in thinking it through, if I just list on `Elements::EVENT_AFTER_BULK_OP` then I can "do the thing" then because it will only be triggered at the end of the save (whether it's one element or thousands).

So I don't need to have an event before the bulk operation starts (though I suspect someone, somewhere may want an `Elements::EVENT_BEFORE_BULK_OP` too).

Obviously, this would only be possible when refactoring things for Craft 5, but it looks great to me, thanks!

(he says, in a vacuum, having never used it in the real world)

brandonkelly:
Yeah exactly. Keeping track of elements as things happen felt like the right approach for a few reasons:

- That will include any nested elements that are modified as a result of the primary elements being saved, which wouldn’t have been known at the outset.
- If there are thousands, keeping a list of modified element IDs wouldn’t really be helpful, since querying for elements by thousands of IDs would be horrible for query performance (and possibly beyond resource limits).

> So I don't need to have an event before the bulk operation starts (though I suspect someone, somewhere may want an `Elements::EVENT_BEFORE_BULK_OP` too).

Yeah maybe. Added it just in case. But in general I suspect `ElementQuery::inBulkOp()` will suffice.

khalwat:
I assume Craft is then going to move to use this for the resave elements stuff, propagation method changes, etc., and in FeedMe, and other bulk element operations ya?

brandonkelly:
Yes, everything in core already uses it in this PR, and we would start using it in Feed Me when porting it to Craft 5.

khalwat:
I wonder if this can be leveraged by the template `{% cache %}` mechanism, too, to coalesce the cache breaking that happens on element changes. Or maybe it already does in some manner?

Anyway, nifty, well done!

bencroker:
This looks great, will definitely be taking advantage of it in Blitz 5, thanks @brandonkelly and @khalwat for raising this!

brandonkelly:
Thanks for taking a look @khalwat and @bencroker! Merged 🚀

janhenckens:
This is a great addition, something I can use in the v5 version of Scout too 👍

michaelrog:
> plugins could potentially stop listening to Elements::EVENT_AFTER_SAVE_ELEMENT/Element::EVENT_AFTER_SAVE altogether, and could go all-in on EVENT_AFTER_BULK_OP.

This sounds like a nice direction, but I'm confused on the event naming. It sounds like these events are going to be a wrapper for _all_ element saves, not just bulk ones.

If I'm reading this right, this is Craft's way of saying:
 - "I'm about to save some number of elements."
 - "I'm done saving some number of elements; You can use this key to find the elements that changed."

In that case, it would be clearer to not mention "bulk" in the names of things, since that's only _one of many_ possible use cases. e.g., maybe more precise:
 - `ElementSaveOperationEvent`, `EVENT_BEFORE_SAVE_OP`, and `EVENT_AFTER_SAVE_OP`
 - `ElementBatchOperationEvent`, `EVENT_BEFORE_BATCH_OP`, and `EVENT_AFTER_BATCH_OP`
 - ...or something in this vein

michaelrog:
Also, curious how this would work in multi-site use cases...

Currently, we can listen for `Element::EVENT_AFTER_PROPAGATE` to do third-party stuff once an element is fully saved and propagated to other sites. Importantly, we can tell which _site_ is relevant for this operation, because the initiating element is accessible directly from the event.

By contrast, these new events _don't_ have the actual elements attached; Instead, they carry a _key_ for querying those elements after-the-fact. So, we would presumably need that key to allow querying _all_ the _site-elements_ that changed during the save operation; segmenting by element ID alone would not be sufficient... right?

michaelrog:
> Keeping track of elements as things happen felt like the right approach for a couple reasons:
> - That will include any nested elements that are modified as a result of the primary elements being saved, which wouldn’t have been known at the outset.

This feels like it might get tricky. A batch operation might save elements of many types, some of which may be third-party... So, how do I know, listening to `Elements::EVENT_AFTER_BULK_OP`, what types of elements changed?

For example:
```
Event::on(Elements::class, Elements::EVENT_AFTER_BULK_OP, function(BulkOpEvent $event) {
  // Fetch all the entries that were affected
  $entries = Entry::find()
    ->inBulkOp($event->key)
    ->status(null)
    ->all();
});```

How do I know, from the event, that I need to build an `Entry` query here?

brandonkelly:
> Eh, I think I misunderstood. This _is_ only for _bulk_ operations. i.e. This functionality _does not_ cover normal run-of-the-mill atomic element saves/deletes.

@michaelrog Your first take was correct. As I said:

> All built-in element (multi-)save and deletion actions will ensure that a bulk operation is active, even if only impacting a single element. So plugins could potentially stop listening to `Elements::EVENT_AFTER_SAVE_ELEMENT`/`Element::EVENT_AFTER_SAVE` altogether, and could go all-in on `EVENT_AFTER_BULK_OP`.

> By contrast, these new events _don't_ have the actual elements attached; Instead, they carry a _key_ for querying those elements after-the-fact. We presumably need that key to facilitate querying _all_ the _site-elements_ that changed during an operation. Capturing element ID alone would not be sufficient... right?

Correct, this is only concerned with which elements were updated, regardless of site. I considered tracking the site, but decided against it since most element saves will propagate across all sites anyway.

> How do I know, from the event, that I need to build an `Entry` query here?

You would just need to check each individual element type you care about. Or continue listening for `EVENT_AFTER_ELEMENT_SAVE` and track which element types are being touched from there – but considering you would need to do that potentially across multiple requests (in the case of batched queue jobs), that may be more trouble than it’s worth.

michaelrog:
> Your first take was correct.

In that case, naming these as "bulk" does feel a bit too specific, and having both "batch" and "bulk" in the nomenclature seems to beg a question about the difference between them. Maybe "batch" would be more precise?

 - `EVENT_BEFORE_BATCH_OP`, and `EVENT_AFTER_BATCH_OP`, `ElementBatchOperationEvent`
 - `EVENT_BEFORE_SAVE_BATCH`, `EVENT_AFTER_SAVE_BATCH`
 - `EVENT_BEFORE_DELETE_BATCH`, `EVENT_AFTER_DELETE_BATCH` (?)

I see where bulk queue jobs are amended to `beginBulkOp()` and `endBulkOp()`, but I _don't_ see where this plays into ordinary element saving. i.e. If I hit "Save" in the control panel, how/where does that action kick off a batch op? (Apologies in advance for being dense here.)

...

> You would just need to check each individual element type you care about.

The challenge is, I may not know ahead-of-time which element types I care about, especially if 3p plugins are involved. So then do I have to check every possible element type?

I guess I could try to query the `elements_bulkops` table using the key, join it into the `elements` table to get type, and select distinct types from that. That seems like a cumbersome hoop to jump... but I guess I don't have a good suggestion for an alternative... 🤔

khalwat:
So if we're hung up on naming... `transaction` might work? Sort of mirroring Yii2's `yii\db\Connection::beginTransaction()` / `yii\db\Connection::endTransaction()` so instead we'd have:

* Events - `Elements::EVENT_BEFORE_ELEMENT_TRANSACTION` &  `Elements::EVENT_AFTER_ELEMENT_TRANSACTION`
* Methods - `$transactionKey = Craft::$app->elements->beginElementTransaction();` & `Craft::$app->elements->endElementTransaction($transactionKey);`
* Query param - `->inElementTransaction($event->key)`

It probably would make more intuitive sense this way... but honestly, the naming doesn't bother me terribly, as long as the functionality is there.

michaelrog:
It occurs to me that _deleted_ elements may be difficult to work with in this context:

_Soft-deleted_ elements can still be queried using `inBulkOp()`... But _hard-deleted_ elements aren't registered at all. (The `elementId` fk prevents them from existing in `elements_bulkops` in the first place.)

e.g. If I hard-delete a bunch of entries, and now need to remove those entries from an external search index, there's no native mechanism for me to batch that, right?

(I could see this being especially frustrating since [Delete For Site behavior for multi-site entries is inconsistent between hard-/soft-delete](https://github.com/craftcms/cms/issues/13645).)

So it seems like the best practice will still be to listen to `EVENT_AFTER_DELETE_ELEMENT` for elements being deleted, rather than trying to handle them in bulk... Am I understanding this correctly?

brandonkelly:
@michaelrog yeah, no way to keep track of hard deletions after-the-fact, including site-specific hard-deletions, so if you care about those, you would still want to listen to `EVENT_AFTER_DELETE` and `EVENT_AFTER_DELETE_FOR_SITE`.

Issue #9867, opened 2021-09-22
Searching users extremely slow in large DB


GaryReckard:
### Description
I've been trying to determine the cause of some slow actions / queries, in particular, search for users on the users index page. I enter in a search term, and it takes 10+ seconds to respond. 

Looking at the debug toolbar, I see the slowest query is the one that searches the `searchindex`, and looks like this:
```
SELECT * FROM `craft_searchindex` WHERE (`keywords` LIKE '%searchterms%') AND `siteId` = 4 AND `elementId` IN (... a list of almost 200k ids ...)
```
I ran that query directly on my DB and it took 8+ seconds to complete.

I also see a query executed right before that one that returns that same list of element ids. It looks like:
```
SELECT `elements`.`id`
FROM (SELECT `elements`.`id` AS `elementsId`, `elements_sites`.`id` AS `elementsSitesId`, `content`.`id` AS `contentId`
FROM `craft_elements` `elements`
INNER JOIN `craft_users` `users` ON `users`.`id` = `elements`.`id`
INNER JOIN `craft_elements_sites` `elements_sites` ON `elements_sites`.`elementId` = `elements`.`id`
INNER JOIN `craft_content` `content` ON (`content`.`elementId` = `elements`.`id`) AND (`content`.`siteId` = `elements_sites`.`siteId`)
WHERE (`elements_sites`.`siteId`=4) AND (`elements`.`archived`=FALSE) AND (`elements`.`dateDeleted` IS NULL) AND (`elements`.`draftId` IS NULL) AND (`elements`.`revisionId` IS NULL)) `subquery`
INNER JOIN `craft_users` `users` ON `users`.`id` = `subquery`.`elementsId`
INNER JOIN `craft_elements` `elements` ON `elements`.`id` = `subquery`.`elementsId`
INNER JOIN `craft_elements_sites` `elements_sites` ON `elements_sites`.`id` = `subquery`.`elementsSitesId`
INNER JOIN `craft_content` `content` ON `content`.`id` = `subquery`.`contentId`;
```

As a test, I put that query as a subquery in the first one, in place of the list of 200k element ids, to get:
```
SELECT * 
FROM 
	`craft_searchindex` 
WHERE 
	(`keywords` LIKE '%searchterms%') AND 
	`siteId` = 4 AND 
	`elementId` IN (
		SELECT `elements`.`id`
FROM (
[...]

PR #14055, opened 2023-12-18
Improve search performance

brandonkelly:
### Description

Improves search performance in three ways:

1. When querying for elements with pre-scored search results, the scoring array is now sliced based on the element query’s `offset` and `limit`, so only the current result set’s element IDs are included in the query.
2. Scored element queries now use a fixed order expression rather than a case expression in the `order by` clause.
3. Non-scored element queries no longer execute a keyword query separately. Instead the keyword query is added a subquery of the main element query within the `where` clause.

### Related issues

- #9867


markhuot:
Oooo, this is an exciting improvement, thanks!

PR #14056, opened 2023-12-18
Copy field values from other sites

brianjhanson:
Continuation of #12436 but instead of handling the copy server-side, the controller will return fragments of HTML that we use to replace fields and immediately save a draft. This simplifies the logic a bit as we can rely on existing "Save Draft" processes to save the values. 

## Description

Adds a new “Copy value from site…” action to translatable fields’ action menus on element edit pages, making it possible to copy the field value from another site over to the current site.

A custom field’s action menu, with a “Copy value from site…” item

Plugin-supplied field types that want to add support for cross-site value copying will need to start implementing the new `CrossSiteCopyableFieldInterface`.


piotrpog:
I think this was "missing piece" of multi-language functionality of Craft CMS. Thanks for adding this.

mmikkel:
> I think this was "missing piece" of multi-language functionality

I concur; really looking forward to seeing it merged. Any chance it'll make it into 5.2? 😇

src/web/assets/cp/src/js/ElementEditor.js
```diff
@@ -589,6 +611,255 @@ Craft.ElementEditor = Garnish.Base.extend(
...
+
+      this.copyModal.on('hide', function () {
+        // return focus to the actions btn
+        const actionMenu = $('#action-menu')
+          .disclosureMenu()
+          .data('disclosureMenu');
+        actionMenu.$trigger.focus();
+      });
+
+      this.addListener($('.copyBetweenSites'), 'submit', 'copyValuesFromSite');
```
brandonkelly: @i-just As far as I can tell, this selector won’t ever match anything. Am I missing something? (Same for line 747.)

src/web/assets/cp/src/js/ElementEditor.js
```diff
@@ -589,6 +611,255 @@ Craft.ElementEditor = Garnish.Base.extend(
...
+          namespace: $btn.data('namespace'),
+        })
+      );
+
+      this.copyHud = new Garnish.HUD($btn, $hudContent);
+
+      this.addListener($('.copyBetweenSites'), 'submit', 'copyValuesFromSite');
+    },
+
+    _buildSitesList: function (sites, hud) {
```
brandonkelly: This method looks to be unused.

src/web/assets/cp/src/js/ElementEditor.js
```diff
@@ -589,6 +611,255 @@ Craft.ElementEditor = Garnish.Base.extend(
...
+      });
+
+      return $ul;
+    },
+
+    createDraftBeforeCopying: function () {
+      // if we're copying a single field (so we have the hud, not the modal)
+      if (this.copyHud) {
+        // check if the field is classed as nested
+        return Garnish.hasAttr(this.copyHud.$trigger, 'data-nested');
```
brandonkelly: Since non-NEM fields are also creating drafts, could we just remove this method and assume that every time we’re copying field values, a draft should be created first?

If so we can also remove `craft\fieldlayoutelements\BaseField::isNested()`.

src/web/assets/cp/src/js/ElementEditor.js
i-just: No, you’re not missing anything. They were used in the older version of this feature, but not anymore. I removed them.

src/web/assets/cp/src/js/ElementEditor.js
i-just: Same as above. That's removed now.

src/web/assets/cp/src/js/ElementEditor.js
i-just: That’s needed to allow copying fields inside of a matrix field that’s set to editable blocks mode to work. 

Replication steps: 
- have at least 2 sites set up
- create a matrix field in editable blocks mode with the propagation method set to “all”
- the matrix field should have an entry type that has, e.g. plain text field set to be translatable
- create a section that’s enabled for both sites; its entry type should use the matrix field we just created
- create an entry in that section, add a block to the matrix field, fill out the plain text field, fully save the entry
- edit the entry in the second site, change the plain text field inside the matrix field so that it’s clear it’s from the other site & fully save
- in either site, edit the entry and copy the plain text field inside the matrix field from the other site

with the code you asked about, it works as expected; without it, it does not

PR #14069, opened 2023-12-21
Protect against empty values

timkelty:
Prevent a potential php error when a `x-craft-license-info` header is present, but empty.
Fixes https://linear.app/craftcms/issue/CLD-408/plugin-store-doesnt-load-on-fresh-install


timkelty:
@brandonkelly we could array_filter in `_normalizeHeaders` if we're sure we're never intentionally expecting an empty value from a header.

PR #14088, opened 2023-12-31
Fix an error with element migrations and no field layout

engram-design:
Because you can call `updateElements` with `$fieldLayout` being `null`, this fragment will throw an error.

In my case, I just want to ensure that the title is copied across for an element, and there's not always guaranteed a field layout (as the user controls that - whether to use a "Form Template" to allow field layouts). This is for Formie.


brandonkelly:
Doh, thanks!

brandonkelly:
5.0.0-alpha.4 is tagged with this fix.

PR #14089, opened 2023-12-31
Add `$dropContentRows` and `$dropContentTable` options to element content migration

engram-design:
Given the option, I'd rather _not_ blow away my custom content rows or tables. This is for Formie, where content is stored in their own content tables `fmc_*`. We're moving content to the native `elements_sites` table, but I'd _really_ like not to blow away content tables _just_ yet as a bit of a backup.


brandonkelly:
Added a class-level `$preserveOldData` property instead, which you can override and set to `true` if you want the old content rows (and table) preserved. (2a06d454fbefd5305ea51def2d9da7b882824ac3)

```php
protected bool $preserveOldData = true;
```

brandonkelly:
Craft 5.0.0-alpha.4 is tagged with that change.

engram-design:
Nice, thanks!

PR #14107, opened 2024-01-04
composer update + monolog v3

i-just:
### Description
Updates monolog to v3.

Note: This can only be done after `webauthn-lib` is updated to v4, so it’s a draft until [cms-1225](https://github.com/craftcms/cms/pull/14106) is reviewed and merged.

### Related issues
[cms-1226](https://linear.app/craftcms/issue/CMS-1226/update-to-monolog-v3)


src/log/ContextProcessor.php
```diff
@@ -84,18 +85,27 @@ public function __invoke(array $record): array
...
         }
 
-        return $record;
+        return new LogRecord(
+            datetime: $record->datetime,
+            channel: $record->channel,
+            level: $record->level,
+            message: $record->message . $message,
+            context: ($this->key === 'context') ? $record->context + $data : $record->context,
+            extra: ($this->key === 'extra') ? $record->extra + $data : $record->extra,
```
timkelty, line 106: This certainly should work…but I'm not seeing (nor do I recall) a reason for `\craft\log\ContextProcessor::key` to exist (seems like its just always `context`).

 I'm tempted to say just drop that prop, and just set `LogRecord::context`.

timkelty (changes requested):
@i-just looks great – just one note. Let me know if you want to talk through it.

src/log/ContextProcessor.php
i-just: I couldn’t figure out what was the use of it either. We don’t set that `$key` anywhere, and it always seems to be `context`. I did some searching and reading, which didn’t seem to give any hints. Since you also think this might not be used, I dropped it.

Issue #14100, opened 2024-01-03
[4.x]: Local image source is deleted before transformation has started


kringkaste:
### Description

After upgrading from 4.4.x to 4.5.13 we have an issue with generating transformations from assets stored on a S3 bucket. We get the following error:

```
[x.x.x.x][1][-][error][craft\errors\ImageTransformException] craft\errors\FsObjectNotFoundException: The file "test-image.jpg" does not exist. in /mnt/dlfnova/htdocs/vendor/craftcms/cms/src/helpers/ImageTransforms.php:220
Stack trace:
#0 /mnt/dlfnova/htdocs/vendor/craftcms/cms/src/helpers/ImageTransforms.php(423): craft\helpers\ImageTransforms::getLocalImageSource(Object(craft\elements\Asset))
#1 /mnt/dlfnova/htdocs/vendor/craftcms/cms/src/imagetransforms/ImageTransformer.php(378): craft\helpers\ImageTransforms::generateTransform(Object(craft\elements\Asset), Object(craft\models\ImageTransform), Object(Closure), NULL)
[...]
```

We debugged the problem on the stage environment and could identify the

```
// And delete it after the request, if nobody wants it.
if (Craft::$app->getConfig()->getGeneral()->maxCachedCloudImageSize == 0) {
    FileHelper::deleteFileAfterRequest($imageSourcePath);
}
```

code block as the source of the issue. Our general config for `maxCachedCloudImageSize` is `0` and the downloaded asset seems to be deleted immediately so the check in line 220

```
if (!is_file($imageSourcePath)) {
    throw new FsObjectNotFoundException("The file \"{$asset->getFilename()}\" does not exist.");
}
```

throw the exception above. When we disable `maxCachedCloudImageSize` in the config, everything works fine and the transformations were generated.

As we can see in the file history, there were changes in 4.5.11. Maybe this is a new issue introduced with these changes.

### Steps to reproduce

1. Set `maxCachedCloudImageSize` to `0`.
2. Have a `asset.getUrl('transformation')` in a template.
3. Render the template.

### Expected behavior

Generate asset transformation.

### Actual behavior

Throws exception.

### Craft CMS version

4.5.13

### PHP version

8.1.18

### Installed plugins and versions

-

i-just:
Hi, thanks for getting in touch and for all the info! 
I can replicate this if I also have `generateTransformsBeforePageLoad` set to `true`. 

[...]

PR #14110, opened 2024-01-05
adjust when we delete file for remove volume when generating transforms

i-just:
### Description
If you have a volume with a remote filesystem (e.g. AWS S3) and you have `maxCachedCloudImageSize` set to `0` and `generateTransformsBeforePageLoad` set to `true` when generating transform not in the control panel, the downloaded file was being deleted before the transform was generated.

Since 4.5.11, the `FileHelper::deleteFileAfterRequest()` calls `ApplicationTrait::onAfterRequest()`, and in the case of generating transforms (not in the CP), the application state at this point is already `STATE_SENDING_RESPONSE` causing the file to be deleted immediately.

### Related issues
#14100

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Fixed a bug where the “Save and add another” element action wasn’t redirecting to a new element edit page, when editing an unpublished draft.
 - Fixed a bug where `craft\helpers\DateTimeHelper::nextYear()` and `lastYear()` weren’t returning the correct dates. ([#14109](https://github.com/craftcms/cms/issues/14109))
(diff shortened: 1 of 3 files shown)


brandonkelly:
@i-just `generateTransform()` won’t know if that’s the _last_ time that the current source image is going to be needed. (There could be multiple transforms being generated for the same asset in that request.)

This might be a simpler fix: https://github.com/craftcms/cms/pull/14110/commits/3651a4216b3c04d6223b507d7e0a275f58a00abe. Can you test and verify it works?

i-just:
@brandonkelly, your solution works great.

PR #14115, opened 2024-01-09
Live conditional fields without autosave

brandonkelly:
### Description

Adds live conditional field support to all element editors, even if drafts/autosaves aren’t supported by the element type (or it’s been disabled via the `autosaveDrafts` config setting).

That includes full edit pages for assets, as well as editor slideouts for assets, tags, and users.

In Craft 5, full edit pages for users will also be included, as they now use the element editor.

### Related issues

- #8099
- #13481


src/controllers/ElementsController.php
```diff
@@ -1417,63 +1417,12 @@ public function actionSaveDraft(): ?Response
...
-                        'id' => $tab->getId(),
-                        'elements' => $elementInfo,
-                    ];
-                }
-
-                $tabs = $form->getTabMenu();
-                if (count($tabs) > 1) {
-                    $selectedTab = isset($tabs[$this->_selectedTab]) ? $this->_selectedTab : null;
-                    $tabHtml = $view->namespaceInputs(fn() => $view->renderTemplate('_includes/tabs.twig', [
-                        'tabs' => $tabs,
```
mezarfatosh, line 1461: edit

mezarfatosh (changes requested):
github.com

PR #14126, opened 2024-01-11
Feature/cms 1221 add a search input to the fields and entry types pages

i-just:
### Description
Adds searching capability to the vue AdminTable for both Data and API modes. 

- the search bar will now show if the `search` property is set to `true` (regardless of whether we’re in Data or API mode)
- if in Data mode, rows will be shown/hidden via adding/removing a `hidden` class depending on whether a match was found in that row
- if in the API mode, search is handled server-side and results are provided back via the `tableDataEndpoint` 
- adds the `searchClear` string property that defines the copy for the clear button (defaults to “Clear”)
- adds the `noSearchResults` sting property that defines the copy to show when no results are found when searching (defaults to “No results.”)
- adds the `searchParams` array property, which is only used in the Data mode; in it, you can define which row attributes are to be searched through (defaults to an empty array)
- switches Settings > Fields and Settings > Entry Types pages from using Data mode to API mode.

Used in the Control Panel on the Settings > Fields and Settings > Entry Types pages.

### Related issues
[cms-1221](https://linear.app/craftcms/issue/CMS-1221/add-a-search-input-to-the-fields-and-entry-types-pages)


packages/craftcms-vue/admintable/App.vue
```diff
@@ -23,19 +23,24 @@
...
-        >
+        <div v-if="search" class="flex-grow texticon search icon clearable">
+          <span class="texticon-icon search icon" aria-hidden="true"></span>
           <input
             class="text fullwidth"
             type="text"
             autocomplete="off"
             :placeholder="searchPlaceholderText"
-            v-model="searchTerm"
-            @input="handleSearch"
```
nfourtythree, line 36: @i-just what was the reason behind these lines needing to change? the `v-model` attribute  and `input` event should handle everything you need so you don't have to pass the value manually.

i-just: As per our chat: 
- I have synced all properties between `packages/craftcms-vue/admintable/App.vue` and `src/web/assets/admintable/src/main.js` (and double-checked that they’re all in the readme file), and I’ve put them all in the alphabetical order; also added a comment to the `App.vue` so that it’s clear that we need to do this (at least for now)
- I have switched back to `v-model` and `@input`
- in terms of `rows` vs `this.$refs.vuetable.tableData` - that’s proving to be a bit more complicated than I expected; not giving up, but I need to jump back to something else first

Issue #14127, opened 2024-01-11
[4.x]: Checkbox field does not show selected option with value '0' in list view in the control panel


lineke:
### Description
I created a field with checkboxes for all the days of the week, with values from 0 to 6 since this was easiest to deal with from javascript. However, when I try to display the selected fieldoptions in the list view, it does not show "Sunday" as selected because it has value 0. If we switch this to 7 for example, it works but then I have to work around this in the javascript code.

### Steps to reproduce

1. Create a checkbox fields with options and values for all days of the week, starting with Sunday (value 0) to Saturday (value 6).
2. Add this field to an entrytype
3. Create a new entry with this entrytype and select multiple days, for example: saturday and sunday
4. Make this field visible in the list view
5. Check to see it only shows 'saturday' and not sunday. Even when you select other days as well, it will not show sunday

### Expected behavior

It should show sunday in the list view as well

### Actual behavior

It does not show sunday

### Craft CMS version

4.5.14 (but also does not work with 4.6.0)

### PHP version

8.0.30

### Database type and version

Mysql 8

### Installed plugins and versions

-

i-just:
Hi, thanks for reporting! I raised a PR for this.

lineke:
Thanks for the quick response and fix!

brandonkelly:
Craft 4.6.1 is out with the fix for this (as well as a related fix for the fields’ condition rules – #14143). Thanks again for reporting!

PR #14128, opened 2024-01-11
check if an option has a value

i-just:
### Description
Adds `hasValue()` method to the `BaseOptionsField`, which should be used to check if selected value(s) have... value (are not `null` and are not an empty string).

element index table with entry that has the following values selected for checkboxes and multi-select: `'0'`, `'1'`, '`2`', `'false'`, `'true'`, `'someString'` and `'0'` for the dropdown.

before:
Screenshot 2024-01-11 at 14 52 40

and after:
Screenshot 2024-01-11 at 14 52 52

### Related issues
#14127

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - Fixed a bug where `craft\helpers\Db::prepareValueForDb()` wasn’t converting objects to arrays for JSON columns.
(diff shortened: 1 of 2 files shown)


brandonkelly:
Looks like something `isValueEmpty()` could be used for ;)

i-just:
I had my reasons not to use `isValueEmpty()`, but I don’t think my justification is strong enough, so I tweaked `isValueEmpty()` and used it instead.

Issue #13341, opened 2023-06-21
[4.x]: Craft sets the timezone too late for localised logs (I think!)


bossanova808:
### Description

Goal: I am trying to get my Craft log files to log in local time rather than UTC time.  This makes it drastically easier to follow logs.  I know not everyone will want this, of course - e.g. with auto log procesing things, might require UTC time.  But I suspect a lot of folks would find it helpful.

My own modules log in local time via a self-built logger write to file thing (which I may move to Monolog later)...which further increases the disparity as I have one set of logs in UTC, and others in local.  I suppose I could convert my own logger to use UTC time to at least have them consistent with each other, but it's just so much easier for me to look at e.g. the local time a Commerce order came in and check against times in the same timezones in the log for what might have caused a warning/error.

Craft already has a general setting for timezone (I use `CRAFT_TIMEZONE=Australia/Melbourne` - and Craft seems to respect this basically everywhere, but it seems like it's happening a bit late in the bootstrapping for the Monolog logger to get set up.

I'd have thought Yii (or PHP) would fall back to the system timezone (which is set correctly), but apparently not:

![image](https://github.com/craftcms/cms/assets/731309/d7ce7138-c878-45c9-89a8-87883f85b613)

In solving this, I worked out that setting the Yii timezone in app.php results in the change I want:

`'timeZone' => App::env('CRAFT_TIMEZONE'),`

...but I am nervous about possible side effects.  

Unfortunately, I was not able to work out a method of doing the equivalent on just the Monolog component via config - which would have been my preferred option.

So - I am running with the app.php change, and seems ok so far, but would like to get the Craft official view on this.

### Craft CMS version

4.4.14

### PHP version

8.2

### Operating system and version

Ubuntu 22.04 (Forge)

### Database type and version

MySQL 8

### Image driver and version

not relevant

### Installed plugins and versions

not relevant

timkelty:
@bossanova808 setting `timeZone` in `app.php` should be a perfectly acceptable solution.

Setting `timeZone` in `app.php`, `date.timezone` in your `php.ini`, and `timeZone` `config.php` in [Craft's `config.php`](https://github.com/craftcms/cms/blob/4.4/src/config/GeneralConfig.php#L2785) all have the same effect: they set [PHP's internal default timezone](https://www.php.net/manual/en/function.date-default-timezone-set.php).

[...]

PR #14130, opened 2024-01-12
Add event when creating log target

timkelty:
### Description
Allows early access to our log targets (and their loggers). This allows configuring the Monolog logger without having to pass every supported prop down through our dispatcher.

### Related issues
- Fixes https://github.com/craftcms/cms/issues/13341
- https://linear.app/craftcms/issue/CMS-676/monolog-direct-configuration-get-current-logger

### Example usage

Changing the timezone use by Monolog via `app.php`:

```
<?php
return [
    'components' => [
            'log' => [
                'class' => \craft\log\Dispatcher::class,
                'on createTarget' => function(\craft\events\LogTargetEvent $event) {
                    $event->target->getLogger()->setTimezone(new DateTimeZone('America/Detroit'));
                }
            ],
    ],
];
```

This is better solution than `\craft\log\Dispatcher::$monologTargetConfig`, but it's probably to late to drop that for v5. I did mark is as deprecated.

### DIFF
--- a/src/events/LogTargetEvent.php
+++ b/src/events/LogTargetEvent.php
@@ -1,0 +1,22 @@
+<?php
+/**
+ * @link https://craftcms.com/
+ * @copyright Copyright (c) Pixel & Tonic, Inc.
+ * @license https://craftcms.github.io/license/
+ */
+
+namespace craft\events;
+
+use craft\base\Event;
+use craft\log\MonologTarget;
+
+/**
+ * Log target event class.
+ *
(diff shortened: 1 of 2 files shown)


brandonkelly:
Rather than add the new event, I pulled in your `Craft::createObject()` changes in bec2776f28bfd28783f40d159bc4df26fafb698b, so now targets can be modified using the DI container.

timkelty:
Yeah that makes sense!

PR #14136, opened 2024-01-15
Feature/cms 1221 add a search input to the fields and entry types pages api mode

i-just:
### Description
An alternative solution to [14126](https://github.com/craftcms/cms/pull/14126).

Adds searching capability to the vue AdminTable when in API mode. 
- the search bar will now show if the search property is set to true
- adds the `searchClear` sting property that defines the copy for the clear button (defaults to “Clear”).
- adds the `noSearchResults` sting property that defines the copy to show when no results are found when searching.
- switches Settings > Fields and Settings > Entry Types pages from using Data mode to API mode.

Used in the Control Panel on the Settings > Fields and Settings > Entry Types pages.

### Related issues
[cms-1221](https://linear.app/craftcms/issue/CMS-1221/add-a-search-input-to-the-fields-and-entry-types-pages)


i-just:
Closing in favour of [#14126](https://github.com/craftcms/cms/pull/14126).

Issue #14138, opened 2024-01-15
[5.x]: Asset thumbnails are stacked when inside asset fields with "display as large thumbnail" mode 


thupsi:
### Description
Asset thumbnails are stacked when inside asset fields with "display as large thumbnail" mode 

### Steps to reproduce

1. Create an asset field and select the  "display as large thumbnail" mode"
2. Assign this field to an entry type
3. Create an entry with this entry type.
4. Add some assets to this field.

### Expected behavior
The asset thumbnails should fit in rows according to available space.

### Actual behavior
The asset thubmnails display stacked (one below the other).

![image](https://github.com/craftcms/cms/assets/19857599/ab59c843-45eb-4199-b46b-c0f68a2ee029)

### Craft CMS version

Craft CMS 5.0.0-alpha.5

### Installed plugins and versions

-

i-just:
Hi, thanks for reporting! I raised a PR to adjust this.

brandonkelly:
Craft 5.0.0-alpha.7 is out now with a fix for this. Thanks again!

PR #14144, opened 2024-01-16
Assets field - display large thumbs in a row

i-just:
### Description
For the Assets field, when View Mode is set to "Large Thumbnails", display the assets in a row, not in a column.

Before:
Screenshot 2024-01-16 at 14 10 23

After:
Screenshot 2024-01-16 at 14 07 12

### Related issues
#14138


brandonkelly:
Fixed in a simpler way in 950b1fbe9125b07a9989b523079320e6414be19d

PR #14165, opened 2024-01-19
allow duplicating of root element

i-just:
### Description
There was an issue preventing the root element (e.g. entry on the main entry index page) from being duplicated.

Checking if the entry has an owner and only setting the `primaryOwner` to be passed to the `duplicateElement()` `newAttributes` param fixes it.

### Related issues
n/a


brandonkelly:
`setPrimaryOwner()` is supposed to accept `null`, so the bug was within that. Fixed in c49109910633a7a0be2e7cd1d476295c97863fbd.

PR #14172, opened 2024-01-22
Feature/cms 1233 multiple tab support for inline matrix blocks

i-just:
### Description
If a matrix field with a view mode of “as inline-editable blocks” has multiple tabs, show them all.

- if there’s more than one tab - show all tabs
- in static mode - show the title bar with the entry type and with working tabs

Screenshot 2024-01-22 at 16 33 34
Screenshot 2024-01-22 at 16 35 11

### Related issues
[cms-1233](https://linear.app/craftcms/issue/CMS-1233/multiple-tab-support-for-inline-matrix-blocks)


gcamacho079:
@brandonkelly @i-just just audited but saw there are more updates since I pulled this down; will add issues to Linear if any remain. 👍🏼

PR #14187, opened 2024-01-24
Color-coded entry types

brandonkelly:
Adds a “Color” setting to entry types.

When an entry type has a color selected, entries of that type will be filled in with that color within cards, chips, and inline-editable Matrix blocks.

Matrix fields set to the “inline-editable blocks” view mode have also been redesigned a bit:

- The titlebar now includes the entry type icon, if it has one
- Matrix blocks with only one entry type no longer show the entry type icon/label in their titlebars.
- Block bodies now have a white background color, to improve the visibility of nested chips/cards/Matrix blocks.
- The “Add” button is now always displayed as a menu, unless there’s only one block type. No more button group listing all types. (This way it’s always consistent regardless of viewport size, and so we can show the block type icons, which would have looked awkward in button group form.)

An entry type’s settings, including the new Color setting

A Matrix field’s Entry Types setting, showing several selected color-coded entry types

A Matrix field set to the inline blocks view mode, where the blocks are each color-coded

A Matrix field set to the cards view mode, where the blocks are each color-coded

An expanded Entry Type menu, listing entry type options with color-coded icons


davidhellmann:
Like the stuff! Question, can we bring in the color and the Icon here (entry type overview) too: 
image

Would be great.

brandonkelly:
5.0.0-alpha.11 is now tagged with that change.

brandonkelly:
@alexanderloewe are you just basing that on the screenshot I posted? I shrunk them down a bit to match global nav icon size in 996e73d81de04fa16c6cdd638a32c496b8830632.

PR #14213, opened 2024-01-26
improved matrix tab height at text only zoom levels

i-just:
### Description
Improves matrix tab height for text-only zoom levels.

Before:
Screenshot 2024-01-26 at 08 17 29

After:
Screenshot 2024-01-26 at 08 17 07

### Related issues
[acc-353](https://linear.app/craftcms/issue/ACC-353/tab-label-containers-should-use-a-responsive-height)


gcamacho079 (approved):
Looks great @i-just!

PR #14214, opened 2024-01-26
added button role, removed tabindex

i-just:
### Description
Tabs in a disclosure menu have a `button` role now.

(Tabs that are anchor tags in the disclosure menu already used the `activate` event listener.)

There's one item left on the list, but I need to double-check that with Lupe - hence, this is only a draft for now.

### Related issues
[acc-355](https://linear.app/craftcms/issue/ACC-355/tabs-inside-disclosure-menu-should-use-button-roles)


i-just:
Double-checked. All good. Can be reviewed :)

PR #14215, opened 2024-01-26
match aria values

i-just:
### Description
Matches aria values linking tabs.

- Tab’s `aria-controls` value now matches the corresponding tabpanel's `id`.
- Tabpanel’s `aria-labelledby` value now matches the corresponding tab’s `id`.

### Related issues
[acc-352](https://linear.app/craftcms/issue/ACC-352/when-viewing-inline-editable-matrix-blocks-with-tabs-aria-values)


gcamacho079 (approved):
This does the trick!

PR #14216, opened 2024-01-26
adjusted .card-grid's list and .card heights to please safari

i-just:
### Description
I adjusted the height of `.card-grid > li` and `.card-grid > li  > .card` to please Safari (it was and still is looking as expected in FF and Chrome).

Before:
Screenshot 2024-01-26 at 11 31 06

After:
Screenshot 2024-01-26 at 11 30 50

### Related issues
n/a (slack chat)


brandonkelly:
Fixed this a different way 🙃  007fb97261f356be9d9ce208a061a3f793e6bcc9

PR #14222, opened 2024-01-26
Add margin around chip/card link so that focus outline is not clipped

gcamacho079:
### Description
Fixes hidden focus outline caused by `overflow: hidden` on parent container


brandonkelly:
Resolved via 0fbe04b1e9966d77180854f5381ac7e3d0ab31dc by using the inner focus ring style.

Issue #14221, opened 2024-01-26
[5.x]: Error renaming plaintext handles in globals


mei-lu:
### Steps to reproduce

1. Create a new global and add a plaintext field to it
2. Try to rename the handle of the plaintext field
3. Save

### Expected behavior

Save without error

### Actual behavior

Invalid field handle: myNewHandle

### Craft CMS version

Craft Pro 5.0.0-alpha.8

### Installed plugins and versions

-

i-just:
Hi, thanks for reporting! I raised a PR for this.

brandonkelly:
Craft 5.0.0-alpha.9 was just released with a fix for this. Thanks again!

Issue #14166, opened 2024-01-19
[5.0.0-alpha.6]: Invalid field handle on re-install


WHITE-developer:
### Description
When trying to setup a clean database with existing project-config, you get an `Exception: Invalid field handle: {customFieldHandle}` when you have used fields where you have renamed the handle inside a fieldlayout.

https://github.com/craftcms/cms/blob/5.0/src/base/Element.php#L5834 here the `CustomFieldBehavior::$fieldHandles` is only returning the field handles as defined in settings -> fields, not the handles you have defined in the field layout editors.

### Steps to reproduce

1. setup a clean Craft 5 project
2. Create a custom field and map this to an entry type
3. start with a clean DB
4. run `./craft install`
5. fill in the required info 
6. when applying the existing project config the Exception will be thrown

### Expected behavior
Existing project config should be getting applied

### Actual behavior
An Invalid field handle exception is thrown

### Craft CMS version

5.0.0-alpha.6

### PHP version

8.3.2

### Operating system and version

Ubuntu 22.04

### Database type and version

8.0.35

### Installed plugins and versions

-

brandonkelly:
I’m not able to reproduce that. Can you please send your Composer and project config files to support@craftcms.com?

lukas-jansen:
Retried it by first rebuilding config files but still got it,
So started from a completely clean project with no plugins installed and just 1 custom field, but still got it.

When not renaming/rehandling the custom field it all works fine

I've send my project yamls and composer files by mail

i-just:
Thanks for all the info and the files. I can replicate this issue. It only happens if the field with the changed handle is part of an entry type that is assigned to a `single` type section. I'm working on a solution for this.

PR #14233, opened 2024-01-29
getting field with the changed handle before CustomFieldBehavior knows about it

i-just:
### Description
An error occurred when applying project config with a `single` section type and an entry type with a field whose handle has been renamed, and when saving a global with a field with a newly renamed handle. 

This was caused by the field not yet loaded into the `CustomFieldBehavior::$fieldHandles`.

### Related issues
#14166 
#14221

### DIFF
--- a/src/base/Element.php
+++ b/src/base/Element.php
@@ -5886,9 +5886,18 @@ JS,
      */
     protected function fieldByHandle(string $handle): ?FieldInterface
     {
+        $fieldWithNewlyChangedHandle = false;
+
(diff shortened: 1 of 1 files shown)


brandonkelly:
Solved in a simpler way via e6d513356da00953fdfd07f02a9403a2804a7843

(This is based on the same thing `saveField()` does for new field handles here:

https://github.com/craftcms/cms/blob/e6d513356da00953fdfd07f02a9403a2804a7843/src/services/Fields.php#L1240-L1241

PR #14235, opened 2024-01-29
Feature/cms 1234 add env var support to sites language settings

i-just:
### Description
Add environment variable support to sites’ language settings and the user locale settings.

It's still a work in progress.

### Related issues
[cms-1234](https://linear.app/craftcms/issue/CMS-1234/add-env-var-support-to-sites-language-settings)


johnwbaxter:
@brandonkelly I don't suppose there is any chance of this being backported to Craft v4 too is there? We're currently having to do some mad stuff to do this to sites in Craft 4!

brandonkelly:
@johnwbaxter There’s a small chance this introduces a BC break, so I wasn’t comfortable doing it for 4.x.

PR #14271, opened 2024-02-02
Remove `TKey` template keyword in `ElementCollection generics

nfourtythree:
### Description
Currently, we are specifying two generics keywords, however we are never needing to specify the first one (`TKey`) anywhere in the code.

Removing it simplifies anything trying to extend or implement the `ElementCollection` class as you do not need to specify both values.

> When using `@extends` or `@var` tags they can take multiple arguments, separated by comma - in the same order the types are defined in the class

This change means we can go from `ElementCollection<array-key, Address>` to simply `ElementCollection<Address>` as we are now only specifying one `@template` keyword.

### DIFF
--- a/src/elements/ElementCollection.php
+++ b/src/elements/ElementCollection.php
@@ -14,9 +14,8 @@ use Illuminate\Support\Collection;
 /**
  * ElementCollection represents a collection of elements.
  *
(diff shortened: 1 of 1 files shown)


timkelty (approved):
I was there!

PR #14308, opened 2024-02-06
Update name of sidebar navigation on account page

gcamacho079:


brandonkelly:
@gcamacho079 That function is also called for other users’ accounts

gcamacho079:
@brandonkelly thanks for pointing that out! I updated it to "Account"

PR #14322, opened 2024-02-07
commerceguys/addressing v2

brandonkelly:
### Description

Updates `commerceguys/addressing` to v2, and adds an “Address Line 3” address field.

### Related issues

- #14318


bojanz:
Great work!

A possible followup might be to introduce usage of the [SubdivisionUpdater](https://github.com/commerceguys/addressing/blob/master/src/Subdivision/SubdivisionUpdater.php#L14) in the record class.

That would auto-update administrative areas both for the 2.x ISO changes and other future renamings (e.g. ISO codes changed for a few Indian provinces, which we'll release in v2.2.0).

The Drupal version of the code looks like this:
```
public function setValue($values, $notify = TRUE) {
    // If a subdivision ID has changed, allow it to be remapped.
    // Primarily covers the 8.x-1.x => 2.0.x updates, where subdivisions
    // started being keyed by ISO code where available.
    if (isset($values['country_code'])) {
      if (isset($values['administrative_area'])) {
        $values['administrative_area'] = SubdivisionUpdater::updateValue($values['country_code'], $values['administrative_area']);
      }
      // Andorra is the only country with remapped localities.
      if ($values['country_code'] == 'AD' && isset($values['locality'])) {
        $values['locality'] = SubdivisionUpdater::updateValue($values['country_code'], $values['locality']);
      }
    }

    parent::setValue($values, $notify);
  }
```

brandonkelly:
Thanks for the suggestion @bojanz :) We’ve just done that here: #14386

WHITE-developer:
Any chance this will come to Craft 4 also?

brandonkelly:
@WHITE-developer We only perform major updates of dependencies for major Craft releases.

Issue #14329, opened 2024-02-08
[4.x]: Duplicate file name issues are not displayed in modal windows


MoritzLost:
### Description

I'm seeing an issue where errors on the `filename` field are not reported when an asset is being edited in a modal window.

When trying to change the filename to a filename that already exists in the same folder, we get an error message below the filename input:

![Screenshot 2024-02-08 at 15 13 20](https://github.com/craftcms/cms/assets/10146880/454c75b7-32d2-4f4b-8892-37a8012bf54e)

But when editing the same asset in a modal window from an entry form, the validation error is not displayed after trying to save:

![Screenshot 2024-02-08 at 15 13 08](https://github.com/craftcms/cms/assets/10146880/c6d9cd5e-671d-41fe-a953-e94731e2f0e9)

I've checked the response in the network request, it correctly includes a `newLocation` error and an `errorSummary` with an appropriate message. Only they're not getting displayed, all we get is the generic **Couldn’t save asset.** message on the bottom left.

### Steps to reproduce

1. Open an asset in a modal window from an entry form.
2. Change the filename to a filename that's already used by another asset in the same folder and try to save.

### Expected behavior

All validation messages should be displayed in a modal context.

### Actual behavior

The validation error message is missing.

### Craft CMS version

4.7.1

### PHP version

8.2

### Installed plugins and versions

-

i-just:
Hi, thanks for reporting! I raised a PR for this.

(Also, error reporting in slideouts has been further improved in v5.)

brandonkelly:
Craft 4.7.3 is out with the fix for this. Thanks again!

PR #14337, opened 2024-02-09
report newLocation error in asset slideout

i-just:
### Description
Report `newLocation` error when editing asset in a slideout.

Screenshot 2024-02-09 at 07 57 28

### Related issues
#14329

### DIFF
--- a/src/elements/Asset.php
+++ b/src/elements/Asset.php
@@ -2666,6 +2666,7 @@ JS;
         return implode("\n", [
             Cp::textFieldHtml([
                 'label' => Craft::t('app', 'Filename'),
+                'attribute' => 'newLocation',
                 'id' => 'new-filename',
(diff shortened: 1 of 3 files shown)


src/elements/Asset.php
```diff
@@ -2674,6 +2674,9 @@ protected function metaFieldsHtml(bool $static): string
                 'required' => true,
                 'class' => ['text', 'filename'],
                 'disabled' => $static,
+                'data' => [
+                    'attribute' => 'newLocation',
+                ],
```
brandonkelly, line 2679: Couldn’t this have just been a top-level `attribute` config, and then no change to `Cp::fieldHtml()` needed?

i-just: Yes, it could. Pushed an update.

PR #14346, opened 2024-02-09
[5.x]: Remove `error` key from response data

bencroker:
This PR fulfils the `TODO` by removing the `error` key from the response data.

Note that a `TODO: review for v5` still exists in the file.


brandonkelly:
Thanks!

PR #14351, opened 2024-02-09
Updates global sidebar to use rem instead of px

gcamacho079:
### Description
When users increase font size by 200%, the badge icon overlaps the navigation icon. This PR updates sizes and widths to use `rem` so that the notification icon doesn't overlap the icon at 200% zoom.

In addition, it changes the breakpoints for the layout to use `rem` instead of `px` widths so that font size zoom triggers a shift to the mobile layout, increasing the screen real estate.


gcamacho079:
@brianjhanson was hoping for your input and a once-over on this one. I noticed that when browser/font zoom triggers the mobile sidebar layout, the `data-sidebar` attribute on `<body>` may still be set to collapsed, so the labels aren’t visible. I was thinking we'd also need to call the expand/collapse methods in the `toggleNav` function - thoughts?

brianjhanson:
@gcamacho079 fixed in [f62e9cb](https://github.com/craftcms/cms/pull/14351/commits/f62e9cb37222490c8302a74b789161c2f00b3488). The rule governing the grid of `nav-item`s wasn't in a media query which caused things to get out of whack.

gcamacho079:
> @gcamacho079 fixed in [f62e9cb](https://github.com/craftcms/cms/pull/14351/commits/f62e9cb37222490c8302a74b789161c2f00b3488). The rule governing the grid of `nav-item`s wasn't in a media query which caused things to get out of whack.

Thank you so much, Brian! 👍🏼

PR #14353, opened 2024-02-10
Change generateUrl signature

timkelty:
### Description
`\craft\helpers\Assets::generateUrl` took a first argument of `BaseFsInterface`, however, if you passed an FS, you could end up with the wrong URL if your asset volume had a subpath.

Furthermore, I don't see any need to pass a volume either, since we're required to pass the asset. So, I removed the param entirely. If you want me to scale that back, I can leave it and just change the typing to `\craft\models\Volume`. I dug around a bit and I've get to see a plugin passing anything but the fs or volume derived from the asset, and I'm not sure why you'd want to.


src/helpers/Assets.php
```diff
@@ -81,19 +81,19 @@ public static function tempFilePath(string $extension = 'tmp'): string
...
      * @throws InvalidConfigException if the asset doesn’t have a filename.
      */
-    public static function generateUrl(BaseFsInterface $fs, Asset $asset, ?string $uri = null, ?DateTime $dateUpdated = null): string
+    public static function generateUrl(Asset $asset, ?string $uri = null, ?DateTime $dateUpdated = null): string
     {
+        $volume = $asset->getVolume();
         $pathParts = explode('/', $asset->folderPath . ($uri ?? $asset->getFilename()));
         $path = implode('/', array_map('rawurlencode', $pathParts));
-        $rootUrl = $fs->getRootUrl() ?? '';
-        $url = ($rootUrl !== '' ? StringHelper::ensureRight($rootUrl, '/') : '') . $path;
```
timkelty, line 96: `Volume::getRootUrl` already ensures trailing slash

PR #14363, opened 2024-02-12
Remove `@property $path`

bencroker:
Removes `@property $path`, which does not seem to exist (possibly confused with the `$pathInfo` property that exists in the parent class).

This should be applied to the `5.0` branch too.


brandonkelly:
Thanks! I think that’s a holdover from Craft 2 🤦🏻

Issue #14370, opened 2024-02-13
[4.x]: Creating a number field with decimal points over 30 causes a database error


peteluffman:
### Description

Creating a number field with decimal points over 30 causes a database error that bricks the site and admin because a DB column is missing.

### Steps to reproduce

1. Create and save new number field 
2. Add field to an entry type (not 100% sure if this step is required)
3. Amend decimal point above 30 (I used 99)
4. Save field

### Expected behavior

Field validation should prevent the SQL query being run.

### Actual behavior

SQL error saying the column could not be created with a length above 30. 

Subsequently, the admin and public facing website becomes fully broken because it can't find the field column in the content table. To recover, I had to manually run a SQL query to create the missing column with a decimal type.

 

### Craft CMS version

4.7.1

### PHP version

8.1.26

### Database type and version

MySQL 8.0.33

### Installed plugins and versions

-

i-just:
Hi, thanks for reporting! I raised a PR for this.

brandonkelly:
Craft 4.7.3 is out with the fix for this. Thanks again!

PR #14371, opened 2024-02-13
number decimals validation rule for mysql

i-just:
### Description
Adds `max` validation to the `Decimal Points` value for the `Number` field so we don’t trigger an SQL error on field save.

This is only an issue in MySQL and won’t be a problem in Craft v5, so the rule is set to only kick in for MySQL and below v5.

### Related issues
#14370


brandonkelly:
Removed the Craft 5 check, and deleted the change in the `5.0` merge commit (25daf8dbc55fb599cb64b712d53d3c7d7ea2c58c).

PR #14386, opened 2024-02-14
Feature/cms 1261 use subdivisionupdater for addresses

i-just:
### Description
As per the suggestion from https://github.com/craftcms/cms/pull/14322, I've hooked up `SubdivisionUpdater`, so that the subdivision IDs get remapped to their ISO code where available.

Tweaked what parents are passed to the `Cp::_subdivisionField()` when using `Cp::addressFieldsHtml()` based on whether they're visible or not.

### Related issues
[cms-1261](https://linear.app/craftcms/issue/CMS-1261/use-subdivisionupdater-for-addresses)


WHITE-developer:
Any chance this will come to Craft 4 also?

PR #14390, opened 2024-02-15
Admin tables adhere to appending `bodyHtml` and `headHtml` when in API mode

nfourtythree:


lukeholder:
This is working great for me. We will need this merged in before the Commerce 5 beta.

brandonkelly (changes requested):
`Craft.appendHeadHtml()` and `Craft.appendBodyHtml()` are `async` functions, so they need to be called with `await`.

lukeholder:
@brandonkelly updates work for me.

PR #14398, opened 2024-02-16
Don't write to .env file if ephemeral

timkelty:
### Description
Allows you to run craft setup in an ephemeral environment


brandonkelly:
What’s the point of running `setup` in the first place then?

timkelty:
When running `craft install`, if ephemeral, you need to make sure you have an ID set, and it isn't the default.

PR #14399, opened 2024-02-16
Fix `@since` version

bencroker:
Fixes the version number, _since_ this class was added in 4.4.0.
https://github.com/craftcms/cms/blob/develop/CHANGELOG.md#440---2023-03-08


brandonkelly:
Good catch, thanks!

PR #14404, opened 2024-02-16
Fixes a type comparison that was returning true unexpectedly

lukeholder:
### Description

Fixes a type comparison that was returning true unexpectedly.

Reproduction:

```twig
{% set mytruval = true %}

{% if mytruval == 'auto' %}
    This will show up.
{% endif %}
```

Twig 3.4.3


brandonkelly:
Thanks! Cherry-picked into `develop`.

PR #14413, opened 2024-02-17
Remove `align-items` CSS property

bencroker:
This PR takes a somewhat heavy-handed approach to fixing the vertical alignment issue created by the markup generated by `Cp::elementChipHtml` in Craft 5.0.0-beta.2, that is affecting the data heading. 

It may be better to target a more specific class, as this may affect other metadata elements, but this is what it looks like...

Before:

Screenshot 2024-02-16 at 19 27 17

And after:

Screenshot 2024-02-16 at 19 27 25

### DIFF
--- a/src/web/assets/cp/src/css/_main.scss
+++ b/src/web/assets/cp/src/css/_main.scss
@@ -4836,7 +4836,6 @@ ul.tree,
       color: var(--medium-text-color);
 
       & > .data {
-        align-items: baseline;
         min-height: auto;
 
         & > .heading,


brandonkelly:
Thanks! I think the heavy-handed approach is probably fine. Checked the “Used by” meta value on entry and field edit pages, and I’d say it’s an improvement there as well.

An entry type’s usages meta

brandonkelly:
Craft 5.0.0-beta.3 is out with that fix.

PR #14414, opened 2024-02-17
Add strict Twig comparison

bencroker:
This PR replaces the loose `==` comparison with a strict `is same as()` comparison (the equivalent of `===` in PHP). This is necessary because in Twig, `true == 'auto'` evaluates to `true`.

So in this example:

```twig
{% extends '_layouts/elementindex' %}

{% set showSiteMenu = true %}
```

`showSiteMenu` is converted to `auto`.

After applying this PR, `showSiteMenu` remains `true`.

Reference: https://twig.symfony.com/doc/3.x/tests/sameas.html

If merged, this PR should be applied to the `5.0` branch as well.


brandonkelly:
@lukeholder beat you to it :) #14404

bencroker:
Yes he did!

brandonkelly:
4.7.4 is out with that fix.

Issue #14427, opened 2024-02-19
[5.0.0-beta.2]: Sort order of matrix blocks (card view) not retained after edit in slideout


wsydney76:
### Description

Matrix field with View Mode: Cards

![image](https://github.com/craftcms/cms/assets/60010597/525f3351-d2b3-48c7-bd1f-f63bfa7b0087)

After editing a matrix block via slideout, it looks correct at first:

![image](https://github.com/craftcms/cms/assets/60010597/d31082c7-92e3-422a-9dbc-172d64e8afc2)

After refreshing the page, the sort order has changed:

![image](https://github.com/craftcms/cms/assets/60010597/8fe2a5bf-045f-4267-b976-1cf8967ac636)

### Steps to reproduce

1. Have a matrix field with View mode: Cards
2. Create an entry with a number of blocks.
3. Edit and save matrix block in slideout. (No provisional draft created for the main entry)
4. Refresh page

### Expected behavior

Sort order is retained.

### Actual behavior

It is not.

### Craft CMS version

5.0.x-dev as 5.0.0-beta.2

### PHP version

8.2.13

### Operating system and version

Linux 5.15.133.1-microsoft-standard-WSL2

### Database type and version

MySQL 8.0.33

### Installed plugins and versions

-

i-just:
Hi, thanks for reporting! I raised a PR for this.

brandonkelly:
Craft 5.0.0-beta.3 is out with that fix. Thanks again!

PR #14430, opened 2024-02-19
when getting nested element, get its sort order too

i-just:
### Description
When a nested element was duplicated to a draft, the draft was created with the highest sort order +1. Then, when it was being applied, the sort order of the canonical element was set to another +1. So if we had a nested element with sort order `1` and another with sort order `2` and we edited the first one and saved it, the new sort order for that element was set to `4`, causing it to go to the bottom of the list.

If we get the sort order when initially getting the element and keep passing it around, the draft created via `duplicateElement` will have the same sort order as its canonical and then be applied with a correct sort order.

### Related issues
#14427

### DIFF
--- a/src/controllers/ElementsController.php
+++ b/src/controllers/ElementsController.php
@@ -13,6 +13,8 @@ use craft\base\ElementInterface;
 use craft\base\FieldLayoutComponent;
 use craft\behaviors\DraftBehavior;
 use craft\behaviors\RevisionBehavior;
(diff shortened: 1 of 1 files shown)


brandonkelly:
Fixed via 5cce1114c4a2b362936ab823d03840c3e6eb7092 (fetching the sort order from `beforeSave()` instead, and only if we actually need to save the ownership data).

PR #14432, opened 2024-02-19
Feature/collapsible details

brianjhanson:
Adds toggle to make all detail sidebars collapsible. 

@benjamindavid and I spent a good chunk of time experimenting with where to place the button, and in the end, this won out, but I'm happy to hear other ideas if there are some. 

https://github.com/craftcms/cms/assets/1843073/b17b034d-bc14-41f6-aad0-d71cad526c87


brianjhanson:
@gcamacho079 head's up, this one affects the front-end

brandonkelly:
@brianjhanson Seems like the button should either go on the very right-hand side of the toolbar area, or below the sidebar (like the global sidebar’s “Toggle Sidebar” button).

brianjhanson:
@brandonkelly updated. I went with to the right of the submit button. I was originally resistant to putting it up there because that space feels kind of sacred to me, but I agree that feels like the most logical place to put it. 

CleanShot 2024-02-22 at 10 48 57@2x

PR #14436, opened 2024-02-21
Feature/cms 1266 error summary in slideouts

i-just:
### Description
Adds the error summary to slideouts.

Unlike the full-page approach, where there’s one error summary area that shows above the tabs, in slideouts, we have an error summary per tab, and each of those only shows the errors in a given tab.

### Related issues
[cms-1266](https://linear.app/craftcms/issue/CMS-1266/error-summary-in-slideouts)


brianjhanson:
@gcamacho079 another one that impacts the front-end. 

CleanShot 2024-02-21 at 11 30 03

gcamacho079:
@i-just I tested this on JAWS/Chrome and NVDA/Firefox, and based on how these read the content (i.e. JAWS doesn’t read automatically, while NVDA does) I think there could be the potential that the visually-hidden text is missed unless we move it to the end of the heading. For example: "Found 3 errors _in this tab; 5 errors in 2 tabs_" where the italicized text would be visually-hidden.

I was also able to trigger the "Found 0 errors" message again by correcting an error in the current tab, while leaving one in another tab (screenshot attached). Also, it still shows an error icon for the current tab.

An error message says 'Found 0 errors' for the current tab, but the tab button shows an error icon and there's an error in another tab.

Based on those two things, I’m still wondering if we should just make the full message visible for everyone, as it makes it crystal clear from that single error box that the other tabs are the issue.

i-just:
Thanks @gcamacho079! 

I changed the heading to: `<h2>Found 1 error in this tab.<span class="visually-hidden">3 errors found in 3 tabs.</span></h2>`. 
I don’t have a strong preference on whether that second part should be visible to all or just hidden, so if you think it should all be visible, please LMK and I can easily change it.

This morning, I was able to reliably replicate that “0 errors” message (It was happening when you fix an error and then save immediately before autosave kicks in). Long story short, I have made a change to ensure we have the draft before attempting to do a full save, and that fixes this issue (and some other quirks).

If you have a sec, give it a quick test and LMK if you think any further changes are needed.

gcamacho079 (approved):

brandonkelly:
Works and looks great!!

PR #14515, opened 2024-03-02
Slightly-darker global sidebar

brandonkelly:
Slightly darkens the global sidebar, to give it a little bit of contrast with the rest of the page.

Screenshot of an Edit Entry page, showing how the sidebar has a slightly darker background than the rest of the page.


AndrewMac:
Thanks for taking the time to revisit this, Brandon. The lack of contrast never sat well with me, so I added a CSS override with the same slightly darker shade. I also found that the sidebar lacked scrollbars if it overflowed the viewport, so I had to add (although it did affect the width of the sidebar), along with CSS to make the global-sidebar__header sticky on scroll. Finally, I added a background color on li.nav-item.sel a:first-child, to make the current selection stand out, as I found it much too easy to miss the teensy black vertical bar that it currently uses … I did keep that, though, as it's an additional signal, but a much more visible selected state suits me much better.

AndrewMac:
Actually, to be fair, I just changed my mind and set the sidebar display to grid, so that I could put the scrolling on the global-sidebar__nav instead and keep the "toggle sidebar" control visible as well.

Issue #13715, opened 2023-09-19
[4.x]: JS Error preventing addition of a new Table field


rob-baker-ar:
### Description

A JavaScript error prevents the ability to define a new "Table" field in the CMS:

![image](https://github.com/craftcms/cms/assets/39988088/20353907-7394-4981-a552-f1777c96c31a)

### Steps to reproduce

1. In CMS > Settings > Fields > [ + New Field ] > New Field Screen
2. Select "Table" from the "Field Type" drop down
3. JS error in console: "VM1408:1 Uncaught TypeError: Craft.TableFieldSettings is not a constructor" (using Chrome `116.0.5845.188`)
4. Cannot use most of the controls on page to define the new field

### Expected behavior

No JS error and the field can be defined.

### Other detail

Clearing caches including "control panel resources" has no effect. This also happens when I disable all our module's custom JS / CSS that would normally be injected into the CMS.

### Craft CMS version

4.5.4

### PHP version

8.1

### Operating system and version

N/A

### Database type and version

N/A

### Image driver and version

N/A

### Installed plugins and versions

- Asset Rev `7.0.0`
- CP Field Inspect `1.4.4`
- Feed Me `5.2.0`
- Field Manager `3.0.7`
- Minify `4.0.0-beta.2`
- Redactor `3.0.4`
- Retcon `2.7.3`
- Retour Redirects `4.1.13`
- SEOMatic `4.0.31`
- Smith `2.0.0`
- Super Table `3.0.9`
- Twig Perversion `3.0.0`
- Twig Profiler `4.0.0-beta.1`
- Wordsmith `4.2.0`

**However, the issue also happens when all plugins are disabled.**

rob-baker-ar:
Upon further testing, it looks like I am getting similar JS errors when selecting most field types in the same drop down. 

It looks like it happens with any field that requires another JS file to be loaded. Looking at the network tab the files that the error claims are missing are in fact being loaded - just not being executed? Or not being execute in time - perhaps a race condition.

Whether it's just me getting this or everyone, I'm not sure.

rob-baker-ar:
This does not happen on a different Craft install we have on the exact same version. I have been checking the differences and on the install where this does not work, all static content (i.e. css / js / etc) is served via a different hostname (which is a subdomain of the main domain the control panel is served from). It is achieved by setting these general config properties: 

```php
[
    ...
    'resourceBasePath' => '[fs path to doc root of alternate hostname web server]',
[...]

PR #14526, opened 2024-03-05
[5.x]: Fix JS modules not loading correctly in the control panel

engram-design:
There's an issue for any field that contains a `<script type="module">` in its assets when loading its field settings. From a new field interface, selecting the field via dropdown will load the HTML/CSS/JS for the field asynchronously. This is notable with Vizy, Hyper, Icon Picker and others, where we use Vue and Vite to handle field settings.

This results in an error when trying to load a script that's a module - the browser can't handle it.

image

This seems to stem from this [change](https://github.com/craftcms/cms/commit/94ed5f9a2bdc80d3eb489bce76df1527d17f5504#diff-b9705a5250928f8dd378beb6ff7f2cc8cd708130275220eb153ee3278d2eb766R1873-R1882) where you fetch the script first via jQuery's `$.getScript()` function, but this doesn't work well with modules. In addition, you're only passing in an array of URLs to this function, whereas we need to know if the URL is a module or not.

I did toy with fetching all attributes of the `<script>` element and use that, but decided to keep it simple.

So firstly, we store the script type and src, and pass into `_loadScripts()`. We then have to swap out jQuery's `$.getScript()` function for our own, but I've tried to keep it almost 1-for-1.

Sorry for the noisy PR, but figured I should rename some variables for clarity. I also didn't touch `this._existingJs` because it's just comparing URLs, not loading them.

### DIFF
--- a/src/web/assets/cp/src/js/Craft.js
+++ b/src/web/assets/cp/src/js/Craft.js
@@ -1916,7 +1916,7 @@ $.extend(Craft, {
       return;
     }
 
-    const scripts = [];
+    const scriptUrls = [];
 
     const nodes = $.parseHTML(html.trim(), true).filter((node) => {
       if (node.nodeName === 'LINK' && node.href) {
@@ -1942,11 +1942,7 @@ $.extend(Craft, {
         }
 
         if (!this._existingJs.includes(node.src)) {
-          scripts.push({
-            type: node.type || 'script',
-            src: node.src,
-          });
-          
+          scriptUrls.push(node.src);
           this._existingJs.push(node.src);
         }
 
@@ -1957,39 +1953,29 @@ $.extend(Craft, {
       return true;
     });
 
-    await this._loadScripts(scripts);
+    await this._loadScripts(scriptUrls);
     $parent.append(nodes);
   },
 
-  _loadScripts: function (scripts) {
+  _loadScripts: function (urls) {
     return new Promise((resolve) => {
-      if (!scripts.length) {
+      if (!urls.length) {
         resolve();
         return;
       }
 
-      const script = scripts.shift();
+      const url = urls.shift();
+      $.ajaxSetup({cache: true});
 
-      const getScript = function(script, callback) {
-        return new Promise((resolve, reject) => {
-          const $script = document.createElement('script');
-          $script.src = script.src;
-          $script.type = script.type;
-      
-          $script.onload = () => resolve();
-          $script.onerror = (error) => reject(error);
-      
-          document.body.appendChild($script);
-        });
-      }
-
-      getScript(script)
-        .then(() => {
-          this._loadScripts(scripts).then(resolve);
+      $.getScript(url)
+        .done(() => {
+          $.ajaxSetup({cache: false});
+          this._loadScripts(urls).then(resolve);
         })
-        .catch((error) => {
-          console.error(`Failed to load ${url}: ${error}`);
-          this._loadScripts(scripts).then(resolve);
+        .fail(() => {
+          console.error(`Failed to load ${url}:`);
+          $.ajaxSetup({cache: false});
+          this._loadScripts(urls).then(resolve);
         });
     });
   },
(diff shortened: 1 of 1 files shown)


brandonkelly:
It’s important that lazy-registered JS is loaded via `$.getScript()`, which uses Ajax + `eval()` rather than adding a new `<script>` tag to the DOM, to workaround browser crossdomain restrictions (see #13715).

Maybe the PR could be adjusted to keep things the way they were, but start ignoring `<script type="module">` tags altogether?

brandonkelly:
Also, we’ve been using `$.getScript()` since 4.5.6. Does this bug not affect Craft 4 for some reason?

engram-design:
Okay, so a more lighter touch would be:

```diff
const nodes = $.parseHTML(html.trim(), true).filter((node) => {
  if (node.nodeName === 'LINK' && node.href) {
    if (!this._existingCss) {
      this._existingCss = $('link[href]')
        .toArray()
        .map((n) => n.href.replace(/&/g, '&amp;'));
    }

    if (this._existingCss.includes(node.href)) {
      return false;
    }

    this._existingCss.push(node.href);
    return true;
  }

  if (node.nodeName === 'SCRIPT' && node.src) {
    if (!this._existingJs) {
      this._existingJs = $('script[src]')
        .toArray()
        .map((n) => n.src.replace(/&/g, '&amp;'));
    }

    if (!this._existingJs.includes(node.src)) {
-      scriptUrls.push(node.src);
+      if (node.type === 'module') {
+        return true;
+      } else {
+        scriptUrls.push(node.src);
+      }

      this._existingJs.push(node.src);
    }

    // return false either way since we are going to load it ourselves
    return false;
  }

  return true;
});
```

So that if this is detected as a module, it bypasses `_loadScripts()` altogether, and returns true to have it part of `nodes` which is then used in `$parent.append(nodes);` but not in `await this._loadScripts(scriptUrls);`

brandonkelly:
Yeah that’s what I was thinking. Can you adjust the PR and verify it works for you?

engram-design:
No worries, done.

brandonkelly:
Is it expected that the `<script type="module">` tag would be omitted, if the same URL is added in a subsequent Ajax request? (Because on the second time, `if (!this._existingJs.includes(node.src))` won’t pass.)

engram-design:
I'm not fully across what `_existingJs` tries to achieve, but I assume it's to prevent scripts from being added to the DOM multiple times?

But we're returning a node not just the URL in that `filter()` call, so the first time it'll return the `<script type="module" src="http://localhost:8080...">` and then it'll check against that `src` whether it's been handled already and not add it or load it. Unless I'm reading that logic wrong?

But I would expect that once loaded, a duplicate script shouldn't be loaded again - not just for script modules?

brandonkelly:
> I assume it's to prevent scripts from being added to the DOM multiple times?

Right.

> But I would expect that once loaded, a duplicate script shouldn't be loaded again - not just for script modules?

I think so, but worth testing, e.g. by creating multiple Hyper fields in succession from the same CP page.

engram-design:
So all good with the field settings, switching back and forth between types seems fine. I've also tested creating multiple fields with the new entry type + add field editor, which takes care of Matrix-like fields and sections, etc.

Similarly, for entries, scripts are only loaded once for multiple Hyper fields.

brandonkelly:
On second thought, decided to just [revert to how we are loading scripts in Craft 4.5.6+](https://github.com/craftcms/cms/commit/c3881705748a78cc7efc2e6bdaf39930fe60e7d2), if that’s working fine. I’m a little concerned with splitting out how we load things depending on `type="script"` vs `type="module"` because there’s still the crossdomain issues to consider, and now we’re also potentially loading things in the wrong order (all `type="script"` first, then all `type="module"`), which might cause dependency issues.

engram-design:
Fair call, I'll give that a quick test. Thanks for making sense of this PR!

brandonkelly:
5.0.0-beta.7 is out now with that change.

PR #14531, opened 2024-03-05
Fix typo in NL translation

MarcelWeidum:
### Description
Typo in NL translation


brandonkelly:
Thanks! I’ve pulled your change in via [CrowdIn](https://translate.craftcms.com/).

PR #14541, opened 2024-03-06
Feature/cms 1091 ability to move entries to a new section

i-just:
### Description
Adds element index action to move entries between sections.

- select entries to be moved and trigger the “Move to” action
Screenshot 2024-03-06 at 12 06 58

- see a modal with a list of compatible sections (sections that have the same entry type(s) as the entries we want to move)
Screenshot 2024-03-06 at 12 07 10

- choose a section to move to
- entries are moved to the new section
Screenshot 2024-03-06 at 12 07 17

**Notes:**
- it’s not possible to move to a single section
- for structures, entries are moved to the end of the new structure as root elements
- if the section entries are moved to has a lower `maxAuthors` limit, no data is changed/lost for the entries, but the next time they’re saved, a validation error will be shown
- when an entry is moved, its revisions and drafts are moved too; if their entry type isn’t allowed by the section we’re moving to, you will still be able to see the draft/revision but not apply/revert to it (the same applies to revisions and draft for entry types that were removed from the section); thanks @brianjhanson for adjusting the awesome `craft-tooltip` to work with disabled elements!
Screenshot 2024-03-11 at 09 59 42
Screenshot 2024-03-11 at 10 00 13

### Related issues
[cms-1091](https://linear.app/craftcms/issue/CMS-1091/ability-to-move-entries-to-a-new-section)


brandonkelly:
Rather than showing the “Couldn’t find any sections that all selected elements could be moved to.” message in the modal, ideally the “Move to” action is just grayed out ahead of time if there aren’t any other sections with the same entry type.

This should be doable via the `validateSelection()` method on the action trigger JS.

gcamacho079 (changes requested):
Just finished a quick scan of the modal contents:
- [x] The section buttons should have an `aria-pressed` value that toggles between true and false (see [internal documentation re: exclusive button groups](https://www.notion.so/pixelandtonic/exclusive-button-group-ed2321f5ffe34bbea5e42e2a93ac3624?pvs=4))
- [x] The Move button's `aria-disabled` state should be toggled when a section is selected
- [ ] The focus indicator for the move icon is under 3:1 contrast ratio against the modal footer background. This is a larger global issue caused by the disabled element opacity.
- [x] The focus indicator for the button is difficult to see since it doesn’t contrast against the dark background. This could benefit from a two-toned focus indicator, like the one used in the global skip link, with a white outline adjacent to the button background.
- [x] Loading spinner that appears in modal content body should have an equivalent status message that is read for screen reader users.

Note that some of these (like the disabled button focus outline) may be larger/more global issues than the scope of this PR.

i-just:
@gcamacho079, as per our chats, all items except the disabled button focus ring are actioned.

To get the last item from the list to work, I added the live region container and `updateLiveRegion` method to all modals (via `Garnish.Modal`) and started using it in my `EntryMover`, which means other modals could start taking advantage of it.

gcamacho079 (approved):

src/controllers/ElementsController.php
```diff
@@ -884,25 +884,52 @@ private function _additionalButtons(
 
         // Apply draft
         if ($isDraft && !$isCurrent && $canSave && $canSaveCanonical) {
-            $components[] = Html::button(Craft::t('app', 'Apply draft'), [
-                'class' => ['btn', 'secondary', 'formsubmit'],
+            /** @phpstan-ignore-next-line */
+            $disabled = $canonical->hasMethod('isEntryTypeCompatible') && !$element->isEntryTypeCompatible();
```
brandonkelly, line 888: ElementsController shouldn’t be giving any special treatment to entries, so we’ll need to rethink how we do this.

src/base/ElementInterface.php
```diff
@@ -881,6 +881,15 @@ public function canDelete(User $user): bool;
...
     public function canDeleteForSite(User $user): bool;
 
+    /**
+     * Returns whether the given user is authorized to move this element.
+     *
+     * @param User $user
+     * @return bool
+     * @since 5.0.0
+     */
+    public function canMove(User $user): bool;
```
brandonkelly, line 891: It seems a little weird to introduce the concept of “moving” to all element types, when entries are the only one that would actually know what to do with it.

src/controllers/ElementsController.php
i-just: ok; addressed here: https://github.com/craftcms/cms/pull/14541/commits/f86e782c22369d5dc5c3dbe61a076855c50c63bd

src/base/ElementInterface.php
i-just: As per our chat, this might end up being a way to go.
But, for now, I have reverted those changes and used a different approach where I added the ability to further customise element chip & card attributes, and I’m adding `data-movable` via those.

gcamacho079:
A few accessibility improvements based on automated and manual testing:

- [x] The `checkbox` role doesn’t communicate the correct interaction pattern. Radio inputs would work better since the selections are exclusive.
  - [x] Radio inputs should be contained in a `fieldset` with a `legend`/`aria-labelledby` that references the title of the modal
  - [x] List markup should be removed as radio group markup should provide the necessary relationship information
- [x] The accessible name of the inputs should match the visible name (i.e., the section name). Right now, the label ID that’s referenced via `aria-labelledby` does not exist, so the name is coming from the input `title`.
- [x] Focus is dropped when an entry is moved. In these instances, focus should probably be moved to the `.elements` container. Ensure programmatic focus can be set by adding `tabindex="-1" to the container.

Issue #14438, opened 2024-02-21
[4.x]: Uploading files via asset modal > upload files and assets index > upload files is broken


i-just:
### Description
Uploading files via asset modal > upload files and assets index > upload files is broken.
Direct upload works as expected.

(Related to: https://github.com/craftcms/cms/issues/14403)

### Steps to reproduce

1. create an Assets field
2. add that field to an entry type
3. create a new entry of that type
4. click “Add an asset”> “Upload files”> select at least one file to upload
5. spinner shows, and there’s an error in the console: `Uncaught TypeError: this.uploader is null`

OR try to upload via Assets > Upload files

### Expected behavior
Able to upload a file this way.

### Actual behavior
Spinner shows, and there’s an error in the console: `Uncaught TypeError: this.uploader is null`. The file actually uploads, but you don’t know that unless you refresh the page and view the assets again.

### Craft CMS version

4 and 5

### Installed plugins and versions

-

Issue #14403, opened 2024-02-16
[4.x]: Asset Browser crashes when quickly switching between sources with different upload permissions


mnlmaier:
### Description
When quickly changing asset sources in the Control Panel's Asset Browser, JS crashes occur when permissions need to be adjusted for the upload functionality.

The Error Stack Trace in the JS Console looks as follows:
```
jquery.js?v=1702329863:2 Uncaught Error: cannot call methods on fileupload prior to initialization; attempted to call method 'option'
    at Function.error (jquery.js?v=1702329863:2:2623)
    at HTMLButtonElement.<anonymous> (jquery-ui.js?v=1702329863:2:2391)
    at Function.each (jquery.js?v=1702329863:2:3002)
    at w.fn.init.each (jquery.js?v=1702329863:2:1482)
    at t.fn.<computed> [as fileupload] (jquery-ui.js?v=1702329863:2:2129)
    at s.constructor.setParams (Uploader.js:43:21)
    at s.constructor.e [as setParams] (Base.js:76:36)
    at s.constructor.onSourcePathChange (AssetIndex.js:296:24)
    at s.constructor.e [as onSourcePathChange] (Base.js:76:36)
    at set sourcePath (BaseElementIndex.js:914:12)
```

I am pretty sure that the solution might just be to check for `this.uploader` before calling `this.uploader.fileupload()`, e.g. by calling it via `this.uploader?.fileupload()` and waiting for it until it's actually defined. still, i'm pretty sure that some more measures need to be taken care of.

In the code: https://github.com/craftcms/cms/blob/develop/src/web/assets/cp/src/js/Uploader.js#L43

### Steps to reproduce
1. Create two asset sources / volumes in the Asset browser
2. Assign the following permissions to a test user: Can upload in volume 1, can only view volume 2
3. Quickly switch between the volumes / sources several times while observing the JavaScript console
4. Notice that the script is crashing and the "upload files" button will not change its `disabled` state
5. Notice that this behavior does not happen when changing sources more slowly.

(Please note that this will be relevant for users which share the same permissions across all asset sources).

### Expected behavior
Permissions for the upload button are adjusted on the fly, including its `disabled` state.

### Actual behavior
`this.uploader.fileupload` will be called in `Uploader.js` while `this.uploader` actually not being defined, which leads to completely crashing the JavaScript.

### Craft CMS version

Craft Pro 4.5.5

### PHP version

8.3.2

### Operating system and version

Darwin 23.2.0

### Database type and version

[...]

PR #14542, opened 2024-03-06
Fix broken JS on Cloud uploader when changing dirs

timkelty:
Following https://github.com/craftcms/cms/pull/14439, a JS error was thrown when navigating directories.
This is because AssetIndex was calling `.fileupload()`, which can't be assumed, as only hte core Uploader uses the jquery uploader plugin.

## Changes
- When selecting source, make `onSelectSource` fire before `onSourcePathChange` to be less awkward.
- With `onSelectSource` fired first, we can be certain out uploader is created by the time we get to `onSourcePathChange`
- Don't call `fileupload` methods if the plugin wasn't instantiated (eg if uploading is disabled)
- Eliminate duplicate inputs by setting`replaceFileInput = false`, which is only needed for very old iframe uploading. 

## History
- this was reported: https://github.com/craftcms/cms/issues/14403
- first fix (@brandon): https://github.com/craftcms/cms/commit/af652dba5d7e58e001350c297963bdf0ff435ab5
- but then it caused this: https://github.com/craftcms/cms/issues/14438
- second fix (@i-just ): https://github.com/craftcms/cms/pull/14439

## Test cases
- permissions
  - create 2 volumes
  - user can do everything for vol1
  - user can only view vol2
  - vol1 has one subfolder
  - go to vol1 > subfolder & upload a file - all good
  - go to vol2 - upload button is disabled and you can’t upload - all good
  - go to vol1, you should be deep linked into the subfolder (cause that’s where you were last); try to upload a file
- Assets page
  - upload button
  - drag to upload
- Assets field
  - select via modal
  - upload directly
  - upload directly via drag


src/web/assets/cp/src/js/AssetIndex.js
```diff
@@ -257,7 +250,6 @@ Craft.AssetIndex = Craft.BaseElementIndex.extend(
           Garnish.hasAttr(this.$source, 'data-can-upload')
         ) {
           this.uploader?.destroy();
-          this.$uploadInput.insertBefore(this.$uploadButton);
```
timkelty, line 260: It is already inserted in `createUploadInputs`, don't do it again.

src/web/assets/cp/src/js/BaseElementIndex.js
```diff
@@ -1990,12 +1990,12 @@ Craft.BaseElementIndex = Garnish.Base.extend(
         this.updateFilterBtn();
       }
 
+      this.onSelectSource();
+
```
timkelty, line 1994: Fire `onSelectSource` before `onSourcePathChange`, so our inputs and buttons are created first.

src/web/assets/cp/src/js/Uploader.js
```diff
@@ -24,23 +24,19 @@ Craft.Uploader = Craft.BaseUploader.extend(
...
       this.$element.on('fileuploadadd', this._onFileAdd);
-
-      // Update our reference to this.$fileInput if it is being replaced on change
-      if (this.uploader.fileupload('option', 'replaceFileInput')) {
-        this.$element.bind('fileuploadchange', this.updateFileInput.bind(this));
-      }
-    },
-
-    updateFileInput: function (event, data) {
-      this.$fileInput = data.fileInputClone || data.fileInput;
```
timkelty, line 35: Not needed with `replaceFileInput = false`

src/web/assets/cp/src/js/Uploader.js
```diff
@@ -24,23 +24,19 @@ Craft.Uploader = Craft.BaseUploader.extend(
...
      */
     setParams: function (paramObject) {
       this.base(paramObject);
-      this.uploader.fileupload('option', {formData: this.formData});
+
+      // Only set params if the uploader has been initialized
+      // It won't be if the input is disabled
+      if (this.uploader.data('blueimpFileupload')) {
+        this.uploader.fileupload('option', {formData: this.formData});
+      }
```
timkelty, line 39: If the user has no upload perms, `fileupload` wont be initialized.

src/web/assets/cp/src/js/Uploader.js
```diff
@@ -128,6 +124,7 @@ Craft.Uploader = Craft.BaseUploader.extend(
       autoUpload: false,
       sequentialUploads: true,
       maxFileSize: Craft.maxUploadSize,
+      replaceFileInput: false,
```
timkelty, line 127: This creates duplicate inputs, which causes problems and is only needed for legacy iframe transport uploading.

PR #14561, opened 2024-03-09
Update tooltip to work on disabled elements

brianjhanson:
Updates the `craft-tooltip` to work for disabled elements. To pull this off, we changed the API just a bit. The biggest change is that the tooltip now grabs the text for the tooltip from `aria-label` rather than from the `innerText`. 

To use the tooltip with disabled elements, you can wrap the element with the `<craft-tooltip`. The previous behavior is still available if you add a `self-managed` property to the tooltip. 

When `craft-tooltip` is used as a wrapper, it will look for an `a`, `button` or `[role="button"]` and use that as the trigger for the tooltip. 

```html
<craft-tooltip aria-label="Text for the tooltip">
  <button type="button" disabled>Some Action</button>
</craft-tooltip>
```

This brings our tooltip more in line with [spectrum's API](https://spectrum.adobe.com/page/tooltip/) and [GitHub Primer](https://primer.style/components/tooltip).

### DIFF
--- a/src/controllers/ElementsController.php
+++ b/src/controllers/ElementsController.php
@@ -886,23 +886,26 @@ class ElementsController extends Controller
         if ($isDraft && !$isCurrent && $canSave && $canSaveCanonical) {
             /** @phpstan-ignore-next-line */
             $disabled = $canonical->hasMethod('isEntryTypeCompatible') && !$element->isEntryTypeCompatible();
-            $btnContent = Craft::t('app', 'Apply draft');
-            if ($disabled) {
-                $btnContent .= ' <craft-tooltip>' .
-                    Craft::t(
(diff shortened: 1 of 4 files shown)


i-just:
Thanks, Brian! 
It works like I’d expect it to, except the `disabled` attribute shouldn’t be added because it prevents the element from being tabbable. Removing `'disabled' => $disabled` (from [here](https://github.com/craftcms/cms/pull/14561/files#diff-aee2e79d39f0844ff1b3633c4c8ad09b7f6a18ff355d1242b07f7d8041431d54R892)) makes it possible to tab into the "Apply draft" button and see the tooltip, but you still can’t activate it, which is exactly what I was after :)

PR #14564, opened 2024-03-11
Add `static` hint in ActiveRecord inherited @method docblock

soulseekah:
Static code analysis tools (namely PHPStan) think that this method is non-static. Adding the static hint as per https://docs.phpdoc.org/3.0/guide/references/phpdoc/tags/method.html helps this.


brandonkelly:
Thanks!

PR #14571, opened 2024-03-11
Accessibility improvements

brianjhanson:
### Description

Accessibility issues from the audit

### Related issues
- [PT-63](https://linear.app/craftcms/issue/PT-63/verification-code-input-doesnt-have-an-accessible-name)
- [PT-64](https://linear.app/craftcms/issue/PT-64/qr-code-key-input-doesnt-have-an-accessible-name)
- [PT-1281](https://linear.app/craftcms/issue/PT-1281/focus-is-not-visible-for-active-option-in-status-menu)
- [PT-80](https://linear.app/craftcms/issue/PT-80/the-status-menu-listbox-contains-children-roles-that-are-not-allowed)
- [PT-1293](https://linear.app/craftcms/issue/PT-129/update-table-in-element-selector-modal-for-keyboard-and-screen-reader)


src/templates/_elements/toolbar.twig
```diff
@@ -36,7 +36,7 @@
                 {% endif %}
             </ul>
             {% if isAdministrative %}
-                {% if showStatusMenu %}<hr class="padded">{% endif %}
+                {% if showStatusMenu %}<hr class="padded" aria-hidden="true">{% endif %}
```
brianjhanson, line 39: @gcamacho079 not sure this is the best fix for [PT-80](https://linear.app/craftcms/issue/PT-80/the-status-menu-listbox-contains-children-roles-that-are-not-allowed) but it works.

gcamacho079: Great solution! 👍🏼 It’s a subtle difference, but `role="presentation"` also quiets the error while saying: “Ignore the usual semantics here.”

brianjhanson: @gcamacho079 is one better than the other? I'm happy to switch it to `role="presentation"` if that's more desirable.

gcamacho079: Yes, let's switch that over! Appreciation!

gcamacho079 (changes requested):
Just a couple things:
- [ ] The QR code input should have a visible label as well
- [ ] For PT-1282, the checkmark and disabled status icons should also inherit the text color

brianjhanson:
> The QR code input should have a visible label as well

Is it okay if this is just an `aria-label` on the SVG element? The QR code contains the same information as the input, it's just an alternate display of it.

gcamacho079 (approved):

PR #14585, opened 2024-03-13
Sidebar accessibility improvements

brianjhanson:
Accessibility improvements for the global sidebar. 

Fixes PT-1292, PT-1291, PT-1290, PT-1287, PT-1289, PT-1288


gcamacho079 (changes requested):
Thank you Brian! Just a couple things:

- [x] Hover state for when the subnav toggle is closed has been resolved. But when the subnav is open, the hover state of the chevron is still under 2.9:1.
- [x] Instead of setting the `aria-label`/`img` on the wrapping span, can those be put on the SVG instead? (Removing the `aria-hidden` on that SVG as well.)
- [x] The "External link" `aria-label` is not translated

brianjhanson:
>  Hover state for when the subnav toggle is closed has been resolved. But when the subnav is open, the hover state of the chevron is still under 2.9:1.

👍  addressed in [8b1a189](https://github.com/craftcms/cms/pull/14585/commits/8b1a1892315ed9112b1597b801e0cb0505d755cb). I almost tweaked the `--ui-control-bg-hover` variable but since 2.9:1 is so close, I'm guessing that color is only an issue because the background of the sidebar is ever so slightly darker. 

>  Instead of setting the aria-label/img on the wrapping span, can those be put on the SVG instead? (Removing the aria-hidden on that SVG as well.)

Removing the `aria-hidden` is a bit tricky because it [comes from the CP helper function](https://github.com/craftcms/cms/blob/167f674cba7a027ae8e40f59f08ce6f2de63bc4a/src/helpers/Cp.php#L2801-L2803) and feels like it should be present in 99% of situations. I did make it `aria-hidden="false"` though. 

> The "External link" aria-label is not translated

We just haven't translated the string yet.

brianjhanson:
@brandonkelly this one's ready for you now. RTL styles have been fixed up. 
CleanShot 2024-03-15 at 10 34 55

PR #14586, opened 2024-03-13
Add options config for backup/restore commands

timkelty:
### Description
The PR allows `\craft\config\GeneralConfig::backupCommand` and `\craft\config\GeneralConfig::restoreCommand` to be an options array to configure how the commands are generated.

#### Goals
- make small tweaks easy (eg ignore a single table). Currently this can only be done via event.
- make it easy to opt-in to postgres's archive format (possibly make this the default for v5?)
- make it clear when things are going to be replaced or merged (eg `ignore-tables`)
- provide a way to tap into command for modification
- be non-breaking

Valid values include:

```
<?php
use craft\config\GeneralConfig;
use mikehaertl\shellcommand\Command as ShellCommand;

    ->restoreCommand([
        'archiveFormat' => true, // pgsql-only

        // example: add an arg
        'callback' => fn(ShellCommand $command) => $command->addArg('--verbose'),
    ])
    ->backupCommand([
        'ignoreTables' => ['foo', 'bar'],
        'archiveFormat' => true, // pgsql-only

        // example: remove an arg
        'callback' => function (ShellCommand $command) {
            $command->setArgs(str_replace('--dump-date', '', $command->getArgs()));

            return $command;
        },
    ]);
```

#### etc…
- I ended up leveraging `mikehaertl\shellcommand\Command` because it was there and were already using it. Easier than stitching strings together.
- To keep backward compat, `ignoreTables` can either come from the `EVENT_BEFORE_CREATE_BACKUP` or from the new `ignoreTables` option. If both are present, the event will take precedence.
- The `callback` could easily be an event instead, but it seemed like it might be a pain to add an event listener just to add an additional arg.


brandonkelly:
> The options array is keyed by the db driver name

This feels like unnecessary complexity to me. No project uses both.

timkelty:
@brandonkelly yep, already nixed it.

src/db/pgsql/Schema.php
```diff
@@ -312,14 +300,4 @@ protected function getIndexInformation(TableSchema $table): array
...
-
-    /**
-     * Returns the PGPASSWORD command for backup/restore actions.
-     *
-     * @return string
-     */
-    private function _pgpasswordCommand(): string
-    {
-        return Platform::isWindows() ? 'set PGPASSWORD="{password}" && ' : 'PGPASSWORD="{password}" ';
-    }
```
timkelty, line 324: Moved to `\craft\db\DbShellCommand`

src/db/mysql/Schema.php
```diff
@@ -382,47 +318,4 @@ protected function findConstraints($table): void
...
-            $contents .= PHP_EOL . 'ssl_key=' . $this->db->attributes[PDO::MYSQL_ATTR_SSL_KEY];
-        }
-
-        FileHelper::writeToFile($this->tempMyCnfPath, '');
-        // Avoid a “world-writable config file 'my.cnf' is ignored” warning
-        chmod($this->tempMyCnfPath, 0600);
-        FileHelper::writeToFile($this->tempMyCnfPath, $contents, ['append']);
-
-        return $this->tempMyCnfPath;
-    }
```
timkelty, line 427: Moved to `\craft\db\DbShellCommand`

src/db/mysql/Schema.php
```diff
@@ -152,79 +155,13 @@ public function createColumnSchemaBuilder($type, $length = null): ColumnSchemaBu
...
-            ' {database}';
-
-        $dataDump = 'mysqldump' .
-            $defaultArgs .
-            ' --no-create-info' .
-            ' ' . implode(' ', $ignoreTableArgs) .
-            ' {database}' .
-            ' >> "{file}"';
-
-        return $schemaDump . ' && ' . $dataDump;
```
timkelty, line 227: Moved to `\craft\db\mysql\BackupCommand`

src/db/pgsql/Schema.php
```diff
@@ -116,29 +125,13 @@ public function getLastInsertID($sequenceName = ''): string
...
-            ' --port={port}' .
-            ' --username={user}' .
-            ' --if-exists' .
-            ' --clean' .
-            ' --no-owner' .
-            ' --no-privileges' .
-            ' --no-acl' .
-            ' --file="{file}"' .
-            ' --schema={schema}' .
-            ' ' . implode(' ', $ignoredTableArgs);
```
timkelty, line 141: Moved to `\craft\db\pgsql\RestoreCommand`

src/db/pgsql/Schema.php
```diff
@@ -148,14 +141,9 @@ public function getDefaultBackupCommand(?array $ignoreTables = null): string
...
     public function getDefaultRestoreCommand(): string
     {
-        return $this->_pgpasswordCommand() .
-            'psql' .
-            ' --dbname={database}' .
-            ' --host={server}' .
-            ' --port={port}' .
-            ' --username={user}' .
-            ' --no-password' .
-            ' < "{file}"';
```
timkelty, line 158: Moved to `\craft\db\pgsql\RestoreCommand`

src/db/DbShellCommand.php
```diff
@@ -0,0 +1,103 @@
...
+        return $this->getCommand()->getExecCommand();
+    }
+
+    /**
+     * Creates a temporary my.cnf file based on the DB config settings.
+     *
+     * @return string The path to the my.cnf file
+     * @throws ErrorException
+     */
+    protected function createDumpConfigFile(): string
```
angrybrad, line 44: Worth renaming to `createMySQLDumpConfigFile()` so it's more explicit?

timkelty: Probably? – only reason I didn't is it seems to have logic in it for handling pgsql, oddly enough.

angrybrad: Hrm... that feels unnecessary, given the first thing it's going to do is throw an exception if it's not MySQL. Probably some leftover code.

timkelty: Renamed and removed pg conditions

brandonkelly:
@timkelty After reading https://github.com/craftcms/cms/issues/14648#issuecomment-2057541574 I’m wondering if it would make more sense to just add a bunch of new top-level config settings for this. Brad is suggesting using a multi-env config in response, which we are sortof moving away from in favor of environment variables, but it doesn’t seem like it will be possible to use environment variables here as currently implemented. (Besides pulling them in manually from `config/general.php` via `App::env()`.)

timkelty:
Closed in favor of https://github.com/craftcms/cms/pull/14897

PR #14595, opened 2024-03-14
RTL Navigation fixes

### LINKED ISSUE #14589: [5.x]: Right to Left subnav issues
### Description

1) While in RTL mode, dots on the subnav are not right-aligned.

2) Main nav highlight bar is also not right aligned.

3) Utility counter badge also not rendering number correctly.

CleanShot 2024-03-14 at 11 10 27@2x

### Craft CMS version

5.0.0.beta-7

### Browsers

Issue seen in Firefox and Chrome

brianjhanson:
Switches to [logical properties](https://css-tricks.com/css-logical-properties-and-values/) to better handle RTL languages. 

Using this over the mixins we have in hopes that we can ditch sass a little easier some day in the future. 

Fixes #14589


brandonkelly:
Whoa, never knew those existed. Let’s go all in on those for Craft 6, and deprecate all the Sass mixins we currently use (e.g. `margin-left()`).

brandonkelly:
The hairline between the sidebar and the rest of the page is still right-aligned:

Screenshot of a portion of the global nav in right-to-left orientation

Issue #14555, opened 2024-03-08
[5.x]: Setting unknown property: craft\mail\transportadapters\Smtp::encryptionMethod


KlingeSean:
### Description
This error is thrown when opening the email SMTP settings page.

### Steps to reproduce

1. open the email SMTP settings page.

### Expected behavior

Being able to open and change the SMTP settings.

### Actual behavior
Recieve an error stating: "Setting unknown property: craft\mail\transportadapters\Smtp::encryptionMethod"

### Craft CMS version

5.0.0-beta.5

### PHP version

8.2

### Operating system and version

Linux 3.10.0-962.3.2.lve1.5.77.el7.x86_64

### Database type and version

MariaDB 10.6.12

### Image driver and version

	GD 8.2.3

### Installed plugins and versions

none

WS-Ramon:
Blijkt dat de volgende velden niet meer bestaan:
```
transportSettings:
  encryptionMethod
  timeout
```

We moesten deze handmatig uit de project.yaml halen.

KlingeSean:
As stated in a comment provided the issue was that after upgrading the project.yaml did not remove these 2 settings from the email settings:
`transportSettings:
  encryptionMethod
  timeout`
  
  Removing those fixes the issue closing this.

brandonkelly:
5.0.0-beta.10 was just released with a fix for this.

PR #14619, opened 2024-03-18
Don't assign unsafe mailer transport settings

timkelty:
Fixes https://github.com/craftcms/cms/issues/14555


src/helpers/MailerHelper.php
```diff
@@ -82,10 +83,19 @@ public static function allMailerTransportTypes(): array
...
     {
-        return Component::createComponent([
+        $component = Component::createComponent([
             'type' => $type,
-            'settings' => $settings,
         ], TransportAdapterInterface::class);
+
+        if ($settings) {
+            if ($component instanceof Model) {
+                $component->setAttributes($settings);
```
brandonkelly, line 92: Should we be passing `false` to `$safeOnly` here?

timkelty: No – the whole point is to go through `setAttributes` _with_ `safeOnly` to avoid errors from stale PC values.

PR #14620, opened 2024-03-18
Allow configuration of SVG sanitizer

timkelty:
### Description
Allows configuration of the SVG sanitizer by accepting `bool|callable`.
A more flexible alternative to https://github.com/craftcms/cms/pull/14492, which exposes a new dedicated for `sanitizeSvgRemoteRefs`.

### Related issues
https://github.com/craftcms/cms/pull/14492

### DIFF
--- a/src/config/GeneralConfig.php
+++ b/src/config/GeneralConfig.php
@@ -2555,6 +2555,16 @@ class GeneralConfig extends BaseConfig
      * ```
      * :::
      *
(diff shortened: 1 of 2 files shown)


cgivord-lg2:
I don't think this is necessary to have a more flexible way than #14492 because the sanitizer library does not offer much options.
Also, IMHO, adding code to the config file is not recommended long term.

PR #14643, opened 2024-03-21
Moves the forgot password link below the field

brianjhanson:
The tab order of this annoyed me, and Lupe says it's also better for zoomed in users. Win win!


thupsi:
> This annoyed me

Haha, it also annoyed me! When tabbing from the username field I started typing the password, only to find out the focus was on the bloody link. I happened all the time 🤣

PR #14675, opened 2024-03-25
allow removing alt text

i-just:
### Description
Adds the `isPrimaryAlt` flag to the `assets_sites` table. 
When you first upload an asset - `isPrimaryAlt` is set to `0`.
When you add the alt text, `isPrimaryAlt` becomes `1` for the site you edited the alt for.
When you edit the asset and amend alt text for another site, its `isPrimaryAlt` also becomes `1`.
`isPrimaryAlt` should never revert from `1` to `0`.

The flag is then used in `AssetQuery::createElement()` to ensure the alt text can be set to an empty value and displayed as such.

### Related issues
[cms-1284](https://linear.app/craftcms/issue/CMS-1284/translatable-alt-text-unable-to-remove-text-for-a-site)

### DIFF
--- a/src/config/app.php
+++ b/src/config/app.php
@@ -4,7 +4,7 @@ return [
     'id' => 'CraftCMS',
     'name' => 'Craft CMS',
     'version' => '5.0.0-beta.11',
-    'schemaVersion' => '5.0.0.20',
+    'schemaVersion' => '5.0.0.21',
     'minVersionRequired' => '4.4.0',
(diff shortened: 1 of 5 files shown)


brandonkelly:
Fixed this by preserving empty string values from `__construct()` and `setAttributesFromRequest()`, and updating `AssetQuery::createElement()` to always override `alt` with `$siteAlt` if the value isn’t `null`. (535ab0d47e13ce595d825913811a454e0a7a75b7)

PR #14729, opened 2024-03-30
Fix `getActionSegments()` docblock

bencroker:
Not sure what happened here, but this PR fixes it.


brandonkelly:
That’s how yii2-apidoc wants links to other class members to be formatted. (See https://docs.craftcms.com/api/v4/craft-web-request.html#method-getactionsegments)

bencroker:
Ah, my bad, thanks!

PR #14730, opened 2024-03-30
Allow a Matrix-based Entry to be accessible via `Craft.MatrixInputEntry`

engram-design:
Rather than a private `const Entry()` variable, it'd be great if that was exposed to other plugins. [Smith](https://github.com/verbb/smith/blob/c2397f92da4206c92162e3fe321cdd6e48c13276/src/resources/src/js/smith.js#L209) in particular uses this to create a new Matrix entry from an existing cloned entry.

It'd be neat not to [duplicate code](https://github.com/verbb/smith/blob/c2397f92da4206c92162e3fe321cdd6e48c13276/src/resources/src/js/smith.js#L266)

### DIFF
--- a/src/web/assets/matrix/src/MatrixInput.js
+++ b/src/web/assets/matrix/src/MatrixInput.js
@@ -108,7 +108,7 @@ import $ from 'jquery';
 
         for (let i = 0; i < $entries.length; i++) {
           const $entry = $($entries[i]);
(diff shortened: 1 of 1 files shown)


brandonkelly:
Renamed it to `Craft.MatrixInput.Entry` for 5.1 (d9bf77ec4618a7ada1f362e319192594a47af838).

PR #14735, opened 2024-04-01
Fix nav template

thomasvantuycom:
### Description

The `_includes/nav` template doesn't pass the `selectedItem` to a nested list. This PR fixes that.


brandonkelly:
Thank you!

brandonkelly:
Craft 5.0.1 is out now with that fix.

PR #14748, opened 2024-04-03
Feature/webauthn update

i-just:
### Description
Stop using deprecated webauthn methods - preparation for webauthn-lib v5 release.

Those changes prevent the passkeys created on the earlier version from working. Hence, they’re not a part of the #14747 fix.

I’m targeting 5.1 at the moment, but we might decide to postpone this even further.


i-just:
closing in favour of https://github.com/craftcms/cms/pull/18545

PR #14781, opened 2024-04-10
Table scroll indicators

### LINKED ISSUE #14765: [5.x]: Horizontal clipping on element indexes
### Description

In element index pages the table can be clipped horizontally with no way to scroll:

image

### Craft CMS version

5.0.2

### PHP version

_No response_

_No response_

### Database type and version

_No response_

### Image driver and version

_No response_

-

brianjhanson:
### Description

Adds a `<craft-scroll-container/>` component that will display shadows around a scrollable element but only used for element table views at the moment. Admittedly this is a lot of code for a relatively small feature, so no hard feelings if we don't want to pull it in. I created this as a web component mostly just for the encapsulation, but it also means we should be able to use it elsewhere but I couldn't find another place I thought would benefit from it as much. 

[CodePen version if you want to play with it](https://codepen.io/brianjhanson/pen/LYvmzPP?editors=0010)

### Related issues
Fixes #14765 

https://github.com/craftcms/cms/assets/1843073/f2af737d-4d36-464a-bc15-e1a96429e0bf

### DIFF
--- a/src/templates/_elements/tableview/container.twig
+++ b/src/templates/_elements/tableview/container.twig
@@ -2,37 +2,45 @@
 {% set instructionId = instructionId ?? "sort-instructions-#{random()}" %}
 
 <div class="tableview{% if context in ['index', 'embedded-index'] %} tablepane{% endif %}">
-    <span class="visually-hidden" data-status-message role="status"></span>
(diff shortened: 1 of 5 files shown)


brianjhanson:
Closing in favor of #14786

PR #14784, opened 2024-04-10
Add public method to get authorization bearer token

timkelty:
### Description
Getting a bearer token from an authorization header is a very common (and somewhat annoying) thing to do.
Our GQL service was already doing this, but not in a reusable way.


src/controllers/GraphqlController.php
```diff
@@ -288,6 +265,25 @@ private function _schema(GqlService $gqlService): GqlSchema
...
+    {
+        $bearerToken = $this->request->getBearerToken();
+
+        if ($bearerToken) {
+            try {
+                $token = $gqlService->getTokenByAccessToken($bearerToken);
+
+                if ($token->getIsValid()) {
+                    return $token;
+                }
```
brandonkelly: Shouldn’t this be

```php
if (!$token?->getIsValid()) {
    throw new BadRequestHttpException('Invalid Authorization header');
}

return $token;

timkelty: Indeed. Doesn't need nullsafe though.
Updated.

PR #14799, opened 2024-04-12
Fix unexpected translations

panlatent:
### Description

This PR solves the problem that element types are not displayed correctly in Chinese language


brandonkelly:
Thanks!

PR #14817, opened 2024-04-16
Improve batched job progress indicator

bencroker:
This PR improves the batched job progress indicator by making it work more like the base job progress indicator, meaning that the percentage and indicator will move from 0% to 100% across all batches, rather than once for each batch. 

In this example, a sendout is being sent to 5 recipients with a batch size of 2 (to help demonstrate the issue).

**Before:**

The progress of 50% is calculated as step 1 of 2 items in the current batch.

screenshot-pvvU3pIR@2x
screenshot-IP2yqgcq@2x

**After:**

The progress of 80% is calculated as step 4 of 5 total items in the job.

screenshot-eV4s9A0X@2x
screenshot-AKWIGtsC@2x

Here’s a real-world example that shows how the percentage and the “step of total” being out of sync can be confusing:

320714358-3df99d17-71c7-4f47-8ffa-71828377c3ab

Source: https://github.com/putyourlightson/craft-campaign/issues/464

Ideally it would be backported to 4.x too.

### DIFF
--- a/src/queue/BaseBatchedJob.php
+++ b/src/queue/BaseBatchedJob.php
@@ -115,7 +115,6 @@ abstract class BaseBatchedJob extends BaseJob
     public function execute($queue): void
     {
         $items = $this->data()->getSlice($this->itemOffset, $this->batchSize);
-        $totalInBatch = is_array($items) ? count($items) : iterator_count($items);
 
         $memoryLimit = ConfigHelper::sizeInBytes(ini_get('memory_limit'));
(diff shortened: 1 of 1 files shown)


bossanova808:
This is great to see - the current behaviour is very confusing (at least in this context, likely any long running, countable batch process).

brandonkelly:
Thanks!! Cherry-picked your commit into the `4.9` branch (8a131a39d00e18555871414f4cd891ca99ca5baa), and also merged that into `5.1`.

brandonkelly:
4.9 and 5.1 are out now with that change.

PR #14826, opened 2024-04-17
Amends for com 165

i-just:
### Description
- If a field layout type was passed via the config, use it when rendering the field layout designer html.
- Adds `Db::parseTimestampParam` method, which parses a query param value for a timestamp column.

### Related issues
This will be needed for the com-165 item.


i-just:
Moving it to a draft as I’ll have at least one more change needed in relation to this feature.

PR #14828, opened 2024-04-18
Replace bitwise operator with logical operator

bencroker:
This PR replaces a bitwise AND operators with a logical AND operator. While the current code does of course work, the use of the bitwise AND operator seems unintentional.

I was unsure which branch to PR this to, so went with `4.x`. Please let me know if there’s a better branch to PR to, or whether it depends on the current minor version, for future reference.

### DIFF
--- a/src/models/VolumeFolder.php
+++ b/src/models/VolumeFolder.php
@@ -139,7 +139,7 @@ class VolumeFolder extends Model
                 'handle' => $volume->handle,
             ];
         } else {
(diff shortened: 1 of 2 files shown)


bencroker:
Removed the change from the `Controller` class, in which the bitwise AND operator _is_ intentional. Thanks, PhpStan!

brandonkelly:
`4.x` is perfect, thank you!

PR #14842, opened 2024-04-22
Add slight delay before opening tooltip on hover

### LINKED ISSUE #14836: [5.x]: Asset filename tooltip overlaps the "Add an asset" button in asset fields
### Description

This is a relatively minor UI annoyance 😊 but – from working with Craft 5 for a little while, I find that it's easy to accidentally trigger the filename tooltip in asset field thumbs when attempting to click the "Add" button, if the mouse pointer is moved downwards and happens to move over the truncated filename for the first thumbnail in the field, en route to the button.

If the user is a _fast clicker_, it's easy to end up on the dedicated asset edit page when attempting to click the "Add" button, as the tooltip overlaps the "Add" button and extends the clickable area for the filename link. And if the user manages to avoid that, it's still a bit of a struggle to make the tooltip disappear, in order to actually be able to click the "Add" button below it.

Some potential solves:

- Force the filename tooltip to always appear above the filename, not below it.
- Add `pointer-events: none` to the tooltip to avoid it catching the click event if/when it's obscuring the "Add an asset" button
- Add a tiny delay before the tooltip appears on filename mouseenter

https://github.com/craftcms/cms/assets/298510/7e9e97fc-08b6-48cb-84b6-04c6499b7ea3

### Steps to reproduce

1. Create an assets field, and select "Large Thumbnails" for the View Mode setting
2. Add the assets field to a field layout, and add a file to it
3. Move mouse downwards towards the "Add an asset" button

### Expected behavior

Not having to fight a tooltip when attempting to add an asset.

### Actual behavior

The tooltip is looking for a fight.

### Craft CMS version

5.0.4

### PHP version

_No response_

_No response_

### Database type and version

_No response_

### Image driver and version

_No response_

-

brianjhanson:
Fixes #14836


brandonkelly:
Element tooltips are pretty buggy on this PR. Looks like maybe it’s triggered if you hover off of an element toward the end of the delay, but before the tooltip has appeared.

https://github.com/craftcms/cms/assets/47792/d3938980-b6cf-455e-9c4c-00d5825dd331

brianjhanson:
@brandonkelly my bad, clearing the timeout should happen on `mouseleave`/`hide` otherwise you can move off the element but the timeout will still fire like you're seeing.

PR #14873, opened 2024-04-26
Fix an error with ElementSelect JS for `selectable = false`

engram-design:
When `this.settings.selectable = false` a JS error will be thrown when removing the element.

```js
Cannot read properties of undefined (reading 'length')
```


brandonkelly:
Thanks!

brandonkelly:
Craft 5.0.6 is out now with that fix. Thanks again!

Issue #14885, opened 2024-04-27
[5.x]: Extra space next to element action buttons


mlathrom:
### Description
There's extra space next to the element action buttons on the entry pages. This is caused by the activity-monitor being empty, but having margin.

### Steps to reproduce

1. Click to create an entry.
2. Scale screen down

### Expected behavior
The action buttons should be flush.

### Actual behavior
There's extra space when the activity-container is empty.

![Screenshot 2024-04-26 223541](https://github.com/craftcms/cms/assets/373889/4dc62b01-acf7-47a7-b104-a6d0725ad721)
![Screenshot 2024-04-27 163558](https://github.com/craftcms/cms/assets/373889/7f22351a-4244-4a71-be34-04e49b675625)

### Craft CMS version

5.05

### PHP version

8.2

### Installed plugins and versions

-

mlathrom:
# Option 1
Simply remove the space. Keep button order.

PR: #14887

![Screenshot 2024-04-26 223608](https://github.com/craftcms/cms/assets/373889/98762d94-db67-4f79-b5cd-d3fd8783c246)


# Option 2
Remove space, but also reverse the order of buttons. It feels right for the button order to be determined by which side they're aligned to.

PR: #14886

![Screenshot 2024-04-27 163404](https://github.com/craftcms/cms/assets/373889/ab9f6bbb-65d8-4a56-9b11-1abf500b18ca)

brandonkelly:
Thanks for reporting, and the PRs! Craft 5.1 is out with a similar solution to your second suggestion (#14898).

PR #14886, opened 2024-04-28
Fix extra space next to element action buttons (Option 2)

mlathrom:
### Description

- Remove extra space next to element entry page action buttons
- Reverse order of action buttons on small screens

#### Issue:
Screenshot 2024-04-26 223541

#### Fix:
Screenshot 2024-04-27 163404

### Related issues
#14885

### Alternative Fix Option
#14887


brianjhanson:
Closing this one in favor of #14898 which is just this, but removes the `:empty { display: none }` because with this solution we can keep the markup around without creating awkward spaces. 

Thanks for the PR!

PR #14897, opened 2024-04-29
Accept closure for backup/restore commands

timkelty:
### Description
- adds `\Closure` as an accepted type for `\craft\config\GeneralConfig::$restoreCommand` and `\craft\config\GeneralConfig::$backupCommand`
  - `function(ShellCommand $command) => ShellCommand`
- adds `\craft\config\GeneralConfig::$backupCommandFormat` (pg only) for using archive format
  - also changes restore to use `pg_restore` to be compatible with create dumps
  - changes extension to be `.dump`

### Example usage

```
<?php
use craft\config\GeneralConfig;

    ->backupCommandFormat('custom')
    ->backupCommand(fn(\mikehaertl\shellcommand\Command $command) => $command
        ->addArg('--clean')
        ->addArg('--if-exists')
    );
```

### Related issues
Replaces https://github.com/craftcms/cms/pull/14586

### Notes
- For MySQL, both the "schema dump" and the "data dump" get passed through the `\Callable`. This would allow someone to have different behavior in either one by looking for `--no-data` or `--no-create-info`.
- It might make more sense to have `backupCommandFormat` be in `\craft\config\DbConfig`, but it felt weird to have it separated from `backupCommand`.

### DIFF
--- a/src/console/controllers/DbController.php
+++ b/src/console/controllers/DbController.php
@@ -214,7 +214,7 @@ class DbController extends Controller
             $db->backupTo($path);
             if ($this->zip) {
                 $zipPath = FileHelper::zip($path);
-                unlink($path);
+                FileHelper::unlink($path);
                 $path = $zipPath;
             }
         } catch (Throwable $e) {
--- a/src/config/GeneralConfig.php
+++ b/src/config/GeneralConfig.php
@@ -7,6 +7,7 @@
 
 namespace craft\config;
 
+use Closure;
 use Craft;
 use craft\helpers\ConfigHelper;
 use craft\helpers\DateTimeHelper;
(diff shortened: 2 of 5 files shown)


src/console/controllers/DbController.php
```diff
@@ -214,7 +214,7 @@ public function actionBackup(?string $path = null): int
             $db->backupTo($path);
             if ($this->zip) {
                 $zipPath = FileHelper::zip($path);
-                unlink($path);
+                FileHelper::unlink($path);
```
timkelty, line 217: Using helper, as theres a chance `$path` will be a directory if you set `backupCommandFormat=directory`.

Issue #14900, opened 2024-04-30
[5.x]: UI and layout issues with cp header and sidebar


mlathrom:
### Description
I think the layout of the header and sidebar on control panel pages could be cleaner. There are also some spacing inconsistencies in at different breakpoints.

### Expected behavior
- Clean layout at all resolutions
- Skip to Entries button should remain in sidebar element
- Toolbar should always line up with main content

### Actual behavior
- "Skip to Entries" button is out of the sidebar on mobile
- Toolbar, action buttons, and show sidebar interface is a bit clunky in it's layout
- The toolbar doesn't line up with the main content in tablet resolution

#### Mobile Layout
Lots of vertical space used.

![mobile-layout](https://github.com/craftcms/cms/assets/373889/93d5a249-8f04-40ae-aad1-8460c93cb562)

#### Mobile Layout
Skip to entries button out of sidebar container.

![mobile](https://github.com/craftcms/cms/assets/373889/0deff251-9e92-4edd-951a-19e581b9ea52)

#### Tablet layout
Toolbar is not aligned with main content. Strange spacing.

![tablet](https://github.com/craftcms/cms/assets/373889/7be26e91-74ec-4538-8813-9e4b4241b5a1)

#### Desktop Layout
Odd spacing for Skip to Entries button

![desktop](https://github.com/craftcms/cms/assets/373889/1647bd13-b748-4374-9017-fced573f6aca)

### Craft CMS version

5.05

### PHP version

8.2

### Installed plugins and versions

-

mlathrom:
@brianjhanson I've created a fix for this. Hope it helps.

#14901

brandonkelly:
Craft 5.10.0-beta.1 is out with changes from #14901 and #14910. Thanks for the feedback and PR!

PR #14901, opened 2024-04-30
Fix UI and layout issues with cp header and sidebar

mlathrom:
### Description
I've optimized the layout and fixed the spacing issues with the cp header and sidebar. I've tried to keep structural changes to a minimum. I think this layout feels much cleaner while maintaining accessibility.

The only minor concession is that the tab index remains the same even though the New Entry button is now visually before the toolbar in mobile layout. But since this is for the mobile layout, it doesn't seem like a big deal.

- Fixed inconsistent padding in sidebar
- Fixed "Skip to Entries" button being out of the sidebar element on mobile
- Turned the element index heading into a dropdown button that reveals the sidebar, rather than having a show sidebar button
- Changed order of toolbar an action buttons to visually flow nicer and conserve vertical space.

#### New Mobile Layout

new-mobile

#### New Mobile Layout - Sidebar opened

new-mobile-2

#### New Mobile Layout - Skip Button

new-mobile-skip-button

#### New Tablet Layout - Fixed spacing

new-tablet

#### Desktop Layout
Remains the same. The Skip to Entries button spacing is fixed.

desktop

### Related issues
#14900

### DIFF
--- a/src/templates/_layouts/cp.twig
+++ b/src/templates/_layouts/cp.twig
@@ -221,12 +221,12 @@ history.replaceState(undefined, undefined, window.location.href.match(/^[^#]*/)[
                                             {% endblock %}
                                         </div>
                                         {% if toolbar %}
-                                            <div id="toolbar" class="flex">
+                                            <div id="toolbar" class="flex" tabindex="2">
                                                 {{ toolbar|raw }}
                                             </div>
                                         {% endif %}
                                         {% if actionButton or additionalButtons %}
-                                            <div id="action-buttons" class="flex">
+                                            <div id="action-buttons" class="flex" tabindex="1">
                                                 {{ additionalButtons|raw }}
                                                 {{ actionButton|raw }}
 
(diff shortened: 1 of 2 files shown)


mlathrom:
I tested this on all of the standard Craft CMS pages and seems to be a pretty big UI improvement.

I also tested this with the Seomatic plugin to make sure it didn't negatively impact a widely used plugin.

One thing I only caught after pushing this pull request is that you can likely remove the following style now:

```
@media only screen and (max-width: 60.8125rem) {
    #header .flex:not(#toolbar) {
        margin-top: 10px;
    }
}
```

It creates weird spacing for the Plugin Store cart icon.

mlathrom:
The only place this seems slightly odd is the Utilities page, where the Heading and button text are the same. But this too seems like a small concession.

image

But the Utilities page deviates from the way the other pages work because it's page title changes based on the selected sidebar menu item.

And I think the fix here is to make the Utilities page consistent with the others.

utilities

brianjhanson:
Closing in favor of #14910

PR #14921, opened 2024-05-02
db/backup/restore refinements

timkelty:
### Description
https://github.com/craftcms/cms/pull/14897 tied the format set for backups (`backupCommandFormat`) with what is used for restoring, which is not ideal.

This PR:
- separates `backupCommandFormat` from what is used for restore
- adds `--format` options for `db/restore` and `db/backup` (for PG)
- For `db/restore`, if no `--format` is passed, it will attempt to detect which pg_restore format should be used (`custom`, `directory`, `tar`)
- Adds flags to the default restore command to allow postgres dumps to work (including Craft Cloud).
- Adds some allowances so the path for `db/restore` can be a directory, which is applicable for pg_restore.


src/db/pgsql/Schema.php
```diff
@@ -164,6 +167,17 @@ public function getDefaultRestoreCommand(): string
...
+            $command
+                ->addArg('--clean')
+                ->addArg('--if-exists')
+                ->addArg('--no-owner')
+                ->addArg('--no-acl')
+                ->addArg('--schema=', '{schema}')
+                ->addArg('--single-transaction')
+                ->addArg('{file}');
+        }
+
```
timkelty, line 180: When using pg_restore, these args need to be on the restore, regardless if the backup was created with them.

src/db/pgsql/Schema.php
```diff
@@ -172,7 +186,7 @@ public function getDefaultRestoreCommand(): string
 
         return $this->_pgpasswordCommand()
             . $command->getExecCommand()
-            . '< "{file}"';
+            . ($this->usePgRestore() ? '' : '< "{file}"');
```
timkelty: Can't use STDIN with pg_restore, because directory is a valid input.

src/console/controllers/DbController.php
```diff
@@ -242,8 +256,8 @@ public function actionBackup(?string $path = null): int
      */
     public function actionRestore(?string $path = null): int
     {
-        if (!is_file($path)) {
-            $this->stderr("Backup file doesn't exist: $path" . PHP_EOL);
+        if (!is_readable($path)) {
+            $this->stderr("Backup path doesn't exist: $path" . PHP_EOL);
```
timkelty, line 260: Account for a possible directory (pg_restore)

timkelty:
See https://github.com/craftcms/cms/pull/14931

Issue #14922, opened 2024-05-02
[5.x] [4.x]: Database backup broken with 5.1.0 and 4.9.0


erinbit:
### Description
The recent release of 4.9.0 and 5.1.0 has a bug which makes MySQL backups fail. This is because in this commit, the logic of when to use `--single-transaction` was reversed. 
https://github.com/craftcms/cms/pull/14897/commits/97da6ca6d5cc73700c332f0a89e9a924cb9651cf#diff-83d39a4592b31403e3830a3100e22c596322d8dd9a240e07f6ecf173ec1d7bb7

This condition should be reversed again, to make mysqldump work, see pull request https://github.com/craftcms/cms/pull/14925

#### Breaking change in detail
Before the commit if mysql version is newer than 5.7.41 or 8.0.32 results in **FALSE** 
```
if (($isMySQL5 && version_compare($serverVersion, '5.7.41', '>=')) ||
     ($isMySQL8 && version_compare($serverVersion, '8.0.32', '>='))) {
     $useSingleTransaction = false;
 }
```

After the commit if mysql version is newer than 5.7.41 or 8.0.32 results in **TRUE**
```
 $useSingleTransaction =
     ($isMySQL5 && version_compare($serverVersion, '5.7.41', '>=')) ||
     ($isMySQL8 && version_compare($serverVersion, '8.0.32', '>='));
```

#### Why do we even check MySQL versions?
This check is done because after those versions, mysqldump will require the FLUSH TABLES privileges to run, which shared hosting usually will not give, as it is considered system privileges. Thus breaking any use of mysqldump. (See https://github.com/craftcms/cms/issues/12557)

Here at [fortrabbit](https://www.fortrabbit.com) hosting we try to work around this by setting our defult `mysqldump.conf` to always disable the GTID option, which also removes the need for FLUSH TABLES privileges on mysqldump versions 5.7.42/8.0.33 or later. (See https://bugs.mysql.com/bug.php?id=109685)
But Craft CMS overrides the default config file with their own config, so this workaround is disabled.

### Steps to reproduce

1. Have a craft install that runs on shared hosting, like [fortrabbit](https://www.fortrabbit.com)
2. Upgrade to or install to Craft CMS 4.9.0 or 5.1.0
3. Open control panel, go to Utilities -> Database Backup
4. Click backup, and you will get this error in the logs
`
[...]

PR #14925, opened 2024-05-02
correct logic for database backup use of single-transaction

erinbit:
### Description
The recent release of 4.9.0 and 5.1.0 has a bug which makes MySQL backups fail. Because in this commit, the logic of when to use `--single-transaction` was reversed. 
https://github.com/craftcms/cms/pull/14897/commits/97da6ca6d5cc73700c332f0a89e9a924cb9651cf#diff-83d39a4592b31403e3830a3100e22c596322d8dd9a240e07f6ecf173ec1d7bb7

#### Breaking change in detail
Before the commit if mysql version is newer than 5.7.41 or 8.0.32 results in **FALSE** 
```
if (($isMySQL5 && version_compare($serverVersion, '5.7.41', '>=')) ||
     ($isMySQL8 && version_compare($serverVersion, '8.0.32', '>='))) {
     $useSingleTransaction = false;
 }
```

After the commit if mysql version is newer than 5.7.41 or 8.0.32 results in **TRUE**
```
 $useSingleTransaction =
     ($isMySQL5 && version_compare($serverVersion, '5.7.41', '>=')) ||
     ($isMySQL8 && version_compare($serverVersion, '8.0.32', '>='));
```

#### Why do we even check MySQL versions?
This check is done because after those versions, mysqldump will require the FLUSH TABLES privileges to run, which shared hosting usually will not give, as it is considered system privileges. Thus breaking any use of mysqldump. (See https://github.com/craftcms/cms/issues/12557)

### Related issues

This fixes: https://github.com/craftcms/cms/issues/14922

### DIFF
--- a/src/db/mysql/Schema.php
+++ b/src/db/mysql/Schema.php
@@ -172,9 +172,10 @@ class Schema extends \yii\db\mysql\Schema
         $commandFromConfig = Craft::$app->getConfig()->getGeneral()->backupCommand;
 
         // https://bugs.mysql.com/bug.php?id=109685
-        $useSingleTransaction =
+        $useSingleTransaction = !(
             ($isMySQL5 && version_compare($serverVersion, '5.7.41', '>=')) ||
-            ($isMySQL8 && version_compare($serverVersion, '8.0.32', '>='));
+            ($isMySQL8 && version_compare($serverVersion, '8.0.32', '>='))
+        );
 
         if ($useSingleTransaction) {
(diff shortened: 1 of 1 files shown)


timkelty:
@erinbit doh! Thanks so much for catching this 😍

brandonkelly:
Craft 4.9.1 and 5.1.1 are out with that fix. Thanks again @erinbit!

PR #14931, opened 2024-05-02
DB backup/restore refinements

timkelty:
### Description
https://github.com/craftcms/cms/pull/14897 tied the format set for backups (`backupCommandFormat`) with what is used for restoring, which is not ideal.

This PR:
- separates `backupCommandFormat` from what is used for restore
- adds `--format` options for `db/restore` and `db/backup` (for PG)
- For `db/restore`, if no `--format` is passed, it will attempt to detect which pg_restore format should be used (`custom`, `directory`, `tar`)
- Adds flags to the default restore command to allow postgres dumps to work (including Craft Cloud).
- Adds some allowances so the path for `db/restore` can be a directory, which is applicable for pg_restore.

### DIFF
--- a/CHANGELOG-WIP.md
+++ b/CHANGELOG-WIP.md
@@ -1,1 +1,10 @@
 # Release notes for Craft CMS 4.10 (WIP)
+
+### Administration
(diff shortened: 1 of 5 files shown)


ezawadzki:
You make my day :100: Thank you

PR #14935, opened 2024-05-03
A11y/core fixes

gcamacho079:
### Description
A few quick fixes

### Related issues
Resolves PT-1665, PT-1689, PT-1670, PT-1667, PT-1690


src/templates/_layouts/components/global-sidebar.twig
```diff
@@ -131,12 +131,13 @@
             </div>
 
             {% if currentUser.admin and devMode %}
-                {% set devModeText = 'Craft CMS is running in Dev Mode.'|t('app') %}
+                {% set devModeText = '<span lang="en">Craft CMS</span> is running in Dev Mode.'|t('app') %}
```
gcamacho079: @brandonkelly we should update all instances of "Craft CMS" and other brand names to include a wrapper with `lang="en"`. Is changing the original translation string the best way of going about this, or should a new one be added?

brandonkelly: Yikes, there are 26 strings with `Craft CMS` in them.

Simply adding `<span>`s to them won’t really work because HTML entities are escaped by default when output by Twig tags. So we’ll need to go through each of them and make sure it’s getting output with a `|raw` filter. And for JS translations, make sure that they’re not getting escaped there either.

It probably makes the most sense to keep the source translation _keys_ the same, so we don’t have to update the message text within the PHP/Twig source files; just the translated messages (what follows the `=>`s in the translation files).

gcamacho079: @brandonkelly sounds good! Just updated the PR per your feedback. The rest are lower priority since only a couple were in the representative sample we used for the audit, so I’ll update those separately.

brianjhanson (approved):

PR #14910, opened 2024-04-30
CP sidebar UI fixes

brianjhanson:
Thanks to @mlathrom again for the majority of the work! I only made a few tweaks. 

Fixes #14900

### Before
CleanShot 2024-04-30 at 13 14 19
CleanShot 2024-04-30 at 13 14 24
CleanShot 2024-04-30 at 13 14 09

### After
CleanShot 2024-04-30 at 13 09 07
CleanShot 2024-04-30 at 13 09 18
CleanShot 2024-04-30 at 13 09 33

PR #14946, opened 2024-05-06
Changes to source navigation and element index UI, user-related document titles

gcamacho079:
### Description

- Reverted a couple changes from #14910:
  - Ensure consistent labeling of the sidebar toggle button across all element index screens (previously fixed in #11169). 
  - Reset tab order in the element index toolbar to ensure predictable focus order

A couple of new changes:
- To communicate context inside an element index on the mobile layout when the sidebar is closed, the source name has been added to the `h1` (see screenshot).
- On the mobile element index layout, the sidebar is kept open on source selection to prevent focus from being dropped (Resolves PT-1303)
- User page document titles have been updated to describe the page purpose (Resolves PT-1701)

Element index screen on mobile. The h1 includes the source name so that users are aware of the active source when the sidebar is initially toggled close.


brianjhanson (approved):
I adjusted the styling just a bit, but otherwise looks good!

CleanShot 2024-05-21 at 11 32 16

gcamacho079:
> I adjusted the styling just a bit, but otherwise looks good!

Thanks, @brianjhanson; I like the adjustments you made! 💯

brandonkelly:
Is this `/` after “My Account” intentional?

Screenshot of the top of the My Account page on mobile

brianjhanson:
@brandonkelly nope. Fixed in https://github.com/craftcms/cms/pull/14946/commits/cd6a015566db038c470005e19ed1965c2349dd74

Issue #14947, opened 2024-05-06
[5.x]: Validation Error thrown "slug cannot be blank" on inline matrix fields with a single required entry type were the title is hidden, the slug is in the URI


brionwolf:
### Description

I realize this is a _**very**_ niche issue, but sharing: craft throws a validation error stating "slug cannot be blank" on matrix fields where Site settings include the entry's slug in the URL Format, the "Min Entries" is set to greater than one, the "View mode" is _As inline-editable blocks_, and the title/slug are hidden fields on the Associated Entry Type.

### Steps to reproduce

1. Create an entry type with various fields
2. Set the Title and slug to hidden, and include logic for creating a title _(Doesn't matter if the logic is just a common string, or dynamic based on custom fields)_.
3. Associate the entry type with a matrix field as the only entry type
4. Set an Entry URI format that includes the entry's slug.
5. Set the Min entries option to a number greater than zero
6. Set the View mode to "As inline-editable Blocks"
7. Assign the matrix field to a section that can be edited and attempt to create an entry

### Expected behavior

- Craft creates the Min-Entries set of entries for the matrix field automatically without errors.

### Actual behavior

- Craft creates a single block _(even if the "Min Entries" is set to more than one)_, throws an error "Could not save entry" _(with 404 errors in Yii console)_ when attempting to autosave, and throws a validation error "Slug cannot be blank" when manually saving.

#### Screencaptures:

![animated screen capture or the craft CMS control panels depicting the above issue](https://github.com/craftcms/cms/assets/5532234/e11f6ece-eddd-4fab-ac88-a43370991bd3)

### Notes/workarounds

- Issue doesn't occur if I do not hide the title field _(Slug field can remain hidden)_.
- Issue doesn't occur if I set the matrix field view to "As cards" or "As an element index" _(where I need to create a new entry for the field either from a slide-out drawer or new tab)_.
- Issue doesn't occur if I use another unique identifier in the URI format _(ie: the entry id)_.
- Issue doesn't occur if I set the "Min Entries" field to 0, while still setting the field to "required" on associated field layouts.

### Craft CMS version

Craft Pro 5.1.0

### PHP version

8.3.4

### Operating system and version

MacOS Sonoma 14.4.1 (23E224)

### Database type and version

MySQL 8.3.0

### Image driver and version

GD 8.3.4

### Installed plugins and versions

- CKEditor | 4.0.4

i-just:
[...]

PR #14954, opened 2024-05-07
don't set cache results to empty array for nested matrix entry with b…

i-just:
### Description
In a matrix in a matrix setup where both matrix fields are in “as inline-editable blocks” mode and the nested matrix has min entries set to at least 1, the title is hidden, and URI format references the slug; if you add a new top-level matrix entry, the draft won’t be saved because of slug validation issue. 

Replication steps:
- create `matrixEt1` entry type with “show the title field” and “show the skug field” disabled
- create `myMatrix1`field, type: matrix, entry type: `matrixEt1`, URI: sth/{slug}, min entries: 1, view mode: blocks
- create a `matrixEt2` entry type with the `myMatrix1` field in it
- create `myMatrix2` field, type: matrix, entry type `matrixEt2`, view mode: blocks
- create a `contentBuilder` entry type with the `myMatrix2` field in it
- create a section that uses the `contentBuilder` entry type
- create an entry in the above section & add an entry (block) to the `myMatrix2`

The issue is triggered because when the entry (block) is added, [the empty value is being cached](https://github.com/craftcms/cms/blob/5.x/src/fields/Matrix.php#L681), causing the nested entry to be an [empty “skeleton”](https://github.com/craftcms/cms/blob/5.x/src/fields/Matrix.php#L986) one without a title or slug, which prevents the slug from being correctly set based on the title attribute.

If you reload the page after adding the entry (block), the draft will start being saved as expected.

This PR excludes nested matrix entries from being cached as empty array when normalising field value from request if it’s a nested entry in the blocks view mode.

Screenshot of the replication setup: 
Screenshot 2024-05-07 at 17 27 42

### Related issues
#14947

### DIFF
--- a/src/fields/Matrix.php
+++ b/src/fields/Matrix.php
@@ -677,7 +677,10 @@ class Matrix extends Field implements
 
         // Set the initially matched elements if $value is already set, which is the case if there was a validation
         // error or we're loading an entry revision.
-        if ($value === '') {
+        // But we should exclude matrix entries in block view nested in matrix entries from this caching:
+        // https://github.com/craftcms/cms/issues/14947
+        /** @var Entry $element */
+        if ($value === '' && !($fromRequest && $this->viewMode === self::VIEW_MODE_BLOCKS && $element->fieldId && $element->saveOwnership)) {
             $query->setCachedResult([]);
         } elseif ($element && is_array($value)) {
             $query->setCachedResult($this->_createEntriesFromSerializedData($value, $element, $fromRequest));


brandonkelly:
@i-just Thanks for figuring this out! I hadn’t been able to reproduce with the steps listed in #14947, but could with these ones.

I was able to resolve it in a different way, by delaying the time when those initial blocks get auto-created until after the Matrix field is fully initialized. (07e944d77a1b9697560c8aef1804ae6f520f3697)

PR #14974, opened 2024-05-10
Allow easier Monolog config

timkelty:
### Description
The most common scenario is to want to add a Monolog handler for a third party service. This is an attempt to make that a bit easier.

As it is, `MonologTarget` is very much geared toward Craft's use of it and is a bit difficult to extend.  One could always just add a target using `samdark\log\PsrTarget`, (which `MonologTarget` extends), but you miss out on the default formatting and processors that Craft applies.

This PR allows `MonologTarget` to be constructed with a Closure for the `logger` property, allowing you to override anything.

##### Before:
```php
<?php
return [
    'components' => [
        'log' => function() {
            $target = new \craft\log\MonologTarget(['name' => 'myTarget']);
            $target->getLogger()->setHandlers([
                (new \Monolog\Handler\StreamHandler('php://stderr', \Monolog\Logger::ERROR))
                    ->setFormatter($target->getFormatter()),
            ]);
        
            $dispatcher = new \craft\log\Dispatcher();
            $dispatcher->targets[] = $target;
        
            return $dispatcher;
        },
    ],
];
```

##### After:
```php
<?php
// After
return [
    'components' => [
        'log' => [
            'targets' => [
                [
                    'class' => \craft\log\MonologTarget::class,
                    'name' => 'myTarget',
                    'logger' => fn(\craft\log\MonologTarget $target) => $target
                        ->getLogger()
                        ->setHandlers([
                            (new \Monolog\Handler\StreamHandler('php://stdout'))
                                ->setFormatter($target->getFormatter()),
                        ]
                    ),
                ]
            ]
        ],
    ],
];
```

##### …or, push a handler onto Craft's default log targets:
```php
<?php
return [
    'components' => [
        'log' => [
            'monologTargetConfig' => [
                'logContext' => false,
                'logger' => fn(\craft\log\MonologTarget $target) => $target
                    ->getLogger()
                    ->pushHandler(
                        (new \Monolog\Handler\StreamHandler('php://stdout'))
                            ->setFormatter($target->getFormatter())
                    ),
            ],
        ],
    ],
];
```


src/log/MonologTarget.php
```diff
@@ -264,6 +285,11 @@ public function setFormatter(?FormatterInterface $formatter): void
         $this->_setLoggerProperty('formatter', $formatter);
     }
 
+    public function getFormatter(): ?FormatterInterface
+    {
+        return $this->formatter;
+    }
```
timkelty, line 345: This should be gettable so you can set the configured formatter when setting your handler.

src/log/MonologTarget.php
```diff
@@ -186,28 +224,11 @@ private function _createLogger(string $name): Logger
...
-            $logger->pushHandler((new RotatingFileHandler(
-                App::parseEnv(sprintf('@storage/logs/%s.log', $name)),
-                $this->maxFiles,
-                $this->level,
-                filePermission: $generalConfig->defaultFileMode,
-            ))->setFormatter($this->formatter));
-        }
+        $handlers = $this->handlers instanceof Closure
+            ? ($this->handlers)($this)
+            : $this->handlers;
```
timkelty: If `$this->handlers` is an array, we take it as-is.

This makes the most intuitive sense to me, however, it _isn't_ how `\craft\log\Dispatcher::$targets` works, where they are appended to the defaults. Happy to change it either way.

src/log/MonologTarget.php
```diff
@@ -141,6 +153,33 @@ public function export(): void
...
+                App::parseEnv(sprintf('@storage/logs/%s.log', $this->name)),
+                $this->maxFiles,
+                $this->level,
+                filePermission: Craft::$app->getConfig()->getGeneral()->defaultFileMode,
+            ))->setFormatter($this->formatter);
+        }
+
+        return $handlers;
+    }
+
```
timkelty: `getDefaultHandlers` is not static because it needs access to `$this->formatter` and `$this->name`.

Its intended use is to be referenced in `handlers` when a Closure is used (when you want to append a handler to the defaults).

src/log/MonologTarget.php
```diff
@@ -217,6 +224,11 @@ public function setName(string $name): void
         $this->_setLoggerProperty('name', $name);
     }
 
+    public function getName(): string
+    {
+        return $this->name;
+    }
+
```
timkelty, line 236: Adds getters for all these so they can be used in the Logger closure

PR #14975, opened 2024-05-10
Show provisional draft data in element chips/cards/tables

brandonkelly:
When an element has a provisional draft, its chips, cards, and index table data will now show the provisional draft’s data, as well as an “Edited” label.

An entry index in table view, showing that one entry has an “Edited” label because it has provisional draft data.

An entry index in card view, showing that one entry has an “Edited” label because it has provisional draft data.

When editing an element via a slideout, its chips, cards, and table data will automatically refresh to reflect the latest changes (including its “Edited” label).


gcamacho079 (changes requested):
This looks great, @brandonkelly! The only change I would suggest is:

- [x] In the card view, it would be helpful to put the status badges in a `ul`.

brandonkelly:
Done! Thanks @gcamacho079

brettburwell:
This is a sneaky awesome update. Bravo! 

It's been fun seeing all of these detail polishing updates drop in the 5.2 release. Keep up the great work.

Issue #15001, opened 2024-05-15
[4.x]: Change in ElementQuery::count effecting `EVENT_BEFORE_PREPARE` events


myleshyson:
### Description
We noticed that after updating to Craft 4.9, one of our queries wasn't running properly. The logic for this particular query depends on the `EntryQuery::EVENT_BEFORE_PREPARE` event. In particular, in depended on the `orderBy` property to be null for for count queries to run properly.

Now however, because of [this](https://github.com/craftcms/cms/commit/0ff4796abd9d073d8c3d15a934e381f90cb403fc) change, the `orderBy` property is not being nulled like it was before with `yiii\db\Query::queryScaler` method.

I'm not sure if this was on purpose? Either way would it be possible to either:
a) keep the queryScaler logic from `yiii\db\Query::queryScaler` for scalar queries (count, min, max, etc) so that things run like they did before the change OR
b) add a property to ElementQuery, something like `queryMethod`, so that events can add that logic back if the query is a scalar query? 

### Craft CMS version

Craft 4.9.0

### PHP version

8.2

yannkost:
I have also noticed that the .count() function does not return the right value for some queries. 

```
{% set searchQuery = craft.entries({ type: availableEntryTypes}).search(query).orderBy('score').limit(5) %}
```
This sample code does return the limit of 5 and not the number of the entries.

yannkost:
Just checked and without the `orderBy('score')` the function call returns the correct number.

i-just:
@myleshyson, in your case, the `EntryQuery::EVENT_BEFORE_PREPARE` used to rely on the `$event->sender->orderBy` returning `null`, but since 4.9, it’s returning an empty string.

@yannkost, in your case, it’s the combination of `search`, `limit` and `orderBy('score')`. 

I raised a draft PR, and we’ll discuss this internally.

brandonkelly:
Craft 4.9.5 and 5.1.5 are out with a fix that covers both of these issues. Thanks for reporting!

lindseydiloreto:
This fix seems to have broken proximity searches for both the [Google Maps](https://github.com/doublesecretagency/craft-googlemaps/issues/111) and [Maps](https://github.com/ethercreative/simplemap/issues/391) plugins.

I've been digging around through the plugins and Craft itself, but can't figure out exactly:
- What benefit did these changes provide?
- How should our proximity search plugins be adjusted accordingly?

Do you have any recommendations?

[...]

PR #15020, opened 2024-05-17
keep the "old" scalar queries behaviour with offset or limit

i-just:
### Description
The performance changes from https://github.com/craftcms/cms/commit/0ff4796abd9d073d8c3d15a934e381f90cb403fc are causing slightly different behaviour when running scalar element queries.

The first reported case points out that before 4.9, in the `EntryQuery::EVENT_BEFORE_PREPARE`, for the scalar queries, the `$event->sender->orderBy` used to return `null`, but now it’s returning an empty string.

The second reported case is a combination of `search`, `limit`, and `orderBy('score')` params. Prior to 4.9, running `count()` on a query that contained all three would still return the total count, but now, it respects the passed-in limit. That’s because, in this case, the prepared subquery contains the IDs of the searched elements and the query that got those IDs respected the set limit. (Using the `offset` param would result in the same non-total count now.)

I raised a PR to go back to the “old” approach when the query contains `offset` or `limit`. 

I’m raising this as a draft in case we want to discuss this further in the CMS meeting.

### Related issues
#15001

### DIFF
--- a/src/elements/db/ElementQuery.php
+++ b/src/elements/db/ElementQuery.php
@@ -1608,6 +1608,10 @@ class ElementQuery extends Query implements ElementQueryInterface
             return count($cachedResult);
         }
 
+        if ($this->offset || $this->limit) {
+            return parent::count($q, $db) ?: 0;
+        }
+
         try {
             return $this->prepareSubquery()->count($q, $db) ?: 0;
         } catch (QueryAbortedException) {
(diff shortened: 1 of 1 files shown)


brandonkelly:
Resolved via dc36833c359e85e3fd9d25b0a3cc6aa90920a640 while still maintaining the performance improvement.

Issue #15015, opened 2024-05-16
[5.x]: "User not authorized to create this element." after cancelling the creation of a block on matrix field


brionwolf:
### Description

- Clicking away or clicking "cancel" while creating an entry in a drawer for a matrix field works as expected (cancelling the creation of that block), but clicking the "add block" button again will throw a "User not authorized to create this element." error.

### Steps to reproduce

1. Create a matrix field, with at least one entry type.
2. Set a `Max Entries` value _(ie: 1 or 2)_.
3. Set the view mode to "As an element index" or "As cards"
4. Add the matrix field to another field layout for testing.
5. Create or edit an entry with the matrix field.
6. Click the button to add a new block to the matrix field
7. Once the entry create drawer opens, cancel the creation of the entry.
8. Repeat steps 6 and 7 until you have done so once-more than the "Max Entries" value.

### Expected behavior

- I should be able to cancel and/or add a block from the matrix field as many times as I want.

### Actual behavior

- Craft throws an error stating "User not authorized to create this element.", and the element cannot be recreated.
- If the `Max Entries` value is greater than 1, one of the "canceled" entries may appear as a draft entry after reloading the page, or attempting to save.
- If the matrix block is nested in the field layout of another matrix block I occasionally have to delete and create the parent block to resolve the error. 

### Screen captures:

![animated image of a craft control panel reproducing an example of the error](https://github.com/craftcms/cms/assets/5532234/9364cf18-cc24-475c-944b-9ef063409872)

### Stack Trace:

```shell
yii\web\ForbiddenHttpException: User not authorized to create this element. in craft/vendor/craftcms/cms/src/controllers/ElementsController.php:2186
Stack trace:
#0 craft/vendor/craftcms/cms/src/controllers/ElementsController.php(194): craft\controllers\ElementsController->_createElement()
#1 [internal function]: craft\controllers\ElementsController->actionCreate()
#2 craft/vendor/yiisoft/yii2/base/InlineAction.php(57): call_user_func_array(Array, Array)
#3 craft/vendor/yiisoft/yii2/base/Controller.php(178): yii\base\InlineAction->runWithParams(Array)
#4 craft/vendor/yiisoft/yii2/base/Module.php(552): yii\base\Controller->runAction('create', Array)
#5 craft/vendor/craftcms/cms/src/web/Application.php(349): yii\base\Module->runAction('elements/create', Array)
[...]

PR #15029, opened 2024-05-20
don't count unpublished drafts towards the total

i-just:
### Description
When getting the total number of entries in a matrix field, don't include unpublished drafts.

### Related issues
#15015


brandonkelly:
I don’t think there’s a good reason to include _any_ drafts in that logic. E.g. if someone had explicitly created a draft on a nested entry from its fullscreen edit page, that shouldn’t count toward the entry count.

So just removed the `drafts(null)` bit and it seems to be working as expected across all view modes.

Issue #15027, opened 2024-05-19
[5.x]: Alt Text Field Cannot Be Bulk Edited


theAdhocracy:
### Description

When using the Bulk Edit mode on an Assets view, the Alternative Text field remains uneditable.

### Steps to reproduce

1. Create any amount of Assets with the Alternative Text field enabled;
2. Create a view that lists those Assets;
3. Press the Edit button to enter Bulk Edit mode.

### Expected behavior

Alternative Text field to be editable.

### Actual behavior

![image](https://github.com/craftcms/cms/assets/7009462/f80f48f1-3e90-47a8-83b5-f6d2539df375)

### Craft CMS version

5.1.4

### PHP version

8.2

### Installed plugins and versions

-

i-just:
Hi, thanks for reporting! I raised a PR for this.

brandonkelly:
This was brought up previously by #14893, and has been added for 5.2 (via aa4d6db7a8c4485853e24fefae513f18fc19eaff).

PR #15031, opened 2024-05-20
allow alt text to be inline editable

i-just:
### Description
Adds support for inline editing of the Alt Text native field.

### Related issues
#15027


brandonkelly:
Already done for 5.2 via aa4d6db7a8c4485853e24fefae513f18fc19eaff :)

PR #15038, opened 2024-05-21
show nested elements in embedded index for revisions

i-just:
### Description
When viewing revisions, the nested entries weren’t showing for matrix fields with element index view mode.

**Replication steps:**
- create a matrix field with view mode: element index
- create an entry type with that matrix field
- create a section with that entry type and have revisions on
- create an entry in that section, add 2 entries to the matrix field and save
- remove one of the nested entries from the matrix field and save
- view the first revision - both nested entries should show in the matrix field, but without this change, none were showing


brandonkelly:
Thanks!

Issue #15037, opened 2024-05-21
[5.x]: User conditional not applied when starting project, inspite it being set in the project config


boboldehampsink:
### Description

A user conditional with the following rules is not applied when starting a project for the first time:

```
userCondition:
              class: craft\elements\conditions\users\UserCondition
              conditionRules:
                -
                  class: craft\elements\conditions\users\GroupConditionRule
                  operator: in
                  uid: 34b06532-29f1-44a6-9acf-ba26f8a7389e
                  values:
                    - 21006657-ecd9-43d2-8813-4cfc3db35488 # **USERGROUP**
              elementType: craft\elements\User
              fieldContext: global
```

### Steps to reproduce

1. Create a usergroup
2. Create an entry type
3. Add a field with the condition as seen above
4. Verify that it works
5. Close down project
6. Start it up from scratch

### Expected behavior

User condition in place

### Actual behavior

No user condition in place, inspite it being saved to PC

### Craft CMS version

5.1.2

### PHP version

8.3

### Database type and version

PostgreSQL

### Installed plugins and versions

-

i-just:
Hi, thanks for reporting! I raised a PR for this.

brandonkelly:
Craft 4.9.6 and 5.1.8 are out now with that fix. Thanks again!

PR #15048, opened 2024-05-22
user group condition selectable when applying pc

i-just:
### Description
Allow the User Group condition to be selectable if we’re applying external project config changes (if the changes are likely to add user group(s).

At the moment, if you’re installing a new project using the existing config (or if you were to add the first user group and a rule that depends on it to an existing project in one go), the User Group condition won’t be added as the `craft\elements\conditions\users\GroupConditionRule->isSelectable()` will always be `false` (because there are no groups for the project yet). Since adding user groups can’t happen earlier on, we need to check if the user groups are likely to be added when applying changes, and if so, allow this rule to be selectable.

### Related issues
#15037

### DIFF
--- a/src/elements/conditions/users/GroupConditionRule.php
+++ b/src/elements/conditions/users/GroupConditionRule.php
@@ -10,6 +10,7 @@ use craft\elements\db\ElementQueryInterface;
 use craft\elements\db\UserQuery;
 use craft\elements\User;
 use craft\helpers\ArrayHelper;
+use craft\helpers\ProjectConfig;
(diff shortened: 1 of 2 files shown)


boboldehampsink:
Any updates on this? I have a site that is going live next week and it needs this fix ;-)

brandonkelly:
Fixed by ignoring `isSelectable()` for all condition rules, whenever applying project config changes, since this also occurs for the “Author Group” entry condition rule, and likely others. (80f555fda302b5aaef108e2529cf2e5f0e7c8fef)

Issue #15058, opened 2024-05-23
[5.x]: Not possible to select Entry as type in Selectable Entries Condition


thomascoppein:
### Description
When adding conditions to Entries Field there is an option to add a Related To condition. 
The placeholder of this when selecting is Entry ID while the selected value in the dropdown is Asset.

### Steps to reproduce

1. Add new field
2. Select Field Type Entries
3. Add a Related To condition

### Expected behavior
I would expect that it is possible to select Entry as type in the dropdown.

### Actual behavior
![Screenshot 2024-05-23 at 11 48 12](https://github.com/craftcms/cms/assets/58168425/85f3f03e-d78e-41cc-bfa0-346fdfbc1288)

### Craft CMS version

5.1.4

### PHP version

8.2

### Operating system and version

macOS Sonoma 14.2.1

### Database type and version

mysql

### Installed plugins and versions

- CKeditor
- Sherlock
- SEOMatic
- Retour
- Field Manager
- Feed Me

i-just:
Hi, thanks for reporting! I raised a PR for this.

brandonkelly:
Craft 5.1.6 is out with a fix for this. Thanks again!

PR #15059, opened 2024-05-23
allow entries and product variants in related to rules

i-just:
### Description
A different approach to getting element types for the `RelatedToConditionRule`. Instead of relying on checking the element type, get all field types of `BaseRelationField`, get the element type for each of those, and allow that type to be used in the "related to" condition.

### Related issues
#15058


brandonkelly:
Good solution!

Issue #15007, opened 2024-05-16
[5.x]: Visual indication of actively selected main navigation pane item can be difficult to visually identify


JamieC24:
### Description

Visual indication of actively selected main navigation pane item can be difficult to visually identify (uses a grey line on the far left towards the edge of the screen. While a visual object has been added to show which of the items is selected, the design is somewhat subtle.

### Steps to reproduce

1. Log in to any install of Craft 5
2. Select an option on the Dashboard (i.e Entries)

### Expected behavior

Ideally, the visual indicator for an active item would be more clear to the user (similar to the hover effect for each of the navigation items)

### Actual behavior

The active item indicator is a grey line on the far left towards the edge of the screen.

### Craft CMS version

Craft CMS Pro 5.1.0

### PHP version

8.2.18

### Operating system and version

MacOS Sonoma 14.4.1

### Database type and version

PostgreSQL 15.8

### Installed plugins and versions

[CK Editor](https://plugins.craftcms.com/ckeditor?craft5) 4.0.4
[Postmark](https://plugins.craftcms.com/postmark?craft5) 3.1.0

gcamacho079:
Hi @JamieC24, thanks for your feedback! We chose the vertical indicator because it meets the requirements for non-text contrast; however, I see how it could be missed very easily since it’s right at the edge of the screen. 

In addition to matching the hover style, we’ve played with a couple of different approaches. **Do you feel like the following changes would help call more attention to the active item?**

- A slight (<3:1 contrast ratio) change in background color 
- Updating the text style to be bold
- Maintaining the vertical indicator
<img width="226" alt="Screenshot 2024-05-31 at 2 30 52 PM" src="https://github.com/craftcms/cms/assets/17525979/125ce71f-ad5d-463d-850a-5e6edeb24522">

brandonkelly:
Craft 5.2.0-beta.6 is out with that change.

PR #15061, opened 2024-05-23
Adds additional styling for active global navigation item

gcamacho079:
… page

### Description
- Makes active global nav items more apparent by using styles that match nested sidebars
- Ties active navigation item styling to `aria-current` values
- Sets global variables for navigation active/hover backgrounds

### Related issues
#15007


brianjhanson:
Thanks @gcamacho079! I adjusted the styling a bit more. I know we're not quite hitting the 3:1 contrast ratio for the active items, so I moved the indicator away from the edge a bit to make that more obvious. 

CleanShot 2024-06-10 at 09 12 25

brianjhanson (approved):

PR #15070, opened 2024-05-25
Move the action button into the page header

brandonkelly:
### Description

Moves the action button/menu into the page header, as opposed to the current page’s breadcrumb.

Screenshot 2024-05-24 at 19-00-25 test - Happy Lager

### Related issues

- #14913


brianjhanson (approved):
I definitely prefer these actions out of the breadcrumb area, but I'm also adding making a better place for this menu and the sidebar toggle to my 6 wishlist

gcamacho079 (approved):

andrewhawkes:
This is much better thanks. My only concern still is the icon is a bit nondescript.

michtio:
A lot better, but imho, the button still doesn't "say" anything. Maybe replacing the ... by "User Actions" or something similar makes more sense still :)

andrewhawkes:
> A lot better, but imho, the button still doesn't "say" anything. Maybe replacing the ... by "User Actions" or something similar makes more sense still :)

Agree it needs some text or at least an icon.

This is a bit "graphics design is my passion" but here 😂:
image

brandonkelly:
The button icon is consistent with every other action menu, and it’s positioned right next to the primary action button, so it should be pretty hard to miss.

I chose not to use a cog icon for action menus because cog icons are used for settings buttons, which isn’t what actions are.

Issue #15069, opened 2024-05-24
[5.x]: Modified matrix entries don't get updated after adding or deleting an entry to the matrix


Liamm3:
### Description
We use a matrix field as our content builder. When we edit an existing entry in the matrix, then add a new one or delete an existing one and then save the entry, all changes are lost.

### Steps to reproduce

1. Create a new matrix field
2. Add some entry types (e.g. 'Plain Text')
3. Add the matrix field to a section
4. Create a new entry in the section
5. Add an entry to the matrix ![area_576](https://github.com/craftcms/cms/assets/29706892/ca338875-56e1-4dc5-a888-57156fd0a307)
6. Save the entry using Ctrl+S
7. Modify the created entry
8. Add a new entry to the matrix  
![area_580](https://github.com/craftcms/cms/assets/29706892/ab13bfcb-fc14-40b7-ac48-7dd1025c2a3b)

9. Save again
10. All the changes are lost
![area_579](https://github.com/craftcms/cms/assets/29706892/c08043ac-3254-49ff-90ca-3ae663547e7f)

### Expected behavior
The modified matrix entry should be saved after saving the entry.

### Actual behavior
Changes on a modified entry are lost, after adding a new entry to the matrix.

### Craft CMS version

5.1.6

### PHP version

8.3

### Operating system and version

Linux

### Database type and version

MariaDB 10.6

### Installed plugins and versions

- CKEditor 4.0.6
- Dumper 5.0.1
- SEO v5.0.0-rc2

i-just:
Hi, thanks for reporting. I was able to replicate this if the autosaving of e.g. the text change wasn’t processed before adding a new entry. If you were to wait until after autosaving is done, all the changes would save correctly. I raised a PR for this.

brandonkelly:
Craft 5.1.8 is out with that fix. Thanks again!

PR #15085, opened 2024-05-28
ensure autosave is processed before adding new entry

i-just:
### Description
Ensure autosaving has a chance to be processed before adding a new entry to the inline editable matrix field.

### Related issues
#15069


brandonkelly:
This could theoretically impact other things calling `ElementEditor:setFormValue()` so I fixed it from there instead. (2db4c86ce044c8bbf6817b06a3620ff12ad0133a)

PR #15097, opened 2024-05-29
Feature/cms 1093

i-just:
### Description
Bumping Yii to `2.0.49` so that we can use `select()` from ` yii\console\Controller` with a [`$default` parameter.](https://www.yiiframework.com/doc/api/2.0/yii-console-controller#select()-detail).

### Related issues
[cms-1093](https://linear.app/craftcms/issue/CMS-1093/prosifymatrix-field-command)


brandonkelly:
Updated to 2.0.50 for Craft 4.10 and 5.2 via 1389fda58b19e6f5781dbced4734794343e4d402.

PR #15110, opened 2024-05-30
Override rebrand path

timkelty:
### Description
Allow override of rebrand path via `CRAFT_REBRAND_PATH` env var.

In cloud, `@storage` is in `/tmp`, so having rebrand there doesn't work. This would allow rebrand to work in Cloud by setting something like:

```
CRAFT_REBRAND_PATH=@root/storage/rebrand
```

### Related issues
https://github.com/craftcms/cloud/discussions/10


src/services/Path.php
```diff
@@ -215,7 +216,9 @@ public function getConfigDeltaPath(bool $create = true): string
      */
     public function getRebrandPath(bool $create = true): string
     {
-        $path = $this->getStoragePath($create) . DIRECTORY_SEPARATOR . 'rebrand';
+        $path = App::env('CRAFT_REBRAND_PATH')
+            ? App::parseEnv('$CRAFT_REBRAND_PATH')
+            : $this->getStoragePath($create) . DIRECTORY_SEPARATOR . 'rebrand';
```
brandonkelly, line 221: Couldn’t this be simplified to just

```suggestion
        $path = App::env('CRAFT_REBRAND_PATH')
            ?: ($this->getStoragePath($create) . DIRECTORY_SEPARATOR . 'rebrand');
```

Or am I missing something here?

timkelty: I used `parseEnv` so it would resolve aliases, so that you could do:

`CRAFT_REBRAND_PATH=@root/storage/rebrand`

Otherwise, on Cloud, you'd have to know to use: `/var/task/storage/rebrand`

Issue #15104, opened 2024-05-30
[5.x]: Multiple conditional fields behave peculiarly when the condition of one assesses the state of the other


JamieC24:
### Description

I have created an Entry Type that uses an Entries field and a URL field.

The Entries field is conditionally set to appear only when the URL field is empty.

The URL field is conditionally set to appear only when the Entries field is empty.

Both fields are visible to begin with (as expected, as both are empty at this point)

When I populate the URL field, the Entries field disappears (as expected)

When I remove the content from the URL field, the Entries field reappears (as expected)

When I select an Entry from the Entries field, the Entries field disappears and the URL field reappears with the content it had before I removed the content from it.

If I refresh the page, neither of the fields are visible (presumably because the state of both is 'has a value' and the fields are both set to appear only when the other has a state of 'is empty')

### Steps to reproduce

1. Log in to any Craft 5.1.7 install
2. Create an Entry Type
3. Save the Entry Type
5. Create a couple of Entries that use the Entry Type you created
7. Edit your Entry Type using Settings -> Entry Types -> [Your Entry Type]
8. Create an Entries field and add it to the "Content" tab of your Entry Type
9. Create a URL field and add it to the "Content" tab of your Entry Type
10. Edit the settings of your Entries field by clicking on the three dots on the far right of the field box and selecting "Settings"
11. Use the "Entry Condition" section to add a rule that maps to your URL field and checks that it "is empty" (the dropdown for this shows as 'relates to' by default)
12. Repeat this process for the URL field, mapping its conditional rule to the Entries field.
13. Save your Entry Type
14. Go to one of the two Entries you created and put something in the URL field (e.g https://www.google.ie/)
15. Remove the text you've just added by highlighting "https://www.google.ie/" and pressing the backspace key
16. Your Entries field should reappear now. Click the "Add an entry" button on it, select the other entry you created and click "Select"

### Expected Behaviour
When you press "Select", a small representation of the entry you chose should appear under the heading for your Entries field, and the URL field should not be visible

### Actual Behaviour
When you press "Select", the Entries field disappears and is replaced by the URL field that has repopulated itself with whatever you populated it with previously (e.g https://www.google.ie/).
If you refresh the page and scroll to where your field was, you should no longer be able to see the Entries field nor the URL field.

### Craft CMS version

Craft CMS Pro 5.1.7

### PHP version

8.2.0

### Operating system and version

MacOS Sonoma 14.5

### Database type and version

PostgreSQL 15.8

[...]

PR #15114, opened 2024-05-31
Bugfix/15104 inline matrix and conditionals

i-just:
### Description
Makes conditionals work inside inline matrix without having to reload the page after creating a draft.

Example:
- You have two fields inside a matrix field with inline-editable blocks view mode: `fieldA` and `fieldB`. `fieldA` should show if `fieldB` is empty. `fieldB` should show if `fieldA` is empty.
- Create an entry that contains this matrix field, add value to `fieldA` and fully save. `fieldB` is no longer showing, which is correct.
- Remove the value from `fieldA` - a provisional draft is created for the owner entry, and the matrix entry is duplicated, which means it now has a different `uid`. However, we’re still using the canonical `uid` when getting the matrix entry, causing the behaviour described in the GH issue.

Passing the new matrix element `uid` from the save draft action to js and using that `uid` when updating the field layout fixes the problem.

### Related issues
#15104


brandonkelly:
> * `fieldA` should show if `fieldB` is empty. `fieldB` should show if `fieldA` is empty.

@i-just Are you sure that’s what you meant? This setup results in both fields being hidden for new entry blocks, as I would expect since neither one has a value initially.

i-just:
Yeah, basically if both fields are empty (don't have a value) then they should both show. If one of them has a value then the other one should be hidden. So,  for a new block, both fields are shown to start with and if you fill one out, the other one gets hidden. 

LMK if I'm still not making sense!

brandonkelly:
Doh, had it backwards.

brandonkelly:
Nicely done! Just changed it to index draft UUIDs by the canonical UUIDs, to be consistent with the ID version.

PR #15117, opened 2024-05-31
Relocate the details toggle

brianjhanson:
Relocates the details sidebar toggle button to the side of the content editor. 

https://github.com/craftcms/cms/assets/1843073/4cad4c82-772b-4199-88a8-9acbb0443788


gcamacho079 (changes requested):
😍 Looks awesome. The only thing I noticed is that at 320px wide viewports, the `.content-pane` overflows the `body` element.

brandonkelly:
Looks great!

The hairline is 4px closer to the sidebar pane than the content pane, so that needs to be centered.

Ideally, both sides can be tightened up, so the distance between the two panes is 24px, to match the distance between the global sidebar and the content pane. So shave off 7px from the left side of the hairline, and 3px from the right side, I think.

brianjhanson:
> Looks great!
> 
> The hairline is 4px closer to the sidebar pane than the content pane, so that needs to be centered.
> 
> Ideally, both sides can be tightened up, so the distance between the two panes is 24px, to match the distance between the global sidebar and the content pane. So shave off 7px from the left side of the hairline, and 3px from the right side, I think.

This difference was intentional to try to make a stronger association between the sidebar and the toggle, but I think you're right and the contrast between the spacing wasn't quite enough to make that obvious, so I've centered it now.

brandonkelly:
Thanks, looks great now!

brandonkelly:
@brianjhanson One more bug: for RTL languages, the icon needs to be inverted:

screenshot of the new sidebar toggle, on a right-to-left interface orientation

Issue #15130, opened 2024-06-03
[5.2.0-beta.1]: Impossible to delete entries with unsaved changes


mmikkel:
### Description

If an entry has unsaved changes, it's not possible to delete it:

https://github.com/craftcms/cms/assets/298510/5b7ca665-e42f-4024-8624-8cbce6e47a12

### Steps to reproduce

1. Select an entry tagged "Edited" in an entry index, and attempt to delete it via the element action

### Expected behavior

The entry should be deleted.

### Actual behavior

The entry isn't deleted, even though Craft's success toast claims it has been 🙂

### Craft CMS version

5.2.0-beta.1

### Installed plugins and versions

None

i-just:
Hi, thanks for reporting! I raised a PR for this.

brandonkelly:
Craft 5.2.0-beta.2 is out with that fix. Thanks again.

PR #15136, opened 2024-06-04
Bugfix/15130 perform action on provisional elements

i-just:
### Description
When performing element actions on provisional drafts, use the canonical ID (not the provisional draft’s ID).

### Related issues
#15130


brandonkelly:
Fixed this via bbd69f9ba08483298c94223fe65642ebf5514e51 – now just setting the `data-id` and `data-draft-id` attributes as if `$element` was the canonical element, if it’s a provisional draft. That way it’s not a breaking change for bulk element actions.

PR #15166, opened 2024-06-07
Adds translations for view mode buttons

gcamacho079:
### Description
Adds translations for:
- ‘Display as cards’
- ‘View settings’

### Related issues
Resolves PT-1767, PT-1765


brandonkelly:
Doh, thanks for catching those!

PR #15193, opened 2024-06-13
Update French translation

scandella:
Change the "Live" status translation to match the "Enabled" switch translation.

### Description
The "status“ mention "Live"  is translated in French by "En direct", which means something like "On air". Not very clear.
I think this mention should be the same as the "Enabled" switch label, considering that this switch determines the status of an entry.
So I suggest translating the "Live" mention by "Activé" which is the same mention for the entry switch.


brandonkelly:
Thanks!

PR #15205, opened 2024-06-14
Correct GraphQL CLI Prompt

AugustMiller:
The `graphql/create-token` action prompts the user for a "Schema name", but it’s actually asking for a name for the new token! 😅


AugustMiller:
This naming issue is actually present in 4.x as well—not sure how to target both branches…

Issue #15209, opened 2024-06-17
[5.x]: Slideout shortcuts broken when the slideout is opened via disclosure menu


thupsi:
### Description
When a slideout is opened via a discolusre menu (e.g. the user selects `edit field` on a FLD item) the slideout shortcuts (e.g `Ctrl+S` for saving, or `ESC` for closing) don't work.

### Steps to reproduce

1. Go to the FLD for an entry type. 
2. Try to edit a field. 
3. The slideout opens. 
4. Try to perform a `save`, or `close (cancel)` on the slideout via the corresponding keyboard shortcut.

### Expected behavior

The action invoked via the shortcut should be performed.

### Actual behavior

On `ESC` nothing happens. On `Ctrl+S` the browser opens the native save dialog.

### Craft CMS version

5.2.0

### Installed plugins and versions

-

thupsi:
I also have a suspicion that this also causes (or is somehow related to) the piling-on of event listeners on multiple disclosure menus on multiple levels of slideouts.

i-just:
Hi, thanks for reporting! I raised a PR for this.

brandonkelly:
Craft 4.10.1 and 5.2.1 are out with a fix for that. Thanks again!

PR #15215, opened 2024-06-17
Bugfix/15209 disclosure menu and shortcuts

i-just:
### Description
When you open a slideout via a disclosure menu, the order of events is:
- clicking the 3 dots menu opens the disclosure dialogue, adds a new UI layer for it, and registers shortcuts
- clicking “edit <type>” opens the slideout, adds a new UI layer for it, registers shortcuts and then removes a UI layer; the intention is that the removed layer should be the one for the disclosure menu we opened, but in fact, it was removing the top layer causing the shortcuts registered for the slideout not to work.

When hiding the disclosure menu, ensure we’re removing that menu’s UI layer and not the very top layer, as at that point, the very top layer might be for something else (e.g. a slideout).

### Related issues
#15209


brandonkelly:
The same bug could have potentially affected Craft 4, so fixed it there instead. We’re already passing the disclosure menu container into `addLayer()` so the only change needed was passing the same container to `removeLayer()`.

PR #15223, opened 2024-06-18
queryScalar - check subquery before changing params

i-just:
### Description
When running a query through our version of `queryScalar`, check if the subquery doesn’t have `distinct`, `groupBy`, `having` or `union` and only overwrite its `select`, `orderBy`, `limit`, and `offset` params if it doesn’t.

### Related issues
#15001

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Added `Garnish.muteResizeEvents()`.
 - Fixed a JavaScript performance degradation bug. ([#14510](https://github.com/craftcms/cms/issues/14510))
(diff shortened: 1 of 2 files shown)


samhibberd:
@brandonkelly does this also impact the `exists()` query method, since upgrading to `4.10.x` which introduced this logic to work off the subquery: https://github.com/craftcms/cms/blob/db4292c5df6e169139dc3118009b1efe3d3f9348/src/elements/db/ElementQuery.php#L1693-L1699
We are running into the same issue with the exists method when a HAVING clause is used (for us with simplemap https://github.com/ethercreative/simplemap/issues/391#issuecomment-2160495458)

brandonkelly:
@samhibberd Craft 4.10.8 and 5.2.10 are out with a fix for that.

Issue #15241, opened 2024-06-24
[5.x]: Collapsing the bottom entry in a structure appends "Nothing yet" to the index view


mmikkel:
### Description

In a structure, if the bottom, top level parent is collapsed and there are no other top level parents that are expanded, Craft injects the "Nothing yet" message:

https://github.com/craftcms/cms/assets/298510/5727fae3-b1b9-43de-935b-815ca59e1b9b

### Steps to reproduce

1. Create a structure, and add some entries at different levels. Make sure the bottom, top level entry has children
2. Collapse all the top level entries, going from top to bottom.
3. When collapsing the bottom, top level entry, observe the "Nothing yet" message appearing.

### Craft CMS version

5.2.3

### Installed plugins and versions

None

i-just:
Hi, thanks for reporting! I raised a PR for this.

brandonkelly:
Craft 5.2.4 is out with a fix for this. Thanks again!

PR #15243, opened 2024-06-25
Bugfix/15241 lazy loading nothing yet message adjustment

i-just:
### Description
When we’re not in a paginated view, and there are already visible elements in the index, don’t show the “Nothing yet” message.

(This only affects v5.)

### Related issues
#15241


brandonkelly:
Fixed via ad19a7054dbd1be5813fb816d9459a6b7992c602 by just checking if the query has an `offset` set.

Issue #15244, opened 2024-06-25
[5.x]: The "Edited" tag in the breadcrumbs bar disappears on page reload


mmikkel:
### Description

When making edits to an existing entry, Craft adds the blue "Edited" tag to the entry pill in the breadcrumbs bar. However, the tag disappears if the page is reloaded, or on subsequent pageloads after navigating away from the provisional draft, and only re-appears after making additional changes to the content.

I'm not sure if the intention is to show the "Edited" tag in the breadcrumbs bar or not (seems kinda redundant to me?), but the fact that it's not consistently shown (or not) feels like a bug :)

https://github.com/craftcms/cms/assets/298510/701c3983-9f4c-427a-8aa3-7962c9d23ef2

### Steps to reproduce

1. Navigate to the edit page for an existing entry
2. Make some edits, wait for the provisional draft to save
3. Confirm that the entry pill now has a blue "Edited" tag next to it, in the breadcrumbs bar
4. Reload the page, and confirm that the "Edited" tag is no longer visible

### Craft CMS version

5.2.3

### Installed plugins and versions

None

i-just:
Hi, thanks for reporting! I raised a PR for this.

brandonkelly:
Craft 5.2.4 is out with a fix for this. Thanks again!

PR #15247, opened 2024-06-25
ensure the correct status label is shown crumbs

i-just:
### Description
Ensure that the “Edited” indicator shows in the crumbs after reloading a provisional draft.

### Related issues
#15244


brandonkelly:
I noticed this also affects the element title, so resolved via 2159e2e513d261f27d89d2b15dd11e234da28e47 instead. Now just passing the original element into `Cp::elementChipHtml()`, and only fetching the canonical for getting its crumbs.

Issue #15301, opened 2024-07-04
[5.x]: "Rename file" (and others) on folder leads to error message


webkp:
### Description
When selecting a folder and trying to do "Rename file" an error appears (independently of the filesystem type [same with S3 or local]). 

Note: 
While I create this issue for "Rename file", the other operations like "Preview file", "Download" and "Copy reference tag" lead to errors too when using them on folders. Only "Move" and "Delete" work as expected.

### Steps to reproduce

![04-07-2024_15-57-54](https://github.com/craftcms/cms/assets/11754639/0fcde578-f5ac-4c08-a2d1-0cae3d5646df)

### Expected behavior
Preferred: Disable and grey out "Rename file" as it is the case with "Replace file" and others
-or-
Being able to rename folders (but then I guess it should simply be "Rename" and not "Rename file". 

### Actual behavior
Operation "Rename file" is available for folders which leads to said error

### Craft CMS version

5.2.5

### PHP version

8.2

### Installed plugins and versions

- craftcms/ckeditor 4.1.0
- mmikkel/cp-field-inspect 2.0.1
- verbb/expanded-singles 3.0.0
- fortrabbit/craft-object-storage  4.0.0
- nystudio107/craft-vite 5.0.0

i-just:
Hi, thanks for reporting! I raised a PR for this.

brandonkelly:
Craft 5.2.6 is out now with a fix for this. Thanks again!

PR #15302, opened 2024-07-04
disallow more actions for asset folders

i-just:
### Description
Disallow “Preview file”, “Download”, “Rename file”, and “Copy reference tag” actions from being performed on folders.

(Tested on v4, and those actions are already disallowed.)

### Related issues
#15301


brandonkelly:
Root issue was that `data-id` attributes are no longer getting included at all on folder chips, so the `is('[data-id=""]')` check wasn’t working anymore (since that is explicitly looking for a `data-id` attribute set to an empty string). Fixed via d56cf0b4ddcd1a6dcfb793d7ffda3bca4a1491e9.

PR #15304, opened 2024-07-05
prevent multiple error icons being added on save

i-just:
### Description
Prevent multiple error icons from being added on subsequent saves.

Before:
Screenshot 2024-07-05 at 08 57 15

After:
Screenshot 2024-07-05 at 08 56 49

### Related issues
#15298


brandonkelly:
Doh, thanks!

Issue #15307, opened 2024-07-05
[5.x]: Editing users in a slideout and saving with cmd+s triggers an error


i-just:
### Description
Editing users in a slideout and saving with cmd+s triggers an error. Saving via the Save button works as expected.

### Steps to reproduce

1. Open user element in a slideout (either from the users index page or Users field)
2. Make a change to one of the fields
3. cmd+s to save
4. notice the warning appear in the top right corner of the slideout

The error is “Element <ID> could not be duplicated because it doesn’t validate.”.
The actual element errors that trigger this are:
- *Username* "u1" has already been taken.
- *Email* "u1@test.com" has already been taken.

### Expected behavior
Be able to change and save the user via a slideout using the cmd+s shortcut.

### Actual behavior
Not able to save the user via a slideout using the cmd+s shortcut.

### Craft CMS version

5.2.5

### PHP version

8.1

### Database type and version

mysql 8

### Installed plugins and versions

-

brandonkelly:
Tracked the bug down to [this change](https://github.com/craftcms/cms/commit/8ccfab112d07cbca56ef65ecbf9f7da15866ca5a) as part of #14436. Fixed by ensuring that the element editor’s `canCreateDrafts` setting is `true` before calling `saveDraft()` on it (260dce5c3d7b7b9e60f50a1f2c4597d9a5a5bdd4).

Tested error summaries across multiple tabs and they still seem to be working, but @i-just maybe you have a better idea of what we need to watch out for with that change?

i-just:
I just tested the solution against the problem it was originally for, and all still works as expected (and so does saving users in a slideout via shortcut).

PR #15308, opened 2024-07-05
Bugfix/15307 slideout save element without drafts via shortcut

i-just:
### Description
When saving an element via slideout with a shortcut, don’t insist on creating a draft if the element doesn’t support drafts.

### Related issues
#15307


brandonkelly:
Fixed via 260dce5c3d7b7b9e60f50a1f2c4597d9a5a5bdd4

PR #15309, opened 2024-07-05
Handle OPTIONS requests for login

brianjhanson:
When we [return null](https://github.com/craftcms/cms/blob/427fabe1d339e720cefefe2b7d7e60234cb04838/src/controllers/UsersController.php#L196) in the `UsersController` we'll miss the [check for if we're dealing with an action request](https://github.com/craftcms/cms/blob/12599c2dca42fc66f5c72ce9ded070ea20a246a4/src/web/Application.php#L301-L304) and end up with a 404. Returning a 404 from the OPTIONS request will end up in an error when trying to login from a cross-origin request (like from a headless app).


vettndr:
@brandonkelly Hi Brandon, is this pr released?

brandonkelly:
@vettndr We ended up deciding to take a different approach, so the change was reverted. I’ll post here when it’s ready.

pixelwerft:
I am struggling with the same issue, any updates on it?

timkelty:
@gettheflow @vettndr The best solution to address these CORS issues is to use Yii's CORS filter. See https://github.com/craftcms/cms/pull/15397#issue-2422964963 for a example of configuring it.

Note: for the `users/login` controller (and any of Craft's native actions), you'll still need the changes from [this PR](https://github.com/craftcms/cms/pull/15397#issue-2422964963), or your `OPTIONS` request will return a 404. For Craft 4, you can test it out by requiring `craftcms/cms:dev-feature/cors as 4.11.0`

PR #15323, opened 2024-07-09
Improve display of plugin notes

brianjhanson:
Updates plugin notes to be displayed along with a replacement reason rather than instead of it.   

CleanShot 2024-07-09 at 11 30 39


brandonkelly:
I don’t want to mess with that logic, since we want the plugin notes to work in Craft 3 as well. If we have both a note and a replacement plugin handle, we can send a consolidated note via the API.

brianjhanson:
> I don’t want to mess with that logic, since we want the plugin notes to work in Craft 3 as well. If we have both a note and a replacement plugin handle, we can send a consolidated note via the API.

I wanted to do it that way initially, but since the `plugin.note` is run through `Craft.escapeHtml()` the link to the plugin will be escaped. 

How were plugin notes added previously? I had to [add the column](https://github.com/craftcms/craftnet/pull/1053/files#diff-e917948216524d1194a48394e056d59fb827157a8fd990a4e42de02957de7b02) and [the property in the response](https://github.com/craftcms/craftnet/pull/1053/files#diff-8b736db30c634d94a155dd2e9d19a9013076efebe78e6f3340a438729578ec96) so I figured the code in Craft was leftover/extra from a previous era.

brandonkelly:
In that case, let’s have the UI either let you set a plugin handle OR a custom note, and say that HTML/Markdown isn’t allowed.

I don’t think we’ve ever actually utilized the `note` field.

PR #15324, opened 2024-07-09
Fixes focus trap issue in slideouts when the sidebar is toggled open initially

gcamacho079:
### Description
Only applies focus trap in situations where the sidebar has modal-like behavior — specifically, when it overlaps the content pane and you have to `ESC` to close.


brandonkelly:
I think I may have beat you to this one (4e4e2e46196c3ab8533f554ff8392216b219e0bf + 449694f58f7d5572c6ac4536c2ab37807cce62e2).

gcamacho079:
@brandonkelly you did! 🥳  I made a different PR to 5.x with an issue I also addressed in this PR (#15332)

PR #15343, opened 2024-07-12
Core changes to make front-end 2FA possible

brianjhanson:
The minimum changes to make 2FA possible from the front-end. I had initially wanted to ship some javascript that would help with this, but the code currently powering the LoginForm and TotpForm very much expects the `Craft` global to be present. I made a few attempts to untangle things but ultimately wanted to get this in so we can get our own implementation for console under our belts first. 

This makes your own implementation _possible_ but not very easy. If you want to go down that path, there are a handful of things you'll have to implement. We have some ideas to make that easier, but they aren't included in this PR.


brianjhanson:
Converted back to a draft so I can work through setup from the front-end

PR #15346, opened 2024-07-13
App type-based configs

brandonkelly:
Adds support for app type-based configs, for `general` and `db` (e.g. `general.web.php`/`general.console.php`).

In addition to returning an array or new `GeneralConfig`/`DbConfig` object, `general` and `db` config files can now return a callable which accepts an existing config object, so it’s possible for app type-based configs to use the fluent syntax on the same config object created by the original config file (`general.php`/`db.php`):

```php
// config/general.php

    ->aliases([
        '@webroot' => dirname(__DIR__) . '/web',
    ])
    // ...
;

// or

return fn(GeneralConfig $config) => $config
    ->aliases([
        '@webroot' => dirname(__DIR__) . '/web',
    ])
    // ...
;
```

```php
// config/general.console.php

return fn(GeneralConfig $config) => $config
    ->addAlias('@web', craft\helpers\App::env('CLI_WEB_URL'))
    // ...
;
```

### DIFF
--- a/bootstrap/bootstrap.php
+++ b/bootstrap/bootstrap.php
@@ -59,17 +59,18 @@ $environment = App::cliOption('--env', true)
 // -----------------------------------------------------------------------------
 
 $configService = new Config();
+$configService->appType = $appType;
 $configService->env = $environment;
 $configService->configDir = $configPath;
 $configService->appDefaultsDir = dirname(__DIR__) . DIRECTORY_SEPARATOR . 'src' . DIRECTORY_SEPARATOR . 'config' . DIRECTORY_SEPARATOR . 'defaults';
-$generalConfig = $configService->getConfigFromFile('general');
+$generalConfig = $configService->getGeneral();
 
 // Validation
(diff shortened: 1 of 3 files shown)


timkelty (approved):
LGTM

Related thought: we could provide some `if` methods for fluent config to facilitate the same kind of thing but in a single file. Could also be handy for environment conditions, eg:

```php
<?php
    ->if(
        true, // expression
        fn(GeneralConfig $config) => $config->devMode(true), // if
        fn(GeneralConfig $config) => $config->devMode(false), // else
    )
    ->ifEnvironment(
        'dev', // CRAFT_ENVIRONMENT=dev
        fn(GeneralConfig $config) => $config->devMode(true)
    )
    ->ifWeb(fn(GeneralConfig $config) => $config->('@web', 'https://example.com'))
    ->ifConsole(fn(GeneralConfig $config) => $config->addAlias('@webroot', '/var/www/html'));
```

Issue #15369, opened 2024-07-17
[5.x]: Setting field values in Entry::EVENT_BEFORE_SAVE doesn't work when $firstSave is true


boboldehampsink:
### Description

I have an Entry::EVENT_BEFORE_SAVE event in which I do:

```
$entry->setFieldValues([
    'latitude' => $coords['lat'],
    'longitude' => $coords['lng'],
]);
```

This works when updating entries, but not for the case where `$entry->firstSave` is `true`.

### Steps to reproduce

1. Create an event listener that sets entry field values.

### Expected behavior

It should work on first save.

### Actual behavior

The first time saving it doesn't work, on updating it does.

### Craft CMS version

5.2.6

### PHP version

8.3

### Database type and version

PostgreSQL 16

### Installed plugins and versions

-

brandonkelly:
Craft 5.2.9 is out with a fix for this. Thanks for reporting!

PR #15372, opened 2024-07-17
Make setting field values via `Entry::EVENT_BEFORE_SAVE` work when `$firstSave` is `true`

i-just:
### Description
Setting field values via `Entry::EVENT_BEFORE_SAVE` doesn’t work when `$firstSave` is `true`.

This happens (in v5 only) because in `_saveElementInternal()`, we no longer always save the element’s content.

Example code:
```
Event::on(
    Entry::class,
    Entry::EVENT_BEFORE_SAVE,
    function (ModelEvent $event) {
        $entry = $event->sender;

        if ($entry->sectionId == 14 && $entry->firstSave) {
            $entry->setFieldValues([
                'number' => 6,
            ]);
        }
    }
);
```

The order of operations is: `ElementsController->actionApplyDraft()` is called, then `Elements->_saveElementInternal()`. At that point, `$element->firstSave` is `false`, so when `EVENT_BEFORE_SAVE` is called, the condition is not met, and the field values are not set. The `$saveContent` is `true` at this point, so content is saved, but since field values weren’t set, this doesn’t result in the desired change.

Then `Drafts->applyDraft()` gets called, then `Drafts->removeDraftData()`, and that’s what sets `$element->firstSave` to `true`. However, when it later calls `Elements->_saveElementInternal()`, it calls it with `$saveContent` set to `false`, so `EVENT_BEFORE_SAVE` is called, the condition is met, and the field values are set, but the content is not actually saved.

### Related issues
#15369


brandonkelly:
The only point of that `saveElement()` call from `removeDraftData()` is to get `draftId` removed from the element. We’re explicitly saying we don’t want content to be saved, so I think `saveElement()` should respect that, even if a plugin changes a field value in the process.

Maybe we should just start communicating the intention by adding a `saveContent` property on the element?

PR #15376, opened 2024-07-18
Add `@phpstan-var` to `mode` enum on ImageTransform model

thomasvantuycom:
### Description

In the commit https://github.com/craftcms/cms/commit/1203286775e19657cef661349dfbd58f772c76bc, several `@phpstan-var` declarations were added for enums. However, the declaration for the `mode` property on the ImageTransform model was omitted. PHPStan flagged this as an issue while I was working on a plugin, but adding the missing declaration resolves the problem.


brandonkelly:
Thanks!

PR #15385, opened 2024-07-18
Add actions menu to table rows for reordering

### LINKED ISSUE #15382: [5.x]: Table entries cannot be drag/dropped to re-order on touch devices
### Description

Using a standard table provides a 'grab handle' to re-arrange rows works well on desktop, but on touch-screen mobile I can't find anything to grab!

In my example, I use this field to list out an events music programme - with one column for the composer and one column for the piece. It's a reasonably nice interface - I know I could use a matrix field here, but that would be a bit more fiddly / visually noisy for my content editors. This works perfectly on desktop / non-touch devices.

I know that there was a similar problem with Matrix fields that was fixed using a drop-down menu (#8490). I understand that this is a problem that hasn't necessarily been fixed well in other software.

### Steps to reproduce

1. Create a table field and begin to add some rows. Try to re-arrange them on a touch device (I've tested Android Chrome on physical hardware and iOS Safari on a simulator).

### Expected behavior

Holding down to drag will re-order - or some other mechanism available to re-order these on mobile.

### Actual behavior

Can't drag / drop to move these.

### Craft CMS version

5.2

### PHP version

8.3

_No response_

### Database type and version

MySQL

### Image driver and version

_No response_

-

gcamacho079:
### Description
This PR introduces an action button with “Move up”/“Move down” buttons to table rows so that reordering is accessible to both keyboard and mouse users without requiring dragging movements; it also makes reordering possible on touch screens.

### Related issues

- Resolves #15382
- Resolves PT-1971


src/web/assets/cp/src/js/EditableTable.js
```diff
@@ -97,29 +97,42 @@ Craft.EditableTable = Garnish.Base.extend(
...
-                  $textarea.focus();
-                }
-              }, 100);
-            }
-          }
+          interactionCallback
         );
+
+        if ($actionMenus.length) {
+          this.addListener($actionMenus, 'keypress,click', interactionCallback);
```
gcamacho079: @brandonkelly I had to add these listeners since the trigger activation doesn’t bubble up to the table. However, when I tried using the `activate` event, I was having issues. It was only getting triggered by mouse clicks but not keyboard activation. Wondering if you have any thoughts about a better way to go about this. 🤔

Issue #15388, opened 2024-07-19
[4.x]: `Element:: EVENT_DEFINE_SIDEBAR_HTML ` not working for `User` elements


tommysvr:
### Description

`EVENT_DEFINE_SIDEBAR_HTML` isn't triggered on User elements, but it works on other elements.

### Steps to reproduce

You can test with:

```
Event::on(User::class, Element::EVENT_DEFINE_SIDEBAR_HTML, static function (DefineHtmlEvent $event) {
  \Craft::dd('User sidebar');
});
```

### Expected behavior

When you load a User you'll see the dd output

### Actual behavior

User page loads as normal

### Craft CMS version

4.10.6

### PHP version

8.2

### Installed plugins and versions

-

brandonkelly:
This is expected because Edit User pages don’t use the new element editor in Craft 4 (whereas user slideouts do).

For Craft 4, you can inject additional HTML into Edit User page sidebars via the `cp.users.edit.details` [template hook](https://craftcms.com/docs/4.x/extend/template-hooks.html).

PR #15392, opened 2024-07-19
trigger define sidebar html and define metadata events on user edit page

i-just:
### Description
Trigger `EVENT_DEFINE_SIDEBAR_HTML` and `EVENT_DEFINE_METADATA` for the full page user edit screen.

At the moment, those events are triggered in the slideout screen but not when editing via a full-page screen. This change makes both screens use those events.

In addition, on the full-page screen, there’s a `cp.users.edit.details` hook available (not used in a slideout).

For v4 only. This already works as expected in v5.

Before:
Screenshot 2024-07-19 at 15 05 47
Screenshot 2024-07-19 at 15 06 05

After:
Screenshot 2024-07-19 at 14 58 22
Screenshot 2024-07-19 at 14 58 42

### Related issues
#15388


brandonkelly:
Don’t love the idea of adding new `craft\elements\User` methods just for this. They can use the template hook in v4 and switch to the event for v5.

PR #15397, opened 2024-07-22
Configurable CORS and site headers

timkelty:
### Description
- [x] Makes minor change to allow [Yii2's CORS filter](https://www.yiiframework.com/doc/api/2.0/yii-filters-cors) to work for actions requests
- [x] Deprecates `allowedGraphqlOrigins` in favor of using CORS filter
- [x] Deprecates `permissionsPolicyHeader` in favor of using headers filter
- [x] Adds 2 new filters: CorsFilter, HeadersFilter
  - These are both for site-only (non-cp) requests, and can be optionally filtered by specific site

Example config: `config/app.web.php`

```php
<?php

return [
    // Attach the CORS filter to the application
    'as corsFilter' => [
        'class' => \craft\filters\Cors::class,

        // optionally filter by site(s)
        'site' => ['siteA', 'siteB'],

        // CORS defaults for all non-CP requests
        'cors' => [
            'Origin' => [
                'https://craft-4-project-cp.ddev.site',
                'https://craft-4-project.ddev.site'
            ],
            'Access-Control-Request-Method' => ['GET', 'POST', 'PUT', 'PATCH', 'DELETE', 'HEAD', 'OPTIONS'],
            'Access-Control-Request-Headers' => ['*'],
            'Access-Control-Allow-Credentials' => true,
            'Access-Control-Max-Age' => 86400,
            'Access-Control-Expose-Headers' => [],
        ],

        // Optionally override any controller action's CORS settings.
        // This is an override only – if the setting is not present above, it will not be added.
        'actions' => [
            'graphql/api' => [
                'Origin' => ['*'],
                'Access-Control-Allow-Credentials' => false,
            ],
        ],
    ],

    // Attach the headers filter to the application
    'as headersFilter' => [
        'class' => \craft\filters\Headers::class,
        'site' => 1,
        'headers' => [
            'foo' => 'bar',
            'Permissions-Policy' => 'interest-cohort=()',
        ],
    ],
];
```

### Related issues
- [x] https://github.com/craftcms/cms/pull/15281#issuecomment-2225423327
- [x] https://github.com/craftcms/cms/discussions/10406
- [x] https://github.com/craftcms/cms/discussions/15350

### DIFF
--- a/src/web/Application.php
+++ b/src/web/Application.php
@@ -636,7 +636,13 @@ class Application extends \yii\web\Application
             try {
                 Craft::debug("Route requested: '$route'", __METHOD__);
                 $this->requestedRoute = $route;
-                return $this->runAction($route, $_GET);
+                $response = $this->runAction($route, $_GET);
+
+                // Return the response for OPTIONS requests that return null
+                // to support the CORS filter: https://www.yiiframework.com/doc/api/2.0/yii-filters-cors
+                return $response === null && $this->getRequest()->getIsOptions()
+                    ? $this->getResponse()
+                    : $response;
             } catch (Throwable $e) {
                 $this->_unregisterDebugModule();
                 if ($e instanceof InvalidRouteException) {
--- a/src/web/Application.php
+++ b/src/web/Application.php
@@ -640,8 +640,8 @@ class Application extends \yii\web\Application
 
                 // Return the response for OPTIONS requests that return null
                 // to support the CORS filter: https://www.yiiframework.com/doc/api/2.0/yii-filters-cors
-                return $response === null && $this->getRequest()->getIsOptions()
-                    ? $this->getResponse()
+                return $request->getIsOptions()
+                    ? ($response ?? $this->getResponse())
                     : $response;
             } catch (Throwable $e) {
                 $this->_unregisterDebugModule();
--- a/CHANGELOG-WIP.md
+++ b/CHANGELOG-WIP.md
@@ -13,9 +13,13 @@
 - Added the `withCustomFields` element query param.
 - Added support for application-type based `general` and `db` configs (e.g. `config/general.web.php`). ([#15346](https://github.com/craftcms/cms/pull/15346))
 - `general` and `db` config files can now return a callable that modifies an existing config object. ([#15346](https://github.com/craftcms/cms/pull/15346))
+- The `allowedGraphqlOrigins` config setting is now deprecated. `craft\filters\Cors` should be used instead. ([#15397](https://github.com/craftcms/cms/pull/15397))
+- The `permissionsPolicyHeader` config settings is now deprecated. `craft\filters\Headers` should be used instead. ([#15397](https://github.com/craftcms/cms/pull/15397))
 
 ### Extensibility
 - Added `craft\config\GeneralConfig::addAlias()`. ([#15346](https://github.com/craftcms/cms/pull/15346))
+- Added `craft\filters\Cors`. ([#15397](https://github.com/craftcms/cms/pull/15397))
+- Added `craft\filters\Headers`. ([#15397](https://github.com/craftcms/cms/pull/15397))
 - Added `Craft.EnvVarGenerator`.
 - `craft\helpers\UrlHelper::cpUrl()` now returns URLs based on the primary site’s base URL (if it has one), for console requests if the `baseCpUrl` config setting isn’t set, and the `@web` alias wasn’t explicitly defined. ([#15374](https://github.com/craftcms/cms/issues/15374))
 - Deprecated `craft\web\assets\elementresizedetector\ElementResizeDetectorAsset`.
--- a/src/config/GeneralConfig.php
+++ b/src/config/GeneralConfig.php
@@ -203,6 +203,8 @@ class GeneralConfig extends BaseConfig
      *
      * @group GraphQL
      * @since 3.5.0
+     * @deprecated in 4.11.0 Use \yii\filters\Cors::class instead
+     * @see https://www.yiiframework.com/doc/api/2.0/yii-filters-cors
      */
     public array|null|false $allowedGraphqlOrigins = null;
 
@@ -3469,6 +3471,8 @@ class GeneralConfig extends BaseConfig
      * @return self
      * @see $allowedGraphqlOrigins
      * @since 4.2.0
+     * @deprecated in 4.11.0 Use \yii\filters\Cors::class instead
+     * @see https://www.yiiframework.com/doc/api/2.0/yii-filters-cors
      */
     public function allowedGraphqlOrigins(array|null|false $value): self
     {
--- a/src/config/GeneralConfig.php
+++ b/src/config/GeneralConfig.php
@@ -203,7 +203,7 @@ class GeneralConfig extends BaseConfig
      *
      * @group GraphQL
      * @since 3.5.0
-     * @deprecated in 4.11.0 Use \yii\filters\Cors::class instead
+     * @deprecated in 4.11.0 Use \yii\filters\Cors::class instead.
      * @see https://www.yiiframework.com/doc/api/2.0/yii-filters-cors
      */
     public array|null|false $allowedGraphqlOrigins = null;
@@ -5370,7 +5370,7 @@ class GeneralConfig extends BaseConfig
      * @return self
      * @see $permissionsPolicyHeader
      * @since 4.2.0
-     * @deprecated in 4.11.0 Use \craft\filters\Headers::class instead
+     * @deprecated in 4.11.0 Use \craft\filters\Headers::class instead.
      */
     public function permissionsPolicyHeader(?string $value): self
     {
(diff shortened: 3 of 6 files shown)


src/web/Application.php
```diff
@@ -112,6 +112,13 @@ public function init(): void
         $this->debugBootstrap();
     }
 
+    public function behaviors(): array
+    {
+        return [
+            'corsFilter' => App::corsFilterConfig(),
+        ];
+    }
+
```
timkelty: This couldn't be applied in `src/config/app.web.php` (via `as corsFilter`), because it needs access to `Craft::$app->getConfig()->getGeneral()` for defaults, and the `as {name}` behavior syntax doesn't appear to support a closure.

src/web/Application.php
```diff
@@ -638,7 +645,10 @@ private function _processActionRequest(Request $request): ?BaseResponse
             try {
                 Craft::debug("Route requested: '$route'", __METHOD__);
                 $this->requestedRoute = $route;
-                return $this->runAction($route, $_GET);
+                $response = $this->runAction($route, $_GET);
+                return $response === null && $this->getRequest()->getIsOptions()
+                    ? $this->getResponse()
+                    : $response;
```
timkelty, line 645: For `OPTIONS` requests that are action requests and return `null`, we need to return the response instead of null, as it has the manipulated response from the filter, otherwise they will proceed to Yii's handling and result in a 404.

vettndr:
Hi Brandon @brandonkelly,
I have a Nuxt.js app which is connected to a CraftCMS instance in a separated domain.

When I try to login from the Nuxt.js application using ```POST actions/users/login``` I'm getting back this error:
```Response to preflight request doesn't pass access control check: It does not have HTTP ok status.```

Is there a way to handle the "preflight request"?

Best

**Comment by timkelty** [comment]
@vettndr The `\craft\filters\Cors` handles `OPTIONS` preflight requests, as well as Cors-related headers.
Implement the filter in your `config/app.web.php` with something like this:

```php
<?php

return [
    'as corsFilter' => [
        'class' => \craft\filters\Cors::class,
        'cors' => [
            'Origin' => [
                'https://my-nuxt-app.com',
            ],
            'Access-Control-Request-Method' => ['GET', 'POST', 'PUT', 'PATCH', 'DELETE', 'HEAD', 'OPTIONS'],
            'Access-Control-Request-Headers' => ['*'],
            'Access-Control-Allow-Credentials' => true,
            'Access-Control-Max-Age' => 86400,
            'Access-Control-Expose-Headers' => [],
        ],
    ],
];
```

vettndr:
> @vettndr The `\craft\filters\Cors` handles `OPTIONS` preflight requests, as well as Cors-related headers. Implement the filter in your `config/app.web.php` with something like this:
> 
> ```
> <?php
> 
> return [
>     'as corsFilter' => [
>         'class' => \craft\filters\Cors::class,
>         'cors' => [
>             'Origin' => [
>                 'https://my-nuxt-app.com',
>             ],
>             'Access-Control-Request-Method' => ['GET', 'POST', 'PUT', 'PATCH', 'DELETE', 'HEAD', 'OPTIONS'],
>             'Access-Control-Request-Headers' => ['*'],
>             'Access-Control-Allow-Credentials' => true,
>             'Access-Control-Max-Age' => 86400,
>             'Access-Control-Expose-Headers' => [],
>         ],
>     ],
> ];
> ```

Hey @timkelty 
thank you for your message!

I tried as you suggested me.... and now I'm facing a ```400 Bad Request``` issue, this is the response:
```
"name": "Bad Request",
"message": "Unable to verify your data submission.",
"code": 0,
"status": 400,
"exception": "yii\\web\\BadRequestHttpException",
"file": "/var/www/html/vendor/yiisoft/yii2/web/Controller.php",
"line": 221,
 ```
I'm pretty sure the credentials I'm using are working in the backend when I try to access on it.

Thanks in advance for your support.

Best

**Comment by timkelty** [comment]
@vettndr your request is failing CSRF validation.

You need to fetch a CSRF token first, then include it along with your request.
See https://craftcms.com/docs/5.x/development/forms.html#ajax for examples

**Comment by vettndr** [comment]
> @vettndr your request is failing CSRF validation.
> 
> You need to fetch a CSRF token first, then include it along with your request. See https://craftcms.com/docs/5.x/development/forms.html#ajax for examples

@timkelty yes, I’m fetching the CSRF token when the app is loaded for the first time and then I’m including it in the POST request.

I don’t know why it’s triggering that 400 exception.

**Comment by timkelty** [comment]
How are you making the request (fetch, axios, etc)?
A common mistake is to not send credentials along with the request (eg [`withCredentials`](https://axios-http.com/docs/req_config), `credentials: include`)

**Comment by vettndr** [comment]
> How are you making the request (fetch, axios, etc)? A common mistake is to not send credentials along with the request (eg [`withCredentials`](https://axios-http.com/docs/req_config), `credentials: include`)

Hi @timkelty 
this is the login request:

```
const { user } = await $fetch(
        `${envs.public.craftUrl}/${authRoutes.login}`,
        {
          method: "POST",
          headers: {
            Accept: "application/json",
            "X-CSRF-Token": csrfToken.value as string,
            "X-Requested-With": "XMLHttpRequest",
          },
          body: {
            loginName: email,
            password: password,
          },
          credentials: "include",
        }
      )
```

as you can see I'm setting the ```credentials```parameter

best

**Comment by timkelty** [comment]
What are the domains involved? Are they subdomains of the same root?

I'm not seeing anything jump out from your example, so if you could, please email support@craftcms.com with as much of the project as possible, specifically:
- composer.json/lock files
- config directory
- any custom modules or plugins
- a DB dump if possible

**Comment by vettndr** [comment]
> What are the domains involved? Are they subdomains of the same root?
> 
> I'm not seeing anything jump out from your example, so if you could, please email [support@craftcms.com](mailto:support@craftcms.com) with as much of the project as possible, specifically:
> 
> * composer.json/lock files
> * config directory
> * any custom modules or plugins
> * a DB dump if possible

Hi @timkelty 
thank you, I will send a request to the team soon.

Anyway I did a little debugging process and I found that in the ```validateCsrfToken($clientSuppliedToken = null)``` function 
the ```$trueToken``` variable is printing different values comparing with ```$this->getCsrfTokenFromHeader()```.

It seems like it's comparing the token I passed in the header with another one.

I would really appreciate to read your thoughts about it.

Best

**Comment by vnali** [comment]
@vettndr @timkelty 

Hi, any updates on this?
I'm also encountering the error "Unable to verify your data submission" when sending a request to the backend through a Vue app.

To investigate, I added some debug output and noticed that when validateCsrfToken() is called, Craft runs loadCsrfToken(), but it doesn't generate the same token as expected.

**Comment by vettndr** [comment]
> @vettndr @timkelty
> 
> Hi, any updates on this? I'm also encountering the error "Unable to verify your data submission" when sending a request to the backend through a Vue app.
> 
> To investigate, I added some debug output and noticed that when validateCsrfToken() is called, Craft runs loadCsrfToken(), but it doesn't generate the same token as expected.


Hi @vnali 
this is happening also in my application just on safari and incognito mode.
I've already asked to the Craft team a little support on this but nothing has changed.

Best

cc @timkelty

**Comment by timkelty** [comment]
Thanks for the nudge, @vettndr.
I found your support ticket – I'll dig in this week and follow up in this thread.

**Comment by vnali** [comment]
@timkelty @vettndr
The mistake on my part was that I didn't include `credentials: 'include'` when fetching the session info. I had only added it to the subsequent requests.
I tested it with some available browsers on Windows, including in incognito mode, and it worked.

If I’m right, it might be worth mentioning this in the documentation [here](https://craftcms.com/docs/5.x/development/forms.html#ajax). @AugustMiller 

```
        const response = await fetch('/actions/users/session-info', {
            credentials: 'include',
            headers: {
              'Accept': 'application/json',
            }
        }); // Or your custom endpoint
        const sessioninfo = await response.json();
```

vettndr:
> credentials: 'include',
>             headers: {
>               'Accept': 'application/json',
>             }

hey @vnali,
is your Vue app in a different domain comparing to the Craft instance?
I added the parameter `credentials: 'include'` ages ago...

vnali:
@vettndr It's the same domain but a different subdomain. I can email you a link that you can check in your browser, if you'd like.

vettndr:
> @vettndr It's the same domain but a different subdomain. I can email you a link that you can check in your browser, if you'd like.

thank you for your answer @vnali 
In my case, I have a different domain, that's why I was asking to the Craft guys a possible solution.

I can check internally to move the frontend under the same domain/subdomain to solve it.

vnali:
@vettndr 
I'm not an expert on this, but since you're seeing this issue in a specific browser, it might be related to how the `SameSite` attribute behaves differently across browsers:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie#browser_compatibility

Also, here's how it's configured in Craft:
https://craftcms.com/docs/5.x/reference/config/general.html#samesitecookievalue

I'd also be interested to know if you've heard anything from the dev team about a possible solution.
Good luck with resolving this issue!

PR #15400, opened 2024-07-22
Relation field improvements

brandonkelly:
### Description

Several improvements for relations in Craft:

- Relation fields are now multi-instance.
- Relation fields now get all the normal Translation Method options. (Existing fields will be set to either “Not translatable” or “Translate for each site”, depending on their previous “Manage relations on a per-site basis” setting.)
- Link fields now store relations when an element is selected—so they now get factored in by `relatedTo` element query params—and the framework is in place for other field types (e.g. CKEditor) to start storing relations as well.

Field types which should store relations (in addition to their normal field data) must implement `RelationalFieldInterface`, and optionally import `RelationalFieldTrait`.

```php
use craft\base\Field;
use craft\base\RelationalFieldInterface;
use craft\base\RelationalFieldTrait;

class MyFieldType extends Field implements RelationalFieldInterface
{
    use RelationalFieldTrait;

    public function getRelationTargetIds(ElementInterface $element): array
    {
        $value = $element->getFieldValue($this->handle);
        return $this->scrapeRelationIdsFromValue($value);
    }

    private function scrapeRelationIdsFromValue(mixed $value): array
    {
        // ...
    }
}
```

### Related issues

- #8497
- #8539
- #11934
- #14728
- #14929

### DIFF
--- a/CHANGELOG-WIP.md
+++ b/CHANGELOG-WIP.md
@@ -1,7 +1,7 @@
 # Release Notes for Craft CMS 5.3 (WIP)
 
 ### Content Management
-- Added the “Link” field type, which replaces “URL”, and can store URLs, `mailto` and `tel` URIs, and entry/asset/category references. ([#15251](https://github.com/craftcms/cms/pull/15251))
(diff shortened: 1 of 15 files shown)


darylknight:
What does "multi-instance" mean?

bencroker:
It means that you can create multiple instances of the same field.

brandonkelly:
Right… now you can drag the same relation field into a field layout multiple times (like you already could with most other field types in Craft 5).

Issue #12453, opened 2022-12-14
[4.x]: Duplicate cp image transforms


SHoogland:
### Description

![image](https://user-images.githubusercontent.com/2057831/207576249-705108d9-bbb7-48c8-9c46-6329d0ce4e82.png)

Somehow 2 of the same transform indexes have been created. and the queue job (generating pending image transforms) keeps iterating over the duplicate items. 

In the job finding the correct image transform and assuming its already there (which it is) but not cleaning up the duplicate.

Looking at the affected transformString its a control panel transform.

I don't really know how to reproduce this yet. but maybe some of you recognize this behavior?

### Craft CMS version

4.3.4

### PHP version

8.1.12

### Operating system and version

Linux 5.10.0-0.bpo.15-amd64

### Database type and version

MySQL 10.6.8

### Image driver and version

Imagick 3.7.0 (ImageMagick 7.1.0-50)

### Installed plugins and versions

Amazon S3	2.0.1
Asset Rev	7.0.0
Blitz Recommendations	2.1.1
CP Field Inspect	1.4.4
Dumper	3.0.1
Elements Panel	2.0.0
Expanded Singles	2.0.0
Feed Me	5.0.4
Field Manager	3.0.4
Freeform	4.0.9
Mailgun	3.0.0
Redactor	3.0.2
Redactor Tweaks	3.0.0
Retour	4.1.9
SEOmatic	4.0.16
Servd Assets and Helpers	3.2.0
Super Table	3.0.7
Typed link field	2.1.5

brianjhanson:
Any chance you're in a multi-server situation? In general, we [only recommend a single queue worker](https://github.com/craftcms/cms/issues/12047#issuecomment-1267811301) to avoid these kinds of issues.

SHoogland:
No, and I don't think it has to do with the queue 🤔 I looked at that issue as well. But I think somehow two of the same transforms are created, and that never gets resolved. which puts the queue in a loop.

brianjhanson:
Are there any clues in the `storage/log` files?

If not, can you send your `composer.json`, `composer.lock`, and a copy of your DB to support@craftcms.com?

PR #15402, opened 2024-07-23
Generate transforms by the specified index from Assets Controller

adrum:
### Description

**TL;DR:** When passing a transform index ID to `/actions/assets/generate-transforms`, the action should generate that exact index.

**Full Story:** We launched a site running 4.10.4 and somehow ended up with duplicate rows in our `imagetransformindex` table. Our template eager loaded transforms, and it yields the second row from the table (id: 21) for a particular transform. When the Assets Controller went to generate the transform by ID, it actually queries the `imagetransformindex` table based on the attributes of the transform, which would yield the first row (id: 20) and not the actual row requested (id: 21). This would yield an ever-growing queue of image transformations that would never flip the `fileExists` to 1 on id 21.

I recognize this solution deals with symptoms and not the root problem, but at least it prevents the queue from infinitely pushing image transforms jobs. I believe this fix is also applicable for 5.x.

We did have multiple processes in our supervisor config, so that was probably the root cause of the duplicate rows. I used these queries to diagnose and clean up the table:

```sql
SELECT 
  assetId,
  format,
  transformString, 
  COUNT(*) as duplicate_count
FROM 
  craft_imagetransformindex
GROUP BY 
  assetId, format, transformString
HAVING 
  COUNT(*) > 1;


DELETE FROM craft_imagetransformindex
WHERE id NOT IN (
  SELECT id FROM (
    SELECT MIN(id) as id
    FROM craft_imagetransformindex
    GROUP BY assetId, format, transformString
  ) as subquery
);

```

### Related issues

https://github.com/craftcms/cms/issues/12453

### DIFF
--- a/src/controllers/AssetsController.php
+++ b/src/controllers/AssetsController.php
@@ -1107,6 +1107,7 @@ class AssetsController extends Controller
     public function actionGenerateTransform(?int $transformId = null): Response
     {
         try {
+            $transformIndexModel = null;
             // If a transform ID was not passed in, see if a file ID and handle were.
             if ($transformId) {
                 $transformer = Craft::createObject(ImageTransformer::class);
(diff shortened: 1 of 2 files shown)


i-just:
Hi @adrum, thanks for this! 
I tweaked your idea a bit to avoid changing the `ImageTransformerInterface->getTransformUrl()` signature, so I’m closing this one in favour of #15477.

PR #15454, opened 2024-07-31
`fields/merge` command

brandonkelly:
### Description

Adds a new `fields/merge` command, which accepts two custom field handles as arguments, and merges one into the other.

The command will do the following;

1. Update any field layouts that contain instances of the outgoing field, and have those instances point to the persistent field instead.
2. Remove the outgoing field.
3. Generate and run a content migration, which calls `afterMergeInto()` on the outgoing field, and `afterMergeFrom()` on the persistent field, giving each field type a chance to react to the merge.

There’s a new `MergeableFieldInterface`, which field types must implement to opt into being mergeable. The interface defines `canMergeInto()` and `canMergeFrom()` methods, giving fields full control over which fields they can be merged with, from both ends of the equation. (By default, the base `canMergeInto()` and `canMergeFrom()` implementations will return `true` if both fields are of the same type, or their `dbType()` results are considered compatible with each other.)

All core field types except Table are mergeable, at least with other fields of the same type – including Matrix, Addresses, and relational fields.

The CLI output for a fields/merge command.<br>

To propagate the change to other environments, just commit the modified project config YAML files, as well as the generated content migration, and then run `craft up` on other environments after pulling in the changes. Be sure to answer `yes` when prompted to run the pending content migration.

### Related issues

- #14554


yeboJoseph:
Could we allow for the second argument to take an array  or more than two arguments?  
Ex:  `fields/merge heading [heading1, heading2, heading3, heading20]` or `fields/merge heading heading1 heading2 heading20`
After updating from Craft 4 to 5, I want to merge certain fields into one. Ex: heading1, heading2, heading3, ...heading20. As of now, I'd have to run this command 20 times. 
`fields/auto-merge` won't work in my instance has it merges _all_ fields with the same settings.

brandonkelly:
(Responded from #16297.)

PR #15467, opened 2024-08-03
Fixed Dutch translations

niektenhoopen:
Minor consistency issues and fixed a typo in "Passkey"

### Description
Fixed the typo in this button:
image

Changed "gemaakt" to "aangemaakt", this is more consistent with the other translations. We write words together when it's possible but "Passkey-verificatie" is allowed to have a dash because it's a combination of an English and Dutch word.


brandonkelly:
Thanks! I’ve made the same edits via [CrowdIn](https://translate.craftcms.com). They’ll be included in the next release.

PR #15468, opened 2024-08-03
Don’t preload defined Twig globals when preloading singles

bencroker:
This PR ignores handles that are defined Twig globals when preloading singles as fallback values. Without this, `Craft::$app->getEntries()->getSingleEntriesByHandle()` is unnecessarily called on values such as `craft`, `currentSite`, `currentUser`, etc. This is a micro optimisation but feels like a better way of loading fallbacks.

If merged, this should be carried over to the 5.x branch as well.


brandonkelly:
Good call, thanks! merged for 4.11 and 5.3.

PR #15472, opened 2024-08-03
fields/auto-merge command

brandonkelly:
### Description

Adds a new `fields/auto-merge` command, which looks for custom fields that are configured identically and are mergeable, and suggests merging them (as a shortcut for calling `fields/merge` manually a bunch of times).

CLI output for the fields/auto-merge command

### Related issues

- #14554


mikolamin:
@brandonkelly there is a typo  !$first->can**m**ergeFrom($other, $reason) instead  !$first->can**M**ergeFrom($other, $reason)
in file cms/src/console/controllers/FieldsController.php:187

brandonkelly:
@mikolamin Thanks for pointing that out! Fixed in 5.3.6.

PR #15474, opened 2024-08-04
4x: Generate transforms by the specified index from Assets Controller

adrum:
### Description

This is just the Version 4 PR of #15402


i-just:
Thanks again! As per #15402, closing in favour of #15477.

Issue #15542, opened 2024-08-15
[5.x]: Link field errors on changing Allowed Link Types


rungta:
### Steps to reproduce

1. Create a new Link field, allow URL link type
2. Add the link field to an entry type
3. Save an element with a URL
4. Edit the Link field, enable any other link type(s) but disable the URL link type
5. Open the element’s Edit page

### Expected behavior

Not sure, but some ideas:
- Warn about this behaviour on the field’s settings page
- Handle the error more gracefully? There will probably be multiple scenarios where the field has data saved in a link type that is no longer enabled or available for various reasons.

### Actual behavior

```
yii\base\ErrorException: Undefined array key "url" in /var/www/html/vendor/craftcms/cms/src/fields/Link.php:330
Stack trace:
#0 /var/www/html/vendor/craftcms/cms/src/web/ErrorHandler.php(79): yii\base\ErrorHandler->handleError(2, 'Undefined array...', '/var/www/html/v...', 330)
#1 /var/www/html/vendor/craftcms/cms/src/fields/Link.php(330): craft\web\ErrorHandler->handleError(2, 'Undefined array...', '/var/www/html/v...', 330)
#2 /var/www/html/vendor/craftcms/cms/src/base/Element.php(6155): craft\fields\Link->normalizeValue('https://example...', Object(craft\elements\Entry))
#3 /var/www/html/vendor/craftcms/cms/src/base/Element.php(4696): craft\base\Element->normalizeFieldValue('legacyUrl')
#4 /var/www/html/vendor/craftcms/cms/src/base/Element.php(4678): craft\base\Element->getFieldValue('legacyUrl')
#5 /var/www/html/vendor/craftcms/cms/src/base/Element.php(2476): craft\base\Element->clonedFieldValue('legacyUrl')
#6 /var/www/html/vendor/yiisoft/yii2/base/ArrayableTrait.php(126): craft\base\Element->craft\base\{closure}(Object(craft\elements\Entry), 'legacyUrl')
#7 /var/www/html/vendor/craftcms/cms/src/base/Element.php(2493): craft\base\Element->traitToArray(Array, Array, false)
#8 /var/www/html/vendor/craftcms/cms/src/web/View.php(675): craft\base\Element->toArray(Array, Array, false)
#9 /var/www/html/vendor/craftcms/cms/src/base/Element.php(3789): craft\web\View->renderObjectTemplate('{{ (_variables....', Object(craft\elements\Entry))
[...]

PR #15544, opened 2024-08-16
Link type no longer allowed

i-just:
### Description
When trying to edit an element with a Link field in the Control Panel and the link type is no longer allowed by the field, set the normalised value to `null` instead of throwing an error.

### Related issues
#15542


brandonkelly:
Resolved via d6ebe930122655c9af198250e97cd6b1d6b23626

Issue #15536, opened 2024-08-14
[5.x]: Able to save entry with a matrix field entry that does not meet validation requirements


jasonlav:
### What happened?
Editors are able to save content to a matrix field that does not pass validation.

### Description
When editing content in a specific order (see steps to reproduce) it is possible to save an entry that contains a matrix field with an entry that does not meet validation.

### Steps to reproduce
1. Create an new section ("Post") and associated entry type ("Blog Post").
2. Create a new matrix field ("Links") and associated entry type ("Link"). The "Link" entry type should have a title and at least one other required field.
3. Add the "Links" matrix field to the "Blog Post" entry type.
4. Create a "Blog Post" entry in the "Post" section with a title of "First Blog Post". Fill out any required fields, leave the "Links" matrix blank. Save it.
5. Edit "First Blog Post." Add a "Link" to the "Links" field. Only fill out the title with "First link", leaving the other required fields blank. Do not save the entry, but click "Back" in your browser.
6. Edit "First Blog Post." Note that "First link" does not appear in the "Links" matrix field. Also note that the "First Blog Post" entry is marked "edited." Save the post.
7. Edit "First Blog Post." "First link" appears in the "Links" matrix field, although the "Link" entry does not pass validation.

### Expected behavior
An entry should not be saved if it does not pass validation, including related matrix field entries.

### Actual behavior
Entry is saved, despite it not passing validation of the matrix field entries.

### Craft CMS version

5.3.4

### PHP version

8.2.22

### Operating system and version

Linux 4.18.0-553.16.1.el8_10.x86_64

### Database type and version

MariaDB 10.6.19

### Image driver and version

GD 8.2.22

### Installed plugins and versions

Amazon S3 2.2.1
Bugsnag 5.0.0
CKEditor 4.1.0

brandonkelly:
Thanks for reporting that! Fixed now for the next release.

brandonkelly:
Craft 5.3.6 is out with that fix. Thanks again!

PR #15548, opened 2024-08-16
Bugfix/15536 nested elements and validation

i-just:
### Description
We’re only showing nested drafts when in element index view mode, so when fetching all uncached nested entries, we should only get drafts in that mode. (This part fixes the reported issue.)

If you create an unpublished draft nested entry that doesn’t validate and then attempt to save the owner, the draft nested entry will be published without being validated, and the errors will only show on the subsequent save. To fix this, duplicate the nested entry as an unpublished draft if the source element is one.

### Related issues
#15536


brandonkelly:
Resolved via 7ed1eeb580f7af9b559f544756819b44e4fb180d – now those nested draft entries will be shown if they exist, and fetched for validation (consistent with the inline-editable blocks view mode).

PR #15569, opened 2024-08-20
Add Craft.cp.announce() and implement for various loading states

gcamacho079:
### Description
Previously, we had live regions that were very localized (i.e., one for element indexes, live regions attached to buttons like the `MultiFunctionBtn`, etc.). This PR attempts to simplify screen reader announcements for status messages by:
- Adding a method for making screen reader accessible announcements
- Setting a default live region when initializing the control panel (`#global-live-region`)
- Adding a live region to the `Slideout` component (in addition to the `Modal` live region implemented by @i-just)
- Listening for the addition/removal of modal layers and modifying the active live region accordingly; this is important because activating a modal layer hides other layers (including the `#global-live-region` and contents of previous layers) from assistive technology.

### Related issues
Resolves PT-2091, PT-2062, PT-2064, PT-2063, PT-2060, PT-2089, PT-2031


i-just (approved):

brianjhanson (approved):
Looks good! It could be useful to split the `announce` function out from the `Craft.cp` object, but that's probably for down the road.

gcamacho079:
> Looks good! It could be useful to split the `announce` function out from the `Craft.cp` object, but that's probably for down the road.

Thank you for reviewing @brianjhanson. Where do you think it would be best to move it to?

brianjhanson:
> > Looks good! It could be useful to split the `announce` function out from the `Craft.cp` object, but that's probably for down the road.
> 
> Thank you for reviewing @brianjhanson. Where do you think it would be best to move it to?

Excellent question and why I think it's for a separate PR. It would probably make the most sense as a separate package but we don't have any of the infrastructure in place for that yet.

PR #15582, opened 2024-08-22
Change jQuery UI version to 1.14.0

bencroker:
It looks like jQuery UI was updated to 1.14.0, at least according to the license file updates in https://github.com/craftcms/cms/commit/3a608f42e9293805b93fd7553973c52d5dcf08ea#diff-5770088c40ad4be0fff29516b741b027fc5ef6a75d7d6728d7417033ca072295

Note: if merged, this should be carried over to the 5.x branch, too.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 4.11.4 - 2024-08-21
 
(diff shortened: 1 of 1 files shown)


brandonkelly:
Doh, thanks!

Issue #15598, opened 2024-08-26
[5.x]: MySQL VIEW causes db/convert-charset command to fail


cweiper:
### Description
After updating our site with a custom plugin to Craft 5, running `php craft db/convert-charset` fails with the following error:

`Converting view_name_here ... Error: SQLSTATE[HY000]: General error: 1347 'db.view_name_here' is not BASE TABLE`

We are using a view to aggregate votes from a submissions table

### Steps to reproduce

1. Create site on Craft CMS 4
2. Create custom plugin that creates a simple MySQL view inside a migration
3. Upgrade to Craft CMS 5
4. Remove the CRAFT_DB_CHARSET and CRAFT_DB_COLLATION settings
5. Run `php craft db/convert-charset`

### Example view creation code
  // Create a view that shows the number of votes for each flavor and order by the number of votes
  $this->execute('CREATE VIEW votes_by_choice AS
      SELECT choiceId, SUM(numberOfVotes) AS numberOfVotes
      FROM votes_submissions
      GROUP BY choiceId
      ORDER BY numberOfVotes DESC');

### Expected behavior
Charset should be converted successfully

### Actual behavior
Command fails with an error

### Craft CMS version

5.3.6

### PHP version

8.2

### Operating system and version

DDEV

### Database type and version

MySQL 8.0

### Installed plugins and versions

- Custom Voting Plugin
"craftcms/ckeditor": "4.0.6",
"craftcms/cms": "^5.0.0",
"craftcms/shopify": "5.1.1",
"nystudio107/craft-retour": "5.0.0",
"nystudio107/craft-seomatic": "5.0.3",
"putyourlightson/craft-blitz": "5.3.1",
"verbb/formie": "3.0.3",

angrybrad:
Thanks for the report - I've created a PR for this here: https://github.com/craftcms/cms/pull/15600

brandonkelly:
Craft 4.12.0 and 5.4.0 are out with that fix. Thanks again!

PR #15600, opened 2024-08-27
Fixed a MySQL error that would occur if you ran `php craft db/convert-charset` with custom views defined in the database

angrybrad:
Issue https://github.com/craftcms/cms/issues/15598

Yii2's `getTableNames()` returns all views, in addition to tables, so we have to do filtering on our end.

Can be merged into 5.x.


brandonkelly:
Resolved via 6ca1e4582accd59f720980c271ef4e652d17b52a instead

Issue #15570, opened 2024-08-20
[5.x]: Entry field queries returning _all entries_ if the `id()` param is used to exclude entry IDs


mmikkel:
### Description

Since upgrading from 5.2.10 to 5.3.4, I'm seeing some odd behaviour with entry queries derived from Entries fields, where adding an `.id(['not', id)` param to the entry field query results in the query **returning every single (non-nested) entry in the site** (except the one(s) with the excluded ID(s)), i.e. completely disregarding the actual relations in the field.

Prior to 5.3.4, this query would return entries related to `entry` in the field `entriesField`, _except_ an entry with the ID `23432234`:

```twig
{% set entries = entry.entriesField.id(['not', 23432234]).all() %}
```

On 5.3.4, the above query will return _all entries in the site_ (except any with the ID `23432234`).

**Crucially**, this only appears to happen if the source entry has been re-saved after upgrading to 5.3.4. I.e. immediately after upgrading to 5.3 things work as before, but once the relation source `entry` is re-saved (either manually or via CLI), the described behaviour manifests.

_I dumped the raw SQL from `entry.entriesField.id(['not', 23432234])` before and after upgrading the site to 5.3.4 (and re-saving the entry) – here's a diff to highlight what changed (likely relevant, the `INNER JOIN relations` part is somehow gone):_ https://www.diffchecker.com/l25MWZtQ/ (the SQL from 5.2.10 is on the left; 5.3.4 on the right).

I'm able to reproduce this on a brand new 5.2.10 install after upgrading to 5.3.4; here's some steps for a reduced test case:  

### Steps to reproduce

1. Install Craft 5.2.10
2. Create a section and an entry type for that section
3. Create an entries field with the handle `entriesField`, and add it to the entry type
4. Create 3 entries in the section; "Entry 1", "Entry 2", "Entry 3"
5. **Add "Entry 2" to the entries field for "Entry 1"**
6. Add the following Twig to the entry template for the section, and visit "Entry 1"'s URL:

```twig
{% set entries = entry.entriesField.id(['not', 23432234]).all() %}

{% for entry in entries %}
    {{ entry.title }}<br/>
{% endfor %}
```

(The element ID in the `.id(['not', ..)` param is completely fictional, and it doesn't matter; the important thing is that the `.id()` param with the `'not'` directive, is used).

As expected, the output should be

```html
Entry 2<br/>
```

7. **Upgrade the site to Craft 5.3.4**
[...]

PR #15613, opened 2024-08-28
ElementQuery `andId` param and method

i-just:
### Description
Adds `andId` method and parameter to the `ElementQuery`. It will be used if the `id` parameter is already set. 

(I’ve not added it to the GraphQL. In GQL, querying a relation field with the `id` param works as expected, even post version 5.3.0)

### Related issues
#15570


brandonkelly:
Resolved #15570 via 2ce4b5babe94f6f9b1d016440ea2c8d9817ead34, so I don’t think there’s a need for this.

PR #15631, opened 2024-08-30
Update app.php

FrDH:
Fixed an odd translation for how the text is used.

### Description
The word "New" is translated as if it was "New class".


brandonkelly:
Thanks!

PR #15646, opened 2024-09-03
Allow DI on ImageTransform

timkelty:
### Description
The goal is to allow developers to write an `ImageTransformer` that builds URLs using properties other than those included in `\craft\models\ImageTransform`, eg `blur`.

This PR does 2 things:
- Replaces new instance creation with `createObject` so `\craft\models\ImageTransform` can be customized through DI.
- Calls `\craft\models\ImageTransform::toArray` without any args, so that all properties are returned.

### Related issues
https://github.com/craftcms/cms/pull/15062

#### Module/Plugin

```php
<?php

namespace modules\testmodule;

use Craft;
use craft\elements\Asset;
use craft\events\DefineAssetUrlEvent;
use craft\helpers\UrlHelper;
use craft\models\ImageTransform;
use yii\base\Event;
use yii\base\Module as BaseModule;

/**
 * @method static Module getInstance()
 */
class Module extends BaseModule
{
    public function init(): void
    {
        parent::init();

        // Defer most setup tasks until Craft is fully initialized
        Craft::$app->onInit(function() {
            Craft::$container->set(ImageTransform::class, [
                'class' => \modules\testmodule\ImageTransform::class,
            ]);


            Event::on(
                Asset::class,
                Asset::EVENT_DEFINE_URL,
                function (DefineAssetUrlEvent $event) {
                    $transform = $event->transform;

                    if (!$transform) {
                        return;
                    }

                    /** @var Asset $asset */
                    $asset = $event->sender;

                    $url = UrlHelper::urlWithParams(
                        "https://foo.com/$asset->path",
                        $transform,
                    );

                    $event->url = $url;
                });
        });
    }
}
```

```php
<?php

namespace modules\testmodule;

class ImageTransform extends \craft\models\ImageTransform
{
    public int $blur = 0;
}
```

#### Template

```twig
{% set transform = {
    mode: 'crop',
    width: 100,
    height: 100,
    quality: 75,
    blur: 40,
} %}
{% set asset = craft.assets().one() %}

{{ asset.getUrl(transform) }}
{{ asset.getImg(transform, ['1.5x', '2.0']) }}
```


src/elements/Asset.php
```diff
@@ -1888,16 +1888,7 @@ public function getUrlsBySize(array $sizes, mixed $transform = null): array
...
-            $sizeTransform = $transform ? $transform->toArray([
-                'format',
-                'height',
-                'interlace',
-                'mode',
-                'position',
-                'quality',
-                'width',
-                'fill',
-            ]) : [];
```
timkelty, line 1968: Getting all attributes instead of specifying, so we pick up any defined by a different class.

Alternatively, we could introduce something like `\craft\models\ImageTransform::getImageAttributes` that returned the attribute names involved in the actual image manipulation.

anhtai9710 (approved):

leevigraham:
@timkelty Finally got round to testing this and it's perfecto!

I forked the 5.x branch and rebased your changes on top. Everything worked fine.

This looks like a simple enough change it could be merged into Craft 4 as well?

image

leevigraham:
@timkelty What's needed to get this merged :)

leevigraham:
@timkelty I had another play with this PR and I can reduce it to just removing the filtering on the `->toArray()` methods and the `$whitelist`.

I can pass through the behaviours custom property by using the `EVENT_DEFINE_FIELDS` event.

```php
Event::on(
    ImageTransform::class,
    Model::EVENT_DEFINE_FIELDS,
    function (DefineFieldsEvent $event) {
        $event->fields['imgixParams'] = 'imgixParams';
    }
);
```

leevigraham:
Yewww!

brandonkelly:
Craft 5.6.0 is out now 🚀

PR #15675, opened 2024-09-05
Fix variable referenced in sidebar keypress handler

gcamacho079:
### Description
The code currently references `ev`, but the parameter is `event`


brandonkelly:
Looks like this got fixed via 8f1062a146c1637a69b9a7dcfcc808a4af80aeae.

PR #15684, opened 2024-09-06
Eager-load fields based on fields actually present in query

brandonkelly:
Refactors how we resolve GraphQL field names to custom fields, so the fields in the element query are referenced rather than the global list of fields. That way any handle overrides can be accounted for.

### DIFF
--- a/src/base/EagerLoadingFieldInterface.php
+++ b/src/base/EagerLoadingFieldInterface.php
@@ -13,7 +13,7 @@ namespace craft\base;
  * @author Pixel & Tonic, Inc. <support@pixelandtonic.com>
  * @since 3.0.0
  */
-interface EagerLoadingFieldInterface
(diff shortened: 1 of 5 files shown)


howells:
@brandonkelly hey Brandon - think I'm running into this issue right now; queries with handle overrides work perfectly in graphiql, but not via my gql API. I assume this is the fix I need, so how can I use it now?

brandonkelly:
@howells This fix only improves the efficiency of queries, and it’s not passing tests yet so I wouldn’t recommend using it.

Craft 5.4.2 fixed a bug related to overriding field handles. If you’re running 5.4.2+ and still experiencing an issue, please report it as a new GitHub issue.

howells:
@brandonkelly no worries I got working again so can't have been related to this; a mystery.

Issue #15678, opened 2024-09-05
[5.x]: Nested addresses owned by a custom field can appear in the wrong context


AugustMiller:
### Description

I think this is two distinct issues:

1. Address elements nested within a custom field on a user can appear in the main "Addresses" view.
2. After deleting an address field, the `fieldId` is not cleaned up, nor are the nested elements deleted.

Running garbage collection didn't seem to help, but this `resave` command took care of it:

```bash
ddev craft resave/addresses --element-id [[offending address ID]] --set fieldId --to "fn(\$element) => null"
```

### Steps to reproduce

1. Create an **Addresses** field;
4. Add it to the **User Profile** field layout;
5. Create a nested address element on a user;
6. Observe that new address in the **Addresses** screen (despite it being owned by a custom field);

Additionally:

1. Delete the custom **Addresses** field;
2. View the **Addresses** screen for the same user;
3. Observe error about a missing field;

### Expected behavior

Addresses owned by custom fields are excluded from the native address list. (I think?)

### Actual behavior

Addresses owned by custom fields leak into native address list.

### Craft CMS version

5.4.1

### PHP version

8.2.20

### Operating system and version

macOS/Docker

### Database type and version

MySQL 8.0.36

i-just:
Thanks for the clear description, August! I raised a PR for this.

brandonkelly:
Fixed in 5.4.2!

PR #15685, opened 2024-09-06
Bugfix/15678 user addresses and fields

i-just:
### Description
**First issue:**
User‘s nested addresses owned by a custom field appear in the main Addresses view when editing that user.

**Fix:**
`User->getAddresses()` should return user's "native" addresses - the ones linked via the `addresses` attribute and not via a custom Addresses field.

The above also fixes the issue where deleting the custom Addresses field caused an error about a missing field when visiting the user's Addresses screen.

**Second issue:** 
After deleting an Addresses field, the `fieldId` is not cleaned up, nor are the nested elements deleted. Garbage collection also doesn't clear those out.

**Fix:**
This happens because there's no foreign key constraint on the `addresses.fieldId` field. 
I added a migration that will delete all the addresses that are linked to a field that's been hard deleted and then add a foreign key constraint to the `fieldId` column. I also adjusted the `Gc` service to run partial elements deletion on Addresses.

### Related issues
#15678


i-just:
I moved it back to draft as there's a way to approach the second problem that is more consistent with what we already have in place.

brandonkelly:
Went with a garbage collection-only approach so we can avoid adding a new migration, and because relying on cascading deletes was only a partial solution anyway (requiring `deletePartialElements()` to finish the job).

Issue #15704, opened 2024-09-10
[4.x][5.x]: Twig CVE in 3.12.0


jeffreyzant:
### Summary
Twig should be updated to 3.14.0 as suggested by composer audit. Is there any reason why currently only the patch releases for twig are installed by ~3.12.0 and we are not using minor versions ^3.12.0?

### Details
```
+-------------------+----------------------------------------------------------------------------------+
| Package           | twig/twig                                                                        |
| CVE               | CVE-2024-45411                                                                   |
| Title             | Twig has a possible sandbox bypass                                               |
| URL               | https://github.com/advisories/GHSA-6j75-5wfj-gh66                                |
| Affected versions | >=3.0.0,<3.14.0|>=2.0.0,<2.16.1|>=1.0.0,<1.44.8                                  |
| Reported at       | 2024-09-09T20:19:26+00:00                                                        |
+-------------------+----------------------------------------------------------------------------------+
```

### Craft CMS version

4.12.1, etc.

### PHP version

8.3

### Installed plugins and versions

-

brandonkelly:
Thanks for bringing that to our attention! Craft 4.12.2 and 5.4.3 are out with Twig 3.14.

PR #15710, opened 2024-09-10
bump twig version

i-just:
### Description
Bump twig version to `~3.14.0`.

### Related issues
#15704


sfsmfc:
@brandonkelly Will this also be fixed in 4.11.x or only in 4.12.x?

brandonkelly:
@sfsmfc Only 4.12.2+ and 5.4.3+. (Both updates are out now.)

PR #15711, opened 2024-09-10
[4.x] Ability to limit the countries available for selection when editing an address

nfourtythree:
### Description
Some projects require the need to limit the list of countries available for selection.

This PR adds and event that, when triggered, can manipulate the countries return. This gives developers control over which are available to content authors.

In the changes I have updated instances where the code where we were calling the addressing libraries core method and replaced it with our own so we have the chance to trigger an event.

The instances replaced are those that end up with the list being shown to a content author. 

The instances of a call to the library's `getList()` method that has not been changed relate to validation of the address element. This call is still there to validate against the entire list of possible countries.

### Related issues
craftcms/commerce#3306


i-just (commented):
Looks good to me!

I would maybe just consider adding a bit more info to the comment for the new `Addresses::EVENT_DEFINE_ADDRESS_COUNTRIES` event around adding new countries/changing country codes. Just a short summary of what we talked about - if you want to add new countries, you will have to use DI and create your own country repository?

i-just (approved):

PR #15720, opened 2024-09-11
Basic HTTP authentication filters

timkelty:
### Description

- Adds `\craft\filters\SiteFilterTrait::$enabled` so all site filters can be easily toggled (eg per environment)
- Adds `\craft\filters\BasicHttpAuthLogin`, providing similar (but more configurable) functionality to `\craft\config\GeneralConfig::$enableBasicHttpAuth`, allowing users to log in as a Craft user with Basic HTTP auth credentials.
- Adds `\craft\filters\BasicHttpAuthStatic`, making it easy to block access to a site with a static username/password
  - defaults to `CRAFT_HTTP_BASIC_AUTH_USERNAME`, `CRAFT_HTTP_BASIC_AUTH_PASSWORD`
- Deprecates `\craft\config\GeneralConfig::$enableBasicHttpAuth`

Both filters use the same `\craft\filters\SiteFilterTrait` as our other filters, meaning they can be limited by site if desired, and they only apply to site requests.

### Example usage

```php
<?php

// config/app.web.php
return [
    // Equivalent functionality to\craft\config\GeneralConfig::$enableBasicHttpAuth
    'as optionalBasicLogin' => [
        'class' => \craft\filters\BasicHttpAuthLogin::class,

        // List of action ID patterns, see \yii\filters\auth\AuthMethod::$optional
        'optional' => ['*'],
    ],

    // Required user login for all site requests, when not in production
    'as requiredBasicLogin' => [
        'class' => \craft\filters\BasicHttpAuthLogin::class,
        'when' => fn() => \Craft::$app->env !== 'production',
    ],

    // Block all frontend requests with default credentials from env vars
    'as basicStatic' => \craft\filters\BasicHttpAuthStatic::class,

    // Block all frontend requests to siteA with configured credentials
    'as basicStaticWithConfig' => [
        'class' => \craft\filters\BasicHttpAuthStatic::class,
        'username' => 'foo',
        'password' => 'secret',
        'site' => ['siteA'],
    ],
];
```


src/filters/BasicHttpAuthStatic.php
```diff
@@ -0,0 +1,61 @@
...
+    public function beforeAction($action): bool
+    {
+        $currentUser = Craft::$app->getUser()->getIdentity();
+
+        if ($currentUser) {
+            return true;
+        }
+
+        if (!$this->username || !$this->password) {
+            throw new InvalidConfigException('Basic authentication is not configured.');
```
timkelty: We could also just `return true` in this case, which would allow you to keep the filter in place and only set `CRAFT_BASIC_AUTH_USERNAME`, `CRAFT_BASIC_AUTH_PASSWORD` in your applicable environment.

timkelty: This is addressed with the addition of the `enabled` prop.

PR #15736, opened 2024-09-13
Only run connection code once

### LINKED ISSUE #15728: [5.x]: Assets Field Type re-arranging asset lagging 5 to 10 seconds in Safari
### Description

I'm seeing an issue with a client where when having a large amount of assets in an Asset field type (50 items definitely is laggy), the moving around of assets when trying to arrange order completely locks up the page - scrolling seems to work still, but interaction on the whole page freezes. Chrome seems to lag a tiny bit, but Safari is almost unusable.

Anyone encountering this? 

Just upgraded to Craft 5 recently and it seems to be since then that this is happening. Went to upgrade the last few minor updates today to see if it fixed things, but no dice.

Included a video to show:

https://github.com/user-attachments/assets/d3b096f2-33c1-4a89-8510-ca1b1a563bcd

### Steps to reproduce

1. Load up 50 assets in an asset field. Drag and drop using the "Reorder" 4 dot diamond.

### Expected behavior

Smooth dragging.

### Actual behavior

Browser locks up for 5 to 10 seconds.

### Craft CMS version

5.4.3

### PHP version

8.3

_No response_

### Database type and version

_No response_

### Image driver and version

_No response_

_No response_

brianjhanson:
I didn't quite get to the bottom of it, but something about our current drag-and-drop implementation will cause `connectedCallback()` to be called a bunch of times when dragging an asset chip. That, mixed with the fact that the code that attempts to determine if the text within a `crat-element-label` is less than ideal means that the browser will crawl to a halt if you have a significant number of assets. 

This PR stores the `desiredWidth` on the component to avoid doing the calculations over and over again. 

This doesn't 100% resolve the issue (especially in Safari) but it's at least better. I think the rest of the issues are deeper in our drag and drop implementation. Since drag-and-drop is notoriously tricky, it might be worth leaning on a library (like [pragmatic drag-and-drop from Atlassian](https://github.com/atlassian/pragmatic-drag-and-drop)) but that's a larger undertaking

Resolves #15728


brianjhanson:
After a bit more testing, this doesn't help as much as I thought. Back to the drawing board for a bit.

PR #15744, opened 2024-09-17
Drag sorting lag

### LINKED ISSUE #15728: [5.x]: Assets Field Type re-arranging asset lagging 5 to 10 seconds in Safari

brianjhanson:
Refactors the `DragSort` file to be a bit more efficient. I spent a good chunk of the day trying to figure out why this issue is surfacing now but couldn't quite get to the bottom of it. In the end, it seems Safari just isn't quite as efficient doing DOM manipulation as chrome is. 

Fixes #15728


src/web/assets/garnish/src/DragSort.js
```diff
@@ -55,12 +53,11 @@ export default Drag.extend(
...
-        this.getHelperTargetX._draggeeOffsetX = this.$draggee.offset().left;
+        const draggeeOffsetX = this.$draggee.offset().left;
+
         return (
-          this.getHelperTargetX._draggeeOffsetX +
-          (this.mouseX -
-            this.mouseOffsetX -
-            this.getHelperTargetX._draggeeOffsetX) /
+          draggeeOffsetX +
+          (this.mouseX - this.mouseOffsetX - draggeeOffsetX) /
```
brandonkelly, line 60: The reason for storing these values as properties on the function is that it drastically improves JS efficiency, because JS doesn’t have to garbage collect a bunch of variables.

brandonkelly:
The performance in Safari is better with this PR, but sorting is a lot buggier in both Firefox and Safari.

Before:

https://github.com/user-attachments/assets/5fcfb389-166d-44b2-998a-2979d42fe014

After:

https://github.com/user-attachments/assets/83dfaeb7-8803-497f-b486-6132b6a9714c

Worth noting that Safari performance is on par with Firefox in Craft 4. So whatever is causing the lag is Craft 5-specific. I don’t think it’s anything to do with the Garnish drag classes.

brianjhanson:
Yeah, I wasn't able to figure out what about 5 is causing the issue. 

At least part of the issue seems to come from `CraftElementLabel.js`, however, even switching that to a `span` doesn't bring performance in line with Chrome. 

Part of it looks to come from side effects related to `FormObserver._checkFormAfterDelay()` but again, even completely commenting that out didn't move the needle much. 

With both of those commented out, the main issue flagged in the performance audits is all of the DOM manipulation, which is what I was hoping to address here. I probably should have included some changes to `CraftElementLabel` but I felt like the performance was in line without them. 

I'll close this for now so I can take another look, but if you have any ideas about things we've changed that might be causing this issue, I'd appreciate the insight.

PR #15772, opened 2024-09-21
Small `StringHelper::toHandle` improvements

lexisother:
### Description

Now also replaces `!`, `,`, `/` & `\` as these are all invalid characters in handles.


brandonkelly:
Thanks for the PR! Fixed this for the next Craft 4 and 5 releases by excluding **all** non-alphanumeric/underscore characters (979e992010e09e02dbe91b2beab81abe87e28dc1).

brandonkelly:
Craft 4.12.4 and 5.4.5 are out with that fix. Thanks again!

Issue #15784, opened 2024-09-24
[4.12.4]: Neo Blocks are not editable after update from 4.12.3


mfell:
### Description

After an update from 4.12.3 to 4.12.4, Neo Blocks are not reacting anymore -> see screen capture video.

The video has bad quality, but you can see at the begining I'm on 4.12.4 and later on I downgreaded to 4.12.3 via composer (DDEV).

https://github.com/user-attachments/assets/6085f4a0-32d8-42f2-bf56-6da8950f9046

### Steps to reproduce

1.

### Craft CMS version

4.12.4

### PHP version

8.1

### Installed plugins and versions

		"craftcms/ckeditor": "3.9.0",
		"craftcms/cms": "4.12.3",
		"doublesecretagency/craft-cpbodyclasses": "2.3.0",
		"doublesecretagency/craft-cpcss": "2.6.0",
		"doublesecretagency/craft-cpjs": "2.6.0",
		"ether/seo": "4.2.3",
		"nystudio107/craft-vite": "4.0.10",
		"robuust/craft-reverserelations": "2.0.0",
		"ryssbowh/craft-restrict-deletions": "2.2.2",
		"spicyweb/craft-embedded-assets": "4.1.2",
		"spicyweb/craft-neo": "4.2.14",
		"verbb/field-manager": "3.0.9",
		"verbb/hyper": "1.3.1",
		"verbb/user-group-field": "3.0.2",
		"vlucas/phpdotenv": "^5.4.0",
		"xpertbot/craft-wheelform": "3.2.1"

AlexShearcroft:
To follow on with the above raised by @mfell it's also affecting other plugins within the Neo field, in particular Redactor and Hyper where they are not rendering within the CMS.

![Screenshot 2024-09-24 at 10 56 32](https://github.com/user-attachments/assets/3ce7b44d-5ab7-4f8a-9c77-9e0efad34fcb)

When checking the console log it returns the following error:
![Screenshot 2024-09-24 at 10 55 18](https://github.com/user-attachments/assets/24541961-a17d-46c4-9778-a2625d21c345)

mfell:
@AlexShearcroft Oh … I hadn’t noticed this yet. Yes, I can confirm, it’s happening in my case as well!

i-just:
Hi, thanks for reporting! I raised a PR for this.

brandonkelly:
Craft 4.12.4.1 and 5.4.5.1 are out with that fix. Thanks again!

[...]

PR #15786, opened 2024-09-24
change the scope of the croppedName variable

i-just:
### Description
Changes the scope of the `croppedName` variable so that it’s available in the next block, too.

(I was only able to reproduce the Neo issue in v4, but it looks like this error could also be triggered in v5.)

### Related issues
#15784


Jelmer-Ketelaar:
Can this PR be reviewed and merged as soon as possible? Our customers are facing ongoing issues

PR #15788, opened 2024-09-24
Drag sorting lag take 2

brianjhanson:
Disables `craft-element-label` and removes `.action-btn` while dragging for slightly better performance. 

Something about drag and drop still doesn't get along with Safari, but this is at least a step in the right direction. I'd like to dig into this again down the road, but for now at least this offers an improvement. 

If this PR doesn't help enough, switching your asset field to "list" seems to perform much better overall.


src/web/assets/cp/src/js/BaseElementSelectInput.js
```diff
@@ -853,7 +879,7 @@ Craft.BaseElementSelectInput = Garnish.Base.extend(
       }, 200);
     },
 
-    selectElements: async function (elements) {
```
brandonkelly: @brianjhanson why did you remove `async`? Everywhere this method is called, a promise is expected.

brianjhanson: @brandonkelly my bad, at one point I saw this in the heap and was wondering if something was waiting on the `async`. I just missed putting it back before pushing it up.

Issue #15797, opened 2024-09-26
[5.x]: Links in error summaries no longer work with Matrix fields


MoritzLost:
### Description

When saving an entry causes validation errors, the error messages in the summary above the form are anchor links that should scroll to the field in question:

![Screenshot 2024-09-26 at 16 35 19](https://github.com/user-attachments/assets/c0374546-43b7-4d4b-94ca-17cc99aa0ce7)

I found that this no longer words if the error is inside a Matrix field. I've only checked Matrix fields that are set to display as inline blocks. Might also be an issue if the field is set to card or element index view.

### Steps to reproduce

1. Create a Matrix field set to inline view and select an entry type with a required field.
2. On a new entry, create an entry in the Matrix field and leave the required field empty.
3. Save the entry.
4. Click on the link in the error summary about the empty required field.

### Expected behavior

The links should always scroll to the field in question, even inside an inline matrix block.

### Actual behavior

Clicking on the links doesn't do anything.

### Craft CMS version

5.4.5.1

### PHP version

8.3

### Installed plugins and versions

-

i-just:
Hi @MoritzLost, thanks for reporting! 

I have tested this against a matrix field in an inline-editable blocks view mode, and it works as expected for the first-level matrix fields. However, it gets confused if the nesting gets deeper. I’ll have a look into that!

MoritzLost:
@i-just You're right, the errors occured in nested Matrix fields one level deep. Forgot to mention that, sorry!

i-just:
Thanks for confirming!

i-just:
I have raised a PR for this. Thanks again for reporting!

brandonkelly:
Craft 5.4.8 is out with that fix. Thanks again!

PR #15811, opened 2024-09-30
anchor summary error to field for nested matrix "blocks"

i-just:
### Description
For a matrix in a matrix where the view mode is set to inline-editable blocks, we need to continue to rewrite the errors all the way up to the element that contains the first inline matrix field. However, when adding a prefix to errors in the summary, we need to ensure only the immediate owner prefix is added. That way, we’re able to reliably anchor from the error summary to the affected field.

### Related issues
#15797


brandonkelly:
It’s expected that recursively-nested fields will get error attribute prefixes that reveal the full path to the element.

For example, locally I have a required Entries field that is nested in two parent Matrix entries, and when I save it without a value, the top-level entry gets an error with the following key:

```
navigation[fe6c1060-e895-4a44-9e15-b0341e6a58f8].elements[7872abb1-22ae-4523-9dde-95b51b239ac8].entry
```

- `navigation` is the first Matrix field
- `fe6c1060-e895-4a44-9e15-b0341e6a58f8` is the `navigation` Matrix field’s entry UUID
- `elements` is the nested Matrix field
- `7872abb1-22ae-4523-9dde-95b51b239ac8` is the `elements` Matrix field’s entry UUID
- `entry` is the required Entries field

With this PR, the error’s key is only:

```
elements[7872abb1-22ae-4523-9dde-95b51b239ac8].entry
```

Which leaves out critical information, as the top-level entry doesn’t know anything about an `elements` field.

i-just:
Thanks for the chat on this one, @brandonkelly! I reverted the previous change and used a different approach.

src/controllers/ElementsController.php
```diff
@@ -1054,7 +1058,21 @@ private function _errorSummary(ElementInterface $element): string
...
+                                    if (!$multiNested) {
+                                        continue 2;
+                                    }
+                                }
+                                // if it's a multi-nested error key for matrix in blocks mode
+                                // manipulate the key to only reference the matrix field, entry and inner field
+                                if ($multiNested && $layoutElement instanceof CustomField) {
+                                    if (
+                                        $layoutElement->getField() instanceof Matrix &&
+                                        $layoutElement->getField()->viewMode === Matrix::VIEW_MODE_BLOCKS
```
brandonkelly: @i-just Do you think it would be save to skip this check, so the same thing works for Neo? If it’s a Matrix field in cards/index view, then there won’t be any inputs to match for the error anyway.

i-just: I tested that, and it won’t affect Neo.

Neo still uses [block indexes](https://github.com/spicywebau/craft-neo/blob/5.2.9/src/Field.php#L856) when prefixing model errors (switching to `$block->uid` makes the prefix look as expected). 
They’re also not using `errorKeyPrefix` when [rendering the blocks form](https://github.com/spicywebau/craft-neo/blob/5.2.9/src/templates/block.twig#L199), which means the error key is set to just the handle of the individual field (e.g. `myPlainTextField`). Passing `{'errorKeyPrefix' : "#{handle}[#{block.uid}]"}` as the 3rd param prefixes the errors as expected. 

With those two changes in place, anchoring to erroring fields inside Neo works as expected, and the lines you mentioned don’t affect that.

Issue #15822, opened 2024-10-02
[5.x]: TypeError exception when trying to edit a single in the control panel


rob-c-baker:
### Description

Clicking a single in the CMS (Entries > Singles > [Entry]) results in `TypeError: Cannot assign null to property craft\elements\db\ElementQuery::$inReverse of type bool in [...]/vendor/yiisoft/yii2/BaseYii.php:557`.

With this stack trace:

```
TypeError: Cannot assign null to property craft\elements\db\ElementQuery::$inReverse of type bool in [...]/vendor/yiisoft/yii2/BaseYii.php:557
Stack trace:
#0 [...]/vendor/craftcms/cms/src/services/Elements.php(3273): yii\BaseYii::configure()
#1 [...]/vendor/craftcms/cms/src/services/Elements.php(3186): craft\services\Elements->_eagerLoadElementsInternal()
#2 [...]/vendor/craftcms/cms/src/elements/db/ElementQuery.php(1815): craft\services\Elements->eagerLoadElements()
#3 [...]/vendor/craftcms/cms/src/elements/db/ElementQuery.php(1850): craft\elements\db\ElementQuery->eagerLoad()
#4 [...]/vendor/craftcms/cms/src/fields/Assets.php(501): craft\elements\db\ElementQuery->one()
#5 [...]/vendor/craftcms/cms/src/fieldlayoutelements/CustomField.php(120): craft\fields\Assets->getThumbHtml()
#6 [...]/vendor/craftcms/cms/src/base/Element.php(3842): craft\fieldlayoutelements\CustomField->thumbHtml()
#7 [...]/vendor/craftcms/cms/src/helpers/Cp.php(620): craft\base\Element->getThumbHtml()
#8 [...]/vendor/craftcms/cms/src/elements/NestedElementManager.php(421): craft\helpers\Cp::elementCardHtml()
#9 [internal function]: craft\elements\NestedElementManager->craft\elements\{closure}()
#10 [...]/vendor/craftcms/cms/src/elements/NestedElementManager.php(420): array_map()
#11 [...]/vendor/craftcms/cms/src/elements/NestedElementManager.php(628): craft\elements\NestedElementManager->craft\elements\{closure}()
#12 [...]/vendor/craftcms/cms/src/web/View.php(1739): craft\elements\NestedElementManager->craft\elements\{closure}()
#13 [...]/vendor/craftcms/cms/src/elements/NestedElementManager.php(580): craft\web\View->namespaceInputs()
[...]

PR #15839, opened 2024-10-04
Bugfix/15822 eager loading properies and custom fields clash

i-just:
### Description
The reported issue will happen if the lazy eager loading kicks in and you have a custom field with a handle that matches the element’s query public property when the property doesn’t accept the value that would come for that handle from the custom field behaviour (e.g. null).

Solution:
- Change the way `ElementQuery->fields()` merges properties and custom fields so that the properties always take precedence.
- Add extra words to reserved field handles.

### Related issues
#15822


brandonkelly:
@i-just I cherry-picked your `ElementQuery.php` change into `4.x` (e834d29d6ed00c9308e303f2c64fc629be9740d7) since that seemed like a good Craft 4 change as well.

Does that solve the issue on its own? If so, I don’t think we need to add all of those additional reserved field handles – so long as the element query functions/params are taking precedence.

i-just:
@brandonkelly Yes, https://github.com/craftcms/cms/commit/e834d29d6ed00c9308e303f2c64fc629be9740d7 solves the issue on its own. I was erring on the side of caution with the additional reserved words.

Issue #15827, opened 2024-10-02
[5.x]: Error in advanced element query "select" when adding column alias on custom field


danbrellis:
### Description
When trying to do an advanced element query calling a custom field in the `select` statement, I get an SQL error if I give the custom field a column alias.

**This works**
```
{% set critters = craft.entries()
  .section('critters')
  .orderBy('title')
  .limit(null)
  .select(['title as t', 'scientificName'])
  .asArray()
  .all() %}
```

This doesn't work
```
{% set critters = craft.entries()
  .section('critters')
  .orderBy('title')
  .limit(null)
  .select(['title as t', 'scientificName as l'])
  .asArray()
  .all() %}
```

Note the `as l` for the `scientificName` custom field.

### Expected behavior
An array of objects with keys `t` and `l`

### Actual behavior
SQL error: `Column not found: 1054 Unknown column 'scientificName' in 'field list'`.

### Craft CMS version

5.4.6

### PHP version

8.3.10

### Database type and version

mysql 8.0

brandonkelly:
This is currently expected, as aliased columns are excluded from being mapped to the actual custom field value expressions:

https://github.com/craftcms/cms/blob/d44a2d4628b02db2290ecff4248ba70f664c070d/src/elements/db/ElementQuery.php#L3389-L3390

That said, I can’t find any reason for that limitation, and it seems to be just erring on the safe side, so I’ve removed that check for Craft 4.13 and 5.5. (39bd1648c61e68f0a260363f5d23e29ebbdd6325)

brandonkelly:
Craft 4.13.0 and 5.5.0 are out now with that change 🎉

PR #15840, opened 2024-10-04
allow aliasing of custom fields in element queries when using asArray

i-just:
### Description
Allow aliasing of custom fields in element queries (when using in conjunction with `asArray`).

Example code to test with:
```
$myEntry = Entry::find()
    ->section('blog')
    ->select(['t' => 'title', 'pt' => 'plainText'])
    ->asArray()
    ->one();
```

This was easier to achieve in v4, though it still required extra work. In v5, that’s harder (though certainly not impossible). Tested against v4 and v5, and with overwritten field handles too.
Feel free to discard it if there’s a reason I didn’t think of why this shouldn’t be implemented.

### Related issues
#15827

### DIFF
--- a/src/elements/db/ElementQuery.php
+++ b/src/elements/db/ElementQuery.php
@@ -3089,6 +3089,17 @@ class ElementQuery extends Query implements ElementQueryInterface
                     }
                 }
 
+                // allow aliasing of custom fields, but only when using asArray
+                // https://github.com/craftcms/cms/issues/15827
+                if (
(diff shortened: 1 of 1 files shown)


brandonkelly:
Custom fields were already included in `$this->_columnMap`; the only problem was that we were ignoring column mappings for columns that had a custom alias. Not really clear why so I’ve removed that check for 4.13 and 5.5 (39bd1648c61e68f0a260363f5d23e29ebbdd6325).

PR #15854, opened 2024-10-08
fix: prevent db/backup from exporting triggeres twice

gopeter:
### Description

If plugins include triggers in their tables, using `craft db/backup` exports them twice, leading into importing problems since triggers have to be imported after the data to prevent their execution.


angrybrad (approved):

brandonkelly:
Thanks for the PR! Cherry-picked your commit into the `4.x` branch instead, as we decided it makes sense to fix it there as well.

brandonkelly:
Craft 4.12.6 and 5.4.7 are out with that fix. Thanks again!

PR #15891, opened 2024-10-14
Handle IDN URL validation

### LINKED ISSUE #15989: [5.x]: The new Linkfield is not supporting some URL's
### Description
https://www.graubündensport.ch

### Steps to reproduce

1. Fill the follwoing url in a linkfield of craft 5: https://www.graubündensport.ch
  - contains a german umlaut
  - is a 302 redirect
3. The entry can't be saved, because the url is said to be invalid

### Expected behavior
The url saves 

### Actual behavior
The entry can't be saved because the url is invalid

### Craft CMS version

5.4.9

### PHP version

8.2

Linux

### Database type and version

_No response_

### Image driver and version

_No response_

-

brianjhanson:
Fixes the validation of URLs with special characters (like `ä`). 

I'm not sure if this is the best fix, so I'm definitely open to suggestions here.

Fixes #15989


timkelty:
@brianjhanson ready for undraft and review, I think?

brianjhanson:
@timkelty yep, my bad. Marked as ready now.

brandonkelly (changes requested):
The JS needs to be updated as well – compare the difference in how the input handles `https://www.graubündensport.ch` vs `https://www.graubundensport.ch`.

brianjhanson:
@brandonkelly updated in 75f46a32f8b8614d00d3da233013960a9bcf9d6b

PR #15904, opened 2024-10-16
Ensure focal point tool can be used without dragging motion

gcamacho079:
### Description
This PR ensures the focal point move can be accomplished via a click versus requiring a dragging motion. To operate:

1. Click the focal point to pick it up. The grabbing cursor and an additional circle around the focal point will indicate the focal point is picked up.
2. Click elsewhere to drop the focal point.
3. To drop the focal point, click on it again.

### Related issues
Resolves PT-2282


brandonkelly:
Nice work!

PR #15961, opened 2024-10-24
Bugfix/nested elements and revisions 2

i-just:
### Description
**Issue 1:**
- create an entry with the addresses field, add an address and save;
- edit the entry again (e.g. change the title) and save so that we have at least one revision
- go to the revision, double-click the address card
- notice that you can still edit all the fields

**Solution:**
I added support for a static display of addresses (e.g. in revisions).

**Issue 2:**
- create an entry with the addresses field, add an address and save;
- edit the entry again (e.g. change the title) and save so that we have one revision
- go to the revision
- click “Revert content from this revision”
- the content is reverted, but if you go to “Revision 2” and try to open the address, an error will be thrown (Invalid element ID)

**Solution:**
I tracked it down to [this change](https://github.com/craftcms/cms/pull/15898/commits/cdb7c8740f4756f6cc244b370cd857e23642f486#diff-5f792cd4c711263de882c2b77648c06b96220acfb045d42bdc14b142a4d9ebe7R2895-R2897). Opted for dropping that bit of code, but if this is not the right way to go, please LMK. 

**Issue 3:** 
- create an entry with a matrix field in the element index view mode and with table view enabled; (revisions enabled too)
- add a few nested entries to the matrix field and save
- make a few changes, saving in between to create a few revisions
- go to a revision that has the nested entries in the matrix field
- notice that: 
    - checkboxes are still showing even though they do nothing
    - the sortable (diamond) icon doesn’t show for nested entries in a matrix field, but extra work was needed to have them not show in the product > variants table
    - if you switch to a table view, the inline editing button shows

**Solution:**
After chatting with Nathaniel (who reported those issues), we opted for:
- hiding checkboxes when in static mode
- ensuring the sortable icon won’t show in the static mode without each input field having to add the same check for static mode individually
- do not initiate inline editing if in static mode

Before:
Screenshot 2024-10-24 at 15 39 01

After:
Screenshot 2024-10-24 at 15 39 14

### Related issues
n/a


brandonkelly:
> **Issue 2:**
> 
> - create an entry with the addresses field, add an address and save;
> - edit the entry again (e.g. change the title) and save so that we have one revision
> - go to the revision
> - click “Revert content from this revision”
> - the content is reverted, but if you go to “Revision 2” and try to open the address, an error will be thrown (Invalid element ID)
> 
> **Solution:** I tracked it down to [this change](https://github.com/craftcms/cms/pull/15898/commits/cdb7c8740f4756f6cc244b370cd857e23642f486#diff-5f792cd4c711263de882c2b77648c06b96220acfb045d42bdc14b142a4d9ebe7R2895-R2897). Opted for dropping that bit of code, but if this is not the right way to go, please LMK.

Looks like [this change](https://github.com/craftcms/cms/commit/e70fd096110fae569f29640b6e05ad43ae55cf2e#diff-5f792cd4c711263de882c2b77648c06b96220acfb045d42bdc14b142a4d9ebe7) also fixed this bug.

PR #15963, opened 2024-10-25
Sortable link types

i-just:
### Description
Makes it possible to sort link types within a Link field’s settings. This then determines the order in which they appear within inputs’ type selects. The type which is set to show first is the default one.

A sortable “Allowed Link Types” setting

### Related issues
- #15659
- #15516


i-just:
@brandonkelly, now that I raised this PR, I wonder if maybe it would make more sense to reuse the existing `src/templates/_includes/forms/checkboxSelect.twig` template instead of creating a copy that does the same but with a few additions. LMK your thoughts when you have a sec, and I can adjust!

brandonkelly:
@i-just Yep agreed it makes sense to combine with `checkboxSelect`. Just made that change and also improved the UI a bit so the disclosure menus are only shown for selected options.

And merged 🚀  Good work!

Issue #15973, opened 2024-10-29
[5.x]: Ordering by number still not fixed when using twig queries.


leonwebcreators:
**What happened?**
When ordering by a number field, the resulted order will be:
6
32
31
30
29
[...]

**This query will result in the above wrong ordered list:**
{% set entries = craft.entries().search('keyword').orderBy('number desc').all() %}

**Ordering with another param works:**
{% set entries = craft.entries().search('keyword').orderBy('weblink desc').all() %}

**Expected behavior**
Able to order by number in Twig

**Actual behavior**
Errors as above

**Craft CMS version**
5.4.9

**PHP version**
8.3

**Operating system and version**
No response

**Database type and version**
PostgreSQL 13/14/15/16

**Image driver and version**
No response

**Installed plugins and versions**
No response

**Related errors (but this one was backend-only)**
#15417

i-just:
Hi, thanks for reporting. I raised a PR for this.

leonwebcreators:
You lord, are a hero. Thanks!

brandonkelly:
Craft 5.4.10 is out with that fix. Thanks again!

PR #15974, opened 2024-10-29
Bugfix/15973 sorting

i-just:
### Description
1. for pgsql, always cast the database value to decimal so that sorting is consistent regardless of where the query comes from
2. for mysql, ensure the text column is cast to char(255) but only for the orderBy clause

I verified that the sorting results for the number and plain text field are the same between the control panel, when querying via Twig or custom php, and when querying via GraphQL.

### Related issues
#15973


bix0r:
Nice! I just ran into this issue today, and this PR fixes my limited scenario. 
php8.3 + postgres:16

PR #15983, opened 2024-10-30
Link label & target customization

brandonkelly:
### Description

Adds new “Show the ‘Label’ field” and “Show the ‘Open in a new tab’ field” settings to Link fields.

The new Link field settings

If either setting is enabled, the link field + label/target settings will be displayed with a light border surrounding them:

A Link field with both new settings activated

`craft\fields\data\LinkData::getLink()` will include the custom label and `target="_blank"` in its returned `<a>` tag when appropriate.

`LinkData::getLabel()` will return the custom label by default, if set. You can also now pass `true` to force the custom label to be returned or `null` if not; or `false` to force the default (link type-provided) label to be returned, ignoring the custom one.

The link target (`null` or `'_blank'`) can be accessed via `LinkData::$target`.

### Related issues

- #15686


thisisjamessmith:
Just a quick note that I think the docs have the true/false parameter of `getLabel()` the wrong way round - might be good to get that corrected:

https://craftcms.com/docs/5.x/reference/field-types/link.html - 

```This is the default behavior when using a link’s link property. If you need to retrieve the label as it was entered (bypassing Craft’s internal fallback logic), use link.getLabel(false).```

AugustMiller:
@thisisjamessmith Good catch—just pushed up a few edits to that paragraph! 💞

PR #15986, opened 2024-10-31
Add HTML5 audio and video player to getPreviewHtml()

urbantrout:
Replaces the generic thumbnail for audio and video files in the sidebar of an assets edit page with an HTML5 audio or video player.


brianjhanson:
Thanks @urbantrout! I wanted to how the audio player sits in the thumbnail so I opened a separate PR [over here](https://github.com/craftcms/cms/pull/16021).

urbantrout:
Thanks!

brandonkelly:
Craft 5.5.0 is out now with that change. Thanks again!

PR #15999, opened 2024-11-01
Use queue for adding matrix entries

### LINKED ISSUE #15980: [5.x]: Matrix initial field not loaded
### Description
Hello !
When a matrix have a one min and max entries inside another Matrix, the field is not initialized, so when the client try to save, we've got an error.

### Steps to reproduce

https://github.com/user-attachments/assets/28b85b38-7ee5-4208-a6be-473b6a3928df

### Expected behavior

Matrix should be initialized

### Actual behavior

Matrix Empty

### Craft CMS version

5.4.9

### PHP version

8.2

_No response_

### Database type and version

_No response_

### Image driver and version

_No response_

-

brianjhanson:
Fixes #15980

For some background, it's important to understand what happens when a matrix field has a `minEntries` setting and is set to inline blocks. In that situation, when the page loads, a bit of javascript will fire off to create the block and save a draft of the current entry. If that matrix field has additional matrix fields within it, the process will be repeated down the tree via JS returned from the initial AJAX request. 

Now, let's say you have a matrix field (`mx1`) set to inline blocks with a `minEntries` of 1. That field is nested within another matrix field (`mx2`) which has a `minEntries` of 2 and is also set to inline blocks. 

In a perfect world, you'd end up with:

```
┌─mx2────────────────────────────────────────────────────────────────────────┐
│                                                                            │
│                                                                            │
│ ┌─mx1───────────────────────────────────────────────────────────────────┐  │
│ │                                                                       │  │
│ │  ┌────────────────────────────────────────────────────────────────┐   │  │
│ │  │                             blocks                             │   │  │
│ │  │                                                                │   │  │
│ │  └────────────────────────────────────────────────────────────────┘   │  │
│ │                                                                       │  │
│ └───────────────────────────────────────────────────────────────────────┘  │
│                                                                            │
│ ┌─mx1───────────────────────────────────────────────────────────────────┐  │
│ │                                                                       │  │
│ │  ┌────────────────────────────────────────────────────────────────┐   │  │
│ │  │                             blocks                             │   │  │
│ │  │                                                                │   │  │
│ │  └────────────────────────────────────────────────────────────────┘   │  │
│ │                                                                       │  │
│ └───────────────────────────────────────────────────────────────────────┘  │
│                                                                            │
└────────────────────────────────────────────────────────────────────────────┘
```

Unfortunately, when the page loads, the top level matrix field will execute two `MatrixInput.addEntry` calls for the `mx2` field, one for each of the `minEntries` it requires. Each of those calls will return additional JS that will execute their own `MatrixInput.addEntry` calls to satisfy the `minEntries` settings of their children. All of these calls will also trigger a `save-draft` call on the parent entry. 

The problem comes in when all these requests get jumbled up. From time to time, the draft will be saved before all the nested entries are ready which causes the nested element editor to clean them up. When that happens and another request expecting that entry to still be around rolls in, things go badly.  

By moving the `.addEntry` calls into the queue (where `save-draft` already lives) the requests stay more orderly. 

An infinite loop example of the new code

https://github.com/user-attachments/assets/b4a266c1-7da2-415d-a80f-13d0a0bd3c30


src/web/assets/matrix/src/MatrixInput.js
```diff
@@ -241,69 +232,78 @@ import $ from 'jquery';
...
-            data: {
-              fieldId: this.settings.fieldId,
-              entryTypeId: this.entryTypesByHandle[type].id,
-              ownerId: this.settings.ownerId,
-              ownerElementType: this.settings.ownerElementType,
-              siteId: this.settings.siteId,
-              namespace: this.settings.namespace,
-              staticEntries: this.settings.staticEntries,
-            },
+        this.elementEditor?.queue.push(async () => {
```
brandonkelly: @brianjhanson This would prevent the entry from getting created if `elementEditor` isn’t set. Would going with `Craft.queue.push(` work instead?

brianjhanson: @brandonkelly yep, seems to work just as well. Thank you!

Issue #15981, opened 2024-10-30
[5.x]: UX Issue on entry title with action buttons


JeanLucEsser:
### Description

When editing an entry, in certain circonstances when the entry title span two lines, action buttons act weirdly and change position. That is true for the French localized version where I guess button labels are not standard width. See attached clip.

### Steps to reproduce

1. Set Craft in French
2. Create an entry with a long title so it spans two lines
3. Make the auto save green check icon appear and disappear

### Expected behavior

The "more ..." action button should be on the same line as the other action buttons.
When the green check icon appears and disappears, the "save" action button should not move.

### Actual behavior

Action buttons change place every time the entry is autosaved.

https://github.com/user-attachments/assets/61173f32-2483-4fa9-9728-fb83e78e812b

### Craft CMS version

5.4.9

### PHP version

8.3

### Operating system and version

MacOS 15

### Database type and version

MySQL 8

### Installed plugins and versions

-

brandonkelly:
Thanks for pointing that out! We’ve improved the header styling for Craft 5.5 (#16001).

brandonkelly:
Craft 5.5.0 is out now with that change 🎉

PR #16001, opened 2024-11-01
Bugfix/15981 entry title adjustments

brianjhanson:
Styling tweaks to the header area to reduce jitter that occurs when the title is long enough to break into a second line.

**Before**

https://github.com/user-attachments/assets/5cb8636b-d71c-47af-988e-deb0a42b6b26

**After**

https://github.com/user-attachments/assets/4751dee3-1748-4a0a-bbc9-1a2d2e1be70e

## Related Issues

- #15981


src/web/assets/cp/src/css/_cp.scss
```diff
@@ -934,6 +934,13 @@ li.breadcrumb-toggle-wrapper {
 }
 
 #page-title {
+  --page-title-columns: 1fr 34px;
```
brianjhanson, line 937: Added a CSS variable in case plugin authors need to tweak the columns up there. I don't think it's a designated extension point though, so maybe it's not necessary.

Issue #15956, opened 2024-10-23
[5.x]: Nested entries edited while in a draft are auto-applied to parent entry


gateszies:
### Description
Our authors are facing a similar problem as discussed over [here](https://github.com/craftcms/cms/discussions/14603). We're using a Matrix-based page builder with nested entries displayed as cards. This has been a huge upgrade for authoring pages, but we've been running into an issue related to drafts. When authors create a draft on the page, they expect that any edits made to the page sections will be included in the memory of that draft. Instead, any edits made to the nested entries are auto-saved to the current version of the page. 
https://github.com/user-attachments/assets/eef32195-f739-4080-b442-6b05b45aa81a

### Steps to reproduce

1. Create a draft on a page
2. Edit a nested entry section
3. Go to the current version of the page and check the edited section
4. The edits made to the edited nested entry section are now applied to the current version of the page

### Expected behavior

When creating a draft on a page, the nested entry content should be tied to that page draft. The revisions should not be auto-applied to the current version of the page until the page draft is applied.

### Actual behavior

Edits made to nested entries while in a page draft are auto-applied to the current version of the page. Especially on a page that exclusively uses a page builder with nested entries, this makes the concept of a page draft non-functional.

### Craft CMS version

5.4.7.1

### Installed plugins and versions

-

gateszies:
It seems the current advice would be to create individual drafts for each nested entry. This has a few problems - the first being that there isn't a way to create a draft in the nested entry drawer view. Our authors are primarily using the slide-out panels to edit elements on a page but the only place to create a draft is if the nested entry is opened in a new tab.
![Image](https://github.com/user-attachments/assets/19d1dbc4-ba70-4768-bd4d-c6a7e4d21bb7)

The bigger issue is the overall draft functionality on nested entries. When an author goes to edit a page with page builder content, they expect that the nested content is in the context of the overall page draft, and therefore shouldn't get applied to the current version unless the page draft is applied. I can see instances where this might not be the case, but in the context of a page builder, it's crucial to have all nested entries tied to overall page draft. 

The below shows what a typical page looks like in this context. Currently the only way to use drafts is to individually open each section in a new tab, create a draft for each, and subsequently manage and apply every draft on every section.
![Image](https://github.com/user-attachments/assets/447827da-b00b-4a85-b865-0fbe24abd2d2)

gateszies:
[...]

Issue #15713, opened 2024-09-10
[5.x]: Matrix Entries status - Live preview


rgauthier4c4:
### Description
When editing an entry and opening the live preview modal, it is not possible to change the status of a matrix block and see the changes in the preview.

### Steps to reproduce

1. Edit an entry with a matrix block
2. Open the live preview
3. The blocks are shown in the preview
4. Change the status of a single block to be disable
5. The block is still showing in the preview

### Expected behavior
The block should disappear from the page.

### Actual behavior
The block is still there.

### Craft CMS version

5.4.1

### PHP version

8.2

### Database type and version

mysql 8.0

### Installed plugins and versions

-

i-just:
Hi, thanks for getting in touch! What view mode is this matrix field using?

rgauthier4c4:
> Hi, thanks for getting in touch! What view mode is this matrix field using?

The field is in "as card" view mode

i-just:
Thanks for confirming! 
I think I see what you mean. When you change the status, you don’t save the nested entry (block), right?

rgauthier4c4:
Exactly, I've played around a little bit an I think I understand the problem now.

Every nested entries needs to be saved/published before his parent can display the change in the live preview.

Live preview is only useful if the entry contains basic fields to display the drafts changes.

i-just:
We’ll discuss this internally.

bliswebagency:
Would just like to add we noticed this while saving draft pages where their nested entries (within a content builder matrix) would be saved with a live status instead of draft. Changing out of card view does save the draft correctly but the card view would be ideal as it can offer a better editing experience in many cases.

Edit: just saw this was already covered under planned improvements on the Craft roadmap page

brandonkelly:
Craft 5.5.0 is out which improves on the nested entry workflow quite a bit, including for Live Preview (#16002).

PR #16002, opened 2024-11-03
Nested element workflow improvements

brandonkelly:
### Description

When saving a nested element within a Matrix/Addresses field in card view, the changes are now saved to a draft of the owner element, rather than published immediately. The changes won’t actually go live until the owner element is saved. If the draft is discarded, the changes will be discarded too.

Nested element cards that contain unpublished edits identify themselves with a blue modified indicator. (Newly-added nested elements have a similar indicator too, now.)

https://github.com/user-attachments/assets/b2dbb210-0b5c-46b3-9b2e-3fed4023a766

This **only** affects Matrix/Addresses fields in cards view. Decided to keep element index view working as-is, since that view is more obviously dealing with separate elements that aren’t really “part of” the owner element’s content, like card view is.

CKEditor will be updated for consistency with these changes, as well.

### Related issues

- #14286
- #14603
- #14727
- #15956
- #16039
- #15713


mark-reason:
I'm still having the same issue on version 5.5.2.

You can see the difference in the following videos. The first has the field set as an inline editable block and the draft changes are reflected in the preview, however in the second video I've got the field set as a card but the changes don't show until I click save at which point they are now published.

https://github.com/user-attachments/assets/ad484c98-ee6d-43b8-90eb-a13e421f6e6a

https://github.com/user-attachments/assets/0bb1ce2c-1d43-490e-af2d-bc9999f5676b

brandonkelly:
@mark-reason It’s expected that Live Preview won’t reflect your changes until you press Save, however the change in 5.5.0 is supposed to be that pressing Save doesn’t actually publish the changes immediately. The only known exception to that is when editing an element type that doesn’t support drafts, like global sets. What is the parent element type you’re editing here?

mark-reason:
It was a matrix field within an entry type. It definitely looks to be publishing the changes straight away, see the video with graphql query after.

https://github.com/user-attachments/assets/448d90ec-4ede-46cb-b2eb-f145fbc95f30

brandonkelly:
@mark-reason Doh, looks like it only worked when you open the editor slideout by double-clicking on the card. I never tested going through the “Edit” action menu item.

Fixed for the next release via cf41c5f0de3018bb590b27d0ebb8371e4af83ef0.

brandonkelly:
@mark-reason Craft 5.5.4 is out with that fix. Thanks again!

Michael-Paragonn:
@brandonkelly I see you mention both the cards view and the element index view; what about the inline-editable blocks view? 
image

(EDIT: Posted a [separate issue here](https://github.com/craftcms/cms/issues/17062))

PR #16005, opened 2024-11-04
Fixed bug where element would disappear in a condition rule on reload.

lukeholder:
### Description

When making a condition rule that uses `BaseElementSelectConditionRule` (like the "Related To" rule) the element would not show up after page reload. This is due to the element query looking for the `$sitesService->getCurrentSite()->id` by default.

Fixed by changing the query to look for all sites. This also mimics the rule’s matching logic, which only matches on element ID.

https://github.com/user-attachments/assets/f4861457-1a9b-45b5-a662-34a5d1ce53a1

### Related Issue

https://github.com/craftcms/commerce/issues/3739


brandonkelly:
Thanks! Fixed on 4.x via 233d23c54227ee2b61faf5a0ed8b6d4b04c99b8d.

brandonkelly:
Craft 4.12.9 and 5.4.10 are out with that fix.

PR #16022, opened 2024-11-05
save nested element draft against newest owner

i-just:
### Description
When using the new `actionSaveNestedElementForDraft()`, a “No element was identified by the request.” error was thrown when trying to save a nested element edited in a slideout in the following case:

- create an owner entry that has a matrix field set to cards view mode
- add an entry to the matrix field & save
- edit the owner entry so that a provisional draft is created (don’t reload the page); 
- open the nested entry in a slideout, make a change so that it autosaves and then try to save via the button at the bottom; the “No element was identified by the request.” error is thrown; the slideout is not closed;
(- if you reloaded the owner entry before editing the nested one, everything worked as expected)

This was happening because when the provisional draft of the owner element was created, and then the nested element was opened in a slideout, and then a change was made to that nested entry (triggering autosave) when that first `element/save-draft` call was made, the provisional draft of the nested entry had its ownership saved against the canonical of the owner entry and not the newly created provisional draft. Specifying the `ownerId` when opening the slideout fixes this problem.

### Related issues
n/a


brandonkelly:
Thanks!

PR #16036, opened 2024-11-07
Show the site menu in the “Related To” condition rule modal.

lukeholder:
Also allows rules that extend `BaseElementSelectConditionRule` to modify the element select config.

### Related issues

https://github.com/craftcms/commerce/issues/3739


brandonkelly:
Thanks Luke! 5.5.0 is out with that change.

Issue #15991, opened 2024-11-01
[5.x]: Users multi-factor auth not possible to reset as admin or from cli


roelvanhintum:
### Description
When a user loses the multi-factor auth secret and backup codes (if they where created), it's impossible for an admin to help the user besides creating a new account. The cli has also no option for this.

### Steps to reproduce

1. Setup multi-factor auth
2. Throw away keys

### Expected behavior

I was looking for some way to help out users without messing around in the database.

### Actual behavior

-

### Craft CMS version

5.4.9

### PHP version

8.3.12

### Operating system and version

Linux 4.4.0

### Database type and version

MySQL 8.4.3

### Image driver and version

GD 8.3.12

### Installed plugins and versions

- Amazon S3	2.2.1
- CKEditor	4.3.1
- Enforce password	3.0.0-beta.1
- ImageOptimize	5.0.3
- SendGrid	3.0.0
- Sentry SDK	3.0.0

i-just:
Hi, thanks for getting in touch! We’ll discuss this internally. 

For now, if you have sufficient privileges, you can log in as that user and turn off 2SV by going to their account page > Password & Verification.

brandonkelly:
Craft 5.5.0 is out now with a new `users/remove-2fa` CLI command that can also help with this (#16053).

PR #16053, opened 2024-11-08
add users/remove-2fa command

i-just:
### Description
Adds `users/remove-2fa` CLI command. 
- It requires the user for whom you’d like to perform the removal (the ID, username, or email address of the user account). 
- You can choose to remove all two-step verification methods or a selected one. 
- As with removing via the Control Panel, if you remove the last non-Recovery Codes method, the Recovery Codes will be removed too.

### Related issues
[cms-1327](https://linear.app/craftcms/issue/CMS-1327/cli-to-remove-2fa)
#15991


timkelty:
@i-just @brandonkelly I pushed a tweak to normalize the action ID to `remove-2fa`. Looks like Yii's sluggification doesn't consider alpha-to-number a break.

i-just:
Thanks @timkelty!

PR #16081, opened 2024-11-11
Hide field modified text from screen reader cursor

gcamacho079:
### Description
Hides the text from the screen reader cursor, while still allowing it to be used for the accessible name calculation when referenced via `aria-describedby`


gcamacho079:
Retaining the current behavior for nested matrix elements

Issue #16087, opened 2024-11-12
[5.x]: Some singles are duplicated when the underlying entry type is merged


MoritzLost:
### Description

We updated one of our sites from Craft 4 to Craft 5. After the update, we merged some redundant entry types. After deploying the change with those merges, we ended up with duplicates of some of the single entries:

![Image](https://github.com/user-attachments/assets/886857b2-984c-4a5b-90a8-5159e3b11676)

The sections `404 Seite`, `Datenschutz` and `Impressum` are all singles. In Craft 4, those all had their own entry type with identical settings. In Craft 5, we merged those three entry types into a generic `page` entry type.

I'm pretty confident the entry type merge is responsible, since all other singles that didn't have their entry type merged weren't duplicated. So it looks like there's a bug that causes Craft to create a new entry type in a single section when the entry types that's used in this section is merged with another one. This issue only happened when merging the update into production, not in the local dev environment.

In addition to fixing this behaviour, I need some guidance on how to clean up those duplicated entries. I can't delete them through the CP since that's not possible for singles. I'll need to nuke those entries directly in the DB, but I'm worried I will leave behind a broken state.

### Steps to reproduce

See above.

### Expected behavior

There should never be more than one entry associated with a section. Merging the entry type used in a single should not cause the single to be duplicated.

### Actual behavior

Upon deploying the changes that contain the merged entry types, an additional entry was created for each single whose entry type was merged. Now we have two redundant entries both associated with the same single.

### Craft CMS version

5.4.10

### PHP version

8.3

### Database type and version

MySQL 8

### Installed plugins and versions

-

brandonkelly:
Any chance you can share your Craft 4 database, as well as your Craft 5 Composer files, so we can try to reproduce on our end? If so please send them to support@craftcms.com.

MoritzLost:
@brandonkelly Done! Let me know if you need any other information or files.

brandonkelly:
Nevermind, we’ve been able to reproduce. We’ll look into it.

brandonkelly:
Craft 5.5.1 is out with a fix for this. Thanks for reporting!

MoritzLost:
@brandonkelly Thanks for fixing this!

I got one production site where the duplication glitch already happened, could you advise me on the best way to delete the superfluous single entries? Probably going to remove those from the database manually? I'm worried about leaving junk or accidentally deleting something that belongs to the existing singles …

brandonkelly:
[...]

PR #16102, opened 2024-11-13
Bugfix/16087 singles and entry types

i-just:
### Description
**Issue 1:**
When saving a single section where the entry types were changed, the existing entry was deleted, and a new one was created in its place and disabled.

This happened if you had one entry type assigned to your single and replaced it with another entry type, and also if you had a single with two entry types, the entry was created for the first entry type on that list, you then removed that first entry type and saved the section. In both of those scenarios, the old single entry would be deleted, along with its content and a new one (with a different ID) would be created and set as disabled.

**Solution:**
When ensuring a single entry, if we can’t find the entry by section and type, do another search without the type. If this time, the entry was found, assign the first new entry type and carry on as normal. This way, the old entry is not deleted, and the content from the old entry is preserved (if e.g. old and new types have some of the same fields).

**Issue 2:**
When running the `entry-types/merge` command to merge an entry type used by a single with another entry type and persisting the other entry type, the content of the single entry is lost (even if both entry types use the same fields with the same handles). (Content of entries of any other section type is preserved.)

This happened because when we’re done applying project config changes for a single, the single entry already has the `typeId` changed to the persisting entry type, and so [this query](https://github.com/craftcms/cms/blob/5.5.0/src/migrations/BaseEntryTypeMergeMigration.php#L47-L52) was not finding the single entry and therefore wasn't re-writing the content.

**Solution:**
When running the `BaseEntryTypeMergeMigration`, also query for the single entries by the persisting entry type id.

**Issue 3:** 
When deploying changes from the `entry-types/merge` command to other environments, singles could get duplicated if their entry type changed.

This happened because when applying project config changes, the entry with the old entry type was deleted, and then a new one was created with the new entry type. This was followed by the old entry type being restored.

**Solution:**
When ensuring a single entry, also look through the trashed entries that were deleted with an entry type. As with the solution for the first issue, if one is found, update its entry type.
When running the `BaseEntryTypeMergeMigration`, run the content updates and restore separately for the singles and account for the fact that the single entry will already have the persisting entry type.

### Related issues
#16087


src/services/Entries.php
```diff
@@ -952,16 +953,41 @@ private function _ensureSingleEntry(Section $section, ?array $siteSettings = nul
...
+            ->status(null);
+
         // If there are any existing entries, find the first one with a valid typeId
         /** @var Entry|null $entry */
-        $entry = Entry::find()
+        $entry = $baseEntryQuery
             ->typeId($entryTypeIds)
-            ->sectionId($section->id)
-            ->siteId($siteIds)
-            ->status(null)
```
brandonkelly, line 961: This seems pretty important – at least `sectionId`. We wouldn’t want to fetch an entry from a different section that happens to share the same entry type here.

brandonkelly:
Oops, referenced the wrong “fixed” issue # in 8c29e20771df93ad4473fe55ac63156b24a3c0ad

ul8:
I've just run into all of these issues in a Craft 5 migration, hope this gets released soon :)

src/services/Entries.php
i-just: Sure, but it’s still being used - I only moved it up a few lines into a variable.

PR #16140, opened 2024-11-18
Redirect edit pages to the previous page on save

### LINKED ISSUE #11261: [4.x]:
### Description
Navigating to CP > settings > fields automatically selects "all fields" however if I then select a field, edit and save it I am sent to the field's field group instead of behaving how I would have expected, i.e. sending me back to the "all fields" list view.

### Craft CMS version

4.0.2

### PHP version

_No response_

_No response_

### Database type and version

_No response_

### Image driver and version

_No response_

-

brianjhanson:
A revival of https://github.com/craftcms/cms/pull/12400 which got closed accidentally.

Updates redirect URLs for elements and fields to return back to the referrer (when available) instead of the Element::getPostEditUrl() URL.

As an added bonus, search queries for searches using the vue AdminTable component will now be persisted to the URL.

Fixes #11261, #11246


packages/craftcms-vue/admintable/App.vue
```diff
@@ -717,6 +717,10 @@
           }
           this.reload();
         }
+
+        const url = new URL(window.location.href);
+        url.searchParams.set('q', this.searchTerm);
```
brianjhanson: @brandonkelly I don't think using `q` here will cause any issues, but if you think it's likely to run into another query param, we can use something else.

brandonkelly: Went with `search` to be consistent with element indexes.

PR #16147, opened 2024-11-19
Make sure folder path is not null

boboldehampsink:
$folder->path can be NULL and will throw an error


brandonkelly:
Thanks for the PR! Craft 4.13.2 and 5.5.2 are out with a (slightly different) fix.

Issue #16158, opened 2024-11-20
[5.x]: `Attempting to save an element in an unsupported site` when creating new nested entry for non primary site


dolphiq-allard:
### Description
Craft 5.5.2 throws an error (`Attempting to save an element in an unsupported site`) when trying to add an entry in a CKEditor field. 

In a multi-lingual site I can add entries within the CKEditor field in the main language, however, in the secondary language, an error is thrown and the entry is not added to the CKEditor field. This error was introduced somewhere between version 5.1.x and 5.5.x. I've added multiple entries within the CKEditor field before upgrading to the latest version.

Error from log files:
```2024-11-20 09:21:57 [web.ERROR] [craft\errors\UnsupportedSiteException] craft\errors\UnsupportedSiteException: Attempting to save an element in an unsupported site. in /var/www/html/vendor/craftcms/cms/src/services/Elements.php:3531
Stack trace:
#0 /var/www/html/vendor/craftcms/cms/src/services/Elements.php(1308): craft\services\Elements->_saveElementInternal()
#1 /var/www/html/vendor/craftcms/cms/src/services/Drafts.php(261): craft\services\Elements->saveElement()
#2 /var/www/html/vendor/craftcms/cms/src/controllers/ElementsController.php(212): craft\services\Drafts->saveElementAsDraft()
#3 [internal function]: craft\controllers\ElementsController->actionCreate()
#4 /var/www/html/vendor/yiisoft/yii2/base/InlineAction.php(57): call_user_func_array()
#5 /var/www/html/vendor/yiisoft/yii2/base/Controller.php(178): yii\base\InlineAction->runWithParams()
#6 /var/www/html/vendor/yiisoft/yii2/base/Module.php(552): yii\base\Controller->runAction()
#7 /var/www/html/vendor/craftcms/cms/src/web/Application.php(350): yii\base\Module->runAction()
#8 /var/www/html/vendor/craftcms/cms/src/web/Application.php(649): craft\web\Application->runAction()
#9 /var/www/html/vendor/craftcms/cms/src/web/Application.php(312): craft\web\Application->_processActionRequest()
#10 /var/www/html/vendor/yiisoft/yii2/base/Application.php(384): craft\web\Application->handleRequest()
#11 /var/www/html/web/index.php(12): yii\base\Application->run()
[...]

Issue #16144, opened 2024-11-19
[5.x]: Invalid Configuration , Invalid owner ID when upgrading to 5.5.1.1


AmsalSugihan:
### Description

Upgrading to 5.5.1.1 throws a Invalid Configuration / Invalid owner ID error when save an entry that have nested matrix.

### Steps to reproduce

1. Run composer update all to update to 5.5.1.1
2. Open the site, or try to edit and save any entry that have any nested matrix field.

### Expected behavior
Normal Saving entry

### Craft CMS version

5.5.1.1

### Installed plugins and versions

-

i-just:
Hi, thanks for getting in touch! I’m not able to reproduce that. Could you please share a full stack trace for this error?

mattwardcc:
I'm experiencing something very similar. It appears to be only happening on Matrix blocks where the `View Mode` option had been set to Cards. 

When I try to save the page I get this error:

```
Invalid Configuration – [yii\base\InvalidConfigException](https://www.yiiframework.com/doc-2.0/yii-base-invalidconfigexception.html)
Invalid owner ID: 410
```

I was getting something similar when viewing the page until I converted all my Matrix field to inline-editable blocks. This at lest allowed me to see the entries in the editor. Just can't save them.

The initial part of the stack trace is

```
1. in /app/vendor/craftcms/cms/src/base/NestedElementTrait.phpat line 162
            if (!$primaryOwnerId) {
                return null;
            }
 
            $this->_primaryOwner = Craft::$app->getElements()->getElementById($primaryOwnerId, null, null, [
                'trashed' => null,
                'preferSites' => [$this->siteId],
            ]) ?? false;
            if (!$this->_primaryOwner) {
                throw new InvalidConfigException("Invalid owner ID: $primaryOwnerId");
            }
        }
 
        return $this->_primaryOwner ?: null;
    }
 
    /**
     * @inheritdoc
     */
2. in /app/vendor/craftcms/cms/src/base/NestedElementTrait.php at line 207– craft\elements\Entry::getPrimaryOwner()
            if (!$ownerId) {
                return null;
            }
 
            // If ownerId and primaryOwnerId are the same, return the primary owner
            if ($ownerId === $this->getPrimaryOwnerId()) {
                return $this->getPrimaryOwner();
            }
 
            $this->_owner = Craft::$app->getElements()->getElementById($ownerId, null, $this->siteId, [
                'trashed' => null,
            ]) ?? false;
            if (!$this->_owner) {
[...]

PR #16162, opened 2024-11-20
Bugfix/16158 creating new nested entry in non primary site

i-just:
### Description
How to reproduce:
- have at least 2 sites set up
- create a channel section with the propagation method set to “all” enabled for both sites;
- with an entry type that contains a matrix field
- the matrix field is in the cards view mode with the propagation method set to “none”
- go to the non-primary site, create an entry in that channel and try to add a new nested element; 
- notice “Attempting to save an element in an unsupported site.” error

Reverted the change from #16136 to fix the reported issue.

Added a different fix for #16136.

### Related issues
#16158 
#16144 
#16176
#16196


yingban:
@i-just thanks!

Issue #16156, opened 2024-11-19
[5.x]: `ApplyFieldSaveEvent` has inconsistent `$event->field` value


lindseydiloreto:
### What happened?

_Sidenote..._ `craft\services\Fields::EVENT_BEFORE_APPLY_FIELD_SAVE` is not included in the [Event Generator](https://craftcms.com/docs/5.x/extend/events.html#event-code-generator).

---

I believe there is a bug in the new [`EVENT_BEFORE_APPLY_FIELD_SAVE`](https://github.com/craftcms/cms/compare/703b601ebae2dc2e4d5196d6dfacd24875f0635b...423aaf777ce7743197dc539be0782713bd0eeee9) event. Specifically, the `$event->field` value is inconsistent based on whether you are:

- (A) Changing the field type in the Control Panel (locally), or
- (B) Applying your Project Config changes.

Here is my [current code](https://github.com/doublesecretagency/craft-googlemaps/blob/d1ea2c7c5743a04b030172a1b33cff0804ff3af9/src/helpers/FieldConversionHelper.php#L172-L193) (shown below via Xdebug).

Locally, (scenario A) the `$event->field` registers as a Google Maps field...

<img width="911" src="https://github.com/user-attachments/assets/3972d27f-532b-4529-9726-dedcb8966ae0">

---

 But if you roll back the database and apply the modified Project Config (scenario B), the `$event->field` registers as an Ether field...

<img width="903" src="https://github.com/user-attachments/assets/87b1e4db-683c-49bb-90d8-d9bee1a997bf">

---

This is tangentially related to a very important question...

⭐️ **What is the best way to determine both the _before_ and _after_ types of the changing field?** ⭐ 

I noticed that `ProjectConfig::EVENT_UPDATE_ITEM` emits a [ConfigEvent](https://github.com/craftcms/cms/blob/d5d42bea20e90ae07c78eedacd6a6238480fdb24/src/events/ConfigEvent.php), which contains both an `$oldValue` and a `$newValue`. Is it possible to introduce something similar to the [ApplyFieldSaveEvent](https://github.com/craftcms/cms/blob/d5d42bea20e90ae07c78eedacd6a6238480fdb24/src/events/ApplyFieldSaveEvent.php)?

### Steps to reproduce

You can get the mostly-functional dev version of this feature here...

```js
"doublesecretagency/craft-googlemaps": "dev-v5-features/address-conversion"
```

1. Install Google Maps & Ether Maps.
[...]

PR #16168, opened 2024-11-20
add outgoing field config to the ApplyFieldSaveEvent

i-just:
### Description
Added `oldConfig` to the `ApplyFieldSaveEvent` to make it easier to compare the changes applied to the field.

In addition to what the OP said:
- when saving the field via the control panel
    - the event only gets triggered once; 
    - the field is of a new type 
    - the config is also for the new type; 
- when applying PC
    - the event gets triggered twice (when updating an existing field); 
    - first time, the field is of the old type, but the config is of the new type; 
    - the second time it’s triggered, both the field and the config are of the new type; 

Regardless of how the field is changed, the `Fields->applyFieldSave()` method still has the field record, which hasn’t been updated yet, which means we still have access to the outgoing field’s config. Getting that config and passing it in the event should give an easier way to compare the changes being applied to the field.

### Related issues
#16156


lindseydiloreto:
Thanks @i-just! 🙏 

Just to clarify... would the `oldConfig` value include the **field type**? Or simply the field configuration data?

I'm slightly confused about the difference between `field` and `config` here, but I assume that the `field` is the entire FieldInterface, whereas `config` is simply the field's configuration (as an array).

Once this PR is accepted, I assume my [event logic](https://github.com/doublesecretagency/craft-googlemaps/blob/d1ea2c7c5743a04b030172a1b33cff0804ff3af9/src/helpers/FieldConversionHelper.php#L172-L193) would change to something closer to this...

```php
// If no field is provided, bail
if (!$field = $event->field) {
    return;
}

// If not a Google Maps Address field, bail
if (!($field instanceof \doublesecretagency\googlemaps\fields\AddressField)) {
    return;
}

// If no original field, bail
if (!$event->oldConfig) {
    return;
}

// If original field wasn't a Maps Map field, bail
if (!$event->oldConfig instanceof \ether\simplemap\fields\MapField) {
    return;
}

// ... migrate the field's data...
```

Am I on the right track?

Will this logic be executed consistently between the original save versus applying the PC change?

i-just:
@lindseydiloreto, the `config` includes the incoming (new) field config data (array) and the `oldConfig` would include the outgoing field config data (array or null). One of the items in both of those arrays is the field’s type. Comparing old and new config should be consistent between the control panel save an applying PC changes. 

As to your code, I’d tweak that to check the `config` type instead of the `field` type, so something along the lines of:

```
// If no field is provided, bail
if (!$field = $event->field) {
    return;
}

// If not a Google Maps Address field, bail
if ($event->config['type'] !== 'doublesecretagency\googlemaps\fields\AddressField') {
    return;
}

// If no original field, bail
if (!$event->oldConfig) {
    return;
}

// If original field wasn't a Maps Map field, bail
if ($event->oldConfig['type'] !== 'ether\simplemap\fields\MapField') {
    return;
}

// ... migrate the field's data...
```

I hope this helps!

brandonkelly:
Just refactored the PR a bit. Removed the `$oldConfig` property, and just made sure that `$field` is always set to the field as it used to be configured.

lindseydiloreto:
Just tested this, and it seems to be working perfectly! Thanks @i-just and @brandonkelly! 🎉 🙏 

Here's the final code we're using to determine whether or not to perform the migration...

```php
// If no field is provided, bail
if (!$field = $event->field) {
    return;
}

// If not currently an Ether "Map" field, bail
if (!($field instanceof \ether\simplemap\fields\MapField)) {
    return;
}

// If no new field config, bail
if (!$newConfig = $event->config) {
    return;
}

// If new config isn't an "Address (Google Maps)" field, bail
if ($newConfig['type'] !== AddressField::class) {
    return;
}

// ... migrate the field's data...
```

I believe it is now behaving consistently between CP fieldtype changes & Project Config fieldtype changes.

Thanks again!

PR #16218, opened 2024-11-27
Updates Image Editor slider to update rotation without requiring dragging, hides graduation list from AT

gcamacho079:
### Description
Makes updating the image rotation achievable with a single pointer without dragging. Also hides the slider graduation list from assistive technology.

### Related issues
Resolves PT-2068, PT-2100


brianjhanson:
@gcamacho079 I made some minor adjustments here.

- Tweaked the style of the graduations so I could make the small ones a bit wider
- Added hover effect on the graduations
- Added a transition to the graduations when it's changed via click (but not via drag)

https://github.com/user-attachments/assets/fd26917d-34d7-48fc-9f86-101110a7ccaf

brianjhanson (approved):

PR #16229, opened 2024-11-28
(fix) Hide phpinfo utility is not available

nitriques:
This simple PR prevents the cp from displaying a page that we know would error out.

On production systems, phpinfo should be disabled anyways.

Let me know if you need me to change anything.

Thanks!

### DIFF
--- a/src/services/Utilities.php
+++ b/src/services/Utilities.php
@@ -69,7 +69,7 @@ class Utilities extends Component
         $utilityTypes = [
             UpdatesUtility::class,
             SystemReport::class,
-            ProjectConfigUtility::class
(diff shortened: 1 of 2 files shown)


brandonkelly:
Thanks for the PR! I fixed this using the existing `isSelectable()` method for the next v4 and v5 releases (52eb5f6fa53f1e8eb5f73d942fb5a9cbfa570580).

nitriques:
@brandonkelly Hey thanks! I did search for this kind of api but could not find it !

brandonkelly:
Craft 4.13.4 and 5.5.4 are out with that change. Thanks again!

PR #16249, opened 2024-12-01
“Allow custom colors” Color field setting

brandonkelly:
### Description

Replaces Color fields’ Presets setting with “Palette”, and adds a new “Allow custom colors” setting.

The new Palette and “Allow custom colors” settings

“Allow custom colors” will be enabled for existing Color fields, and disabled for new Color fields by default.

Color fields without any Palette colors will continue presenting how they always have, as a manual color input.

Otherwise, they will present as a color select input, similar to the one within entry type settings, filled with the preset colors.

An expanded Color field

If “Allow custom colors” is enabled, a “Custom color…” option will be listed at the end. When that’s selected, a color input will be displayed below.

A Color field with a custom color value

### Related issues

- #15107
- #16233
- #14488


gcamacho079 (changes requested):
- [x] The Selectize dropdown does not have an accessible name. Because the dropdown is immediately adjacent to the `legend` and it makes visual sense that it would be the name of the dropdown, you can reference the `legend` ID from the input's `aria-labelledby` attribute to give it an accessible name.
- [x] When the “Custom color…” option is selected, tabbing back into the Selectize dropdown causes the custom color part of the form to disappear, affecting keyboard usability. To recreate:
  - Have the custom color option selected
  - From the previous focusable element, tab forward through the UI so that focus lands on Selectize dropdown
  - Continue tabbing. Keyboard focus should have skipped past the custom color inputs, and you have to tab backwards to interact with them
  **Expected behavior:** keyboard focus should land on the custom color inputs immediately after the Selectize.
- [x] The lack of visible, adjacent labels that match the custom color inputs can cause issues for voice control users. I suggest adding visible, adjacent labels to both the "Color picker" and "Hex value" inputs that exactly match, or are part of, each input's accessible name. ([See WebAIM contrast checker for potential layout](https://webaim.org/resources/contrastchecker/))
- [x] At 320px wide screens, the Selectize dropdown overflows the body container

gcamacho079 (approved):
Looks great!

PR #16265, opened 2024-12-03
Feature/cms 660 show read-only settings when `allowAdminChanges` is disabled

i-just:
### Description
Shows read-only settings when `allowAdminChanges` is disabled.

- switched `.revision-notice` to a generic `.content-notice` class so that the styles can be reused for other purposes too
- when admin changes are disallowed, the “Settings” main nav item and all the subpages (except for the plugin settings) are shown, with all the fields and forms being disabled and with a notice at the top of every page
- plugins can opt into having their settings shown when `allowAdminChanges` is disabled; to do so, they need to ensure none of the settings are editable in such case and set `public bool $hasReadOnlyCpSettings` to `true`

Further info are in [Brandon's comment](https://github.com/craftcms/cms/pull/16265#issuecomment-2576459583) below.

Screenshot 2024-12-06 at 07 47 21

### Related issues
[cms-660](https://linear.app/craftcms/issue/CMS-660/show-read-only-settings-when-allowadminchanges-==-false)


Mosnar:
Love this! It might be worth adding a "Learn More" button to a KB article explaining why settings can't/shouldn't be changed in production. Newer Craft developers or over-confident clients might read the current disclaimer as instructions to simply turn on "allowAdminChanges" and potentially get themselves in a mess.

gcamacho079 (changes requested):
@i-just awesome! I pinged you on the Notion page for this feature as I had some follow-up questions.

i-just:
thanks @gcamacho079! I actioned all your comments. Extra info on the Notion page. LMK if anything doesn't make sense.

gcamacho079:
Thanks for knocking out those changes, @i-just! A couple more things I noticed that should also get the `disabled` attribute:

- [x] The `.menubtn` for “Selectable {{ element }} condition” on Entries and Assets field settings pages
- [x] The “Static Rows” field (and the conditionally-revealed fields) on Table field settings
- [x] Info toggle buttons inside disabled table fields (for example, in the “Site Settings” table heading on the Matrix field settings page)

src/helpers/Cp.php
```diff
@@ -2917,6 +2930,7 @@ public static function layoutElementSelectorHtml(
                 'has-custom-width' => $element->hasCustomWidth(),
                 'has-settings' => $element->hasSettings(),
             ],
+            'disabled' => true,
```
brandonkelly: @i-just What’s the reason for this?

brandonkelly:
I [changed things up a bit](https://github.com/craftcms/cms/pull/16265/commits/1a81b86c1a03fabd2ef0870adb15d6581a9f25a3) so now all configurable components have a `getReadOnlySettingsHtml()` method, which by default will return the same thing as `getSettingsHtml()` except with all inputs disabled (similar to `Field::getStaticHtml()`).

Fields, filesystems, and mail transport adapters now override `getReadOnlySettingsHtml()` whenever possible, with the exception of `BaseOptionsField` and `BaseRelationField` (since in those cases, there’s no way to explicitly override `getReadOnlySettingsHtml()` without leaving 3rd party subclasses in the dust, so it’s better to wait until v6 before updating those).

Plugins similarly now have a `getReadOnlySettingsResponse()` method, and a `$hasReadOnlyCpSettings` property. `craft\base\Plugin::getReadOnlySettingsResponse()` will return `$this->getSettingsResponse()` with inputs in the settings HTML disabled (like `ConfigurableComponent::getReadOnlySettingsHtml()` and `Field::getStaticHtml()`), and its `init()` method will automatically set `$this->hasReadOnlyCpSettings = true` if `getSettingsResponse()` hasn’t been overridden.

The point of these changes is to limit the need to check `allowAdminChanges`, in case there’s a reason to render any of these settings without taking `allowAdminChanges` into effect. For example, a form plugin might have a need to render a Plain Text field’s settings even when `allowAdminChanges` is disabled (because maybe the field isn’t getting saved to the project config), and should still be able to call `$field->getSettingsHtml()` and get the normal non-disabled response.

src/helpers/Cp.php
i-just: It was adding a `disabled` attribute to the fields inside the field layout designer’s `.fld-field-library`. But with the changes from https://github.com/craftcms/cms/pull/16374, it’s no longer needed (plus, it looks like I forgot to make it dynamic). I took that out.

PR #16276, opened 2024-12-04
Add "not" operators to text condition rules

kabel:
### Description

Adds the ability to use a `NOT` query modifier on text-based field filters in the element index. This allows users to filter elements by a custom text field with a value that:

* does not begin with the value
* does not end with the value
* does not contain the value
* does not equal the value

Example, using an entry with a custom field labeled "Country", looking for non-US entries:

image

image

### Related issues

N/A


kabel:
This would be a nice follow-up to 077dc2d and #18734 that made it into 5.10.

PR #16279, opened 2024-12-05
darken the code colour in installer failed message

i-just:
### Description
Darken the colour of the `code` tag in the installer failed message.
 
Before:
Screenshot 2024-12-05 at 11 36 28

After:

Screenshot 2024-12-05 at 11 39 37

### Related issues
n/a


i-just:
“PHP 8.3 on pgsql” test should run smoothly once #16280 is merged in.

brianjhanson:
@i-just I hope I'm not stepping on your toes, but I tweaked this a bit. Since the type around the `code` is already monospace, I feel like we don't need to call as much attention to it.

PR #16280, opened 2024-12-05
`_deleteOrphanedFkRows` wrap column name in quotes

i-just:
### Description
In the new `Gc->_deleteOrphanedFkRows()` method, the column name needs to be wrapped in quotes for PostgreSQL.

### Related issues
n/a


brandonkelly:
Doh, missed this and already made the same change via f73a913679d816544e6692f2ff3b8ae60007dfee.

PR #16307, opened 2024-12-10
Updates the aria-label attribute in the getLink method

gcamacho079:


brandonkelly:
Thanks for catching that!

PR #16309, opened 2024-12-10
Stacking issues

### LINKED ISSUE #16298: [5.x]: `z-index` bug with autosuggest inputs
### What happened?

When an autosuggest input appears near the bottom of a page that occupies the full height of a window, I'm observing a visual "stacking" glitch with a couple elements in the footer:

z-index stacking bug

- The edition badge appears to use `relative` positioning without any corresponding `top` / `left` / etc. values;
- The  "Go" arrow is given an explicit z-index by the browser [due to it's use of `opacity`](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_positioned_layout/Understanding_z-index/Stacking_context);

### Craft CMS version

 5.5.4

### PHP version

8.2

_No response_

### Database type and version

_No response_

### Image driver and version

_No response_

_No response_

brianjhanson:
Fixes some stacking issues in Safari

Fixes #16298


brandonkelly:
Thanks!

PR #16332, opened 2024-12-13
Allow `DateTimeHelper::humanDuration` to support custom language localization

kabel:
### Description

This helper is exposed through the twig filter `|duration` and it would be helpful to pass a custom language to this function without having to override the site language.

### Related issues

N/A

### DIFF
--- a/src/helpers/DateTimeHelper.php
+++ b/src/helpers/DateTimeHelper.php
@@ -762,7 +762,7 @@ class DateTimeHelper
      * @return string
      * @since 4.2.0
      */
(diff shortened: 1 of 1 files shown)


brandonkelly:
Thanks for the PR! Added via 761208f7d36886cb29aa2a8bcb95b9d08044b9bf instead (for Craft 4.14 and 5.6), as it needed a test case.

brandonkelly:
Craft 4.14.0 and 5.6.0 are out with that change. Thanks again!

PR #16355, opened 2024-12-23
Redirects via config file

timkelty:
### Description
Adds configurable redirects via `config/redirects.php` that are only parsed just before serving a 404, so there isn't any added work for normal route/controller requests.

By default, redirects are 302 (temporary) and match the full request path (`Craft::$app->getRequest()->getFullPath()`) with leading and trailing slashes removed. Yii's [URL rule pattern match replacement](https://www.yiiframework.com/doc/guide/2.0/en/runtime-routing#url-rules) can be used in `to`/`from`.

Matching logic can be overridden via `\craft\web\RedirectRule::match` callback. When `match` is used, `to`, `from`, and `caseSensitive` are ignored.

Also provides a `\craft\web\ErrorHandler::EVENT_BEFORE_REDIRECT` event, should any plugins want to know about a handled redirect.

```php
<?php

return [
    // Path match (case-insensitive by default)
    'redirect/from' => 'redirect/to',

    // Path match with Yii URL Rule named parameters
    // https://www.yiiframework.com/doc/guide/2.0/en/runtime-routing#named-parameters
    'redirect/from/foo/<bar:{slug}>' => 'redirect/to/<bar>',

    // Path match (case-sensitive)
    [
        'from' => 'redirect/FROM/<year:\d{4}>/<month>',
        'to' => 'https://redirect.to/<year>/<month>',
        'caseSensitive' => true,
    ],

    // Custom match callback
    [
        'match' => function (\Psr\Http\Message\UriInterface $url): ?string {
            parse_str($url->getQuery(), $params);

            return isset($params['bar'])
                ? sprintf('redirect/to/%s', $params['bar'])
                : null;
        },
        'statusCode' => 301,
    ],
];
```


michtio:
@timkelty am I correct in assuming this only allows for `path matches` and not `url matches` e.g. in case of a multisite setup with different domain names?

timkelty:
> @timkelty am I correct in assuming this only allows for `path matches` and not `url matches` e.g. in case of a multisite setup with different domain names?

`match` gives you full control if want to match on anything else:

```
        'match' => function (\Psr\Http\Message\UriInterface $url): ?string {
           if ($url->getHost() === 'mysite.com') {
               return $url->withHost('en.mysite.com');
           }

           return null;
        },
```

michtio:
Thank you for claryfing @timkelty!

PR #16415, opened 2025-01-10
Field action menus

brandonkelly:
### Description

Adds action menus to custom fields within element edit pages, with “Edit field” (if admin changes are allowed) and “Copy field handle” items:

The action menu for a custom field labeled “Featured?”

(Both of these actions require that the current user is an admin, but the underlying support for field action menus is not admin-exclusive.)

~Since all fields will now have a “Copy field handle” action, the “Show field handles in edit forms” admin user preference has been removed.~

**Update:** The “Show field handles in edit forms” preference is [back](https://github.com/craftcms/cms/commit/5ac9c4a636763128fcbd8f204266d4e14b2056ea). When it’s enabled, field action menus won’t include a “Copy field handle” item.

Action menus are populated from an `actionMenuItems` array within the config passed to `craft\helpers\Cp::fieldHtml()`. Field layout elements can register items via `craft\fieldlayoutelements\BaseField::actionMenuItems()`. And custom field types can register additional action items via their `getActionMenuItems()` method.


mmikkel:
...I have some mixed feelings about this one 😅 

Love to see the "Edit field" action finally added to element edit forms. I think the field actions menu is a good idea, with a lot of potential use cases.

The "Copy Field Handle" action specifically, feels like a step backwards to me though. To simply _view_ a field's handle, devs now have to click twice to open a modal. Copying the handle requires a third click inside the modal, after moving the cursor to the center of the viewport. Compared to the inline field handles we've had for years, that feels kinda cumbersome.

If retaining the inline field handles is off the table, maybe it'd be possible to render the field handle inline in the field actions menu – without requiring additional clicks and that modal? I think that could hit the sweet spot in terms of surfacing the handles in an effective enough way, whilst still helping to streamline this stuff (i.e. "field actions").

brandonkelly:
@mmikkel A couple people internally had strong opinions on that as well! Just [brought the preference back](https://github.com/craftcms/cms/commit/5ac9c4a636763128fcbd8f204266d4e14b2056ea), and when it’s enabled, field action menus will stop showing a “Copy field handle” item.

mmikkel:
@brandonkelly Fantastic! Thanks Brandon.

PR #16453, opened 2025-01-19
Entry type relabeling

brandonkelly:
### Description

Adds the ability to override entry types’ names and handles on a per-section/field basis.

A Matrix field’s settings with two entry types selected. The first entry type is selected, and a slideout is open showing Name and Handle fields with overridden values.

The UI for it works identically to custom fields within field layout designers: double-click on an entry type (or press the “Settings” action menu item), and a slideout will open showing a chip with the original entry type details, plus “Name” and “Handle” fields. Setting those fields will only affect the entry type’s name/handle within the context of the section/field you’re editing.

Plus, the `global_entry_types` and `entrify_matrix_blocks` Craft 5 migrations have been updated to preserve sections’/Matrix fields’ original entry type handles, for any handles that need to be renamed for uniqueness. (They’ve also been updated to stop changing entry types’ names for uniqueness, since entry type names dropped the unique requirement in Craft 5.3.)

The `entry-types/merge` command has also been updated to preserve the outgoing entry type’s name and handle, for any usages that support the overrides.

### Related issues

- #14779
- #15924
- #16444


rob-c-baker:
This looks like great news - I'm hoping it means that for sites already upgraded, with integers appended to entry type handles all over the place, that can now be undone with the entry type merge command? I am guessing it depends on the detail of each case and whether the outgoing type is the one to keep?

brandonkelly:
@rob-c-baker You could already use the `entry-types/merge` command to reduce the number of entry types, where that makes sense.

PR #16484, opened 2025-01-22
Make label available on color field

riasvdv:
### Description

This adds the palette's label to the `ColorData` so it can be retrieved in twig using `{{ fieldName.label }}`


brandonkelly:
Doh, missed this! Already made the same change via #16492.

PR #16497, opened 2025-01-23
Add AnimationBlocker to prevent autoplay of GIFs and other potentially animated filetypes

gcamacho079:
### Description
- Pauses all GIFs added to the control panel by adding a canvas of the first frame over the image. Also does this for `.webp` images, which could be animated.
- Adds a play/pause button when the image size allows for comfortable placement

### Related issues
Resolves PT-25


gcamacho079:
@brandonkelly the user preference has been removed. 👍🏼

src/web/assets/animationblocker/AnimationBlockerAsset.php
```diff
@@ -0,0 +1,36 @@
...
+     * @inheritdoc
+     */
+    public $sourcePath = __DIR__ . '/dist';
+
+    /**
+     * @inheritdoc
+     */
+    public $depends = [
+        JqueryAsset::class,
+    ];
```
brianjhanson: Could we get rid of the jQuery dependency?

src/web/assets/animationblocker/src/AnimationBlocker.ts
```diff
@@ -0,0 +1,320 @@
...
+  private static minToggleableWidth: number = 100;
+  #imageAddedObserver = AnimationBlocker.createImageAddedObserver();
+  private static playIcon: string =
+    '<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 384 512"><!--! Font Awesome Pro 6.5.2 by @fontawesome - https://fontawesome.com License - https://fontawesome.com/license (Commercial License) Copyright 2024 Fonticons, Inc. --><path d="M73 39c-14.8-9.1-33.4-9.4-48.5-.9S0 62.6 0 80V432c0 17.4 9.4 33.4 24.5 41.9s33.7 8.1 48.5-.9L361 297c14.3-8.7 23-24.2 23-41s-8.7-32.2-23-41L73 39z"/></svg>';
+  private static pauseIcon: string =
+    '<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 320 512"><!--! Font Awesome Pro 6.5.2 by @fontawesome - https://fontawesome.com License - https://fontawesome.com/license (Commercial License) Copyright 2024 Fonticons, Inc. --><path d="M48 64C21.5 64 0 85.5 0 112V400c0 26.5 21.5 48 48 48H80c26.5 0 48-21.5 48-48V112c0-26.5-21.5-48-48-48H48zm192 0c-26.5 0-48 21.5-48 48V400c0 26.5 21.5 48 48 48h32c26.5 0 48-21.5 48-48V112c0-26.5-21.5-48-48-48H240z"/></svg>';
+
+  protected get imageAddedObserver() {
+    return this.#imageAddedObserver;
+  }
```
brianjhanson: Do we need this getter? It doesn't look like anything uses it, and if we want the `imageAddedObserver` to be available we could just make it public.

brianjhanson (commented):
@gcamacho079 Looking good! Just a few comments and I pushed up a few minor styling tweaks. 

The biggest thing I noticed is that the animated images will play the first time they're inserted into the dom after you upload it (here's a quick video https://share.cleanshot.com/qXWpkXNh).

I think that's because when those are inserted into the DOM, they get an action URL that doesn't have `.gif` or `.webp` in the `src` or `srcset`. I'm not sure what the best way around it is. The only thing I can think of would be to output some kind of `data-filename` on the `img` so we can use that to determine if we should stop the animation.

I suppose another option would be to refactor this into a web component like Shoelace has. Then we could use PHP to determine if an image is animated or not and either output a `<craft-animated-image/>` wrapper or not. I'm not sure how involved that change would be though. It could be larger than we want it to be.

src/web/assets/animationblocker/AnimationBlockerAsset.php
gcamacho079: Absolutely!

src/web/assets/animationblocker/src/AnimationBlocker.ts
gcamacho079: Good call, I may have had a purpose for this at one point, but I'll remove it for now. 👍🏼

gcamacho079:
> @gcamacho079 Looking good! Just a few comments and I pushed up a few minor styling tweaks.
> 
> The biggest thing I noticed is that the animated images will play the first time they're inserted into the dom after you upload it (here's a quick video https://share.cleanshot.com/qXWpkXNh).
> 
> I think that's because when those are inserted into the DOM, they get an action URL that doesn't have `.gif` or `.webp` in the `src` or `srcset`. I'm not sure what the best way around it is. The only thing I can think of would be to output some kind of `data-filename` on the `img` so we can use that to determine if we should stop the animation.
> 
> I suppose another option would be to refactor this into a web component like Shoelace has. Then we could use PHP to determine if an image is animated or not and either output a `<craft-animated-image/>` wrapper or not. I'm not sure how involved that change would be though. It could be larger than we want it to be.

@brianjhanson thank you for catching that, and for your style updates. 👍🏼 I'll poke around and see what route makes sense. We could also do a hybrid approach of wrapping the image in a web component during the AnimationBlocker initialization.

src/base/Element.php
```diff
@@ -4181,6 +4181,7 @@ public function getThumbHtml(int $size): ?string
                     'sizes' => sprintf('calc(%srem/16)', $size),
                     'srcset' => sprintf('%s %sw, %s %sw', $thumbUrl, $size, $this->thumbUrl($size * 2), $size * 2),
                     'alt' => $this->thumbAlt(),
+                    'craft-animated' => $this->getIsAnimated(),
```
brianjhanson: I don't see a big reason to namespace this since it's already a data attribute, but I'm not super passionate about it.

brianjhanson (approved):
Just one minor "take it or leave it" note. Otherwise, looks good!

src/base/Element.php
gcamacho079: I always get a little paranoid thinking someone else might be using that particular data attribute, but I can pull it out. 👍🏼

gcamacho079:
@brianjhanson all done! 👍🏼 Thanks for the review!

brandonkelly:
Nice work!

PR #16516, opened 2025-01-24
Start using Rector for QA

samsonasik:
| Q             | A
| ------------- | ---
| QA?  | ✔️

This PR apply Rector as part of QA check, start with:

- use php5.6 preset for starter to ease review.
- Only apply to `lib` and `src` only (for now)
- Using `--dry-run` to show the diff output if there is expected change by rector, to proces, we can run:

```bash
vendor/bin/rector --dry-run
```

This still using Rector 1.x since composer.json still require phpstan 1.x, Rector can be updated to v2 once phpstan also updated to v2.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,3 +24,13 @@ jobs:
     secrets:
       token: ${{ secrets.GITHUB_TOKEN }}
       slack_webhook_url: ${{ secrets.SLACK_CRAFT_WEBHOOK_URL }}
+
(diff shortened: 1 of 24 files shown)


samsonasik:
Ready for merge 👍

TomasVotruba (approved):

angrybrad:
I made a few changes in https://github.com/craftcms/cms/pull/16600 so I will close this one out. Thanks!

PR #16545, opened 2025-01-27
Craft 4 Backport: ensure the db connection is closed

joshuapease:
Backporting commit 4097887 from the 5.x branch.

Bug Ticket:
- https://github.com/craftcms/cms/issues/16546

### Description

I work on a Craft 4 site with a large test suite that often runs into these `Too Many Connections` errors. 

My workaround has been to split my tests up into smaller groups. But it looks like [this commit from the Craft 5 branch](https://github.com/craftcms/cms/commit/40978874f80929181de2986395e71974413f6802) resolves my issue. 

Is there any chance that we can backport this into Craft 4? It would really help me get our test suite and Codeception upgrade in order before switching to Craft 5 in the coming months. 

CleanShot 2025-01-27 at 11 03 05@2x

### DIFF
--- a/src/test/Craft.php
+++ b/src/test/Craft.php
@@ -164,6 +164,9 @@ class Craft extends Yii2
 
         parent::_before($test);
 
+        // transaction events are registered now, so it's ok to open the connection
+        \Craft::$app->db->open();
+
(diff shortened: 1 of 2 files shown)


brandonkelly:
Thanks!

Issue #16086, opened 2024-11-12
[5.x]: Applying project config changes will sometimes revert the config folder to an early state


MoritzLost:
### Description

I'm still seeing the issue described in #15012. I can't reliably reproduce it, but sometimes, running `up` or `project-config/apply` will _not_ apply the config, but will instead reset the config to an earlier state (essentially running `project-config/write`).

I've encountered this in two scenarios:

1. When switching between branches with different config states in a local dev environment (in `dev` mode with `allowAdminChanges` set to `true`).
2. When deploying to a live server (in `production` mode with `allowAdminChanges` set to `false`).

Not sure if those are the same issue, or two unrelated issues.

I don't know what causes this or how to reproduce it. Just now I saw this while deploying an update from Craft 4 to Craft 5. The environment doesn't allow admin changes, and the deployment script only applies the most recent state from GitHub, here's the relevant part:

```bash
# Flush all caches
$FORGE_PHP craft cache/flush-all --interactive=0
$FORGE_PHP craft clear-caches/all --interactive=0

# Update source code from repository
git pull --ff-only origin $FORGE_SITE_BRANCH

# Install Composer dependencies
$FORGE_COMPOSER install --no-dev --no-progress --no-interaction --prefer-dist --optimize-autoloader

# Apply project config and any pending migrations
$FORGE_PHP craft migrate/all --no-content --interactive=0
$FORGE_PHP craft project-config/apply
$FORGE_PHP craft migrate --track=content --interactive=0
```

Now I can see in the deployment log that the Composer installation ran successfully, and the first set of migrations (before the `project-config/apply` step) were also executed correctly. But after that, the `project-config/apply` command resulted in an error:

```
29 migrations were applied.

Migrated up successfully.
Your project config files were created for different versions of Craft and/or plugins than what’s currently installed.

Craft CMS is installed with schema version of 5.3.0.2 while 4.5.3.0 was expected.
Navigation is installed with schema version of 2.1.0 while 2.0.7 was expected.
Super Table is installed with schema version of 4.0.0 while 3.0.0 was expected.

Try running `composer install` from your terminal to resolve.
```

Checking the files on the server, I could see that the `config/` folder was reset to the state before the deployment started. After running `git reset --hard` and `git clean -df`, I restarted the deployment. This time, it finished successfully.

So something caused Craft to write out the existing config instead of applying it. Either as part of the `migrate/all --no-content` step, or at the beginning of the `project-config/apply` step.

### Steps to reproduce

[...]

PR #16574, opened 2025-01-29
check if schema version has changed against the old version

i-just:
### Description

**A rough outline of what’s going on:**

This will happen if `src/services/ProjectConfig->areChangesPending(true)` in the `src/console/controllers/MigrateController->runAction()` returns `true`, in which case we [“allow project config changes, but don’t overwrite the pending changes”](https://github.com/craftcms/cms/blob/5.6.2/src/console/controllers/MigrateController.php#L186-L192). When testing I had both `devMode` and `allowAdminChanges` set to `false`.

**In the above case:**
- the `migrate/all --no-content` goes through fine until `EVENT_AFTER_REQUEST` gets triggered; 
- that then makes the `ProjectConfig->flush()` get triggered (it can be enqueued by, e.g. `ProjectConfig->set()` which [calls `ProjectConfig->_saveConfigAfterRequest()`](https://github.com/craftcms/cms/blob/5.6.2/src/services/ProjectConfig.php#L529)). 
- `flush()` [calls `writeYamlFiles()`](https://github.com/craftcms/cms/blob/5.6.2/src/services/ProjectConfig.php#L420), which writes to files the [config we got from `getCurrentWorkingConfig()`](https://github.com/craftcms/cms/blob/5.6.2/src/services/ProjectConfig.php#L1621), which at the end of `migrate/all` can still have “old” `version` and `schemaVersion` values

Current working config can still have the “old” `schemaVersion` value because of [those lines](https://github.com/craftcms/cms/blob/5.6.2/src/services/Updates.php#L369-L371) (and [those](https://github.com/craftcms/cms/blob/5.6.2/src/services/Plugins.php#L862-L864) for the plugins).

(In terms of 4 to 5 upgrade, the first thing that calls `src/services/ProjectConfig->set()` is `craft/migrations/m230524_220029_global_entry_types->safeUp()`.
With 5.5.x to 5.6.x, there’s the `src/migrations/m241125_122914_add_viewUsers_permission` that also adjusts the project config.)

#### Steps to reproduce

(php 8.2; mysql 8.0)
1. set up a fresh install with just Craft CMS v5.5.10; set `devMode` and `allowAdminChanges` to `true`; ensure you're on a Pro version;
2. create a user group (`group1`) with permission to access the control panel and to edit users
3. take a backup of the database
4. change the cms requirement to 5.6.4 & run composer update
5. run `ddev php craft migrate/all --no-content` 
6. at this point, all works as expected
7. run `ddev php craft project-config/touch`
8. make a copy of the `config/project` folder
9. set `devMode` and `allowAdminChanges` to `false`
10. import database backup (one with version 5.5.10)
11. run `ddev php craft migrate/all --no-content`
12. after the command has finished, view the contents of `config/project/project.yaml` - the `system.schemaVersion` will be set to `5.5.0.0`, and it should be `5.6.0.2`

Now you can use steps 8-12 to keep reproducing this over and over

### Related issues
#16086


src/services/Updates.php
```diff
@@ -366,7 +367,7 @@ public function updateCraftVersionInfo(): bool
         // Only update the schema version if it's changed from what's in the file,
         // so we don't accidentally overwrite other pending changes
         $projectConfig = Craft::$app->getProjectConfig();
-        if ($projectConfig->get(ProjectConfig::PATH_SCHEMA_VERSION, true) !== $info->schemaVersion) {
```
brandonkelly, line 369: This prevents `system.schemaVersion` from getting updated when you’re updating for the first time (e.g. from a dev environment), since the value will match what’s in the `info` table at that point.

i-just: Yes, you’re right. I must have got confused when testing. I do wonder if that check is needed, though. The `ProjectConfig->_setInternal()` method will check if the value has changed anyway, so dropping that check would mean that the value would only get updated when needed anyway (first update and when deploying further).

Those checks were added [a while ago](https://github.com/craftcms/cms/commit/68b05761b439b0bc42bf4ebc3572ceb8cbdfc2ae), and it looks like back then we weren’t checking if the value has changed before setting it.

brandonkelly: Yeah, maybe not needed as of af448b70336.

Still, not sure I’m following how #16086 could occur because of these checks. Were you able to reliably reproduce it? If so can you update the PR with steps to reproduce?

i-just: Sure! I updated the description with the steps to reproduce. 

I also pushed the change to let the pc `_setInternal()` figure out if the schema version value has changed.

brandonkelly:
Thanks! I was able to resolve the main issue just by preventing `migrate` commands from writing out their project config changes when there are pending YAML changes. (653c9c17aae4eb91e7309dca4d8c07330d7cbeec)

The changes in this PR are probably worth doing as well, though, just to simplify the code.

PR #16575, opened 2025-01-29
Fixed custom option values not showing up in element index table

lukeholder:
### Description

Custom values were not showing up in element table preview:

Before

CleanShot 2025-01-29 at 21 38 08@2x

After

CleanShot 2025-01-29 at 21 31 23@2x

(the field called 'mydropdown' is a checkbox input 😅)

### Related issues

https://github.com/craftcms/commerce/issues/3853


brandonkelly:
Thanks Luke!

Issue #16577, opened 2025-01-29
[4.x]: 4.14.2 fails tests on Codeception 5 / Yii2 module 1.1.6+


qrazi:
### Description
Updated our project from Craft CMS 4.14.1 to 4.14.2. Codeception tests fail immediately with a 255 exit code. Error to be found in `./storage/logs/phperrors.log`.

This is caused by using `codeception/module-yii2` >=1.1.6. See the change: https://github.com/Codeception/module-yii2/compare/1.1.5...1.1.6.

I'll create a PR as well with a suggested solution to the code.

Error:
```
PHP Fatal error:  Declaration of craft\test\CraftConnector::startApp(): void must be compatible with Codeception\Lib\Connector\Yii2::startApp(?yii\log\Logger $logger = null): void in /app/vendor/craftcms/cms/src/test/CraftConnector.php on line 135
```

### Steps to reproduce

1. Install Craft CMS 4.14.2
2. Add testing by following documentation: https://craftcms.com/docs/4.x/testing/codeception.html
3. Also add `codeception/module-yii2` as this is required by Craft CMS test tooling

### Expected behavior
Example test runs successfully.

### Actual behavior
Example test fails with exit code 255 and no further error on CLI, error in phperrors.log

### Craft CMS version

4.14.2

### PHP version

8.2.6

### Operating system and version

Docker container running Alpine Linux 3.18.0

### Database type and version

Docker container MySQL 8.0.29

### Image driver and version

GD 8.2.6

### Installed plugins and versions

- Amazon S3 	2.2.1
- Blitz 	4.23.10
- Breadcrumb 	2.0.0
- Calendar 	5.0.13
- CKEditor 	3.11.0
- Compress 	4.0.1
- Craftagram 	3.0.0
- Design Tokens 	1.1.1
- Doxter 	5.0.6
- Element API 	4.2.0
- Elements Panel 	2.0.0
- Embedded Assets 	4.2.1
- Feed Me 	5.9.0
- Formie 	2.1.40
- Hide Admin 	2.0.0-beta.1
- HTTP Basic Authentication 	dev-patch_2_fix_CVE-2024-51736
- Image Base64 	1.1.1
- Imager 	dev-dev-craft4
- Linkit 	4.0.4.1
- Maps 	4.0.8
- Mix 	v1.7.2
- Navigation 	2.1.1
- Neo 	4.2.24
- Password Policy 	4.1.0
- PDF Generator 	3.0.0
- Retour 	4.1.22
[...]

PR #16580, opened 2025-01-29
refactor: make test integration compatible with newer versions of `codeception/module-yii2`

qrazi:
### Description
Only extend method signature. Pass-through all method arguments to parent. This satisfies both version ranges of `codeception/module-yii2` <= 1.1.5 && >= 1.1.6. 

Using Codeception 5 requires newer Yii2 module. Testing documentation does not restrain Codeception version to 4.x. So making the code work like this seems pragmatic.

### Related issues
https://github.com/craftcms/cms/issues/16577

### DIFF
--- a/src/test/CraftConnector.php
+++ b/src/test/CraftConnector.php
@@ -130,11 +130,20 @@ class CraftConnector extends Yii2
      * We'll open the connection after all of the transaction listeners are
      * registered.
      *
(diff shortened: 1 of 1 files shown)


brandonkelly:
Thanks!

brandonkelly:
Craft 4.14.3 is out now with that change.

PR #16587, opened 2025-01-29
Resets focus target on modal/slideout close when trigger was in a disclosure menu

gcamacho079:
### Description
Also fixes a bug where the slideout container was still in the DOM after close


gcamacho079:
@brandonkelly one thing outstanding: when you hit "Save" in the `AssetImageEditor`, focus returns to the disclosure trigger but is dropped shortly after, around when the image re-renders. I'm having a hard time finding the source of the issue - do you have any ideas?

brandonkelly:
@gcamacho079 You were just missing this bit 🙃  https://github.com/craftcms/cms/pull/16587/commits/ec415c3be4f9d9c731c60da8a9365970691f67d8

Issue #16588, opened 2025-01-30
[5.x]: Custom Date Fields not respecting site timezone in CSV exports


watarutmnh:
### Description

When exporting entries to CSV, custom Date fields are exported in UTC format, while system fields (dateCreated, dateUpdated) are correctly exported in the site's configured timezone (JST/UTC+9).

### Steps to reproduce

1. Configure Craft CMS site timezone to 'Asia/Tokyo' in config/general.php
2. Create a custom Date field (with "Show Time Field" disabled)
3. Create an entry and set a date value
4. Export entries to CSV using the Expanded format

### Expected behavior

Custom Date fields should be exported in the site's configured timezone (JST/UTC+9), matching the behavior of system fields.

### Actual behavior

- System fields (dateCreated, dateUpdated): Correctly exported in JST (UTC+9)
Example: 2025-01-22T15:08:50+09:00
- Custom Date fields: Exported in UTC
Example: {"date":"1974-03-16 15:00:00"}

### Additional Notes
- This issue was observed after updating to 5.6.1, which fixed timezone handling for system fields in CSV exports (PR #16447)
- The Control Panel correctly displays all dates in JST
- Database correctly stores dates in UTC

### Craft CMS version

Craft CMS version: 5.6.1

### PHP version

PHP version: 8.4

### Database type and version

Database: MySQL 8.0.40

### Installed plugins and versions

-

i-just:
Hi, thanks for reporting! I raised a PR for this. It targets 4.15 and 5.7, as that’s a slight change in behaviour.

brandonkelly:
Craft 4.15.0 and 5.7.0 are out with that change.

PR #16629, opened 2025-02-04
Expanded element export & custom date fields

i-just:
### Description
When running a raw export, all dates (attributes and fields) should be exported in UTC, just like they’re stored in the DB (that’s already the case).

When running an expanded export, all dates (attributes and fields) should be exported in the system’s timezone (in the ISO8601 format, e.g. 2025-02-04T08:58:45+01:00), or, if the date field has “show timezone” turned on, the date should be exported in the selected timezone.
The attributes are already exported in this way. This PR ensures this also happens for the custom fields.

### Related issues
#16588


timeverts:
@brandonkelly, this change was fairly significant and 'broke' quite a few of our Craft sites that were using custom date field values. Should this change not have been part of a major release instead (e.g. 6.x)?

brandonkelly:
@timeverts How exactly did it break things?

Issue #16640, opened 2025-02-06
[5.x]: Matrix Blocks are not saved to new languages of an entry


kamejosh:
### Description

I maintain a website based on craft which relies heavily on nested Matrix Blocks (it's Craft 5 latest). And has multiple Languages/Sites. This works fine when I create a new Entry, I add a new Block and this block is now available on all different sites. 
But when I add a new Language/Site and I also want to have all entries available for this site this does no longer work.
First I go in to all sections and make the new site available there, next I go to an entry and add the entry for the new site.

Now when I save the Entry is available for the new Site but it doesn't have all the block elements. This would be manageable if not for another issue: 
When I now create the block elements on the new Site of the Entry, the same blocks are also created in all other languages, which is not what I want, I want all languages/sites to follow the same structure, like when I setup a new entry.

The only workaround once this has happened, is to deleted every block and recreate everything. Which is not great.

There is another way to prevent this from happening but it's not very intuitive, once I add the new site and before I hit save for the first time, I need to mark every block element as changed (go into the deepest block and add a space or flip a switch) when now everything is marked as changed (blue border on the left) the new site also gets all the content elements. But this is not the way I think this is supposed to be done.

my sections have "Let each entry choose which sites it should be saved to"

![Image](https://github.com/user-attachments/assets/1d23979b-c15b-4518-90dd-f3dacd79773b)

my matrix blocks have "Save entries to all sites the owner element is saved in"

![Image](https://github.com/user-attachments/assets/f0f65c62-2ae2-4125-9d78-8636fa71daba)

my elements have "Translate for each site" where applicable

![Image](https://github.com/user-attachments/assets/5ccab203-eb60-42a4-a5e1-331c045db128)

### Steps to reproduce

1. Have a multi language site

![Image](https://github.com/user-attachments/assets/f0dc138d-51ff-4497-a1ac-f1fa1d69546d)

2. create an entry with nested matrix blocks in a single language
3. add a second language to the entry
4. save the entry

### Expected behavior

The nested matrix blocks are saved to the new language

### Actual behavior

The nested matrix blocks are sometime saved to the new language

new language entry: 

![Image](https://github.com/user-attachments/assets/e39c825e-6d88-4e55-9644-ed38a803d61e)

existing language entry: 

![Image](https://github.com/user-attachments/assets/24ed8c8a-0471-4775-84a6-c867f850bf86)

### Craft CMS version

Craft Pro 5.5.6.1

### PHP version

 8.3.16

[...]

PR #16649, opened 2025-02-07
Bugfix/16640 drafts and propagation

i-just:
### Description
1. Don’t set the initial delta value for the Link field with an empty value, as that will cause the field to get marked as modified even if it wasn’t (e.g. if you change the value of a different field)
2. If you edit an entry nested in a matrix field and then propagate the parent element to another site, only the edited nested entry was propagated to the new site (in the blocks and cards view modes).

**Steps to reproduce:**
on a clean installation:
- set up at least 2 sites (`siteA`, `siteB`)
- create a Plain Text field (`plainText`)
- create a `myEntryType` entry type; it should have the default title field and the `plainText`
- create `contentBuilder` Matrix field, entry type: `myEntryType`; view mode: inline-editable blocks; propagation: all
- create a section (e.g. `blog`) with an entry type (e.g. `article`) containing the default title field and the `contentBuilder` field; enable that section for both sites; propagation method: let each entry choose
- create an entry in the `blog` section for `siteA`; add two `myEntryType` “blocks”, populate the title and plain text for both; fully save the entry;
- edit the entry again, edit `plainText` for one of the “blocks”, then, under “status”, enable it for `siteB` and fully save
- go to `siteB` and observe that only the edited “block” was propagated

Note: this was also happening if you turned off autosaving drafts

The issue was further exacerbated by an empty Link field as those were getting marked as modified even if they were not.

It’s also reproducible with a matrix field in cards view mode.

### Related issues
#16640


brandonkelly:
Nice job!

I was curious why this wasn’t a bug in Craft 4.

Turns out it was, just harder to reproduce: with the same setup, you have to modify a Matrix block’s nested field *and* add a new site before the draft gets autosaved. The easiest way to reproduce is using JavaScript. With the Status field expanded to show the site select input, and a Matrix block’s nested Plain Text field selected in the DOM inspector, you can run the following JS code:

```js
$($0).val('edit');
$('[aria-labelledby="add-site-label"]').val('2').trigger('change');
```

And then the same buggy code path will occur, and only the modified Matrix block will get propagated to the new site.

So I cherry-picked the Matrix fix into 4.x (3c692b45f4384cbb43764233d6d24bf4bcfb708a).

Then I looked into why the Matrix field’s delta data was getting submitted on both autosave requests in Craft 5, but only the first autosave request in Craft 4 (when editing the entry at a normal human speed).

That turned out to be due to a0e98b3c56c4ea3e1232b1f2ac49dddb8a186ca5: when we’re passing `this.$container.data('modified-delta-names')` to `Craft.findModifiedDeltaNames()`, the array is passed by reference so  Matrix sub-field’s delta name ends up getting appended to `this.$container.data('modified-delta-names')` and then gets tacked on to each subsequent save request. So the modified delta names get added onto each time you edit a new field, rather than each autosave only posting the fields that have changed since the last autosave, like it’s supposed to work.

That core bug sortof exists in Craft 4 as well, although it’s inconsequential there. So I fixed on 4.x as well (fbe700fd3e7c5c56eec115e0bee3bf6f1eb0a5a0).

That leaves just the Link field bug. I’m not sure why we were calling `setInitialDeltaValue()` for that field, and not seeing any negative side effects from removing it, so went ahead and merged that change too.

PR #16650, opened 2025-02-07
Remove fallback rand() with try catch on StringHelper

samsonasik:
### Description

Since requirement is php 8.2, fallback rand() seems not needed. see https://3v4l.org/siVcW


samsonasik:
Ready for merge 👍

brandonkelly:
Not clear why that was added in the first place in 1480c5fa952d02dd05a758ac5814651179b6ef54.

PR #16656, opened 2025-02-07
Import map improvements

brianjhanson:
Our [initial implementation](https://github.com/craftcms/cms/pull/16414) doesn't work so well with slideouts attempting to us the feature. In that scenario, when the slideout is loaded, it's corresponding `importmap` will be returned as a part of `headHtml` and injected into the page. Unfortunately, that map won't be parsed and any imports referencing it will fail.

Instead, we need something a bit more global. This PR is a first draft at that. We now have a `View::getImportMap()` method that will collect all imports registered via a `View::EVENT_REGISTER_JS_IMPORT_MAP` event and return the array.

## Open questions
- Can we make this process any nicer?
- Should this method be on the `View` or would `AssetManager` be more appropriate?
- Should we include a twig function for outputting the import map?

### DIFF
--- a/src/events/RegisterJsImportEvent.php
+++ b/src/events/RegisterJsImportEvent.php
@@ -1,0 +1,13 @@
+<?php
+
+namespace craft\events;
+
+use craft\base\Event;
+
+class RegisterJsImportEvent extends Event
+{
+    /**
+     * @var array import entries to be added to the import map.
+     */
+    public array $imports = [];
+}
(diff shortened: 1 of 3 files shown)


brianjhanson:
Turns out we don't need this, plugin authors will just need to call `registerJsImport` early in the request (like `registerTwigExtension`). 

```php
if (Craft::$app->getRequest()->getIsCpRequest()) {
    $assetManager = Craft::$app->getAssetManager();
    $bundle = $assetManager->getBundle(MyAssetBundle::class);

    $view->registerJsImport('@namespace/package', $assetManager->getAssetUrl($bundle, 'index.js'));
}
```

PR #16676, opened 2025-02-11
[4.x][5.x] Fixed condition rules losing their condition when being created

nfourtythree:
### Description
Tracking through the creation of a condition rule. The condition is set in the config when [creating a rule on a condition](https://github.com/craftcms/cms/blob/5.x/src/base/conditions/BaseCondition.php#L110) (mostly applicable in `5.x`) however the condition [can be lost when creating the new config for the rule](https://github.com/craftcms/cms/blob/4.x/src/services/Conditions.php#L90-L115).

This means when the new rule class is created it does not have the condition available.

This PR is applicable for both `4.x` and `5.x`.

### Related issues
craftcms/commerce#3890


brandonkelly:
@nfourtythree Thanks! Craft 4.14.5 and 5.6.6 are out with this fix.

PR #16680, opened 2025-02-11
Enable Rector php7.1 set

samsonasik:
### Description

Applied rule: `ListToArrayDestructRector` from php 7.1 set to change `list()` to array destruct


samsonasik:
Ready to merge 👍

brandonkelly:
Thanks!

PR #16685, opened 2025-02-11
Enable Rector php7.2 set

samsonasik:
### Description

Applied rule: `GetClassOnNullRector` from php 7.2 set to change get_class() with null argument to use self::class.

### DIFF
--- a/rector.php
+++ b/rector.php
@@ -19,4 +19,4 @@ return RectorConfig::configure()
         // autoload may need to be bootstrapped to early load some child classes
         RemoveExtraParametersRector::class,
     ])
-    ->withPhpSets(php71: true);
+    ->withPhpSets(php72: true);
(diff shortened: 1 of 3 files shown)


samsonasik:
Ready to merge 👍

brandonkelly:
Neither of those changes make sense; we’d never want to fall back on `self::class` in those cases. And the values will never be `null` in those cases to begin with.

samsonasik:
in that case, docblock `@var` seems needed to enforce type check, or use assert(), I will update

samsonasik:
I updated to use `@var` to verify that it never null https://github.com/craftcms/cms/pull/16685/commits/2ec37c5f15b9336d742ea1d47dcdbaad6cddc95f

samsonasik:
Ready to merge 👍

brandonkelly:
Thanks!

PR #16687, opened 2025-02-12
Enable Rector php7.3 set

samsonasik:
### Description

- [x] Ensure array docblock correctly set on usage on string functions when it always string data
- [x] replace is_array() and instanceof Countable to `is_countable()` as function already exists.
- [x] Ensure cast to string for mixed typed needle on strpos


samsonasik:
Ready to merge 👍

PR #16704, opened 2025-02-14
Enable Rector php 7.4 set

samsonasik:
### Description

enable php 7.4 set, applied changes:

- change closure to arrow function
- add default null on nullable properties
- Use null coalescing operator ??=


samsonasik:
Ready to merge 👍

PR #16711, opened 2025-02-16
Enable Rector php 8.0 set

samsonasik:
### Description

Applied changes:

- use str_contains/str_starts_with/str_ends_with
- use static::class over get_class($this)
- remove unused variable cactch on try catch
- use $object::class over get_class($object)
- apply constructor promotion
- make implements \Stringable with __toString() method class.

### DIFF
--- a/src/auth/sso/BaseExternalProvider.php
+++ b/src/auth/sso/BaseExternalProvider.php
@@ -334,7 +334,6 @@ abstract class BaseExternalProvider extends BaseProvider
     /**
      * Normalize a callback
      *
-     * @param mixed $callback
      * @param string|null $defaultClass
      * @return callable|null
      * @throws \yii\base\InvalidConfigException
--- a/src/auth/sso/BaseExternalProvider.php
+++ b/src/auth/sso/BaseExternalProvider.php
@@ -334,6 +334,7 @@ abstract class BaseExternalProvider extends BaseProvider
     /**
      * Normalize a callback
      *
+     * @param mixed $callback
      * @param string|null $defaultClass
      * @return callable|null
      * @throws \yii\base\InvalidConfigException
--- a/src/base/Field.php
+++ b/src/base/Field.php
@@ -43,7 +43,7 @@ use yii\db\Schema;
  * @author Pixel & Tonic, Inc. <support@pixelandtonic.com>
  * @since 3.0.0
  */
-abstract class Field extends SavableComponent implements FieldInterface, Iconic, Actionable
+abstract class Field extends SavableComponent implements FieldInterface, Iconic, Actionable, \Stringable
 {
     use FieldTrait;
 
--- a/src/base/Field.php
+++ b/src/base/Field.php
@@ -30,6 +30,7 @@ use craft\validators\UniqueValidator;
 use DateTime;
 use Exception;
 use GraphQL\Type\Definition\Type;
+use Stringable;
 use yii\base\Arrayable;
 use yii\base\ErrorHandler;
 use yii\base\InvalidArgumentException;
@@ -43,7 +44,7 @@ use yii\db\Schema;
  * @author Pixel & Tonic, Inc. <support@pixelandtonic.com>
  * @since 3.0.0
  */
-abstract class Field extends SavableComponent implements FieldInterface, Iconic, Actionable, \Stringable
+abstract class Field extends SavableComponent implements FieldInterface, Iconic, Actionable, Stringable
 {
     use FieldTrait;
 
--- a/src/base/MemoizableArray.php
+++ b/src/base/MemoizableArray.php
@@ -173,6 +173,7 @@ class MemoizableArray implements IteratorAggregate, Countable
      *
      * @param string $method
      * @param string $key
+     * @param mixed $value
      * @param bool $strict
      * @return string
      */
--- a/src/base/MemoizableArray.php
+++ b/src/base/MemoizableArray.php
@@ -35,11 +35,6 @@ use IteratorAggregate;
 class MemoizableArray implements IteratorAggregate, Countable
 {
     /**
-     * @var array Array elements
-     */
-    private array $_elements;
-
-    /**
      * @var callable|null Normalizer method
      */
     private $_normalizer;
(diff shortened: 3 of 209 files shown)


samsonasik:
Ready to merge 👍

src/auth/sso/BaseExternalProvider.php
```diff
@@ -334,7 +334,6 @@ protected function populateUser(User $user, array $data): User
     /**
      * Normalize a callback
      *
-     * @param mixed $callback
```
brandonkelly: This seems wrong? (And other changes like it)

samsonasik: it already typed mixed, so `@param` is useless, except it has description

samsonasik: @brandonkelly do you want me to skip these change? If yes, I can skip from config, eg: skip `MixedTypeRector`

brandonkelly: If it’s only removing `mixed` param docs, then yes. If we’re going to remove `@param` docs that don’t add any value over what’s in the code, we should be doing it for all types, not just `mixed`.

samsonasik: @brandonkelly ok, I rolled back mixed type change https://github.com/craftcms/cms/pull/16711/commits/4f79c7a0f5efe7f68ab4fe8803eb11e7d7365f88 👍

samsonasik:
@brandonkelly  Ready to merge 👍

src/base/MemoizableArray.php
```diff
@@ -57,13 +52,12 @@ class MemoizableArray implements IteratorAggregate, Countable
     /**
      * Constructor
      *
-     * @param array $elements The items to be memoized
+     * @param array $_elements The items to be memoized
```
brandonkelly: This is a breaking change. Someone might be specifying the argument by name.

```php
new MemoizableArray(elements: [
    // ...
]);
```

samsonasik: @brandonkelly sure, I updated to skip when property vs param name are different on next commits

https://github.com/craftcms/cms/pull/16711/commits/de6fed397636a70821b072a994c20b7c8a25b6a6

and

https://github.com/craftcms/cms/pull/16711/commits/cadbf1b315548bd9f97b34c8a4acf91282d00a95

 👍

samsonasik:
@brandonkelly Ready to merge 👍

src/base/Field.php
```diff
@@ -43,7 +43,7 @@
  * @author Pixel & Tonic, Inc. <support@pixelandtonic.com>
  * @since 3.0.0
  */
-abstract class Field extends SavableComponent implements FieldInterface, Iconic, Actionable
+abstract class Field extends SavableComponent implements FieldInterface, Iconic, Actionable, \Stringable
```
brandonkelly: What is the point of adding this? Just to be explicit? Seems unnecessary. At the very least, it should be imported so we don’t get that ugly `\`.

samsonasik: so it can exactly enforce string return on `__toString()`, see this change for example https://github.com/craftcms/cms/pull/16711/files#diff-5072cd57908a25217bfd7147397671cf40c6cf5c81d522e2061ca27a8ed49d67

sure, i can import that on next commit :)

samsonasik: @brandonkelly I imported Stringable to use statement https://github.com/craftcms/cms/pull/16711/commits/0ef22bea0d202a8ee6ec97d63443155669314c24 👍

samsonasik:
@brandonkelly Ready to merge 👍

src/base/NestedElementTrait.php
```diff
@@ -371,16 +371,11 @@ public function addInvalidNestedElementIds(array $ids): void
...
-                $this->setPrimaryOwner(reset($elements) ?: null);
-                break;
-            default:
-                parent::setEagerLoadedElements($handle, $elements, $plan);
-        }
+        match ($plan->handle) {
+            'owner' => $this->setOwner(reset($elements) ?: null),
+            'primaryOwner' => $this->setPrimaryOwner(reset($elements) ?: null),
+            default => parent::setEagerLoadedElements($handle, $elements, $plan),
+        };
```
brandonkelly: I’m surprised this is even allowed, since `match()` will return the matching result, but `setOwner()`, `setPrimaryOwner()`, and `setEagerLoadedElements()` each have a `void` return type. (So you couldn’t do `$foo = $this->setOwner(…)` without getting a PHP error, for example.)

Even if it is allowed, I don’t think we should be using `match()` anywhere where the return value isn’t used.

samsonasik: that's allowed, but sure, I skipped only void usage 👍 https://github.com/craftcms/cms/pull/16711/commits/045613fef983f51c1158cb0dbd232c6fd93bc8d5

samsonasik:
@brandonkelly Ready to merge 👍

src/cache/ElementQueryTagDependency.php
```diff
@@ -18,20 +18,14 @@
...
-    public ?ElementQuery $elementQuery = null;
-
     /**
      * Constructor
      *
      * @param ElementQuery $elementQuery
      * @param array $config
      */
-    public function __construct(ElementQuery $elementQuery, array $config = [])
+    public function __construct(public ?ElementQuery $elementQuery, array $config = [])
```
brandonkelly: This changes the signature. `null` was previously not allowed to be passed to `$elementQuery`.

samsonasik: The promotion property is  based on property type, so you choose: to have nullable or non-nullable property?

brandonkelly: We should not be making any type changes as part of this PR, either to properties or method arguments.

samsonasik: sure, skipped property promotion for this file https://github.com/craftcms/cms/pull/16711/commits/2ce4511b49c8611ddbac10ea6180bdafedbbca9e 👍

samsonasik:
@brandonkelly Ready to merge 👍

samsonasik:
I am closing for now as it has too many conflict right now, and I don't have time to do it for now.

PR #16724, opened 2025-02-19
fix Getting unknown property: craft\console\Request::actionSegments from queue jobs

jodeba:
## Description
actionSegments are only available when Craft::$app->getRequest()->isSiteRequest

So this fixes an error that occurs when:
- code is executed in a queue job
- a draft Entry was created from the drafts service
- serialized array is passed as a value to the matrix field
- you want to save the element (draft)

For example from this line:
https://github.com/digitalpulsebe/craft-multi-translator/blob/faa15ae8cff04d0bef92e07d0194c233ea0096d6/src/services/TranslateService.php#L81

## Related issues
https://github.com/digitalpulsebe/craft-multi-translator/issues/40

### DIFF
--- a/src/fields/Matrix.php
+++ b/src/fields/Matrix.php
@@ -1541,7 +1541,7 @@ JS;
                     $entry->getPrimaryOwnerId() === $element->getCanonicalId() &&
                     // this is so that extra drafts don't get created for matrix in matrix scenario
                     // where both are set to inline-editable blocks view mode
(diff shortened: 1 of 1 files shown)


brandonkelly:
Thank you!

brandonkelly:
Craft 5.6.8 is out with this fix. Thanks again.

jodeba:
Thank you for the quick response!

Issue #16753, opened 2025-02-21
[5.x]: Multibyte characters do not survive `db/backup` unless db charset is explicitly set to `utf8mb4` in config


mmikkel:
### Description

When backing up the database using the `db/backup` CLI command or Craft's "Database Backup" utility, multibyte characters (such as emojis) used in native fields (title fields, alternative text) or PlainText custom fields, will be exported as a `'?'` (question mark) _unless `CRAFT_DB_CHARSET` is explicitly set to `utf8mb4`_.

Note: I have not set/configured the `backupCommand` config setting.

### Steps to reproduce

1. Make sure the database is using `utf8mb4` charset and `utf8mb4_0900_ai_ci` collation (i.e. Craft 5's defaults).
2. Also make sure that the charset **isn't** explicitly set via the `CRAFT_DB_CHARSET` env var or in a `config/db.php` file
3. Create an entry, and add a fire emoji (🔥) in its title.
4. Backup the database using the `db/backup` CLI command, or the "Database Backup" utility
5. Re-import the database dump from the `db/backup` command
6. Confirm that the 🔥 in the entry's title has been replaced by a question mark `?`

Repeat the process, but this time make sure there is a `CRAFT_DB_CHARSET=utf8mb4` environment variable, before backing up the database. This time, the fire emoji will be intact.

### Expected behavior

Multibyte characters in native and plaintext fields should be preserved when Craft dumps a database using the `utf8mb4` charset and `utf8mb4_0900_ai_ci` collation.

### Actual behavior

Multibyte characters in native and plaintext fields are exported as `?` unless `CRAFT_DB_CHARSET` is set to `utf8mb4`.

### Craft CMS version

5.6.9.1

### PHP version

8.2.27

### Operating system and version

DDEV v1.24.2

### Database type and version

MySQL 8.0.40

### Installed plugins and versions

None

i-just:
Hi, thanks for reporting! I raised a PR for this.

brandonkelly:
Craft 5.6.10 is out now with that fix. Thanks again!

PR #16762, opened 2025-02-24
default backup command and default charset

i-just:
### Description
When getting the default backup command, use `getCharset()` to get the `default-character-set` value so that for MySQL `utf8` is normalised to `utf8mb4`.

### Related issues
#16753


src/db/mysql/Schema.php
```diff
@@ -204,7 +204,7 @@ public function getDefaultBackupCommand(?array $ignoreTables = null): string
             ->addArg('--dump-date')
             ->addArg('--no-autocommit')
             ->addArg('--routines')
-            ->addArg('--default-character-set=', Craft::$app->getConfig()->getDb()->charset)
```
timkelty, line 207: Seems like ideally we'd want to make `\craft\config\DbConfig::$charset` protected and add `setCharset` so this code would work either way?

The only reason why the current method is `\craft\config\DbConfig::charset` and not `\craft\config\DbConfig::setCharset` is because of fluent config conventions. So, if we add it, one can just reference the other.

i-just: Ok, but what about the environment overrides? Currently, they rely [on the properties being public](https://github.com/craftcms/cms/blob/5.x/src/helpers/App.php#L156). Or am I missing a trick here?

timkelty: ah shit that's annoying.

PR #16768, opened 2025-02-24
Stop scrolling body when element selector is open

### LINKED ISSUE #16759: [5.x]: Body behind scrollable modals should not scroll
### Description
When I scroll within a modal containing a list of entries, the body behind scrolls which can cause some nasty UX.

https://github.com/user-attachments/assets/76383299-740a-463f-ace6-e730e0b8142b

### Craft CMS version

5.6.9.1

### PHP version

_No response_

_No response_

### Database type and version

_No response_

### Image driver and version

_No response_

-

brianjhanson:
Mimics the no-scroll behavior of [slideouts](https://github.com/craftcms/cms/blob/d2366a4fca28951fe753ec8293d4d38b97a7f22b/src/web/assets/cp/src/js/CpScreenSlideout.js#L444) for element selector modals

### Related issues
Fixes #16759


brandonkelly:
Does the issue only affect element selector modals? I assume the changes should be made on `Garnish.Modal` and `Craft.Slideout`.

brianjhanson:
Yeah, I think you're right. I was nervous about changing it too broadly, but I don't see anywhere we're using `Garnish.Modal` where we _don't_ want to lock the body. 

As for Slideout, that [already has code](https://github.com/craftcms/cms/blob/79e353914d097788ae587a38f62086d085daf798/src/web/assets/cp/src/js/Slideout.js#L243-L247) to manage the no-scroll.

PR #16770, opened 2025-02-24
Fix action btn missing focus outline

gcamacho079:
### Description
When the user preference for “Always show focus outlines” is toggled on, the actions button doesn’t receive a focus outline

(Thanks @i-just for raising this issue 🙏🏼 )


brandonkelly:
Thanks!!

PR #16783, opened 2025-02-27
Maintain field settings when switching field types

brandonkelly:
When switching field types, any field settings which are defined by the same base class are now preserved.

Applies to options fields (Checkboxes, Dropdown, Multi-select, and Radio Buttons) and relational fields.


alexanderbuergin:
❤️

PR #16796, opened 2025-02-27
Don't limit element cache by config cacheDuration

timkelty:
### Description
Limiting the element cache duration by `\craft\config\GeneralConfig::$cacheDuration` (defaults to a 1 day) can result in element caches being set with unexpectedly short expiry.

It seems better and more flexible to just let it `stopCollectingCacheInfo` return a `null` duration, and leave it to the caller to limit the value if desired.


mmikkel:
The way I'm reading this change, template caches created with `{% cache %}` and containing element queries will no longer default to expire as per the `cacheDuration` config setting?

If that's the case, this feels like a pretty major behavioural change IMO.

timkelty:
> The way I'm reading this change, template caches created with `{% cache %}` and containing element queries will no longer default to expire as per the `cacheDuration` config setting?
> 
> If that's the case, this feels like a pretty major behavioural change IMO.

@mmikkel In that case, the template cache [would be set with a `null` duration,](https://github.com/craftcms/cms/blob/20f8452151ea351525d601ddf3a59eaf2ecb1336/src/services/TemplateCaches.php#L260-L260), which would still cause it to [default to `\craft\config\GeneralConfig::$cacheDuration`](https://github.com/craftcms/cms/blob/20f8452151ea351525d601ddf3a59eaf2ecb1336/src/helpers/App.php#L867-L867).

mmikkel:
Aha! Good to know, thanks @timkelty – and pardon the noise 🙂

timkelty:
Decided to target this to `6`, as it technically is a breaking change in this scenario:

An element has an expiry _longer_ the configured `$generalConfig->cacheDuration`

- The current (odd) behavior is to _limit_ that to `$generalConfig->cacheDuration`.
- The breaking change will just allow the duration to be set to what the element dictates.
- In either case, default template cache is not affected because the cache component has a configured default of `$generalConfig->cacheDuration`.

timkelty:
> Aha! Good to know, thanks @timkelty – and pardon the noise 🙂

No sweat – it made us dig deeper and realize it [technically is a behavior change still](https://github.com/craftcms/cms/pull/16796#issuecomment-2691095523), so will wait for 6.

PR #16805, opened 2025-03-01
Field editability conditions

brandonkelly:
### Description

Groups the existing field layout element conditions into a fieldset labelled “Visibility Conditions”, and adds a new “Editability Conditions” fieldset below it for custom fields, with “Current User Condition”.

The two condition fieldsets

When a custom field has editability conditions, they will be matched against the current user. If the user isn’t a match, the field will be displayed statically (read-only).

### Related issues

- #15903
- #16780


AugustMiller:
Safe to assume that the "visibility" of a field is still a prerequisite for editability, in the underlying controller? Say I were to POST directly to `elements/save`—historically, only fields that are “visible” are actually populated with request data.

(It looks to me like fields must be visible _and_ editable, now, to be picked up, due to the implicit `showInForm()` check against user conditions—they're just applied at different stages.)

I think our current documentation on [read-only field inputs](https://craftcms.com/docs/5.x/extend/field-types.html#read-only-mode) should cover this, but I will expand the list of circumstances that a field might be rendered in static mode. ✌️

brandonkelly:
@AugustMiller Correct, a field must be both visible and editable to accept POST changes.

PR #16812, opened 2025-03-02
Adds a `db/repair` CLI command

angrybrad:
On MySQL, this runs `OPTIMIZE TABLE` on each table in the database.

https://dev.mysql.com/doc/refman/8.4/en/optimize-table.html

On PostgreSQL, this runs `ANALYZE VERBOSE` on each table in the database.

https://www.postgresql.org/docs/current/sql-analyze.html

These run on the individual tables on separate statements to minimize the amount of table locking that happens.

You can alternatively run these using the `mysqlcheck -o`, but we’re avoiding that because we can’t be guaranteed it’s available on the installation or that it’s the correct version. PostgreSQL doesn’t have an equivalent CLI tool for this.

Tested on MySQL and PostgreSQL in DDEV and Craft Cloud.

Can be merged into Craft 5.

Relevant docs to update (@AugustMiller)
https://craftcms.com/docs/4.x/installation.html (and for 5)
https://craftcms.com/knowledge-base/troubleshooting-performance-issues#the-database


src/console/controllers/DbController.php
```diff
@@ -82,6 +82,45 @@ public function options($actionID): array
         return $options;
     }
 
+    /**
+     * Repairs all tables in the database.
+     *
+     * Note that this can cause table locking, which could interfere SQL being executed.
```
deleugpn: interfere with?

src/console/controllers/DbController.php
```diff
@@ -82,6 +82,45 @@ public function options($actionID): array
...
+     * @see https://www.postgresql.org/docs/current/sql-analyze.html
+     */
+    public function actionRepair(): int
+    {
+        if (!$this->_tablesExist()) {
+            $this->stdout('No existing database tables found.' . PHP_EOL, Console::FG_YELLOW);
+            return ExitCode::OK;
+        }
+
+        if ($this->interactive && !$this->confirm('Are you sure you want to repair all tables from the database?')) {
```
deleugpn, line 101: Is it possible to reach to the database name here (without querying the actual db, just config-wise) and embed it into the message?

deleugpn (approved):

src/console/controllers/DbController.php
angrybrad: Sorry - still learning English.

src/console/controllers/DbController.php
angrybrad: Possible - but we don't do it in any other console commands mainly because 99.9% of all Craft installs connect to a primary database and there's no confusion.

PR #16816, opened 2025-03-03
initialise slug generation when the entry type has changed and the slug is empty or temp

i-just:
### Description
steps to reproduce: 
- have a section with two entry types: `et1` and `et2` (both have the native title field visible); 
- open the edit page for a brand new (fresh) entry, don’t fill out any content, switch to the `et2` entry type, and start filling out the title; the slug doesn’t get auto-generated; 

When saving a draft and the entry is fresh, and the entry type is changed, check if the slug is still empty or temporary, and if so, initialise the `SlugGenerator`.

### Related issues
n/a


brandonkelly:
Thanks! I moved the logic over to `TitleField` instead, so the JS code will get executed every time a Title field is (re-)rendered.

PR #16819, opened 2025-03-03
Nested element duplication

brandonkelly:
### Description

Adds a “Duplicate” action to nested elements in card view, as well as nested entries within Matrix fields in inline-editable block view.

A Matrix field in card view, with three nested entries. The action menu is expanded for the second entry, revealing a “Duplicate” action.

A Matrix field in inline-editable block view, with three nested entries. The action menu is expanded for the second entry, revealing a “Duplicate” action.

### Related issues

- #14178
- #15773
- #16784


alexanderbuergin:
Very cool

Coysh:
I've just updated to 5.7 - and don't seem to have this option. Any ideas why not?
CleanShot 2025-04-16 at 15 49 23
CleanShot 2025-04-16 at 15 51 54

brandonkelly:
@Coysh has your field reached its Max Entries setting? Otherwise not sure. Please write into support@craftcms.com with your Composer files and a database backup, and we can look into into it from there.

Coysh:
@brandonkelly - yep that was it. Max entries. Thanks for pointing out an obvious user-error!

PR #16848, opened 2025-03-07
Fix Cp helper class for frankendphp worker mode 

giorgiopogliani:
### Description
I was testing frankenphp worker mode with Craft CMS and I found a problem that I don't think has a solution without some changes to Craft.

Basically the `craft\helpers\Cp` class has a static property of the requested site and it's only set once. If it's already set the property is returned. This is causing the sites dropdown to basically not work without restarting the worker.

The property It's also private and not nullable, so I can't do much from outside.

The changes:
• Static properties cannot be unset, so I updated the type hint of the `$_requestedSite` to be nullable. The function `requestedSite()` returns a nullable anyway, so maybe, it might even make sense.
• Then I updated the if in the `requestedSite()` function to set the variable only when it's null instead of uninitialized.
• I added a public function to clear the requested site, setting it to null. In this way I can call the function from outside at the start of each request to be sure that the correct site will be used.

Thanks.

### DIFF
--- a/src/helpers/Cp.php
+++ b/src/helpers/Cp.php
@@ -107,7 +107,7 @@ class Cp
      * @var Site|false
      * @see requestedSite()
      */
-    private static Site|false $_requestedSite;
+    private static Site|null|false $_requestedSite;
 
     /**
      * Renders a control panel template.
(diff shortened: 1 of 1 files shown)


brandonkelly:
Thanks! Renamed to `reset()` and merged into the `5.7` branch via c379f83c1ac3ff0c59b723fe3a1098743c6aaa53.

brandonkelly:
Craft 5.7.0 is out with that change.

Issue #16872, opened 2025-03-12
[5.x]: Changing the propagation method for a matrix field results in existing contents being sorted incorrectly


MoritzLost:
### Description

Kind of a follow-up to #16740 

Different project, but again, I have a multi-site with two sites and a matrix field inside _another_ matrix field. I changed the propagation method of the nested matrix field from `Save entries to all sites the owner element is saved in` to `Only save entries to the site they were created in`. As expected, an `ApplyNewPropagationMethod` queue job is created and deduplicates the existing contents for all entries on both sites.

After the queue job has finished, the sort order of the entries in the matrix field is incorrect in one of the sites. All entries got mixed around seemingly randomly. This only happened in one of the sites. In the primary site, which is also the site where all entries were created initially, the order of the matrix entries is still correct. But on the second site, the matrix entries are shuffled around.

This time, this happened in my dev environment and I can reliably reproduce it.

### Steps to reproduce

See above.

### Craft CMS version

5.6.11

### PHP version

8.3

i-just:
Hi, thanks for reporting! I raised a PR for this.

MoritzLost:
@i-just Thanks! Do you know when the next release with that bugfix will land? I'm stuck on a client request with this …

brandonkelly:
Craft 5.6.12 is out with that fix. Thanks again!

PR #16885, opened 2025-03-13
when element belongs to a field, orderBy elements_owners.sortOrder

i-just:
### Description
When applying a new propagation method, the elements are grabbed ordered by ID, ascending, which can lead to nested elements changing their order after the job is done.
Sorting elements by `elements_owners.sortOrder`, if they belong to a field, mitigated this.

Steps to reproduce:
1. clean v5 installation
2. create 2 sites
3. create a matrix field (e.g. cards view mode) with an entry type that can contain just the title; propagation method: all;
4. create a section (e.g. channel) with an entry type that uses the matrix field from the previous step; the section should be enabled for both sites, propagation method: all
5. create one entry in that section; in the matrix field, create 3 entries and call them something that clearly describes their order (e.g. `test 1`, `test 2`, `test 3`); drag `test 3` above `test 2` so that the sort order doesn’t match the ID order; fully save the root entry
6. take a backup of the database
7. edit the matrix field, change the propagation method to none and save
8. let the `ApplyNewPropagationMethod` job finish
9. edit the root entry and check the order of the entries in the matrix field; it’ll be the correct order for the primary site (`test 1`, `test 3`, `test 2`) and incorrect (ordered by ID) for the second site (`test 1`, `test 2`, `test 3`)

If you now import the backup, switch to this PR and go through steps 7-9, the order will be the same (correct: `test 1`, `test 3`, `test 2`) for both sites.

### Related issues
#16872


brandonkelly:
Ordering by ID is important so the element batches stay consistent across queue jobs, if there’s enough elements that the job needs to be batched (#12638).

The `sortOrder` property should have been persisting for the duplicated elements. It wasn’t because the `$otherSiteElements` query wasn’t applying `$this->criteria` to the query, so the query was missing its `$queryId` param, preventing `NestedElementQueryTrait` from joining in the `elements_owners` table, etc.:

https://github.com/craftcms/cms/blob/1cde6408dab3ce7f0e4b51a0868540b6aac69cc5/src/elements/db/NestedElementQueryTrait.php#L201-L202

Fixed by ensuring `$this->criteria` gets applied to that query, too. (60884b9fb73ad669d78cd8cbfaf6cb5039044772)

Issue #16875, opened 2025-03-12
[5.x]: relatedToCategories on multisite does not work when one category is disabled on other site


remoflury:
### Description
I got a multisite craft cms 5 instance using graphql. I got an entry type, which i can assign one ore multiple categories to it. Those categories can be be disabled or enabled on each site individually. When filtering for an entry, which has a category disabled on **another** site, the response will be an empty array, even if there are clearly entries on this site with this category

I am implementing a frontend filter of entries, which sends a query somewhat like this on change of the filter:
```
query teamQuery($siteHandle: [String], $onlyFilters: Boolean!, $teamMemberIds: [QueryArgument]) {
  entries(
    site: $siteHandle
    section: "team"
    id: $teamMemberIds
    relatedToCategories: {id: ["28235"]}
  ) @skip(if: $onlyFilters) {
    id
    title
    ... on team_Entry {
      teamCategories {
        id
        title
      }
    }
  }
}
```

### Expected behavior
Show all filtered entries of the current site with the categories (enabled on the selected site).

### Actual behavior
Show all entries of current site with the categories. but if the filtered category is disabled on another site, no entry is shown.

### Craft CMS version

5.6.11

### PHP version

8.3.17

### Database type and version

mysql

### Installed plugins and versions

-

i-just:
Hi, thanks for reporting! It looks like this will happen if the category is disabled for the primary site. I raised a PR for this. 

If you need this to work straight away, you can modify the `relatedToCategories` part to use any site, like so: `relatedToCategories: [{id: ["12345"], site: "*"}]`.

brandonkelly:
I think this is working as expected. `relatedToCategories` will execute an element query to identify the categories which should be related, and supports all normal category criteria params. So it should work like other category queries in GraphQL, and ignore disabled categories by default.

Rather than passing `site: "*"`, you could pass `status: null`, which would cause the category query to ignore category statuses.

Alternatively, if the categories are related via a Categories field, you can use the field-provided param, which ignores status.

```graphql
query teamQuery($siteHandle: [String], $onlyFilters: Boolean!, $teamMemberIds: [QueryArgument]) {
  entries(
    site: $siteHandle
    section: "team"
    id: $teamMemberIds
    myCategoriesField: [28235]
  ) @skip(if: $onlyFilters) {
    id
    title
    ... on team_Entry {
      teamCategories {
        id
        title
      }
    }
  }
}
```

(Replace `myCategoriesField` with the actual field handle.)

PR #16886, opened 2025-03-13
GQL, relatedTo* and default site

i-just:
### Description
When using `relatedTo*` in GQL we should default to all sites, as opposed to the primary site.

Steps to reproduce:
- clean v4 installation
- create 2 sites (`siteA` and `siteB`)
- create a category group which contains a category (`cat1`) that is disabled for the primary site (`siteA`) and enabled a non-primary site (`siteB`)
- create a Categories field (`myCategories`) for that group
- create a section (`gh16875`) with an entry type which contains the `myCategories` field; this section is enabled for both sites, propagation method all (but also tested with none)
- in that section, create an entry which has the `cat1` category in it; save;

This query in twig will return the entry, and the content of the `myCategories` will be empty.

```twig
{% set results = craft.entries()
  .site('siteA')
  .section('gh16875')
  .myCategories({'id': <category_id>})
  .all() %}
```

If you switch to `siteB` the same query will return the entry and the `cat1` category in the `myCategories` field.

In GQL, this query:

```graphql
query myQuery {
  entries(
    site: "siteA"
    section: "gh16875"
    relatedToCategories: [{id: ["<category_id>"]}]
  ) {
    id
    title
    ... on gh16875_default_Entry {
      myCategories {
        id
        title
      }
    }
  }
}
```

will return an empty array for both `siteA` and `siteB`.

With this PR, the GQL query will return the same results as the twig.

### Related issues
#16875

### DIFF
--- a/src/gql/base/RelationArgumentHandler.php
+++ b/src/gql/base/RelationArgumentHandler.php
@@ -38,6 +38,9 @@ abstract class RelationArgumentHandler extends ArgumentHandler
         foreach ($criteriaList as $criteria) {
             /** @var ElementQuery $elementQuery */
             $elementQuery = Craft::configure(Craft::$app->getElements()->createElementQuery($elementType), $criteria);
+            if ($elementQuery->siteId == null) {
+                $elementQuery->site('*');
+            }
             $idSets[] = $elementQuery->ids();
         }
 


brandonkelly:
I think this is already working as expected – see https://github.com/craftcms/cms/issues/16875#issuecomment-2725406910.

PR #16888, opened 2025-03-13
Element copy/paste support

brandonkelly:
### Description

Adds the ability to “copy” entries and addresses, and paste them elsewhere.

There are three ways to trigger an element copy:

- via a “Copy [type]” element action available in chips, cards, and edit screen toolbars
- via a “Copy” bulk element action within element indexes
- via a “Copy all [type]” Matrix field action (when in cards or inline-editable blocks view modes)

An element chip with its action menu opened, revealing a “Copy entry” item within it.

When an element(s) is copied, a special toast notification will appear that identifies the copied element(s), and “Paste [type]” buttons will appear throughout the control panel wherever pasting the copied element(s) is supported.

A snippet of the Craft control panel, showing an element copy notification hovering over an Addresses field. The Addresses field has a “Paste address” button with special styling to match the notification.

The notification will persist across all browser tabs and across page loads, until either the element is pasted somewhere, or the notification is explicitly closed.

Other tidbits:

- It’s possible to copy addresses between separate Addresses fields, or even between an Addresses field and a user’s addresses.
- It’s possible to copy entries between separate sections or Matrix fields, or even between a section and a Matrix field, so long as the target section/field allows the copied entry type(s).
- It’s possible to copy entries to different sites.

### Related issues

- #14178
- #14469
- #15880


thupsi:
Excellent! Thank you so much Brandon!

gcamacho079 (changes requested):
Everything is looking great so far, @brandonkelly. The only issue I found manually testing is that after activating the Paste button, focus is dropped when the button is removed. (I tested this for inline editable Matrix items only, but I expect it probably happens in the other contexts you mention.) In these cases, we should make sure that focus is moved to a logical element, like the first focusable item in the pasted content.

PR #16907, opened 2025-03-18
Bugfix/copy text amends

i-just:
### Description
Add the `value` attribute to the `craft-copy-attribute` element when the button is created via `Craft.ui.createCopyTextBtn()`.
When getting the value for the `CraftCopyAttribute` via the attribute, add an extra check to prevent a js error.

This was also causing the vue tables, like entry types table, to have very tall rows.

Before:
Screenshot 2025-03-18 at 08 24 46

After:
Screenshot 2025-03-18 at 08 24 34

### Related issues
Follow-up to #16896.


brianjhanson (approved):

brandonkelly:
Doh, missed this and made a similar fix via de2f6cd4ba16b82ab48bdbb231718c16726da6a3.

PR #16908, opened 2025-03-18
Feature/cms 1373 warn when moving large asset folders

i-just:
### Description
If moving more than 50 items (assets + folders - not counting assets in those folders) or more than 50 MB of assets (including the assets in the folders), warn the user about the move beforehand.

Screenshot 2025-03-18 at 10 10 39

Otherwise, show a toast notification after the move is complete, with an Undo button. 

https://github.com/user-attachments/assets/6d20535e-028a-48e7-880c-450e4bd8699e

If any conflicts were encountered during the initial move, the Undo button won’t be available (even if some assets/folders were moved).

### Related issues

- #16749
- #16854


brandonkelly:
Let’s only show the “Undo” button if there weren’t any conflicts to begin with.

If there are conflicts, you already get a chance to cancel the action within the conflict resolution dialog.

i-just:
Thanks! Done and the PR description updated.

Issue #16778, opened 2025-02-26
[5.x]: Can't save entry because provisional draft already exists


rgauthier4c4:
### Description

Sometimes when a user is trying to save an entry in the CMS, an error occured with the following message: "A provisional draft already exists for element/user 1513/1126"

It happens on two different projects that our clients are filling their website content in staging. Both project were up to date in the cms version.

The first time it happened, the user couldn't save his draft or publish the entry, but another user could save the same entry. 

As for my temporary fix, I'm running the garbage collector by command line and the faulty draft is delete (php craft gc/run)

I haven't been able yet to reproduce the steps leading to the bug on my local environnement but I will let you know if I can find more details about this.

![Image](https://github.com/user-attachments/assets/3d608f22-eff6-48dd-beb3-fe02c2c516e5)

### Craft CMS version

5.6.6

### PHP version

8.2

### Database type and version

mysql 8.0

### Installed plugins and versions

-
Amazon S3            2.2.1
Blitz                        5.9.10
CKEditor                 4.5.0
Code Field               5.0.0
CP Field Inspect      2.0.4
Dummy Data            5.0.5
Embedded Assets   5.4.3
Guide                        5.2.0
Hyper                       2.2.3
Image Resizer          4.0.2
Matrix Extended      v4.2.0
Retour                       5.0.7    
SendGrid                   3.0.0
SEOmatic                  5.1.11
Splashing Images     5.0.3
Sprig                          3.5.2
Twigpack                   5.0.0

brandonkelly:
The next time it happens, please send a database backup and your Composer files into support@craftcms.com and point out the affected user/entry.

LucExitable:
@brandonkelly, We have a customer who's also affected by this. I've just emailed a database dump and composer files to support@craftcms.com with extra info about the specific element / user.

Thanks for looking into this!

brandonkelly:
Craft 5.6.12 is out with a fix for this. Thanks for your help!

PR #16911, opened 2025-03-18
Bugfix/16778 provisional draft already exists

i-just:
### Description
**Steps to reproduce:**
1. clean 5.6.11 install
2. create a `matrix` Matrix field with an entry type (`matrixEt`) - it doesn’t have to have any fields in it, just the native title
3. create a `blog` section with an `article` entry type containing the `matrix` field
4. create an entry in the `blog` section (title: “root”), add one entry to the `matrix` (title: “test 1”) field; fully save each slideout and root entry;
5. open the edit page for entry from step 4;
6. change the title to “root 2”
7. edit “test 1” (dbl-click) 
8. change the title to “test 1 edit”
9. close the “test 1” slideout without saving
10. reload the “root 2” page
11. edit “test 1” (dbl-click) 
12. change the title again and notice that the autosave returns an error about the existing provisional draft.

**Issue:**
In step 8, the newly created provisional draft has its owner set to the canonical version of the root entry, not the provisional draft.

After ensuring that the nested element draft is owned by the owner provisional draft, the `NestedElementManager->deleteNestedElements()` needs to account for when we’re deleting an owner that’s a provisional draft. In that case, the nested elements that are also provisional drafts and only belong to the owner provisional draft also have to be deleted.

Huge thanks to @tommysvr for providing a video that made compiling the above steps so much easier.

### Related issues
#16778


brandonkelly:
There are still a couple lingering bugs we need to work through:

- The fix didn’t work retroactively for existing nested provisional drafts.
- If any nested elements were fully saved to duplicates of themselves, they were no longer getting deleted when the owner provisional draft was discarded.

So for now, just fixed this by making `elements/save-draft` stop throwing an exception if there’s an existing provisional draft. Instead it will log a warning and delete the previous provisional draft.

PR #16924, opened 2025-03-19
add `isNewSite` to `ResaveElements` job

i-just:
### Description
Same fix as [here](https://github.com/craftcms/cms/commit/7ce8b10dd7fa64ada460efd614e3474e35ae6ee6) but for when we’re adding site to a section, so that [this code](https://github.com/craftcms/cms/blob/4.14.11/src/fields/BaseRelationField.php#L583-L597) doesn't kick in.

**Steps to reproduce:**
- clean 5.6.12 install
- create 2 sites
- create a section `relations` with an entry type that can have just the default title; enabled for both sites
- create at least one entry in the `relations` section
- create an Entries field, non-translatable
- create a section `blog` with an entry type that contains that entries field added twice; the section should only be enabled for the primary site
- create an entry in the `blog` section, fill out the title and fully save
- edit the entry from the previous step, add an entry to the second Entries field & save
- edit the `blog` section and enable it for the second site
- wait for the resave job to complete
- view the `blog` entry in the primary site - all is good; switch to the second site - the relation is showing in the first field, not the second one; if you check the `elements_sites` table for this `elementId`, the `content` column will be `null` for the site we just enabled for this section.

I wasn’t able to reproduce it on v4, but since we previously added this mechanism for v4, I opted to do this bit here. If it was a wrong call, LMK and I can target only v5.

### Related issues
n/a; noticed while working on https://github.com/craftcms/cms/pull/16919

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Release Notes for Craft CMS 4
 
+## Unreleased
+
+- Fixed a bug where relations weren’t always propagating to newly-added sites for sections correctly. ([#16924](https://github.com/craftcms/cms/pull/16924))
+
 ## 4.14.11.1 - 2025-03-19
 
 - Fixed a performance degradation bug that occurred when working with Categories or Entries fields with “Maintain hierarchy” enabled. ([#16920](https://github.com/craftcms/cms/issues/16920))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Release Notes for Craft CMS 4
 
-## Unreleased
+## 4.14.11.1 - 2025-03-19
 
 - Fixed a performance degradation bug that occurred when working with Categories or Entries fields with “Maintain hierarchy” enabled. ([#16920](https://github.com/craftcms/cms/issues/16920))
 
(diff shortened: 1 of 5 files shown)


brandonkelly:
Doesn’t look like this fix is complete? `ResaveElements::$isNewSite` is unused.

Assume its value needs to be passed onto `$item->isNewForSite` within `processItem()`?

Let’s have the property be renamed to `$isNewForSite` too – since in this case the site itself isn’t new; it’s just new for the elements being resaved.

i-just:
Yeah, sorry, my bad. I played with the `resaving` attribute and didn’t redo all the changes :/ 

I opted for `ResaveElements::$hasNewSite`, which is then passed to `$item->isNewSite`. 

The `Element::$isNewForSite` is already used to check if the element is new for the site we’re working on, and the condition we want to bypass kicks in when `Element->isNewForSite` is `true`, so I didn’t want to mess with it.

Alternatively, this could use the `Element::$resaving` property with a check for it added [here](https://github.com/craftcms/cms/blob/4.14.11/src/fields/BaseRelationField.php#L583-L597), but it would change how the resave job behaves at the moment, so I’m not sure it’s a good idea.

brandonkelly:
> Alternatively, this could use the `Element::$resaving` property with a check for it added [here](https://github.com/craftcms/cms/blob/4.14.11/src/fields/BaseRelationField.php#L583-L597), but it would change how the resave job behaves at the moment, so I’m not sure it’s a good idea.

I do think that’s actually a better option here. Went ahead and made that change.

PR #16939, opened 2025-03-21
TTR is ignored for `BaseBatchedJob` that implement `RetryableJobInterface`

timkelty:
### Description
It is not uncommon for jobs to implement `yii\queue\RetryableJobInterface`.
- Feed Me: https://github.com/craftcms/feed-me/blob/6.x/src/queue/jobs/FeedImport.php#L22
- Blitz: https://github.com/putyourlightson/craft-blitz/blob/develop/src/jobs/GenerateCacheJob.php#L18

In that scenario, the Job's `getTtr` [will be ignored](https://github.com/craftcms/cms/blob/4.15/src/queue/BaseBatchedJob.php#L154) and prefer `BaseBatchedJob::$ttr`.

To remedy this in a non-breaking way, `BaseBatchedJob` now implements `yii\queue\RetryableJobInterface`.

Note: here is where Yii accounts for `yii\queue\RetryableJobInterface` when pushing a job: https://github.com/yiisoft/yii2-queue/blob/master/src/Queue.php#L155-L159

### DIFF
--- a/src/queue/BaseBatchedJob.php
+++ b/src/queue/BaseBatchedJob.php
@@ -12,6 +12,7 @@ use craft\base\Batchable;
 use craft\helpers\ConfigHelper;
 use craft\helpers\Queue as QueueHelper;
 use craft\i18n\Translation;
+use yii\queue\RetryableJobInterface;
 
 /**
  * BaseBatchedJob is the base class for large jobs that may need to spawn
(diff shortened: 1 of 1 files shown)


brandonkelly:
`getTtr()` only provides the _default_ TTR for jobs. With this change, if a job is pushed with an explicit TTR, it would only be respected for the first job, and then `getTtr()` would take precedence for subsequent batches.

Without the change, the only current semi-negative side effect is that if for some reason, a job wanted to define a different TTR for each batch, it wouldn’t be able to do that, since `getTtr()` will only be called once, when queuing the first job, only if a specific TTR wasn’t specified. However it’s unknowable by the time the job is executing, whether the TTR was user-supplied of came from `getTtr()`. So we should err on the side of just being consistent with whatever the first job’s TTR was, regardless of how it was set.

PR #16947, opened 2025-03-24
Make BaseBatchedJob TTR aware

timkelty:
### Description
In addition to memory usage, also tracks running time and compares against TTR


src/queue/BaseBatchedJob.php
```diff
@@ -138,11 +143,18 @@ public function execute($queue): void
             $this->itemOffset++;
             $i++;
 
-            // Make sure we're not getting uncomfortably close to the memory limit, every 10 items
-            if ($startMemory !== null && $i % 10 === 0) {
-                $memory = memory_get_usage();
-                $avgMemory = ($memory - $startMemory) / $i;
-                if ($memory + ($avgMemory * 15) > $memoryLimit) {
+            // Make sure we're not getting uncomfortably close to the memory limit or TTL, every 10 items
+            if ($i % 10 === 0) {
```
brandonkelly: Why not just do this every time?

timkelty: I was just integrating it with the existing code, but we probably should.
Why bother for memory? Is `memory_get_usage` a taxing op?

src/queue/BaseBatchedJob.php
```diff
@@ -139,14 +146,22 @@ public function execute($queue): void
...
+                    $avgMemory = ($memory - $startMemory) / $i;
+                    if ($memory + ($avgMemory * 15) > $memoryLimit) {
+                        break;
+                    }
                 }
             }
 
+            // Make sure we're not getting uncomfortably close to the TTL.
+            $runningTime = microtime(true) - $start;
+            if ($runningTime + static::TTR_BUFFER_SECONDS > $this->ttr) {
```
AugustMiller: I think this is comparing microseconds to seconds? `runningTime` will probably always end up being greater than `ttr`.

```php
$scaledTtr = $this->ttr * 1_000_000;
```

What do you think about scaling it linearly, too?

```php
// If we ran for another 50% of the currently-elapsed time, would we go over?
if ($runningTime * 1.5 > $scaledTtr) {
  // ...
}
```

AugustMiller (dismissed):

src/queue/BaseBatchedJob.php
timkelty: @AugustMiller it is still seconds, just with a microsecond accuracy.

timkelty: > What do you think about scaling it linearly, too?

What would be be gaining?

src/queue/BaseBatchedJob.php
brandonkelly: > Why bother for memory? Is `memory_get_usage` a taxing op?

Not that I know of.

src/queue/BaseBatchedJob.php
AugustMiller: I don't know exactly why I proposed this estimation logic—I guess parity with the memory calulation would make the most sense (a rough average of iterations so far, i.e: `$runningTime + ($runningTime / $i) > $this->ttr`) ?

Any fixed value assumes that each iteration takes less than that amount of time. As an example (using the current 10s threshold): if the first three items took ~15s each, and the TTR was 60s, we would start the fourth item (because there are more than 10 seconds remaining), and then it would _likely_ get terminated. If the goal is to increase resilience against a cutoff time, we should be hyper-cautious (like 50% of TTR elapsed), or use the information we have to dynamically use more of the time…

That example could be addressed with something like…

```php
// If the next iteration took twice as long as the current average time, would it go over?
if ($runningTime + (($runningTime / $i) * 2) > $this->ttr) {
    break;
}
```

src/queue/BaseBatchedJob.php
timkelty: K, killed it for both 🤷‍♂️

src/queue/BaseBatchedJob.php
timkelty: Ah ok, that makes sense.

timkelty: Ok, made that change. 🎩

brandonkelly:
> Allows `batchSize` to be `null`, allowing batching to be solely determined by TTR.

@timkelty did you test this? `craft\base\Batchable::getSlice()` (called from `execute()`) requires `$limit` to be an int.

timkelty:
> @timkelty did you test this? `craft\base\Batchable::getSlice()` (called from `execute()`) requires `$limit` to be an int.

Apparently not well enough…I guess PHPstan was happy because we don't have `strict_types=1`…
Anyway, [fixed now](https://github.com/craftcms/cms/pull/16947/commits/1d9ba9e703d8cb5d211f0f2aeaed5f610c2edc7f).

src/queue/BaseBatchedJob.php
```diff
@@ -106,18 +106,24 @@ final protected function totalItems(): int
      */
     final protected function totalBatches(): int
     {
-        return (int)ceil($this->totalItems() / $this->batchSize);
+        return (int)ceil($this->totalItems() / $this->batchSize());
+    }
+
+    final protected function batchSize(): int
+    {
+        return $this->batchSize ?? 1;
```
brandonkelly: This will effectively cause `batchSize` to default to `1`, meaning a new job will be spawned for every item. Which doesn’t seem consistent with the PR description.

My vote is to just revert the nullable `batchSize` change.

timkelty: Yeah, I guess our docs could just say to pass 1 if you want it to be controlled by ttr.

brandonkelly: Setting to `1` would have the same effect: a new batch per each individual item.

I don’t think there’s a realistic way to have an unlimited batch size that is fully left up to the TTR and memory consumption. We still need to know how many items to query at the beginning of `execute()` to iterate over for the current batch.

timkelty: oh duh, I was confusing batchSize and totalBatches.

PR #16955, opened 2025-03-25
Feature/pt 2432 ensure the order of headings is semantically correct

gcamacho079:
### Description
Updates heading structure in field layout designer and adds a skip link when followed by the card view designer.


src/models/FieldLayoutTab.php
```diff
@@ -54,7 +54,10 @@ public static function createFromConfig(array $config): self
     public function labelHtml(): string
     {
         return
-            Html::tag('span', Html::encode($this->name)) .
+            Html::tag('span', Html::encode($this->name), [
+                'role' => 'heading',
+                'aria-level' => '3',
+            ]) .
```
brandonkelly: @gcamacho079 Any reason not to just make this an `<h3>` tag?

gcamacho079: @brandonkelly I can definitely make it an `<h3>` tag. This method required no CSS changes so I thought I’d give it a try that way. 👍🏼

gcamacho079:
@brandonkelly resolved and updated the CSS/JS

PR #16972, opened 2025-03-27
Multi-element type eager loading maps

brandonkelly:
### Description

Adds support for `ElementInterface::eagerLoadingMap()` to return mappings for multiple element types, or not specify element types at all if they aren’t known.

```php
// existing format
return [
    'elementType' => MyType::class,
    'map' => [
        ['source' => 100, 'target' => 101],
        // ...
    ],
];

// multiple element type-specific mapping groups
return [
    [
        'elementType' => MyType1::class,
        'map' => [
            ['source' => 100, 'target' => 101],
            // ...
        ],
    ],
    [
        'elementType' => MyType2::class,
        'map' => [
            ['source' => 102, 'target' => 103],
            // ...
        ],
    ],
];

// types defined per mapping
return [
    'map' => [
        ['elementType' => MyType1::class, 'source' => 100, 'target' => 101],
        ['elementType' => MyType2::class, 'source' => 102, 'target' => 103],
    ],
];

// no types specified
return [
    'map' => [
        ['source' => 100, 'target' => 101],
        // ...
    ],
];
```

Needed to fix a bug where nested elements’ `getOwner()` and `getPrimaryOwner()` methods weren’t working properly if they had been loaded alongside other elements that didn’t share the same owner type (fully explained in #16960).

### DIFF
--- a/CHANGELOG-WIP.md
+++ b/CHANGELOG-WIP.md
@@ -109,6 +109,7 @@
 - Added `craft\web\View::registerIcon()`.
 - Added `craft\web\assets\codemirror\CodeMirrorAsset`.
 - `craft\base\Element::fieldLayoutFields()` now has an `editableOnly` argument.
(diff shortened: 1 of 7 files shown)


olivierbon:
That solves my use case 👍

i-just (approved):

Issue #16763, opened 2025-02-24
[5.x]: Cannot copy billing or shipping address on order edit screen


philipboomy:
### Description

It would be nice to be able to copy the address on the order edit screen. We some times reply to customers where we list their address for confirmation. We cannot copy paste the address anymore. We have to click the edit of the address and then copy paste 1 input field at a time. I am not sure if this is a setting change or on purpose.

nfourtythree:
Hi @philipboomy 

Thank you for your message.

Commerce implements the address card preview as dictated by Craft itself. So for this we will need a figure out a solution that works in the core of Craft (you can see the same behaviour if you look at addresses on a user).

Will transfer this issue across to the Craft repo so we can track any progress.

If you urgently require a solution for you could add the following code in a custom module's `init()` method. 

```php
if (Craft::$app->getRequest()->getIsCpRequest()) {
    Craft::$app->getView()->registerCss('.order-address-display .card { -webkit-user-select: initial; -moz-user-select: initial; -ms-user-select: initial; user-select: initial; }');
}
```

Thanks!

brandonkelly:
Related: #16632

boboldehampsink:
I added this custom CSS via @nfourtythree's suggestion so that the card's contents are selectable but not firing the click event to open the drawer, but the title and button still do:

```css
#orderDetailsTab div[model-name="billing"] {
  pointer-events: none;
}

#orderDetailsTab div[model-name="billing"] button,
#orderDetailsTab div[model-name="billing"] a {
  pointer-events: auto;
}

.order-address-display .card {
  user-select: auto !important;
}
```

brandonkelly:
Card body contents will be user-selectable in Craft 5.7 (#16983).

brandonkelly:
Craft 5.7.0 is out with that change.

PR #16983, opened 2025-03-29
Card styling/UX improvements

brandonkelly:
### Description

Cards, chips, and inline-editable Matrix blocks are now styled more consistently:

- Each have a consistent background color (either gray or based on the entry type color), with a subtle 1px border around it, and subtle box shadow.
- Cards and inline-editable Matrix blocks have titlebars that have a slightly darker background color, which include the icon and entry type name (if applicable) on the left, and action items on the right.
- Card thumbnails are now styled more consistently with large chips, and are now right-aligned, so card content lines up regardless of whether there’s a thumbnail.

Various element cards and chips

A couple UX improvements are included as well:

- Nested element cards’ titles are no longer hyperlinked. Instead, an edit button is included in the toolbar, which opens the slideout editor. <kbd>Command</kbd>/<kbd>Ctrl</kbd>-clicking on an edit button opens the element’s edit page in a new window (if it has one).
- Card body text is now user-selectable.

### Related issues

- #15651
- #16632
- #16763
- #16185
- #16186
- #16670


iainhenderson:
Brilliant thanks @brandonkelly

i-just (changes requested):
Only a few of issues that I noticed:

- when you create a new nested entry in a matrix field (view mode cards) for an existing or brand new root entry, once the nested entry is saved, you can’t use the actions menu (3 dots); the edit icon triggers two edit slideouts, reordering doesn’t seem to work (via diamond icon); once you reload the page (or save the root entry), all of the above starts working as expected
- matrix in cards view mode (with the standard title field), when you duplicate a nested entry via actions menu > duplicate, it shows the title as hyperlinked; if you reload, the hyperlink is gone; same with copy and paste; this doesn’t happen when duplicating in the element index view mode;
- if you’ve chosen to show Entry Type in the card and you’re logged in as a non-admin, the Entry Type name is hyperlinked, but you don’t have permissions to edit Entry Types; maybe it would be best not to hyperlink it for non-admins?
- if we’re hyperlinking the Entry Type in the card, should we also hyperlink the Section (to link to the Section’s settings)?
- on the element index page in table view mode, there seems to be a large gap between the last item on the list and the footer

BTW, I love that you can click on the name of the entry type to edit it straight away (e.g. from the field edit page).

Once released, the CKEditor should be updated so that we don’t show the edit button and don’t hyperlink the title in the nested element cards.

brandonkelly:
Thanks @i-just!

> * when you create a new nested entry in a matrix field (view mode cards) for an existing or brand new root entry, once the nested entry is saved, you can’t use the actions menu (3 dots); the edit icon triggers two edit slideouts, reordering doesn’t seem to work (via diamond icon); once you reload the page (or save the root entry), all of the above starts working as expected

Good catch. Looks like I introduced that in fd65d23b3312d7cce8e5b8670db66f7e154e3289 for #16893, so fixed on the 5.7 branch.

> * matrix in cards view mode (with the standard title field), when you duplicate a nested entry via actions menu > duplicate, it shows the title as hyperlinked; if you reload, the hyperlink is gone; same with copy and paste; this doesn’t happen when duplicating in the element index view mode;

Fixed!

> * if you’ve chosen to show Entry Type in the card and you’re logged in as a non-admin, the Entry Type name is hyperlinked, but you don’t have permissions to edit Entry Types; maybe it would be best not to hyperlink it for non-admins?

Doh, yeah. Updated all components’ `getCpEditUrl()` methods to return `null` for non-admins.

> * if we’re hyperlinking the Entry Type in the card, should we also hyperlink the Section (to link to the Section’s settings)?

Agreed. Updated `section` attributes to render as chips.

> * on the element index page in table view mode, there seems to be a large gap between the last item on the list and the footer

Are you referring to the 24px top margin on the footer? If so that’s intentional and hasn’t changed in 5.7 or this PR.

The top margin of the control panel footer

> Once released, the CKEditor should be updated so that we don’t show the edit button and don’t hyperlink the title in the nested element cards.

Yep. Actually it would be safe to update now. The instance config passed to `app/render-elements` just needs to include:

```js
showEditButton: false,
hyperlink: false,
```

Those values will just be ignored if running < 5.7.

i-just:
Thanks!

Looks like the edit icon still opens up two slideouts (when you create a new nested entry in a matrix field (view mode cards) for an existing or brand new root entry).

> Are you referring to the 24px top margin on the footer? If so that’s intentional and hasn’t changed in 5.7 or this PR.

I meant the space between the last item and the footer. It looks about 15px taller on this PR vs 5.x branch. I measured the space between the bottom of the title and the start of the footer, and it’s around 30px on 5.x and 45px on this branch. (The `#footer` `margin-block` is 9px on 5.x and 5.7 and 24px on this branch.)

5.x
Screenshot 2025-03-31 at 13 34 28

this branch
Screenshot 2025-03-31 at 13 35 28


> Yep. Actually it would be safe to update now. The instance config passed to app/render-elements just needs to include

Yup, I already have it done locally :) I’ll PR it at some point today.

src/helpers/Cp.php
```diff
@@ -688,39 +730,67 @@ public static function elementCardHtml(ElementInterface $element, array $config
...
-            }
-        }
+        $thumb = $element->getThumbHtml(120);
+        $icon = $element instanceof Iconic ? $element->getIcon() : null;
+        $title = $element->getCardTitle();
 
         $html = Html::beginTag('div', $attributes) .
+            Html::beginTag('div', ['class' => 'card-titlebar']) .
+            Html::beginTag('div', [
+                'class' => ['flex', 'flex-nowrap', 'flex-gap-xs'],
```
gcamacho079:
```suggestion
                'class' => ['flex', 'flex-nowrap', 'flex-gap-s'],
```
The current spacing seems a little close to me; would this work?

gcamacho079 (approved):

brianjhanson (approved):
For a bit I tried to figure out if we could make matrix blocks and cards share the same rendering path, but it looked like that was a bigger job than was worth taking on right now.

thupsi:
Hi! I've noticed some bugs with the new styles:

image

1. The active tab in the "Row" block has a white background, instead of a yellow tint.
2. The "Column" blocks, having no color set, are inheriting the yellow shade for their titlebar.
3. The text in those blocks also inherits the text color from the parent "Row" block, so it's not gray but dark brown/yellow instead (might not be immediately noticeable in the screenshot). 

In addition to the above I would like to make a proposal to **add "White" to the available colors for entry types**. Since the default now is implicitly "Gray", it would enable keeping the palette muted but have visual grouping and hierarchy by lightness as opposed to color, like in the example bellow:

image

src/helpers/Cp.php
brandonkelly: Sure, done!

brandonkelly:
> Looks like the edit icon still opens up two slideouts (when you create a new nested entry in a matrix field (view mode cards) for an existing or brand new root entry).

@i-just Doh, missed that. Fixed now!

> * The active tab in the "Row" block has a white background, instead of a yellow tint.
> * The "Column" blocks, having no color set, are inheriting the yellow shade for their titlebar.
> * The text in those blocks also inherits the text color from the parent "Row" block, so it's not gray but dark brown/yellow instead (might not be immediately noticeable in the screenshot).

@thupsi Thanks for testing! Fixed.

thupsi:
@brandonkelly I found another small CSS color bug. Text in CKEditor fields inherits the text color of the containing block:
image

brandonkelly:
@thupsi Thanks, posted that on CKEditor’s repo: https://github.com/craftcms/ckeditor/issues/390

iamtompickering:
@brandonkelly not sure if it's just me being a bit picky (apologies if I am!), but the font sizing of Blocks as inline-editable blocks feels a little off now to me. I made a few minor CSS changes using a CP CSS plugin to make a few adjustments. 

Before: 
CleanShot 2025-04-13 at 15 27 06@2x

CleanShot 2025-04-13 at 15 38 49@2x

After: 
CleanShot 2025-04-13 at 15 43 52@2x


CleanShot 2025-04-13 at 15 39 53@2x


The sizing adjustments I made feel a little more in keeping with the spacing of everything else around it in my opinion. Just thought I'd share my opinions, hope that's okay!

brandonkelly:
@iamtompickering Thank for the feedback. I prefer the smaller font size for entry type labels, as a way to set it apart from the tab labels, but I agree the tightened-up inline-editable blocks looks a bit better. So just made that change for the next release. (ea36d2225a1d1d711f4d051be7335597cef6cb69)

iamtompickering:
@brandonkelly completely fair and understand the logic behind it! I'm sure I'll learn to love the smaller font size!

PR #16985, opened 2025-03-30
“In-bound” Color Field Previews

AugustMiller:
Currently, color fields display a random hex value when previewed in the element card attribute designer UI.

This alters the behavior to use palette colors, when available!

❗ Targeting 5.7, as other Color field changes are going out in that release.

---

Curious if this (and `getStaticHtml()`) could benefit from the addition of color labels in #16492…


brandonkelly:
> Curious if this (and `getStaticHtml()`) could benefit from the addition of color labels in #16492…

Yep, good call, done.

Issue #887, opened 2017-01-31
Better Asset Permissions on Asset Fields, etc.


angrybrad:
> Created by: David Ritter (david.ritter@spinthis.com) on 2015/05/18 19:35:25 +0000
> Votes at time of UserVoice import: 16

-----------------------------------

Craft's permissions are pretty flexible, except when it comes to assets. 

When editing a field that allows users to choose an asset, all asset sources will appear in the modal, even ones the user has no View Source permission on. This happens on Redactor fields as well. 

The main Assets manager gets it right—and only shows the sources a user has permission to view or edit—but when it comes to editing entries or when choosing assets, everything is shown, which is not ideal.

The majority of the time, the current setup works well since most users probably want to be able to see others' content. But in  other cases, you simply don't want some users seeing other assets.

Just an example workflow. Big company launch, kind of hush hush. You have a few designers. Some of them are working up some designs. Some pages aren't ready to go live yet. The current situation makes it impossible to do something like that. 

I asked about this on support, Andris mentioned "This is to avoid situations where an Assets fields would be assigned a single source that would not be usable by some users, rendering the field unusable."

That's only part of the problem.  I would vote to disable/grey the field or let them pick another asset from a source they do have access to view. You don't want to be creating asset fields and locking down folders for every asset field, that gets to be ridiculous.

I've seen this come up a few times on SE where we need to be able to "jail" or silo a user to their own assets or at a least ones they are only supposed to see.

* http://craftcms.stackexchange.com/questions/8667/is-replace-file-a-potential-security-hole/8683#8683
* http://craftcms.stackexchange.com/questions/5046/why-cant-users-upload-files-in-this-asset
* http://craftcms.stackexchange.com/questions/10151/restricted-access-to-assets-in-redactor

iainhenderson:
I've just run into problems with this behaviour as well.

The assets fields I'm having problems with appear in blocks as part of a 'page builder' style matrix.  

Currently the only way I can think of to properly lock down these sources is to duplicate the matrix, entry and structure multiple times with slightly different asset source settings for  each of the user groups, and then controlling access to those structures.

That would mean creating and maintaining 308 additional fields. Which as Brad stated is a bit ridiculous.

oyvindostmo:
[...]

PR #16991, opened 2025-03-31
Add the “show unpermitted sections” lightswitch for Entry fields

ThomasDeMarez:
### Description
As already implemented for the Asset field, this PR would add an extra field "Show unpermitted sections" to the Entry field's config. This way Sections can be hidden for users that do not have permission to view any Entries in specific Sections.

- This implementation is setup in a non breaking way by enabling the setting by default.
- The native Link field has also been updated to support this new setting.

### Related issues
https://github.com/craftcms/cms/issues/887

---

**Entry Field**
Link field

**Link Field**
Link field

### DIFF
--- a/src/fields/BaseRelationField.php
+++ b/src/fields/BaseRelationField.php
@@ -1373,11 +1373,16 @@ JS, [
             ]);
         }
 
+        if ($showUnpermitted = $this->getShowUnpermittedHtml()) {
+            $showUnpermittedHtml = $showUnpermitted;
+        }
+
         return [
             'field' => $this,
             'elementType' => $elementType::lowerDisplayName(),
             'pluralElementType' => $elementType::pluralLowerDisplayName(),
             'selectionCondition' => $selectionConditionHtml ?? null,
+            'showUnpermitted' => $showUnpermittedHtml ?? null,
         ];
     }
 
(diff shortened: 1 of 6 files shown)


brandonkelly (changes requested):
Thanks for the PR! I like the idea, but I don’t think we should be doing this from `BaseRelationField`; it should be confined to `craft\fields\Entries`, similar to how Assets’ fields “Show unpermitted volumes” and “Show unpermitted files” settings are implemented. We also may want to split this up into two settings (like Assets fields) – “Show unpermitted sections” and “Show unpermitted entries”.

ThomasDeMarez:
The reason I implemented this on BaseRelationField is so other plugins, like Solspace’s `EventFieldType`, could easily override the base function and introduce their own “Show unpermitted Events” lightswitch. It could also be reused for any other native Craft field that extends BaseRelationField.

If you’d still prefer to implement this specifically within `craft\fields\Entries`, I’m happy to move it there. Let me know what works best!

brandonkelly:
I still think it makes more sense to do it on the child classes, since these settings are dealing with things that are pretty element type-specific.

ThomasDeMarez:
Updated as requested.

brandonkelly:
Thanks! Merged for Craft 5.7 via #17034.

brandonkelly:
Craft 5.7.0 is out with those new settings. Thanks again!

PR #17007, opened 2025-04-01
Bump package versions

brianjhanson:


brandonkelly (changes requested):
@brianjhanson Looks like you need to build + run prettier for the checks to pass.

brianjhanson:
I failed at rebasing this onto 4.15, so ended up opening https://github.com/craftcms/cms/pull/17020 instead.

Issue #16248, opened 2024-11-30
[5.x]: Changes to nested elements don't result in a revision being created until the parent is saved


ryansupercool:
### Description

When editing and saving a nested element a revision is not created until after you save the parent despite the changes to nested element being applied to the canonical element. This can be observed when using a Matrix field in table view mode (inline mode does not exhibit this behaviour). We also see this issue within Commerce when managing Products/Variants, as well as in our own nested elements.

### Steps to reproduce

Requires a Matrix field with oneEntry type and field. The Matrix field should be set to "Display as an element index" and "Display in a table".

1. Create a new Entry and save.
2. A a new Matrix Entry and save, note a revision has been created.
3. Edit the Matrix Entry and save, note a revision hasn't been created. If you were to view the parent Entry on the frontend you will observe that the Matrix Entry edit is published.
4. Save the parent Entry, note a revision has now been created.

At the 1:00 minute mark the changes to the Matrix Entry are "live" so it doesn't appear to be a UI bug with the parent Entry not tracking changes. Craft does know something has changed as it will create the revision with the edits to the Matrix Entry when resaving the parent.

### Expected behavior

When performing step 3 a revision is created.

### Actual behavior

When performing step 3 a revision isn't created.

https://github.com/user-attachments/assets/331901c7-a693-41f1-b687-7df92fa860ad

### Craft CMS version

5.5.4

### PHP version

8.2.24

### Operating system and version

Linux 6.11.6-orbstack-00279-g28c6c77332e6

### Database type and version

MySQL 8.0.36

### Image driver and version

Imagick 3.7.0 (ImageMagick 6.9.11-60)

brandonkelly:
This is working as expected. When you use the element index view mode, the nested entries aren’t really treated as part of the entry content in the way they wood when you use the inline-editable or cards views.

That said, we should at least make it possible for nested entries to track their own revisions. I’ve added that to our Linear project.

ryansupercool:
To confirm it's expected that edits to nested elements that are saved don't result in a revision being created on the parent until the parent itself being saved even if there have been no changes on the parent itself?

1. Create parent and save
2. Create nested element and save the parent. At this point you have one revision on the parent.
3. Edit the nested element and save. At this point you still have one revisiosn on the parent and the nested element changes are published.
4. Save the parent without making further changes to the parent. You now have two revisions on the parent. The first revision is the nested element being created at 2, the second revision contains the edit from 3.

[...]

PR #17008, opened 2025-04-01
Nested entry revisions

brandonkelly:
### Description

Adds an “Enable versioning for entries in this field” setting to Matrix fields. When enabled, nested entries will get new versions created for them each time they are saved, regardless of whether they were saved directly on their own edit page, or via a Matrix field’s input.

### Related issues

- #16248


thupsi:
Nice! 🙂 I've been experimenting with using matrix fields as nested sections, but the lack of nested entry versioning was sometimes a deal breaker.

PR #17021, opened 2025-04-03
Only clear CP resources if they exist

timkelty:
### Description
There are some environments (e.g. Craft Cloud) `where \craft\config\GeneralConfig::$resourceBasePath` may not exist at all. In that case, we don't want a "The dir argument must be a directory" error thrown.

### DIFF
--- a/src/utilities/ClearCaches.php
+++ b/src/utilities/ClearCaches.php
@@ -183,7 +183,7 @@ class ClearCaches extends Utility
                     }
 
                     $basePath = Craft::getAlias($basePath);
-                    if ($basePath !== false) {
+                    if ($basePath !== false && file_exists($basePath)) {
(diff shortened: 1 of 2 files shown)


src/utilities/ClearCaches.php
```diff
@@ -184,7 +184,7 @@ public static function cacheOptions(): array
                     }
 
                     $basePath = Craft::getAlias($basePath);
-                    if ($basePath !== false) {
+                    if ($basePath !== false && file_exists($basePath)) {
```
timkelty, line 186: Checking `file_exists` and not `is_dir`, so that you _will_ get an error if `\craft\config\GeneralConfig::$resourceBasePath` is pointing to a non-dir file.

PR #17046, opened 2025-04-09
clarify the replace file action comment

i-just:
### Description
@timkelty, as per our chat, I updated the comment to clarify the circumstances under which the `assets/replace-file` action would proceed based on the filename only.

### Related issues
n/a


timkelty (approved):
😍

PR #17054, opened 2025-04-09
Async csrf improvements

brianjhanson:


src/templates/_special/async-csrf-input.twig
```diff
@@ -1,19 +1,39 @@
...
+      return input;
+    }
+
+    function handleFocusIn(e) {
+      const asyncInput = e.currentTarget.querySelector('craft-csrf-input');
+      if (asyncInput) {
+        fetchCsrfData().then(({csrfTokenName, csrfTokenValue}) => {
+          const input = createHiddenInput({name: csrfTokenName, value: csrfTokenValue});
+          asyncInput.replaceWith(input);
+        });
```
timkelty, line 25: I think we also need to defer/debounce submit events that come in between `focusin` and injecting the input.

Over a slow connection, you could interact with the form and submit before the CSRF has been injected, resulting in a CSRF error.

leevigraham: Maybe firing a `csrfloaded` event that bubbles up to the form would allow js to disable / enable the form submission.

PR #17065, opened 2025-04-11
Added version 7 to symphony/http-client

cornedor:
### Description

Symfony 6.0 has been end-of-life for a while, I'm having issues with installing packages that are using symfony/http-client 7+ now. For example, https://github.com/stape-io/stape-sgtm-php


brandonkelly:
Thanks for the PR! I’ve pulled your commit into the 4.15 and 5.7 branches.

brandonkelly:
Craft 4.15.0 and 5.7.0 were just released with this change. Thanks again!

PR #17074, opened 2025-04-12
Adds a db/repair CLI command to repair and optimize database tables

angrybrad:
Runs `OPTIMIZE TABLE` on each table for MySQL or `ANALYZE VERBOSE` for each table in PostgreSQL.

Both can cause table locking while they are running.

Can be merged into 5.x, too.


brandonkelly:
Seems redundant. #16812

Issue #17090, opened 2025-04-15
[5.x]: Table Fields with Static Row option don't update their content when rows are changed


MoritzLost:
### Description

We use a Table Field with the **Static Rows** option and a **Row heading** column to allow users to enter social media links in a simple and concise interface, without having to create too many fields manually:

![Image](https://github.com/user-attachments/assets/35170281-4103-4bbe-b70f-6a369ecf436e)

However, we have noticed that it's nearly impossible to change the definition of the table fields once some content has been entered, because existing contents are not updated when the table field changes.

### Steps to reproduce

1. Create a table field and configure it as seen in the screenshot above.
2. Add it to an entry type and create some entries. Fill all rows with content.
3. Change the order of two rows. The values in all existing entries won't match their heading any more.
4. Delete a row, or add a row and move it between the existing rows. In either case, existing entries will again not match their heading any more.

### Expected behavior

When the **Static Rows** option is used, changes to the structure of the table field should be reflected in existing contents. In particular if it's used in combination with a **Row heading** column.

### Actual behavior

Basically, the only modification to a table field using those options is to add rows at the bottom. You can't re-arrange, delete or add rows (above existing rows) without screwing up existing contents.

### Craft CMS version

5.6.11

### PHP version

8.3

brandonkelly:
Sorry for the delay, but we finally got this fixed for the Craft 4.17 and 5.9.

After updating, you will need to resave your Table field so the default rows get UUIDs assigned to them, then resave any elements that use that field (e.g. with `resave/entries --with-fields=myTableField`).

MoritzLost:
@brandonkelly Thanks, Brandon!

brandonkelly:
Craft 4.17.0 and 5.9.0 are out with that fix. Thanks again!

PR #17150, opened 2025-04-22
Bugfix/17090 static table row order

i-just:
### Description
When the “Table” field is in “static rows” mode, keep track of row ids specified in the “Default Values” table in the field’s settings and, when editing an element with that field, use them to display the rows in the same order.

This will only start taking effect once the “Table” field that is in “static rows” mode and the elements containing it are resaved.

### Related issues
#17090


src/controllers/FieldsController.php
```diff
@@ -360,6 +360,11 @@ public function actionSaveField(): ?Response
             $fieldUid = null;
         }
 
+        $settings = $this->request->getBodyParam('types.' . $type);
+        if (isset($oldField) && property_exists($oldField, 'maxRowId')) {
+            $settings['maxRowId'] = $oldField->maxRowId;
+        }
+
```
brandonkelly: FieldsController should not be giving special treatment to any one field type like this.

src/fields/Table.php
```diff
@@ -91,6 +92,11 @@ public static function valueType(): string
      */
     public ?array $defaults = [[]];
 
+    /**
+     * @var int|null The highest rowId used by the field's default rows so far
+     */
+    public ?int $maxRowId = null;
```
brandonkelly: We could probably simplify things and drop `maxRowId` altogether if we just assign random numbers/strings as the row IDs. Maybe UUIDs using `StringHelper::UUID()` / `Craft.uuid()`?

brandonkelly (changes requested):

src/fields/Table.php
i-just: OK, I switched to UUIDs and removed the `maxRowId`. Once the static table field and the elements using it are resaved, the correct order will be maintained.

PR #17179, opened 2025-04-25
Updated dutch translation for "Critical", 

NathanReul:
in this context it should not be used as an adjective in dutch.

Elaboration:
"A critical update" => "Een kritisch**e** update" (As adjective)
"This update is critical" => "Deze update is kritisch" (Not as adjective)


brandonkelly:
Thank you!

Made the same change via [CrowdIn](https://translate.craftcms.com) and pulled it in for the next Craft 4 and 5 releases.

PR #17189, opened 2025-04-26
Add ElementHelper::EVENT_BEFORE_RENDER_ELEMENT event

leevigraham:
Adds an event that allows developers to override the `_partial/` template rendering for elements.

### Description

I'm currently implementing this in a custom module by adding a new Behaviour to Neo's `Block` and `BlockQuery`  classes. The behaviour adds a `renderComponents()` method which in turn does custom rendering. This event would allow me to remove that functionality completely and simplify the implementation.

Additionally it would allow other developers to do more complex rendering in their own PHP classes before rendering templates.

### Related issues

https://github.com/craftcms/cms/discussions/17188

### DIFF
--- a/src/events/BeforeRenderElementEvent.php
+++ b/src/events/BeforeRenderElementEvent.php
@@ -1,0 +1,39 @@
+<?php
+/**
+ * @link https://craftcms.com/
+ * @copyright Copyright (c) Pixel & Tonic, Inc.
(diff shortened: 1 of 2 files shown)


brandonkelly:
Thanks! Moved the rendering logic and event over to `Element::render()` (013f83c97843353eac19155adec52288acd6afb7).

PR #17191, opened 2025-04-28
remove left over console logs

i-just:
### Description
Removes leftover `console.log` from `src/web/assets/cp/src/js/Slideout.js` and `src/web/assets/cp/src/js/TagSelectInput.js`.

### Related issues
n/a


brandonkelly:
Doh, thanks!

PR #17220, opened 2025-05-05
Return URLs saved into session are now sanitized

angrybrad:
Make sure no PHP or HTML code can be saved to session when setting a returnUrl on a request.

Can be merged into Craft 5.


amanion-cisa:
[CVE-2025-35939](https://www.cve.org/CVERecord?id=CVE-2025-35939)

weotch:
@brandonkelly @angrybrad Is Craft 3.x vulnerable to this issue as well?

angrybrad:
@weotch it is affected, but it is also a trivial “vulnerability”, that, as far as we know, cannot be exploited by itself.

Plus, Craft 3 has been out of active support for 2+ years, and out of security support for over a year.

https://craftcms.com/knowledge-base/supported-versions#end-of-life-versions

weotch:
Gotcha, thanks for the quick reply!

PR #17242, opened 2025-05-08
changed entry type and the modified indicators

i-just:
### Description

**Steps to reproduce:**
- create an entry type `et1` with a plain text field in it
- create a second entry type `et2` with a matrix field (in any view mode) in it
- create a section with both entry types, make sure the `et1` is the first on the list
- create a new entry in that section (don’t save)
- switch entry type to `et2`
- click “New entry” to add a new nested entry to the matrix field
- note that the blue modified indicator doesn’t show for the matrix field

If you were to reload the page after switching the entry type, the indicators would show as expected.

This PR fixes the issue by allowing the delta names to get registered when the entry type is changed and updating the `data-delta-names` on the `main-form` container.

### Related issues
n/a


brandonkelly:
This was also a problem with any custom field which was initially hidden but then later shown, via conditions.

Fixed for the next 4 and 5 release via d6870a1bb4e7ef8b78a26b1b99dbdb9f5e134ac4.

The `ElementEditor.js` changes were a little less involved than this PR though, but seem to get the job done. Not sure if I’m missing something?

PR #17243, opened 2025-05-08
fix: route delete button pointer cursor

patrikkaprinay:
### Description
This PR fixes the button cursor on the delete button when editing a route (at path [craft]/admin/settings/routes)

SCR-20250508-rqwj-2


brandonkelly:
Thanks for the PR!

The real issue is that those Delete “buttons” were `<a>`s instead of `<button>`s. Fixed that for the next release, which also resolved the cursor issue.

brandonkelly:
Craft 5.7.6 is out with that fix. Thanks again!

PR #17252, opened 2025-05-12
[5.8] Add `Json` type to Craft's GraphQL type definitions

nfourtythree:
### Description
This PR adds the `JSON` to Craft's GraphQL inbuilt types.

Along with this, user queries have been updated so that the `preferences` field now returns a JSON object rather than a string.

Before:
CleanShot 2025-05-12 at 10 21 55@2x

After:
CleanShot 2025-05-12 at 10 22 11@2x


nfourtythree:
Closing as this is not applicable.

Issue #17249, opened 2025-05-09
[5.x]: Conditionally hidden Title field is required


mmikkel:
### Description

If a _required_ Title field is conditionally hidden, the element will fail validation with the error message "Title cannot be blank.".

This works differently from required custom fields, which are not be included in validation when they are conditionally hidden.

_My use case here is an entry type that will be used both in an inline-editable Matrix field, and for standalone section entries, where the title field should be hidden for the former, but visible and required for the latter._

![Image](https://github.com/user-attachments/assets/21829ae6-092b-4827-9ee8-0b0593cafb1b)

![Image](https://github.com/user-attachments/assets/d3c185ad-cbbc-497e-96ec-bd241a94bb71)

### Steps to reproduce

1. Create an entry type
2. Add the native Title field, and make it required
3. Add an "Entry Condition" to the Title field, ensuring it becomes hidden (e.g. "Level is greater than 1000000000")
4. Create an entry and try to save it

### Expected behavior

Like custom fields, conditionally hidden Title fields should not be included in validation.

### Actual behavior

Elements with empty, required Title fields will fail validation.

### Craft CMS version

5.7.5

### Installed plugins and versions

None

i-just:
Hi, thanks for reporting! I raised a PR for this.

brandonkelly:
Craft 5.7.6 is out with that fix. Thanks again!

PR #17253, opened 2025-05-12
Bugfix/17249 title field conditionals format and requiredness

i-just:
### Description
Only validate the entry’s title field as required if it’s both required and the field is set to show in the form (not conditionally hidden).

If the entry’s title field is present in the field layout but conditionally hidden, update the entry’s title based on the default title format before saving.

### Related issues
#17249


brandonkelly:
I reverted the change to `updateTitle()`, because I realized that could cause unexpected changes to entry titles. E.g. if you have an entry type where the Title field is visible for some users but not others, you wouldn’t want the title reverting to the Default Title Format each time a user who can’t see it edits the entry. (7cd2ee5568d44e5ede62f254bee95fd574f14603)

Issue #17246, opened 2025-05-09
[5.x]: Changing an asset's file extension changes it's mimetype


joepagan:
### Description

Changing an asset's file extension changes it's mimetype

### Steps to reproduce

1. Change the extension where the asset is originally a `jpg` to a `tiff` in the asset manager
2. run `$asset->getMimeType()`
3. returns `image/tiff`
4. When it's has a tiff extension, test in terminal `file --mime-type -b filename.tiff`
5. returns `image/jpeg`
6. change back to jpg
7. returns `image/jpeg`
 
### Expected behavior

mime type should have always returned `image/jpeg` regardless of file extension change

### Actual behavior

mime type changes if the extension changes

### Craft CMS version

5.7.3

### PHP version

8.2

### Installed plugins and versions

-

brandonkelly:
Thanks for reporting that!

I’ve made a couple changes for Craft 5.8 which will make `getMimeType()` more reliable going forward: (#17254)

- Craft will store assets’ real MIME types as part of the asset data, for all newly-uploaded assets. (Unfortunately not really possible to do this retroactively, without downloading all assets’ content from remote filesystems, which would be very expensive.)
- Locally-stored assets’ `getMimeType()` methods will now call `FileHelper::getMimeType()` (which looks at the actual file data, as opposed to `FileHelper::getMimeTypeByExtension()`), as a fallabck for older assets where the MIME type wasn’t saved on upload.

joepagan:
Nice one cheers Brandon! 

Just on your PR notes could you clarify this?

>When Asset::getMimeType() is called without a transform applied, it now returns the stored mimeType value if it’s set, or the volume’s getMimeType() method as a fallback.

Are you saying here if a transform is applied to an asset it will return the transform's mime type? Or it will return the volume's fallback?

brandonkelly:
It will use the transform format’s MIME type (same as it is now).

joepagan:
@brandonkelly do I need to do something special to access that? 

Where original image is a tiff:

```php
$transformParams['format'] = 'jpg';
$transformedAsset = $asset->setTransform($transformParams); // I know this should manipulate the existing $asset model but just incase it didn't

// Check mime type of the transform:
$asset->getMimeType(); // returns image/tiff
$asset->getMimeType($transformParams); // returns image/tiff
$transformedAsset->getMimeType(); // returns image/tiff
$transformedAsset->getMimeType($transformParams); // returns image/tiff
```

[...]

PR #17254, opened 2025-05-12
Asset MIME type improvements

brandonkelly:
### Description

- Adds a new `mimeType` column to the `assets` table, so assets can start keeping track of their MIME types when known to a higher level of confidence than `FileHelper::getMimeTypeByExtension()` offers.
- Any time asset contents touch the server (on upload, or during indexing for local filesystems/if we had a reason to download the file), we now set the asset’s MIME type based on `FileHelper::getMimeType()` while we have a local path to the asset. If that returns `null` and the asset was just uploaded, the MIME type specified by the request is used as a fallback (via `UploadedFile::$type`, or the MIME type specified in the `data` URL).
- When `Asset::getMimeType()` is called *without a transform applied*, it now returns the stored `mimeType` value if it’s set. Otherwise, `FileHelper::getMimeType()` or `getMimeTypeByExtension()` will be called, depending on where the volume’s filesystem implements `LocalFsInterface`.

### Related issues

- #17246


i-just (approved):

i-just:
@brandonkelly, would we want to also try to get the mime type during asset indexing?

brandonkelly:
@i-just Good call. Just updated it so we are storing the MIME type during indexing, if it’s a local filesystem or we had a reason to download the file.

Issue #17259, opened 2025-05-13
[5.x]: Long chip titles overflow container


leevigraham:
### What happened?

Long urls in chips overflow their container.

The container should grow in size.

This is due to a fixed height on the chip.

![Image](https://github.com/user-attachments/assets/fb9f410c-85eb-46f3-b633-84fbb5cb75b0)

### Craft CMS version

5.x

### Installed plugins and versions

-

i-just:
Hi, thanks for reporting! I raised a PR for this.

brandonkelly:
Craft 5.7.6 is out with that fix. Thanks again!

PR #17261, opened 2025-05-14
Bugfix/17259 link field styling

i-just:
### Description
Allow the URL chip in the link field to grow in height.

Before:
Screenshot 2025-05-14 at 07 43 40

After:
Screenshot 2025-05-14 at 07 43 56

Prevent the Element chip in the link field from overflowing the container.

Before:
Screenshot 2025-05-14 at 07 44 27

After:
Screenshot 2025-05-14 at 07 44 17

### Related issues
#17259


leevigraham:
@i-just **very** minor thing… but should the chip have a small amount of block padding?

brandonkelly:
@brianjhanson would it be possible to give URL chips the truncate + tooltip treatment like element chips?

brianjhanson:
> @brianjhanson would it be possible to give URL chips the truncate + tooltip treatment like element chips?

I don't see why not, I'll put a PR together

Issue #17270, opened 2025-05-15
[5.x]: Card label disappears when dragging entry with a truncated title


tommysvr:
### Description

When dragging and dropping a Matrix entry in card view, the title disappears for cards that have a truncated title

### Steps to reproduce

1. Create multiple Matrix entries in a field with the View More set to 'As cards' and in a grid
2. Drag and drop an entry with a long title that gets truncated
3. The title disappears

### Expected behavior
Title remains visible

### Actual behavior
Title disappears, but is visible again after saving the main entry.

There is an warning in the browser console: `No label link found in craft-element-label`

https://github.com/user-attachments/assets/e5bf3f94-df6e-4d5e-bac9-19b66bb866c1

### Craft CMS version

5.7.6

### PHP version

8.2

i-just:
Thanks Tommy! I raised a PR for this.

brandonkelly:
Craft 5.7.7 is out with that fix.

PR #17272, opened 2025-05-15
put the label-link back into craft-element-label before disconnecting

i-just:
### Description
Fixes an issue where moving an element with an overflowing label to a new position (via dragging or disclosure menu) would cause the label to disappear after the move.

### Related issues
#17270


i-just:
Info that might be helpful in the future:
Using [Element.moveBefore()](https://developer.mozilla.org/en-US/docs/Web/API/Element/moveBefore) as mentioned [here](https://developer.mozilla.org/en-US/docs/Web/API/Web_components/Using_custom_elements#lifecycle_callbacks_and_state-preserving_moves) instead of `$element.insertBefore()` works like a charm in Chrome, but it’s not supported in Firefox or Safari, so it’s not an option at this time.

PR #17277, opened 2025-05-17
“is of type” Link field condition operator

brandonkelly:
### Description

Adds an “is of type” operator to Link fields’ condition rules.

An entry condition builder, with a condition rule for a Link field. Its operator is set to “is of type”, and there’s a select input with “Entry” selected.

Couple things to note:

- Existing Link field condition rules won’t have the new operator, because the underlying condition rule class has changed. Those rules will need to be removed and re-added to get it.
- Only elements that have been saved on Craft 5.5.0+ will match “is of type” Link rules. To ensure all Link field data is stored in the updated format, run the following command, setting `--with-fields` to a comma-delimited list of your Link field handles:
  ```sh
  php craft resave/all --with-fields=foo,bar,baz
  ```

### Related issues

- #17276

### DIFF
--- a/CHANGELOG-WIP.md
+++ b/CHANGELOG-WIP.md
@@ -2,6 +2,7 @@
 
 ### Content Management
 - Matrix fields set to the “inline-editable blocks” view mode now have “Expand all blocks” and “Collapse all blocks” actions. ([#17141](https://github.com/craftcms/cms/pull/17141))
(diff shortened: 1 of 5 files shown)


darylknight:
Does this have to be run just locally or also on the live site?

brandonkelly:
@darylknight It would need to be run on production as well (or you could restore a local backup where you ran it).

Issue #17278, opened 2025-05-17
[5.x]: Allow querying for single SectionEntryUnion in GraphQL


denisyilmaz:
### Description
This is a suggestion to add the ability to query for single SectionEntryUnion entries via GraphQL. 

The reason behind this is that we love to use the `[SectionEntryUnion]` query as it has beneficial type inference for typescript projects compared to the `[EntryInterface]` query. 

Unfortunately for "single" section types we currently have to unwrap the array just to get a single entry variable to work with like this: 
```ts
const homepage = this.page()?.data?.homepageEntries?.at(0)
``` 

The linked commit adds the missing `EntryResolver::resolveOne` to the `sectionLevelFields()` function, but I'm not sure if this will be enough to get this feature implemented:

https://github.com/denisyilmaz/cms/commit/c20d49f460f81c65a92f2ebc96571875cd539739

It would allow us to query like this:

```gql
query GetHomepage {
  homepageEntry {
    __typename
  }
}
```

### Suggested Changes
```php

foreach (GqlHelper::getSchemaContainedSections() as $section) {
            $singleName = "{$section->handle}Entry";
            $singleTypeName = "{$section->handle}SectionEntryQuery";
            $singleSectionQueryType = GqlEntityRegistry::getEntity($singleTypeName);

            if (!$singleSectionQueryType) {
                $entryTypesInSection = [];

                // Loop through the entry types and create further queries
                foreach ($section->getEntryTypes() as $entryType) {
                    if (isset($entryTypeGqlTypes[$entryType->id])) {
                        $entryTypesInSection[] = $entryTypeGqlTypes[$entryType->id];
                    }
                }

                if (empty($entryTypesInSection)) {
                    continue;
                }

                $arguments = EntryArguments::getArguments();

                // Unset unusable arguments
                unset(
                    $arguments['section'],
                    $arguments['sectionId'],
                    $arguments['field'],
                    $arguments['fieldId'],
                    $arguments['ownerId'],
                );

                foreach ($section->getEntryTypes() as $entryType) {
                    $arguments += $gqlService->getFieldLayoutArguments($entryType->getFieldLayout());
                }

                // Create the section query field
                $singleSectionQueryType = [
                    'name' => $singleName,
                    'args' => $arguments,
                    'description' => sprintf('Entry within the “%s” section.', $section->name),
                    'type' => Type::listOf(GqlHelper::getUnionType("{$section->handle}SectionEntryUnion", $entryTypesInSection)),
[...]

PR #17279, opened 2025-05-17
feat(gql): allow querying of single SectionEntryUnion

denisyilmaz:
### Description
This is a suggestion to add the ability to query for single SectionEntryUnion entries via GraphQL. The reason behind this is that we love to use the `[SectionEntryUnion]` query as it has beneficial type inference for typescript projects compared to the `[EntryInterface]` query. 

Unfortunately for section types "single" we have to unwrap the array just to get a single entry variable to work with like this: `const homepage = this.page()?.data?.homepageEntries?.at(0)`. This commit adds the missing `EntryResolver::resolveOne`. 

### Related issues
#17278

### DIFF
--- a/src/gql/queries/Entry.php
+++ b/src/gql/queries/Entry.php
@@ -97,6 +97,53 @@ class Entry extends Query
         $gqlService = Craft::$app->getGql();
 
         foreach (GqlHelper::getSchemaContainedSections() as $section) {
+            $singleName = "{$section->handle}Entry";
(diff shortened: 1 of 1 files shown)


brandonkelly:
Thanks for the PR! I refactored the surrounding code a bit in 3b56cbef210dc2fa109974655ef310915a7c542b / 21e009c6ab00bce30beaab57877bd70c60d130f9 first, which made implementing `<section>Entry` queries a lot simpler (73da221c636286cd4b7e2917a20ad84b96718444).

Issue #17290, opened 2025-05-20
[5.x]: Filtering entries by same field with different handle does not discriminate


leonardigabriele:
### Description

In Craft 5 it is possible to use, in the same entry type, the same field definition multiple times changing Label and Handle. This is very handy not having to create a number of fields which have the same shape just to use it more than once in the entry.
I have an entry type with 3 fields of type "Assets" and original handle "file", with different final handles, let's say "file1", "file2", file3" for the sake of simplicity. 
What happens is that if I filter the entry type by "file1" > "is empty" only the entries where ALL 3 FIELDS ARE EMPTY get returned. The fields get treated as they are one in the context of the filters.
I think this is quite a severe bug which should be fixed as soon as possible, at the moment we are unable to provide the client layouts with the correct data because the bug extends to queries too.

### Steps to reproduce

1. Create "Assets" field
2. Create Entry type with the field used multiple times with changed handle. es. "file1", "file2", "file3"
3. Create some entries with some of the fields populated
4. Filter entries in the CMS UI or doing queries targeting when one of the fields "is empty"

### Expected behavior

I should get the entries where that specific field is empty.

### Actual behavior

I only get entries where ALL FIELDS with the same original field definition are empty.

I tested this with "Assets" type fields but I don't know if the bug extends to other field types, worth a check.

### Craft CMS version

5.7.5

### Installed plugins and versions

-

brandonkelly:
Thanks for reporting that! I was able to fix, however the fix requires that any elements which have multi-instance relation fields created before Craft 5.3.0 be resaved. So I made the fix for Craft 5.8.0, with a note about that (#17295).

PR #17295, opened 2025-05-20
Better condition support for multi-instance relation fields

brandonkelly:
### Description

Improves condition rules for multi-instance relation fields, so they factor in the individual instance data, rather than the global field as a whole.

> [!NOTE]
> Any elements that have multi-instance relation fields, which haven’t been saved since Craft < 5.3.0, will not be matched accurately by relation conditions. To ensure all relation field data is stored in the updated format required by relation conditions going forward, run the following command, setting `--with-fields` to a comma-delimited list of your relation field handles:
>
> ```sh
> php craft resave/all --with-fields=foo,bar,baz
> ```

### Related issues

- #17290

### DIFF
--- a/CHANGELOG-WIP.md
+++ b/CHANGELOG-WIP.md
@@ -3,6 +3,9 @@
 > [!NOTE]
 > Elements with Link fields created before Craft 5.5.0 should be resaved to take advantage of the new “is of type” condition rule operator. ([#17277](https://github.com/craftcms/cms/pull/17277))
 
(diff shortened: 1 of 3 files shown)


peteeveleigh:
@brandonkelly could you clarify what "Multi instance relations fields" means? Would that be an Entries field type that allows more than one related Entry?

brandonkelly:
@peteeveleigh As in, relation fields that have been added to the same field layout multiple times.

PR #17313, opened 2025-05-23
Feature/cms 1418 delete entries for site delete other users entries for site

i-just:
### Description
Adds “Delete entries for site” and “Delete other users’ entries for site” permissions (in addition to the existing “Delete entries” and “Delete other users’ entries”). Those are now used when determining if a user should see and be allowed to delete an entry for the site via the element index page or entry edit page.

There’s a migration that automatically assigns the new permissions to any users/groups that already have “Delete entries”/“Delete other users’ entries” permissions.

### Related issues
[CMS-1418](https://linear.app/craftcms/issue/CMS-1418/delete-entries-for-site-delete-other-users-entries-for-site)


src/base/Element.php
```diff
@@ -4160,12 +4160,7 @@ protected function destructiveActionMenuItems(): array
         // Permissions
         $canDeleteDraft = $isDraft && !$this->isProvisionalDraft && $elementsService->canDelete($this, $user);
         $canDeleteCanonical = $elementsService->canDelete($canonical, $user);
-        $canDeleteForSite = (
-            $isMultiSiteElement &&
-            count($propSiteIds) > 1 &&
-            (($isCurrent && $canDeleteCanonical) || ($canDeleteDraft && $isNewSite)) &&
-            $elementsService->canDeleteForSite($this, $user)
-        );
```
brandonkelly: These conditions are still important – otherwise entries with only one site will still get the “Delete for site” action.

src/services/UserPermissions.php
```diff
@@ -558,17 +558,35 @@ private function _entryPermissions(array &$permissions): void
                                     'type' => $pluralType,
                                 ]),
                             ],
+                            "deleteEntriesForSite:$section->uid" => [
```
brandonkelly: The new user permissions should only be added for sections with the propagation method set to `custom`.

brandonkelly (changes requested):

src/base/Element.php
i-just: That’s now adjusted. Plus, I aligned the delete for site element index action with this a bit more.

src/services/UserPermissions.php
i-just: done

PR #17316, opened 2025-05-23
Feature/draft validation

jamie-s-white:
### Description
This PR adds the "validate drafts" action to the hamburger menu for elements:

Screenshot 2025-05-23 at 15 39 44

This allows for admins to fully validate an element that is not enabled, or to more easily see validation errors at the top of the page before they commit to saving and publishing their updates:

Screenshot 2025-05-23 at 15 41 59

### Related issues
We had an issue recently where we changed alt text on images to be mandatory, and our client was then tripped up when trying to publish updates because the contextual error messages for related assets was not clear for them until they tried to publish the entry.

We wanted to add a "half way house" where they could see these errors in a much more usable and accessible fashion without needing to commit to a full save or publishing of the element.


brandonkelly:
Thanks for the PR! Made a few changes and merged into the `5.8` branch (c6ec0c0a01b18c35641fab512c01a92e2313a083).

brandonkelly:
Craft 5.8.0 is out with that action. Thanks again!

Issue #17337, opened 2025-05-27
[5.x]: Password reset has no rate or count limit


sebastian-lenz:
### Description

The password reset function has no rate limit and therefore can be exploited, e.g. to cause work loads on the server or send spam to users.

### Steps to reproduce

1. Visit the control panel login page
2. Click on "Forgot password"
3. Enter a username or email-address
4. Reload and immediately repeat step 1

### Expected behavior

The system should prevent the user from immediately retriggering the reset password mail. 

### Actual behavior

The system sends without any rate or count limits reset password mails.

### Craft CMS version

5.7.7

### PHP version

8.3.21

### Operating system and version

Linux 6.8.0-59-generic

### Database type and version

MySQL 8.0.42

### Image driver and version

Imagick 3.8.0 (ImageMagick 6.9.12-98)

sebastian-lenz:
As Craft stores a timestamp whenever a verification token is generated on the user record we can actually pretty easily implement a cooldown timer that prevents the reset password mail from being sent repeatedly. I've created a fork that adds a cooldown setting and implements a simple check to prevent the missuse.

bastiaanwuisman:
We've got a client also running into this issue. Any chance this will be picked up in an upcoming update? If not, we'd probably have to implement something similar to the linked PR in our project. Thanks!

brandonkelly:
Craft 4.17.0 and 5.9.0 are out, which now impose a rate limit on `users/send-password-reset-email` requests.

PR #17338, opened 2025-05-27
Add a password reset cooldown

sebastian-lenz:
### Description

Add a password reset cooldown to prevent the possibility of exploiting the password reset function

### Related issues

https://github.com/craftcms/cms/issues/17337

### DIFF
--- a/src/config/GeneralConfig.php
+++ b/src/config/GeneralConfig.php
@@ -2027,6 +2027,20 @@ class GeneralConfig extends BaseConfig
     public string $partialTemplatesPath = '_partials';
 
     /**
+     * @var int The duration in seconds of the cooldown timer after a reset password mail has been sent.
+     *
+     * Prevents other from exploiting the reset password functionality by applying a phase within it is not possible to
+     * trigger another password reset mail. Set the duration to zero to disable the cooldown.
+     *
+     * ::: code
+     * ```php Static Config
+     * ->passwordResetCooldownDuration(300)
+     * ```
+     * :::
+     */
+    public int $passwordResetCooldownDuration = 60;
+
+    /**
      * @var string|null The query string param that Craft will check when determining the request’s path.
      *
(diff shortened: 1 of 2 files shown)


brandonkelly (changes requested):
Couple issues with this:

- I don’t think we should be repurposing `verificationCodeIssuedDate` for this.
- It could introduce a new user enumeration vector, as the way it’s coded, you’d get an exception thrown immediately whenever you submit a reset-password form for a valid username/email multiple times, but not for invalid usernames/emails.

Given the second point, it might make more sense to simply set a rate limit on the `users/send-password-reset-email` action, regardless of which username/email is provided.

sebastian-lenz:
Hi @brandonkelly — thanks for looking into this!

It seems like we're considering two different approaches here, so it would be helpful to know which direction you'd prefer before I update my pull request.

> I don’t think we should be repurposing `verificationCodeIssuedDate` for this.

My intention was to avoid introducing a new database column just to track password resets. `verificationCodeIssuedDate` is currently only set via `Users::setVerificationCodeOnUser`, which is only called by `Users::getEmailVerifyUrl` and `Users::getPasswordResetUrl`. That’s why I thought it would be safe to reuse it for implementing a cooldown timer. That said, if you'd rather keep it separate, I’m happy to add a dedicated column instead.

> It could introduce a new user enumeration vector [...]

I've updated the pull request so that, when the cooldown is active, it now fails silently instead of throwing a new exception. I also added an error boundary around the call to `Users::_getUserRecordById` to ensure no internal details are leaked. Unless `preventUserEnumeration` is disabled, I don’t believe this introduces a new enumeration vector — but let me know if you see something I missed.

> [...] it might make more sense to simply set a rate limit on the `users/send-password-reset-email` action [...]

That’s a good idea too. From what I can tell, Yii doesn’t provide built-in support for request rate limiting, so we’d likely need to implement it ourselves. If that’s the preferred route, I’d be glad to work on a separate PR for that.

PR #17339, opened 2025-05-27
Feature/copy link field urls

brianjhanson:
### Description

Adds a "Copy URL" action item to text link fields. The main goal is to give users an easier way to see full URLs (https://github.com/craftcms/cms/pull/17261). We initially thought about truncating the text and providing a tooltip as we do with element chip titles, but that had some issues with accessibility. After playing with various methods of avoiding truncation in these link fields, this felt like the best option. 

CleanShot 2025-05-27 at 11 01 29@2x

The "Copy URL" option will open up the standard copy text dialog. 

CleanShot 2025-05-27 at 11 01 31@2x

### Related issues
https://github.com/craftcms/cms/pull/17261#issuecomment-2887446926


brandonkelly:
Looks great, thanks!

PR #17358, opened 2025-05-28
Updates Image Editor so that cropper move and resize are accessible for SR and keyboard users

gcamacho079:
### Description
Adds visually hidden buttons to the Image Editor that allow screen reader and keyboard users to resize and move the cropping rectangle.

Still outstanding (but planned for a future release):
- Visual UI for allowing users to resize and move the cropping rectangle without using a dragging motion.


brianjhanson (approved):
Amazing! Even as someone who doesn't navigate via keyboard very often I found it pretty intuitive to use. I do kind of wish there was a way to move the handles by a larger increment if I wanted to. Typically I'd expected something like shift + arrow to do that, but I noticed doing that will constrain the crop to the current aspect ratio instead (in the "unconstained" mode). 

In the end, I felt like I was able to move the handles around fast enough that it wasn't a huge issue, but at the same time I could see it being useful.

gcamacho079:
> I do kind of wish there was a way to move the handles by a larger increment if I wanted to. Typically I'd expected something like shift + arrow to do that, but I noticed doing that will constrain the crop to the current aspect ratio instead (in the "unconstained" mode).

Thanks Brian! I agree that would be a great feature but I also got stuck at the stage of realizing it would interfere with the aspect ratio part of things. I'll brainstorm some ways we can make that happen in the future.

PR #17365, opened 2025-05-29
Add key to `registerHtml`

brianjhanson:
Ensures the `fetch` call will only happen once per page load


AugustMiller:
Are we seeing this happening in the wild? 😮 It's typically safe to use `{% js %}`, `{% css %}`, `{% html %}`, and the underlying `craft\web\View` "register" methods with the same strings multiple times… in each case, the fragments are held under `md5` hashes of their content, which I would expect to be identical!

brianjhanson:
It looks like you're right @AugustMiller! I didn't realize the register methods were so smart.

PR #17370, opened 2025-05-30
Upgrade PHPStan (and Rector) to v2

riasvdv:


bootstrap/console.php
```diff
@@ -12,6 +12,7 @@
 ini_set('display_errors', 'stderr');
 
 // Make sure they're running PHP 8+
+/** @phpstan-ignore-next-line */
 if (PHP_VERSION_ID < 80002) {
```
riasvdv: Is this check still needed as Craft requires PHP 8.2 through composer?

bootstrap/web.php
```diff
@@ -8,6 +8,7 @@
  */
 
 // Make sure they're running PHP 8+
+/** @phpstan-ignore-next-line */
 if (PHP_VERSION_ID < 80002) {
```
riasvdv: Still needed?

src/base/Element.php
```diff
@@ -6180,7 +6181,10 @@ protected function statusFieldHtml(): string
     protected function notesFieldHtml(): string
     {
         // todo: this should accept a $static arg
-        /** @var static|DraftBehavior $this */
+        /**
+         * @var static|DraftBehavior $this
+         * @phpstan-ignore varTag.nativeType
```
riasvdv, line 6225: PHPStan seems to have issues with recognizing behaviors and combining the types. This check is [a v2 addition](https://github.com/phpstan/phpstan-src/commit/a69e3bc2f1e87f6da1e65d7935f1cc36bd5c42fe) so I've opted to ignore it where it causes issues currently instead of a global ignore

src/base/LogTargetTrait.php
```diff
@@ -22,6 +22,7 @@
  * @mixin Target
  * @author Pixel & Tonic, Inc. <support@pixelandtonic.com>
  * @since 3.1.23
+ * @phpstan-ignore trait.unused
  */
 trait LogTargetTrait
```
riasvdv, line 28: Is this used somewhere?

bootstrap/console.php
brandonkelly: Nope, that’s safe to remove.

bootstrap/web.php
brandonkelly: Nope

src/base/LogTargetTrait.php
brandonkelly: Doesn’t seem to be anymore in core, but possibly still in use by Craft installs which are still using a non-Monolog-based log target. (see #4127 for context on why it was added.)

Safe to deprecate in 5.8 and remove in v6 though, I’d say. Everyone should be using Monolog now – right @timkelty?

PR #17391, opened 2025-06-03
Update how we draw column separators

### LINKED ISSUE #17389: [5.x]: Styling Issue Markdown in Matrix Block
### What happened?

Found a minor styling issue: There appears to be a strange border-left on UI elements within a matrix block.

Image

### Craft CMS version

5.7.8.1

### PHP version

_No response_

_No response_

### Database type and version

_No response_

### Image driver and version

_No response_

-

brianjhanson:
Updates how we draw column separators when using a multi-column field layout. With this change, we'll only draw it when two elements with classes starting with `width-` are next to each other. 

Fixes https://github.com/craftcms/cms/issues/17389

### DIFF
--- a/src/web/assets/cp/src/css/_cp.scss
+++ b/src/web/assets/cp/src/css/_cp.scss
@@ -1002,12 +1002,6 @@ li.breadcrumb-toggle-wrapper {
   grid-template-columns: var(--page-title-columns);
   align-items: center;
   gap: var(--s);
-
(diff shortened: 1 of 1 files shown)


brandonkelly:
Resolved by hiding the borders entirely within Matrix blocks, for all elements that the border is initially added in by `_cp.scss`. (cc073176ad9f65f8194607706b5554c339261a42)

PR #17413, opened 2025-06-09
Namespaces describedBy and labelledBy used in autosuggest

gcamacho079:
### Description
Fixes an issue where the IDs referenced weren’t present on the page due to a lack of namespacing


gcamacho079:
@brandonkelly because we’re having to use this method for more than input IDs — and sometimes arrays of IDs as well — I wonder if it’s worth (a) updating the namespacing method to `namespaceId` and (b) adding a method that takes multiple IDs. 🤔

brandonkelly:
@gcamacho079 Where are you seeing this? Multiple space-separated IDs within `aria-describedby` and `aria-labelledby` _should_ already be getting namespaced properly, via `craft\helpers\Html::namespaceHtml()` / `namespaceAttributes()`.

https://github.com/craftcms/cms/blob/372fd6cda65088d153dfcf0a0838dbedcd0c142f/src/helpers/Html.php#L896-L925

gcamacho079:
@brandonkelly one place I'm seeing this is the "Base Path" field in the Filesystem settings. In the screenshot below, the `aria-labelledby` and `aria-describedby` IDs don't have a namespace:

Screenshot of the Base Path field highlighted in DevTools, with aria-describedby and aria-labelledby attributes that don't match the IDs of the label or instructions

brandonkelly:
Ahh, I see, those properties get passed to the Vue component via JS, so not actually present in the HTML that gets sent to `Html::namespaceHtmlHtml()`.

Issue #17404, opened 2025-06-05
[5.x]: `Actionable` Menu Items are Leaky


AugustMiller:
### Description

Fields can contribute action menu items that should appear in field layouts; the Matrix field uses this to provide a **Copy all entries** item.

However, `Field`s themselves are `Actionable`, meaning when they’re rendered as a chip, the field layout element-destined actions can appear out-of-context:

![Image](https://github.com/user-attachments/assets/605ba664-77ab-4573-95f4-e3dd9eab0ade)

I believe this only _currently_ manifests with Matrix fields… but it points to a broader design/intent issue with `getActionMenuItems()`.

### Steps to reproduce

1. Create a section and entry type;
2. Create a Matrix field and add it to the field layout;
3. Open the **Settings** slideout from the field layout element;
4. Open the action menu on the field chip at the top of that view;

### Expected behavior

Only actions intended for use with the field (**Field settings** and **Copy field handle**) are shown.

### Actual behavior

All items returned by the field are shown.

### Craft CMS version

5.7.10

### PHP version

8.2.28

### Operating system and version

macOS / DDEV

### Database type and version

MySQL 8.x

i-just:
Thanks August! PRs raised!

brandonkelly:
Fixed in 5.7.11!

PR #17415, opened 2025-06-10
getActionMenuItems and actionMenuItems can be dependent on an element

i-just:
### Description
Allow `$element` and `$static` to be passed to the field’s `getActionMenuItems()` and `actionMenuItems()` so that action menu items can be added depending on whether the field is in an Element context or not.

As it’s potentially a breaking change, we might actually have to target v6?

### Related issues
#17404
More info & context: https://github.com/craftcms/cms/pull/17416

### DIFF
--- a/src/base/Field.php
+++ b/src/base/Field.php
@@ -542,9 +542,9 @@ abstract class Field extends SavableComponent implements FieldInterface, Iconic,
     /**
      * @inheritdoc
      */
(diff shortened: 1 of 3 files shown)


brandonkelly:
> As it’s potentially a breaking change, we might actually have to target v6?

Yeah, this needs to wait.

PR #17424, opened 2025-06-11
Content Block field

brandonkelly:
### Description

Adds a new “Content Block” field type, which provides a nested field layout that can be pulled into other element types’ field layouts.

A field layout designer for a Content Block field, with “Image”, “Is decorative?” and “Alt Text” fields

A Content Block field’s input, showing the three fields it was configured with

There’s also a “View Mode” setting, which determines how the content block layout should be presented within parent field layouts:

The View Mode setting, with options “Grouped”, “In a pane”, and “Inline”

When “Inline” is selected, nested fields will appear right alongside other fields in the parent field layout, without any obvious indication that they’re associated with a nested element.

### Development

To access field values within a Content Block field, prefix them with the Content Block field’s handle:

```twig
{{ entry.myContentBlockField.myNestedField }}
```

Ditto for GraphQL:

```graphql
...on myEntryType {
  myContentBlockField {
    myNestedField
  }
}
```

### Related issues

- #14015
- #14600
- #15626
- #17133
- #15699


FalkoJoseph:
This is gonna be great! 🎉

trijammer:
@brandonkelly I love this, thank you. When retroactively applying grouping to a bunch of existing fields (or matrix fields) used by many entries containing data already, is there any straightforward way to migrate the data?

brandonkelly:
@trijammer You can copy top-level field values into a Content Block field using `resave/*` commands, e.g.:

```sh
php craft resave/entries --set myContentBlockField \
  --to "fn(\$entry) => ['fields' => ['subFieldA' => \$entry->fieldA, 'subFieldB' => \$entry->fieldB]]"
```

Or if the data is already nested within a Matrix field with Min/Max Entries set to `1`:

```sh
php craft resave/entries --set myContentBlockField \
  --to "fn(\$entry) => ['fields' => ['subFieldA' => \$entry->matrixField[0]->fieldA ?? null, 'subFieldB' => \$entry->matrixField[0]->fieldB ?? null]]"
```

BenHMatrix:
Question not specifically related to the Content Block field, but @brandonkelly is there a way to have the native alt text field display alongside Asset fields without having to do something like like the first example here? It would mean my content editors are a lot more likely to fill out the alt text field if it was right there and they didn't have to go to the asset.

brandonkelly:
@BenHMatrix Yep, you can choose to have assets’ alt text included in their card view (from the volume’s settings), and set your Assets field to show the cards view.

samuel-clara:
With multiple "Content blocks" support for the same entry this would be perfect ❤️

trijammer:
> With multiple "Content blocks" support for the same entry this would be perfect ❤️

+1!
It would be great, to avoid another layer of matrix blocks + entry types.

PR #17426, opened 2025-06-11
Generated fields

brandonkelly:
### Description

Adds the ability to define “generated fields” alongside field layouts. Generated fields are special element values that are generated on element save, based on a given Twig template.

A “Generated Fields” table, with an “Alt” field defined.

Generated fields can be accessed just like custom fields from an element, via their handle, both within Twig and GraphQL. They also get their own element query and GraphQL query arguments, condition rules, and card attributes.


thupsi:
Nice! Would it be possible for generated fields to be selectable in the Card Layout Editor? It's the one feature of the Preparse plugin I would miss!

khalwat:
https://github.com/nystudio107/craft-code-editor :)

brandonkelly:
> Nice! Would it be possible for generated fields to be selectable in the Card Layout Editor? It's the one feature of the Preparse plugin I would miss!

Yep, added that today :)

PR #17447, opened 2025-06-15
Wrap plain text preview in .code class if monospace

leevigraham:
### Description

"Monospace" `PlainText` fields display the input with the `code` class. 

Card previews and table cell values should also reflect the `monospace` config.

image

image

### Related issues

https://github.com/craftcms/cms/pull/2636


brandonkelly:
Thanks! Merged for 5.8.

PR #17467, opened 2025-06-19
Translate prefix and suffix in Number input.twig

7ochem:
### Description

If I add a suffix in the settings of a number field, let's say "items", when I want the user to enter a number of "items", then this is not translated. In my case, the admin user has set the language to Dutch. Now the field name and instructions for instance are all translated to Dutch when I add those to the `translations/nl/site.php` translations file. However any postfixes or suffixes are not translated and remain in English, just the way I entered it. In my opinion this should also be translatable so that all the texts involved in the form are in the preferred language of the user.

I'm targeting the 4.x branch here as I'm encountering this in a 4.x setup. I hope you can merge this forward to 5.x too.

### Related issues

None

### DIFF
--- a/src/templates/_components/fieldtypes/Number/input.twig
+++ b/src/templates/_components/fieldtypes/Number/input.twig
@@ -16,8 +16,8 @@
         id: descriptionId,
         class: 'visually-hidden',
         text: [
-            hasPrefix ? 'Value prefixed by “{prefix}”.'|t('app', {prefix: prefix}),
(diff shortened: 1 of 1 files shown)


brandonkelly:
Thanks for the PR! Merged for the next Craft 4 and 5 releases.

brandonkelly:
Craft 4.15.7 and 5.7.11 are out with that fix. Thanks again!

Issue #16443, opened 2025-01-16
[5.x]: Nested entries missing from older revisions


rungta:
### Description

Older revisions appear to be missing some nested entries (via a matrix field).

### Steps to reproduce

We have not been able to consistently reproduce this behavior, but we have at least one instance of this on a live production site where a revision created post upgrading to Craft 5.5.8 appears to be missing (most of) its nested entries (via a matrix field).

There were also a large number of “Resaving entries” queue tasks running in parallel (due to the upgrade from 5.3.6 to 5.5.8) so unsure if this bug occurred due to some race condition between the resaving task and a user-initiated entry save operation.

While investigating this issue we noticed that revisions created on older versions of Craft (5.3.6) were also missing some nested entries, but reporting it as a new bug because this behavior still seems to be present in 5.5.8.

### Expected behavior

Older revisions retain all the nested entries that belonged to the matrix field when the revision was created.

### Actual behavior

Certain nested entries seem to be missing.

1.  Revision 57 (screenshot from a backup taken at 2025-01-12 12:30pm)
    ![Image](https://github.com/user-attachments/assets/51b210f7-bf8f-4c9e-96c3-60b166fe3f19)

2. Same revision 57 is missing the fourth nested entry (Our Voices, Upcoming Events) after a few edits were made. (screenshot from a backup taken one hour later at 2025-01-12 1:30pm)
    <img width="980" alt="Image" src="https://github.com/user-attachments/assets/379cd884-2986-486d-8614-c0c2ea109b3f" />

3. Same revision 57 is missing 5 of the 6 nested entries after a few more edits were made, and the entry was also restored at one point from revision 60. (screenshot from current state on production)
    <img width="980" alt="Image" src="https://github.com/user-attachments/assets/c68b9527-c7c7-4de4-8137-10ad556fae78" />

### Craft CMS version

5.5.8

### PHP version

8.2.27

### Operating system and version

Ubuntu 22.04 LTS

### Database type and version

MariaDB 10.11.10

### Image driver and version

Imagick 3.7.0 (ImageMagick 6.9.11-60)

### Installed plugins and versions

- Calendar Links 	4.0.0
- Child Me! 	2.2.3
- CKEditor 	4.4.0
- Colour Swatches 	5.1.0
- Element Index Defaults 	4.0.0
- Formie 	3.0.16
- Hyper 	2.2.2
- iCalendar 	5.0.0
- Image Resizer 	4.0.1
- Imager X 	5.1.0
[...]

PR #17473, opened 2025-06-20
force create new nested element revisions when matrix has enableVersioning set to false

i-just:
### Description

Steps to reproduce:
- set `->maxRevisions(3)`
- create a matrix field in e.g. cards mode, revisions disabled for the field (default)
- create a section with an entry type that uses the matrix field from the step above
- create an entry in that section; fully save & continue editing
- add one nested entry (title “one”) - save (& continue editing)
- add one nested entry (title “two”) - save (& continue editing)
- add one nested entry (title “three”) - save (& continue editing)
- view the latest revision (revision 3) - it correctly has nested entries “one” and “two”
- add another nested entry (title “four”) - save (& continue editing)
- view the latest revision (revision 4) - it only has nested entries “two” and “three”; nested entry “one” is missing

The nested elements aren’t duplicated for every revision, so the same nested element can belong to revisions 1, 2 and 3.

When there are enough revisions to trigger the `PruneRevisions` job and the owner is being deleted, nested entries that belong to it are hard deleted.

Let's say you have:
- current entry with:
    - owner “test” (id: 791)
    - nested entry: “one”, (id: 795)
    - nested entry: “two”, (id: 800)
    - nested entry: “three” (id: 805)
- and revisions:
    - rev 2: with nested entry: “one” (id: 797)
    - rev 3: with nested entries: “one” (id: 797) and “two” (id: 802)
- at this point, everything works as expected
- you add a nested entry: “four” (id: 810) & save the owner
- rev 4 is created and `PruneRevisions` job deletes rev 2; rev 2 has nested entry “one” (id: 797) in it, so it’s hard deleted too, which causes it to be missing from rev 3 and rev 4

**Solution:**
When the Matrix field has `enableVersioning` set to `false`, ensure that each owner revision has its own set of nested entries.

Tested with and without `enableVersioning` with cards and element index view modes.
(This solution doesn’t work retrospectively.)

### Related issues
#16443

### DIFF
--- a/src/elements/NestedElementManager.php
+++ b/src/elements/NestedElementManager.php
@@ -23,6 +23,7 @@ use craft\enums\Color;
 use craft\enums\PropagationMethod;
 use craft\events\BulkElementsEvent;
 use craft\events\DuplicateNestedElementsEvent;
+use craft\fields\Matrix;
 use craft\helpers\ArrayHelper;
 use craft\helpers\Cp;
 use craft\helpers\Db;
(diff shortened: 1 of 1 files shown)


brandonkelly:
Fixed this via 1b6bf1e742175e76f362373dede2e2a1573ae11d instead, which just checks whether nested revisions belongs to any other owners before deleting them, and if so, reassigns them instead.

PR #17497, opened 2025-06-26
Inline search + element creation for relational fields

brandonkelly:
### Description

Relational fields with a single element source now have a “Show the search input” setting (enabled by default). When enabled, their inputs will get a search input for quickly selecting elements without having to go through the element selection modal.

An Assets field with the new search input, with “a” entered into it, and showing five search results.

Entries fields with a single section source _and_ without a selection criteria condition also support inline entry creation.

An Entries field with “Content Strategy” entered into the search input, showing a “+ Content Strategy” option, indicating that the entry doesn’t exist but can be created.

### Related issues

- #11379
- #12721
- #13174


AndrewMac:
Hi @brandonkelly, just checking: I can get the search field to appear if I have a single source selected, or if I have "Restrict assets to a single location" enabled, but in that latter case I can't seem to get the search field to appear if I also have "allow subfolders" checked. Is that the expected behaviour? That's fine if it is; just wondering, that's all …

brandonkelly:
@AndrewMac It’s showing up when “Restrict assets to a single location is enabled on my end.

Is your field’s “Show the search input” setting enabled?

brandonkelly:
@AndrewMac Sorry nevermind, that is in fact a bug in in 5.8.0, fixed for 5.8.1 via #17548.

AndrewMac:
Thanks so much! I know you probably get this all the time, but the way you and your team respond to issues and get fixes in place so quickly is one of the reasons I so much want to stick with Craft CMS. Of course, the amazing dedication to continually improving things is also another ;-)

AndrewMac:
Hi @brandonkelly, this is generally working now, but I have a few fields where it still fails. 
The conditions that cause the search box to disappear are:

Restrict assets to a single location: TRUE
AND path/to/subfolder is set to a Twig variable

It works if path/to/subfolder is empty, or is a simple string
It's when I use a Twig variable that the problems occur

Again, I'm not sure if this is something that can be fixed or not, but it would be great if it could be!

brandonkelly:
@AndrewMac The only way I can reproduce that is if either: A) the subpath can’t be determined yet (e.g. it’s set to `{title|kebab}` and there’s no title yet; or B) no folder exists for the dymanic subpath yet.

In both cases, it’s expected that the search input won’t show because we either don’t know where to search yet, or there’s no folder, so obviously no assets to be searched yet.

AndrewMac:
Thanks @brandonkelly, I had another look and found that the problem is fixed when I use the double-brace {{ currentSite.group }}, which works, instead of the single-brace { currentSite.group }, which doesn't.

Interestingly enough, I also found that the single-brace version does work if "Restrict assets to a single location" is unchecked with only one source selected. But I can get what we need with the double-brace anyway, so no issue.

My fault, and sorry for taking up your time with this!

brandonkelly:
Ah yeah – the single brace syntax is short for `{{ object.______ }}` (where `object` is the element being saved). But `currentSite` is a global variable, not a property of the element.

`{site.group}` would work though, and is probably closer to what you actually want.

PR #17504, opened 2025-06-27
Show safe errors on GQL interface

marten-wirelab:
`\GraphQL\Error\UserError` [explicitly states](https://github.com/webonyx/graphql-php/blob/master/src/Error/UserError.php) that it is safe to show to the client, so we don't need to replace it with a generic error message, even if dev mode is disabled

### Description
On one of our platforms, we let an external party develop an integration that will be using the GQL interface. It would greatly improve their developer experience if validation errors are shown instead of the generic error message that is now shown when dev mode is disabled. 
This PR accomplishes that goal by only exposing client-safe errors through the interface.

### DIFF
--- a/src/services/Gql.php
+++ b/src/services/Gql.php
@@ -1296,7 +1296,8 @@ class Gql extends Component
             }
 
             // If devMode enabled or exception is safe to show, substitute the original exception here.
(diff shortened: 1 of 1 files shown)


brandonkelly:
Thanks for the PR! Merged for Craft 5.8.

brandonkelly:
Craft 5.8.0 is out with that change. Thanks again!

PR #17509, opened 2025-06-27
Collapsible header adjustments

brianjhanson:
Adjustments to the collapsible sources styling based on [some feedback](https://github.com/craftcms/cms/pull/17226#issuecomment-3009837073). 

At the moment there's a slight regression in that we no longer show nested sources when the parent is collapsed. The bug pointed out by @mmikkel exposed some flaws in my hiding logic. 

| Before | After |
|:-----|:-----|
| CleanShot 2025-06-27 at 16 48 24@2x | CleanShot 2025-06-27 at 16 48 24@2x |


mmikkel:
Nice work @brianjhanson 🎉 

FYI, I'm still seeing the bug described in https://github.com/craftcms/cms/pull/17226#issuecomment-3009837073, where headers without any enabled sources beneath them are visible:

5.7.11 (only headings with enabled sources show):

CleanShot 2025-06-30 at 20 22 52

5.8.x-dev (site "Swedish" has no enabled sources below the "News" heading, but it still displays):

CleanShot 2025-06-30 at 20 28 23

PR #17574, opened 2025-07-08
Fixed PHPStan error related to `UrlManager`

mspirkov:
### Description

I see that the UrlManager is being overridden for the base Yii application. Therefore, I moved the PHPDoc annotations from the Web Application to the `ApplicationTrait`


brandonkelly:
Thanks for the PR!

PR #17591, opened 2025-07-10
Stop hyperlinking chips & cards by default

brandonkelly:
### Description

Updates chips and cards to be not hyperlinked by default, except on element indexes.

### Related issues

- #17080


thupsi:
Hi @brandonkelly ! I don't know if this is on purpose but this also removed the hyperlinking of titles in "Index" view mode for matrix fields, where IMO it makes _much more sense_ for the entry titles to remain hyperlinked.

PR #17634, opened 2025-07-18
 More strict selected styles

### LINKED ISSUE #17633: [5.x]: Styling of menu subitems
### Description
Not that important, but saw all subitems with the same name? get active styling 😄 

### Steps to reproduce

1. Visit Settings in some plugin
2. Open the sub items from other plugins
3. See that all Settings are marked active

### Expected behavior
Only the current active page should show active

### Actual behavior

Image

### Craft CMS version

5.8.7

### PHP version

_No response_

_No response_

### Database type and version

_No response_

### Image driver and version

_No response_

- Commerce
- Navigation
- Retour
- Formie
- ...

brianjhanson:
Fixes #17633

### DIFF
--- a/src/templates/_layouts/components/global-sidebar.twig
+++ b/src/templates/_layouts/components/global-sidebar.twig
@@ -101,7 +101,7 @@
                             {% if item.subnav %}
                                 <ul class="nav-item__subnav" id="{{ item.id }}-subnav" data-state="{{ item.sel ?  'expanded' : 'collapsed' }}">
                                     {% for itemId, item in item.subnav %}
-                                        {% set itemIsSelected = selectedSubnavItem is defined and selectedSubnavItem == itemId -%}
+                                        {% set itemIsSelected = selectedSubnavItem is defined and selectedSubnavItem == itemId and item.url in craft.app.request.fullPath -%}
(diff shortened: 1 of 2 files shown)


src/templates/_layouts/components/global-sidebar.twig
```diff
@@ -101,7 +101,7 @@
                             {% if item.subnav %}
                                 <ul class="nav-item__subnav" id="{{ item.id }}-subnav" data-state="{{ item.sel ?  'expanded' : 'collapsed' }}">
                                     {% for itemId, item in item.subnav %}
-                                        {% set itemIsSelected = selectedSubnavItem is defined and selectedSubnavItem == itemId -%}
+                                        {% set itemIsSelected = selectedSubnavItem is defined and selectedSubnavItem == itemId and item.url in craft.app.request.fullPath -%}
```
brianjhanson, line 104: We need both conditions. The first will unintentionally match any subnav item with the same ID (usually something like "settings") while the second unintentionally matches subnav items that have similar IDs (like `commerce/inventory` and `commerce/inventory-locations`). Together they seem to only match the actually active subnav item.

Issue #17628, opened 2025-07-17
[5.x]: Unable to see the "Show the search input" toggle on Assets field


benjaminkohl:
### Description

I was attempting to disable the search input on an Assets field and learned that there is a toggle for this but it does not appear for me. I cannot determine the appropriate combination of settings to get that toggle to appear but I also feel that it should always be an option to disable it because it takes up space and is not always convenient or necessary.

Here is a screen capture from my Discord discussion with mmikkel.

<img width="833" height="210" alt="Image" src="https://github.com/user-attachments/assets/7ff6a356-560c-4663-8318-96f2b4f1177d" />

### Steps to reproduce

1. Open Assets field settings form.
2. Search for the "Show the search input" toggle.

### Expected behavior

See the "Show the search input" toggle.

### Actual behavior

Not seeing the "Show the search input" toggle.

### Craft CMS version

5.8.6

### PHP version

8.3

### Installed plugins and versions

-

benjaminkohl:
That bot scared. I thought it was linking me to an existing issue and I was like NO WAY.

i-just:
Hi, thanks for getting in touch!

The “Show the search input” toggle will only show if you have explicitly selected a single source for the field. Selecting “All” while having only a single source doesn’t count as having selected a single source, and the toggle (intentionally) won’t show in this case.

That said, I believe there’s a bug there for the Assets-type field. The toggle should also show if you have “Restrict assets to a single location” turned on, but currently, it doesn’t. I raised a PR for this.

mmikkel:
> The “Show the search input” toggle will only show if you have explicitly selected a single source for the field. Selecting “All” while having only a single source doesn’t count as having selected a single source, and the toggle (intentionally) won’t show in this case.

The bug being reported is that the search input will render in edit forms for Assets fields that has "All" sources selected, if there is only a single source (i.e. Volume) in the system. 

And since the "Show the search input" toggle (correctly) doesn't show in the field's settings when "All" sources are selected, there is no way to disable the search input (aside from selecting the single volume under Sources instead of "All", which will display the toggle).

I checked, and the same behavior does *not* occur with Entries fields. I also tested with your PR and that fix does not affect this behavior (so I guess there were two separate bugs related to this 😊)

To reproduce:

1. Have 1 Volume in the system
[...]

PR #17640, opened 2025-07-21
show search input should show when restrict location is on (for assets field)

i-just:
### Description
Fixes a bug where the “Show the search input” field was not shown for the Assets field when a single source was selected via the “Restrict assets to a single location”.

Fixes a bug where the search input was shown on the element edit page when the Assets field had Sources “All” selected, with only one source available to select.

### Related issues
#17628


brandonkelly:
Resolved both issues via ecc16cdfad3ec589fda50410aa7cf2cf98c6059e instead, so the Assets-specific logic is contained to Assets field files.

Issue #17644, opened 2025-07-21
[5.x]: Integrity constraint violation: 1062 Duplicate entry '126007-3' for key 'elements_sites.idx_djwxcsldvwxgnblmmmgsfddjesvzsrwwqyjv'


patrick-long:
### What happened?
Experienced a duplicate entry error when attempting to apply a draft to an entry that contained different site translations between the current revision and the draft version.

### Description
When working with multiple users who are operating within a draft workflow, if a user creates a draft of an entry that has sites added to it after the draft has been created, there is potential for the draft to throw an error when a user later tries to apply the draft:

```
Integrity constraint violation – [yii\db\IntegrityException](https://www.yiiframework.com/doc-2.0/yii-db-integrityexception.html)

SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '126007-3' for key 'elements_sites.idx_djwxcsldvwxgnblmmmgsfddjesvzsrwwqyjv'
[...]

PR #17650, opened 2025-07-22
Bugfix/17644 site settings and drafts

i-just:
### Description
Fixes an integrity constraint violation error that can be thrown when updating the `elements_sites` table when applying a draft that belongs to fewer sites than the canonical.

### Related issues
#17644

### DIFF
--- a/src/services/Elements.php
+++ b/src/services/Elements.php
@@ -3278,7 +3278,9 @@ class Elements extends Component
         $dirtyFields = $element->getDirtyFields();
 
         // Get the element's site record
-        if (!$isNewElement && !$element->isNewForSite) {
+        if ($isNewElement && $element->isNewForSite) {
+            $siteSettingsRecord = null;
+        } else {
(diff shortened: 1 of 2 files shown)


src/services/Elements.php
```diff
@@ -3278,7 +3278,9 @@ private function _saveElementInternal(
         $dirtyFields = $element->getDirtyFields();
 
         // Get the element's site record
-        if (!$isNewElement && !$element->isNewForSite) {
+        if ($isNewElement && $element->isNewForSite) {
```
brandonkelly: @i-just Is the change in logic intentional here?

If this was just meant to clean the code up a bit, the exact reverse condition would be

```php
if ($isNewElement || $element->isNewForSite) {
```

i-just: It was intentional. First, I wanted to drop the condition altogether and attempt to grab the site settings in all cases. Anyway, testing it again this morning, switching to `($isNewElement || $element->isNewForSite)` is a better call. I just pushed that change.

PR #17690, opened 2025-07-30
Implement the orientation fieldset using the radio group field macro

gcamacho079:
### Related issues
Resolves PT-2223, PT-2087, PT-2066


gcamacho079:
@brandonkelly this change also removes the portrait and landscape SVGs, but I'm not sure how much they enhance the UI. Open to your thoughts!

brandonkelly:
@gcamacho079 Could we switch it to a button group, using the same markup Button Group fields use (without visible labels)? We could use the `image-portrait` and `image-landscape` icons.

gcamacho079:
@brandonkelly ready for your review again!

PR #17691, opened 2025-07-30
Use translation for alert icon alt

gcamacho079:


gcamacho079:
@brandonkelly I also noticed while testing this that the string `Found {num, number} {num, plural, =1{error} other{errors}}` is not translated in the Spanish translation file.

PR #17699, opened 2025-08-01
Allow `CreateFieldLayoutFormEvent::static` to be modified via its event

engram-design:
### Description
I'd like the ability for some plugins [Workflow](https://github.com/verbb/workflow/issues/805#issuecomment-3138826775) to be able to made an entry static under certain conditions.

All that's required is the local variable to be set via the event.


brandonkelly:
Thanks Josh! Craft 5.8.13 is out with that change.

PR #17712, opened 2025-08-05
Prevents narrowing of mapped fields based on type or section

lthurston:
### Description

Sometimes, you want to write a query resolver that has a handful of possible conditions for matching entries. For instance, let's say I have a resolver that I want to use to fetch both entryType1 and entryType2, or sometimes just one or the other. entryType1 has a field on it called `foo`. Both entry types have a field called `bar`.

I might have a condition that's like:

```
$query->andWhere(['or', ['foo' => 'baz'], ['bar' => 'qux']]);
```

This works fine! Yay!

But, if I do this:

```
$query->type('entryType2');
$query->andWhere(['or', ['foo' => 'baz'], ['bar' => 'qux']]);
```

Ruh roh! In this case `foo` doesn't get mapped to a JSON-embedded field in `content`. Because it's kind of annoying to have to juggle which entries have which fields, I've added:

```
$query->withAllFieldLayouts(true);
$query->type('entryType2');
$query->andWhere(['or', ['foo' => 'baz'], ['bar' => 'qux']]);
```

Now it JUST WORKS.

### Related issues

Kind of related to my work is this: https://github.com/craftcms/cms/blob/5.x/src/elements/db/ElementQuery.php#L2806-L2808

It will bypass a missing field if it's wrapped in an OR condition (hey that's what I'm trying to do!), but it fails to map the field in $_columnMap. 

I think my work closes a loop that was started here, but not finished.

### DIFF
--- a/src/elements/db/EntryQuery.php
+++ b/src/elements/db/EntryQuery.php
@@ -123,6 +123,14 @@ class EntryQuery extends ElementQuery implements NestedElementQueryInterface
     public mixed $authorId = null;
 
     /**
+     * @var bool A flag that forces the mapping of all custom fields regardless of $typeId or $sectionId
+     * .
+     * @used-by withAllFieldLayouts()
+     * @used-by fieldLayouts()
+     */
+    public mixed $withAllFieldLayouts = false;
+
+    /**
      * @var mixed The user group ID(s) that the resulting entries’ authors must be in.
      * ---
      * ```php
@@ -242,6 +250,9 @@ class EntryQuery extends ElementQuery implements NestedElementQueryInterface
                 break;
             case 'authorGroup':
                 $this->authorGroup($value);
+                break;
+            case 'withAllFieldLayouts':
+                $this->withAllFieldLayouts($value);
                 break;
             default:
                 $this->nestedTraitSet($name, $value);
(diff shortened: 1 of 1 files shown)


lthurston:
@brandonkelly I thought you might find this one useful too. If I need to do more to make the case for this, or implement it differently, I'm happy to do that.

AugustMiller:
I imagine this problem exists for more than just entries (any element query that can be narrowed by a `FieldLayoutProvider`), so if it indeed requires a fix in core, the property and method should porbably be on `craft\elements\db\ElementQuery`.

lthurston:
>  I imagine this problem exists for more than just entries 

That certainly makes sense, but I can't speak to whether it would be useful in a more general context. My approach to resolving it was: 1) locate where the code that narrows the field mappings takes place, 2) add an additional condition for that code, 3) add an interface for the developer to disable that narrowing, all of which I was able to implement within the EntryQuery class. If a similar narrowing of mapped fields takes place for Elements, Entities, Elephants, Eulogies then I would expect it to take place on an ancestor class to Entries.

brandonkelly:
> ###
> 
> Kind of related to my work is this: https://github.com/craftcms/cms/blob/5.x/src/elements/db/ElementQuery.php#L2806-L2808
> 
> It will bypass a missing field if it's wrapped in an OR condition (hey that's what I'm trying to do!), but it fails to map the field in $_columnMap.
> 
> I think my work closes a loop that was started here, but not finished.

That code is for handling top-level element query params, e.g. `craft.entries().myCustomField('foo')` – see #17084 for context.

Given that you’re already sortof wading in the deep end with `andWhere()`, I’m thinking a better solution would be for you to stop using the `type` param, and restrict entries to the entry type using `andWhere` instead.

```php
$entryType = Craft::$app->entries->getEntryTypeByHandle('entryType2');

$query
    ->andWhere(['entries.typeId' => $entryType->id])
    ->andWhere(['or', ['foo' => 'baz'], ['bar' => 'qux']]);
```

That will give you the same resulting SQL, but since the `type` param was never used, `EntryQuery::fieldLayouts()` won’t end up limiting the returned layouts based on the entry type.

lthurston:
Yeah, that makes sense. Because I am doing this in the context of a query resolver (specifically a subclass of the entry query resolver), I'll have to handle the type argument w/o pushing that into the type or section arguments. Which is fine, I can reuse code from the entry resolver.

I'll give this a go, appreciate the feedback!

Issue #17713, opened 2025-08-06
[5.x]: 5.8.13 causes craft\errors\FieldNotFoundException


jamesmacwhite:
### Description

After upgrading to 5.8.13, the following stack trace is raised related to Custom Fields.

The following occurred after the Composer update command was run and triggered post-update-cmd commands like `migrate/all` and `project-configt/apply` via CLI.

The following stacktrace was raised.

The Scout plugin is referenced in the stacktrace but I think it might more to do with the entry query behaviour, which I believe has a specific change in this release possibly https://github.com/craftcms/cms/issues/17709

The Scout plugin has not been updated in a couple of months and downgrading back to 5.8.12 does not trigger this error. If this is a Scout plugin bug I'm more than happy to raise it there, but it seems like a upstream change in `craftcms/cms` is possibly the trigger.

```
@php craft install/check && php craft migrate/all --interactive=0 || exit 0
Exception 'craft\errors\FieldNotFoundException' with message ''

in /var/www/html/vendor/craftcms/cms/src/fieldlayoutelements/CustomField.php:187

Stack trace:
#0 /var/www/html/vendor/craftcms/cms/src/models/FieldLayout.php(1137): craft\fieldlayoutelements\CustomField->getField()
#1 [internal function]: craft\models\FieldLayout->craft\models\{closure}(Object(craft\fieldlayoutelements\CustomField))
#2 /var/www/html/vendor/craftcms/cms/src/models/FieldLayout.php(1136): array_map(Object(Closure), Array)
#3 /var/www/html/vendor/craftcms/cms/src/models/FieldLayout.php(953): craft\models\FieldLayout->_customFields()
#4 /var/www/html/vendor/craftcms/cms/src/Craft.php(393): craft\models\FieldLayout->getCustomFields()
#5 /var/www/html/vendor/craftcms/cms/src/Craft.php(241): Craft::_fields()
#6 /var/www/html/vendor/craftcms/cms/src/Craft.php(204): Craft::_autoloadCustomFieldBehavior()
#7 [internal function]: Craft::autoload('craft\\behaviors...')
#8 /var/www/html/vendor/yiisoft/yii2/di/Container.php(507): ReflectionClass->__construct('craft\\behaviors...')
#9 /var/www/html/vendor/yiisoft/yii2/di/Container.php(385): yii\di\Container->getDependencies('craft\\behaviors...')
[...]

PR #17714, opened 2025-08-06
prevent throwing an exception if layout contains non-existent field

i-just:
### Description
Following the changes from https://github.com/craftcms/cms/commit/e80ca1b60bf4daa8958e6a7d0801ec8ab6f427f0, if you have a field layout that (for some reason) references a field which no longer exists, you’ll get a `FieldNotFoundException` when attempting to apply project config or access some areas of the control panel.

Before the change mentioned above, we were checking if the field exists and throwing an exception from within the `craft\fieldlayoutelements\CustomField::setFieldUid()` method. That exception was then caught, and we carried on. Moving the check to the `craft\fieldlayoutelements\CustomField::getField()` means there are plenty of places where it won’t be caught.

Checking if the field exists and throwing the exception from the `craft\fieldlayoutelements\CustomField::setFieldUid()` method fixes the issue.

@brandonkelly, I’m not able to reproduce the error that the related commit fixed (on a version prior to 5.8.13), so I wasn’t able to test if this PR affects that issue.

### Related issues
#17713

### DIFF
--- a/src/fieldlayoutelements/CustomField.php
+++ b/src/fieldlayoutelements/CustomField.php
@@ -232,6 +232,12 @@ class CustomField extends BaseField
     {
         $this->_fieldUid = $uid;
         $this->_field = null;
+
+        if (($field = Craft::$app->getFields()->getFieldByUid($uid)) === null) {
+            throw new FieldNotFoundException($uid);
+        } else {
+            $this->setField($field);
+        }
     }
 
     /**


brandonkelly:
Closing in favor of 26c510022211e9a9357f3357f79a1c51d0969512 + 7cbc686800719d58b5ebbe11d154b5b90d72e140.

The point of e80ca1b60bf4daa8958e6a7d0801ec8ab6f427f0 is to avoid calling `getField()` from `setFieldUid()`, so we can avoid a `FieldNotFieldException` from being thrown when instantiating the field layout – so it doesn’t cause trouble when installing Craft with existing project config data, where not all fields may actually be created yet.

PR #17730, opened 2025-08-09
Option to return a template in controller

alex-brukhty:
### Description
Option to return a template in controller.
Handy in frameworks like HTMX and custom JS implementation for partial rendering.


brandonkelly:
This would introduce a slew of potential security vulnerabilities, as there’s no way to ensure the passed template is actually intended to be rendered and returned via this `template` param.

PR #17731, opened 2025-08-10
Display grouped “New entry” buttons as a button group

brandonkelly:
### Description

Matrix fields with multiple entry type groups will now show a group of buttons for new entry creation, where each entry type group gets its own menu button.

A button group with “General”, “Media”, and “Misc” menu buttons, with “Media” expanded to reveal “Gallery” and “Image” nested options.

When the viewport isn’t wide enough to support the expanded button group, a single “New entry” button will be shown instead, as before.

### Related issues

- #15895


gcamacho079 (changes requested):
I've suggested some updates in this PR: https://github.com/craftcms/cms/pull/17735

superflausch:
Maybe a similar approach could also work for the context menus on the blocks themselves. Those menus get messy as well when there are many different entry types. See my comment here: [link](https://github.com/craftcms/cms/discussions/17600#discussioncomment-14285593).

davidhellmann:
I like it and therefore don't dare ask. Could there still be an option to specify whether to use a single dropdown or a group dropdown? I'm asking because some of my authors like the "OLD" approach more. And of course, I still want to have the groups, but I just want to be able to choose which view I want.

ryanleichty:
At first I thought this would be a nice upgrade, but I actually prefer the old iteration with a single dropdown. It requires a single click to see every option available, whereas now you have n+1 clicks in order to discover what options are available. It would be great to have a setting to control how these are displayed, whether grouped as separate dropdown menus or as a single dropdown like before.

PR #17735, opened 2025-08-11
Update button group to use aria-describedby instead of overriding ari…

gcamacho079:
…a-label; add ariaDescribedBy property to createButton config


gcamacho079:
@brandonkelly I just made a quick PR using `aria-describedby` to provide supplementary info regarding what the button does (i.e. "Create entry").

This is more bulletproof for several reasons, one of which is that it doesn't introduce a mismatch between the visible label and and accessible name. As a result, I wasn’t sure if we needed the `Create new {group} {type}` translation, or if we could just stick with the existing `Create {type}` translation.

brandonkelly:
@gcamacho079 Thank you!

PR #17736, opened 2025-08-11
Add a setting to Slideouts for focus trigger delay

gcamacho079:
### Description
This PR adds a setting to Slideouts for introducing a slight delay to focusing the trigger element, and implements it for the `ElementEditorSlideout`.

### Reproducing the issue
1. On the 5.x branch, go to My Account > Addresses
2. Click the "Edit" button for an existing address (either via keyboard or mouse)
3. Click somewhere inside the slideout footer (to activate `:focus-visible` styles when the modal is closed)
4. Use the `Tab` key to get to the "Save" button then hit `Enter`

**Expected (this branch):** A visible focus outline should appear on the "Edit" button
**Actual (5.x):** Focus is dropped, and you have to hit `Tab` to refocus the Edit button

I couldn't figure out exactly what is causing the focus to drop after the slideout closes, but introducing a short delay before focusing the trigger element seems to solve the issue.


brandonkelly:
@gcamacho079 Would it make sense to just always have a delay, rather than adding a setting for it?

gcamacho079:
> @gcamacho079 Would it make sense to just always have a delay, rather than adding a setting for it?

@brandonkelly That seems reasonable! I've removed that setting.

PR #17762, opened 2025-08-21
Enhance writeToFile checks

timkelty:
### Description
`file_put_contents` doesn't give us specific errors, so this PR adds a couple more checks to the `writeToFile` helper to first:

- check is the destination is writable
- check if there is sufficient disk space


src/helpers/FileHelper.php
```diff
@@ -461,6 +461,27 @@ public static function writeToFile(string $file, string $contents, array $option
             }
         }
 
+        if (!static::isWritable($file)) {
+            throw new ErrorException("The file path \"$file\" is not writable.");
+        }
+
+        if (function_exists('disk_free_space')) {
```
timkelty, line 468: Some hosts explicitly disable this function.

PR #17767, opened 2025-08-22
Fix German translation: 'move' instead of 'delete'

Megafry:
Correct German translation to 'verschieben (move)' instead of 'löschen (delete)'.


brandonkelly:
Thank you!

PR #17779, opened 2025-08-25
Customizable entry index pages

brandonkelly:
### Description

Gives entries the ability to divide their sources into multiple index pages, via the Customize Sources modal.

The “Customize Sources” modal for entries, showing that three pages have been defined: “Entries”, “Categories”, and “Tags”.

Each page gets its own spot in the global CP navigation.

The global sidebar showing that “Entries”, “Categories”, and “Tags” items are displayed right alongside “Dashboard” and “Assets”.

Additionally, the `entrify/categories` and `entries/tags` commands have been updated to automatically assign newly-created sections in “Categories” and “Tags” entry index pages, respectively.

### Related issues

- #13307
- #11340
- #14650
- #15162
- #14609
- #9781


thupsi:
Yay! Been waiting a long time for this 🙂

Maybe the content icon should change to something else now, to reflect the full spectrum of content types? Maybe something more abstract, similar to the "categories" icon?

brandonkelly:
The newspaper icon feels pretty generic to me, but open to other suggestions!

brandonkelly:
`pen-nib` could work:

The global sidebar with the “pen-nib” icon used for the Content section

mmikkel:
> `pen-nib` could work:

I don't really have a problem with the newspaper and don't really see the big reason to change it – but, it'd throw my hat in the ring for something like `pen-line` or `pen-to-square` instead of that one 🙂 Feels maybe a bit more contemporary (and do young people even know what a fountain pen is, or what it looks like)? 🫠

Anyway, whatever the icon ends up being, can't wait to give this PR a spin! 🎉

thupsi:
Yeah, like @mmikkel says, I think the pen nib is too retro. I do prefer `pen-line` and `pen-to-square` to that, but still all these don't mean "content" to me.

For a more broad symbolism I would suggest something like `cubes`, `cube`, `diagram-project`, `circle-nodes`, `chart-diagram`. Or even use `sitemap` (and use something with folders for the categories icon, until they go away completely).

thupsi:
My main reasoning is that the icon should represent content in the classic sense, but also, equally, globals and categories. It's indeed a little difficult to encompass all these, I know!

thupsi:
Hm, even `rectangle-list` would work IMO. It kinda looks like a newspaper, but also like a digital piece of content. 

And.. I'm stopping now, sorry 🙂

gcamacho079 (changes requested):
Awesome feature! Here are some notes from an accessibility audit on the modal updates:

- [x] The "New page" button and "New source" disclosure menus should be moved out of the footer. Their current placement in the footer leads to unexpected focus order, since having them in the footer introduces the expectation that they would be located immediately before the "Cancel" button in the focus order. In addition, they are very distant from the columns/lists when zoomed. Recommend either:
    - Moving the buttons next to the `h2` elements, or
    - Adding buttons directly after the last `.customize-sources-item`, similar to what's done in the Field Layout Editor
- [x] `.customize-sources-item` elements should use `li` wrapped in `ol`
- [x] Keyboard users can’t move the sources to another page. I recommend adding these options to the Actions disclosure menu or opening a modal (similar to the "Move to..." action in the element index footer disclosure menu)
- [x] Instead of adding new translations for "Page actions" and "Source actions", for consistency with other action menus in the UI, you can use `aria-describedby` on the Actions button to reference an ID on each item's `.label` element
- [x] Each `.cs-sidebar` element should reference the sidebar heading via `aria-labelledby`, so that each navigation landmark has a unique accessible name
- [x] When adding a new page or source, the `Craft.cp.announce()` method should be called with a success message
- [x] The modal should be unusable at a minimum screen size of 320px. I recommend updating the view at small viewport widths to only show a single screen at a time, with the ability to navigate back and forth through the flow. Focus should be managed so that when the screen changes, keyboard focus is moved to an appropriate element in the new screen. (See screenshot below)

Mockup of three screens in the modal. The first only shows Pages and its associated buttons. An arrow points to the next mockup, which shows only the Sources items and has a button at the top to go back to pages. A final arrow points toward the last screen, which shows the Settings for the selected source and has a button at the top to go back to sources.

(Thanks @brianjhanson for brainstorming ideas!)

ryansupercool:
This is looking really nice.

This may not be the right place to discuss this, and it's not clear if this is the "Unified Content View" in its final form or just a step towards it, but how do you see Custom Elements fitting into the unified view, if at all?

We've a plugin that adds various Custom Elements to Craft for Event management. In Craft 4, those elements sit under Events in the sidebar, along with "Entries", "Globals", "Categories", "Assets" etc (the same as Commerce).

With the move of Globals and Categories to Entries, and now "Content", those Custom Elements are beginning to feel a little out of place, after all, it's all "content".

Could/should Custom Elements be able to put themselves in the mix of "Content"? Could a solution here be to hoist those new nested sources to the top level in the sidebar, allowing for a Craft 3 like setup?

thupsi:
I've been playing with this feature (couldn't wait for it to be released 🙂) and I like it a lot! Way to go Brandon & team! 🎉

Some quick suggestions & observations:

**1. Can you make the top-level "Content" label configurable?** 
For example, in some projects it would make sense to have the tree like so:

```
Site (instead of "Content")
├── Content
├── Taxonomies
└── Settings 
```

**2. Optional icons for pages.**
It would really help when the main nav is collapsed.

**3. Pages with only one source should not display the sidebar / sources list.**
It looks reduntant and repetitive if the page has the same name with the solitary source. In fact, if a page contains only a _single_, it's  even more reduntant and repetitive (the title is repeated 3 times), so...

**4. Pages with only a single should just display the single's edit form.** 
That for me would be ideal, and it also would be nice if _even sources with only one single_ displayed the single's edit form like described in https://github.com/craftcms/cms/discussions/17760. But that one maybe is out-of-scope for this PR.

**5. The default "Entries" page should not be removable, and/or we should have a way to bring back no-longer accessible non-custom sources.**
For example, if I delete "entries", all my non-custom sources disappear. The same if I move a non-custom source to a new page and later delete that page.

Feel free to ignore all of the above 🙂

thupsi:
Continuing with unsolicited suggestions:

IMO, the "Content" grouping feels a little _too opinionated and forced_ , which is a bit weird for Craft, where almost everything follows a blank slate philosophy.

Yes, _all entries are content_, but taxonomies and global settings maybe are not, at least not in the mind of all users. The fact that all these are now "entries" under the hood, is a technicallity that shouldn't dictate the author facing UI. I would argue that, for certain projects, there is enough semantic separation between the various "Pages" to warrant their own place as top level nav items. Therefore, I would love to see an option to have the **"Content" section ungrouped**, or even **allow additional "Sections"** outside of "Content".

I can see how that could be a awkward due to missing icons for pages/sections but see my suggestion above for optional page  icons.

UPDATE WITH EXAMPLE:
Notice how, in the screenshot bellow, it looks strange, from an author's perspective, that "Settings" is inside "Content" (while "Assets" is not):

image

ryanleichty:
Anything that gets us closer to [Sanity's structure tool](https://www.sanity.io/docs/studio/structure-tool) and flexibility I'm a fan of 🎉

brandonkelly:
Just pushed up some changes:

- Entry index pages are now added to the global nav at the top level, not nested under “Content”.
- Pages now get customizable icons (newspaper is used by default).
- Several accessibility improvements per @gcamacho079’s feedback.

The PR title/description has also been updated to reflect those.

gcamacho079 (changes requested):
## Accessibility
- [x] When trying to create a page with an empty page name field, nothing happens. The Page Name field should be marked required and an error should show if the user tries to submit the form with an empty Page Name field.
- [ ] The explicit `tabindex` attribute can be removed from “Back to…” buttons.
- [x] Keyboard focus is dropped when removing a page from the list of Pages. I recommend moving focus to an Action menu button for the previous/next element. (Note: this should be tested in cases where (a) the page is the last one in the nav list, and (b) when there is no previous/next item in the list. In cases where the last item is removed, focus can be moved to the “New page” button.)

## General UX feedback
### Breadcrumbs
The breadcrumb structure for custom index pages is a bit confusing for me. For example, I created a "Blog Post" index page with only a single source:

A Blog Post custom index page with a single source called Blog

One of the entries inside has a breadcrumb structure that I find unexpected based on the "Blog Post" index structure.

A blog post entry with a breadcrumb that includes a Content breadcrumb above the Blog Post level despite the index being a top-level nav item, and a disclosure menu that includes a Singles source, which isn't present in the index

### Navigation
This issue could potentially introduce WCAG violations, so depending on how we choose to resolve this, @AugustMiller may need to mention it in the Docs.

**Description:** Admins can create pages that conflict with pre-existing navigation items. For example, you could create a "Settings" page that conflicts with the existing top-level Settings link. They could be slightly differentiated by setting a different icon, but because icons are marked as decorative in the code, they would have identical accessible names and be impossible for AT users to differentiate.

A custom sources modal that shows a Settings index page has been added. This has created a navigation link in the global sidebar that looks identical to the preexisting Settings nav item.


**Suggestion:** Do not allow the creation of Pages with names that conflict with pre-existing navigation items, or add instructions that warn against doing so.

### Drag handles for smaller viewports
Should the drag handles be removed on smaller viewports (i.e., when only a single pane shows at a time)? I accidentally dropped something into the void because there wasn't a drop target. 🤔

src/web/assets/cp/src/js/CustomizeSourcesModal.js
```diff
@@ -129,262 +178,307 @@ Craft.CustomizeSourcesModal = Garnish.Modal.extend({
...
-    return this.$sidebarToggleBtn.attr('aria-expanded') === 'true';
+    return this.sourceContainers[pageName];
   },
 
   addSource: function (sourceData, isNew) {
-    const $item = $('<div class="customize-sources-item"/>').appendTo(
-      this.$sourcesContainer
-    );
-    const $itemButton = $('<div class="customize-sources-item__btn"/>')
+    const pageName = sourceData.page ?? this.selectedPage.name ?? '__DEFAULT__';
```
AugustMiller, line 441: Missing a check for `this.multiPage`? Modal blows up on categories index, when trying to access `selectedPage`. 😉

brandonkelly: Doh, [fixed](94f573d0cefca927cf1a1db0567d6f8dcdb28a4f).

thupsi:
While testing this in a dev environment, I encountered a bug where the sorting of the index "pages" gets stuck, so I can't get them reordered.

So, I do this: 

image

And I get this:

image


Unfortunately I didn't have time to figure out what triggers this behaviour, so just reporting without steps to reproduce.

thupsi:
Aaaaand just found another small bug 🐛:

Content page list items on the global sidebar get an id in the form of  `nav-content-<my-page>`. When the name of the page is non-latin though (tested in greek), the `<my-page>` part is ommited, resulting in multiple list items with an id of `nav-content`.

brandonkelly:
@thupsi Thanks for reporting those! Fixed now.

thupsi:
Hey @brandonkelly, thanks for the fixes! I found though that now _the order of the sources themselves_ in the sidebar can get stuck.

Also, the bug that Lupe reported about the breadcrumbs is still there (don't know if it was fixed and there's been a regression), meaning that sources that belong in index pages other than the current one are visible in the breadcrumbs.

brandonkelly:
@thupsi Fixed now!

thupsi:
Ahem 😅
image

brandonkelly:
@thupsi That’s fixed now too :)

thupsi:
Almost there 😅!

I noticed that disabling a source results in displaying `Entries` in the breadcrumbs of that source's entries, instead of the custom page's title.

I hereby promise to not report any more bugs until after 5.9 is released 🙂

brandonkelly:
@thupsi Thanks, fixed! And please feel free to continue reporting issues if you come across any ❤️

thupsi:
> @thupsi Thanks, fixed! And please feel free to continue reporting issues if you come across any ❤️

I'll take you up on that then 🙂

It's a breadcrumbs bug yet again!



image

*Here are my custom pages, `Content` and `Theming`. There's no other entry index page.*

image

*Observe the `Content` custom page with 3 sources.*

image

*Now I'm in an `Artworks` entry and, sure enough, there's the `Artworks` breadcrumb nav item.*

image

*However, when I press that `Artworks` link, I end up in a generic `Entries` page (a place previously thought to have vanished from this earth) where all my enabled sources, from all custom pages, reside.*

**That's the bug!** 🐛



**And now for something else**...
 that might be a bug or not, but sure looks weird for editors.

image

*Let us examine this humble `Theming` custom page, with a custom source that displays a specific single.*

image

*When I go to `Theme settings` I have this breadcrumb structure: `Content`>`Singles`>`Theme settings`*

Only I (the mighty dev) can understand why it's like this: The `Theme settings` single belongs to the (disabled) `Singles` source, that is inside the "Content" custom page. 

For the editor though it does not make any sense: The `Singles` source is nowhere to be found in the CP, and, perhaps more importantly, she is currently editing an entry in the `Theming` page, not the `Content` page. 

The above situation makes me reluctant to distribute singles as appropriate between custom pages, because the breadcrumbs, for some of them, will always appear incorrect.

One possible way to resolve this is described in https://github.com/craftcms/cms/discussions/18090 by yours truly.

Thank you for your attention to this matter! (●'◡'●)

brandonkelly:
> _However, when I press that `Artworks` link, I end up in a generic `Entries` page (a place previously thought to have vanished from this earth) where all my enabled sources, from all custom pages, reside._

I’m not able to reproduce that. Can you send a database backup and your Composer files into support@craftcms.com so we can look into it?

> _Let us examine this humble `Theming` custom page, with a custom source that displays a specific single._

How exactly are you displaying a specific single as its own source? That’s not possible in the core functionality.

thupsi:
> I’m not able to reproduce that. Can you send a database backup and your Composer files into [support@craftcms.com](mailto:support@craftcms.com) so we can look into it?

Done!

> How exactly are you displaying a specific single as its own source? That’s not possible in the core functionality.

It's very possible to do with core functionality 🙂: Custom source that is configured to only show entries from a single section. See screenshot:

image

As far as I know that's the only way to currently emulate a "Global Set". Sadly, it's a bit of a UI/UX downgrade from a real Global Set, an issue I touch upon in https://github.com/craftcms/cms/discussions/17760, in my usual long-winded way!

brandonkelly:
Sorry, of course 🤦🏻

arentsen:
Great feature! How do I get the content dropdown? I only see a separate entry all on the main level for each page I add. 

And how do I select categories and tags? I can't find them in the filter field.

thupsi:
> Great feature! How do I get the content dropdown? I only see a separate entry all on the main level for each page I add.

There's no dropdown, each page is right on the main nav.

> And how do I select categories and tags? I can't find them in the filter field.

You can entrify categories and tags via the CLI.
https://craftcms.com/docs/5.x/reference/cli.html#entrify

SkermBE:
@thupsi, not sure if this is a mock-up or something I'm not finding.. but this is the feature I'm really missing before moving my globals to entries. I need a way to quickly show/navigate between my global (singles) and at this point in time it not possible.

I have a few singles that act as a global page, putting them in a tree-layout would kine fix this so they are alway visible to navigate to. 

@brandonkelly, can we have the option to create tree-like custom index pages and/or have an improvement for singles where they act more like globals where you can edit the page while still navigate to other pages quickly. 

image



> Continuing with unsolicited suggestions:
> 
> IMO, the "Content" grouping feels a little _too opinionated and forced_ , which is a bit weird for Craft, where almost everything follows a blank slate philosophy.
> 
> Yes, _all entries are content_, but taxonomies and global settings maybe are not, at least not in the mind of all users. The fact that all these are now "entries" under the hood, is a technicallity that shouldn't dictate the author facing UI. I would argue that, for certain projects, there is enough semantic separation between the various "Pages" to warrant their own place as top level nav items. Therefore, I would love to see an option to have the **"Content" section ungrouped**, or even **allow additional "Sections"** outside of "Content".
> 
> I can see how that could be a awkward due to missing icons for pages/sections but see my suggestion above for optional page icons.
> 
> UPDATE WITH EXAMPLE: Notice how, in the screenshot bellow, it looks strange, from an author's perspective, that "Settings" is inside "Content" (while "Assets" is not):
> 
> image

brandonkelly:
@SkermBE That was requested here: #17760

Issue #17803, opened 2025-08-31
[5.x]: Can't create fresh entries via graphql


Latex88:
### Description

It seems that creating brand new entries directly via GraphQL is not possible. The createDraft mutation requires a canonical entry ID, so it can only be used to create a draft of an existing entry. The only way I’ve found to create a “fresh” entry is to use the regular save_{sectionHandle}_{typeHandle} mutation and set the entry status to disabled.

Steps to Reproduce
	1.	Attempt to run the createDraft mutation without providing a canonical entry ID.

Expected Behavior

I should be able to create a new, unsaved entry as a draft (without needing an existing canonical entry).

Actual Behavior

The createDraft mutation requires a canonical entry ID and fails without one.

### Craft CMS version

5.8.15

### Installed plugins and versions

-

brandonkelly:
Thanks for pointing that out! We’ve made it possible to create new entries in an unpublished draft state for Craft 5.9 (#17805).

Latex88:
Thanks!!!

brandonkelly:
Craft 5.9.0 is out with that change 🎉

PR #17805, opened 2025-09-01
Feature/17803 gql create unpublished draft

i-just:
### Description
Allows for an unpublished draft of an Entry to be created via GraphQL mutation. 

To do so, you need to use the `save_{sectionHandle}_{typeHandle}_Draft` mutation with a new `asUnpublishedDraft` parameter set to `true`. 

In order for this to work, the `draftId` parameter is only mandatory when `asUnpublishedDraft` is not present or is not set to `true`.

### Related issues
#17803


i-just:
moving to a draft to work on tests

PR #17815, opened 2025-09-03
the key for table columns should be 'attr' too

i-just:
### Description
Selecting a custom field on the element index page (via View > Table Columns checkboxes) wasn’t working as expected. The custom fields were showing on the list, but selecting one to show in the table wasn’t actually showing the corresponding column in the table.

It looks like when introducing lazy-loading of element source attribute info (https://github.com/craftcms/cms/commit/835207d7ddf0f4517b3aabac7ab390375af28abf), we accidentally changed the name of the key that holds the custom field uid from `attr` to `key`. This caused the input value of the custom field checkbox to be `1` instead of `field:<uid>`, leading to this issue.

### Related issues
n/a; reported by @riasvdv


brandonkelly:
Doh, thanks!

PR #17816, opened 2025-09-03
Playwright tests package

i-just:
### Description
This PR adds the `craftcms-playwright` package and a set of e2e tests for the cms repo.

The package is responsible for booting a ddev environment, installing and setting up Craft CMS, initiating tests and tearing down the environment afterwards.

It comes with a set of fixtures (helpers) that help perform certain repeatable test tasks (`packages/craftcms-playwright/playwright/fixtures`).

The `packages/craftcms-playwright/README.md` contains installation and usage instructions.

The cms tests are located in the `tests-playwright/tests` directory.

It uses the same data fixtures mechanism as our Codeception tests. I had to make minor tweaks to the fixtures code, so that logging works here too (`src/test/ActiveFixture.php`, `src/test/fixtures/elements/BaseElementFixture.php`; commit: https://github.com/craftcms/cms/commit/b6407e68ab63a21550f9367b1712311108c7a444).

co-authored with @nfourtythree

@gcamacho079, thanks for reviewing and already adding tests!

### Related issues
[cms-602](https://linear.app/craftcms/issue/CMS-602/playwright-tests-package)


brandonkelly:
@i-just Is there a reason the `tests/_craft/storage/runtime/.gitignore` file was removed? Looks like that’s the reason for the PHPStan error.

i-just:
@brandonkelly, no reason; looks like it was removed accidentally. The directory containing it was moved around in the first commit on this branch (12195cf2ea8c20d26b4040aec86206ca263cb14b) and then accidentally removed during one of the subsequent re-shuffles (cfe38c8d4dd295c851fadf7e23e0f3af24888fdc).

I added that file back in and the tests are passing.

Issue #17824, opened 2025-09-04
[5.x]: Allow self relations still clickable


thomascoppein:
### Description
When creating an entry field with the Allow self relations setting disabled, the current entry still appears selectable. Even though it is displayed as greyed out, clicking on it adds it as a relation.

### Steps to reproduce

1. Create a section and an entry type (e.g. Article).
2. Create a field that allows selecting entries from that section.
3. Add the field to the entry type (Article).
4. Create and save a new entry.
5. In the entry, click Add entry on the created field.
6. Notice that the current entry appears greyed out.
7. Click the greyed-out entry → it becomes selected.

Additional scenario:

8. Remove the self-selected entry from the field in the main entry view (without saving entry).
9. Click Add entry again on the created field.
10. Notice that the current entry in the list is no longer greyed out, and can be selected without restriction.

Before clicking:
<img width="301" height="65" alt="Image" src="https://github.com/user-attachments/assets/610d0cc2-38a1-4069-a328-5735c6234c79" />
After clicking:
<img width="314" height="62" alt="Image" src="https://github.com/user-attachments/assets/ec2a7978-54a7-4974-84ef-242f685bbc3a" />
After removing selected entry from field and clicking Add entry again:
<img width="297" height="60" alt="Image" src="https://github.com/user-attachments/assets/0dfaea40-75d2-4320-a4bd-1a41d1e33732" />

### Expected behavior

- When Allow self relations is disabled, the current entry should never be selectable.
- It should always remain greyed out in the entry selection modal, regardless of whether it was previously added and removed.

### Actual behavior

- The current entry can still be selected when it appears greyed out.
- After removing the self-relation from the field, the entry is no longer greyed out in the selection modal and becomes fully selectable again.

### Craft CMS version

5.8.16

### PHP version

8.3

### Operating system and version

macOS

### Database type and version

MySQL

### Installed plugins and versions

- Formie
- ...

i-just:
Hi, thanks for reporting! I raised a PR for this.

brandonkelly:
Craft 5.8.17 is out with a fix for this. Thanks again!

PR #17826, opened 2025-09-04
elements with disabled class shouldn't be selectable

i-just:
### Description
Elements with the “disabled” class shouldn’t be selectable. 

The code added in this PR is already present in the v4 codebase.

### Related issues
#17824


brandonkelly:
Looks like this bug was introduced by #17053.

PR #17834, opened 2025-09-08
Fix twig error in de-CH translation

ursbraem:
A bracket error resulted in the ICU Plural rules not being applied and raw twig being dumped in the cp when alt text was missing on a related asset – in the "Swiss german" translation.

before:

ScreenShot 2025-09-08 at 4  34 06

after:

ScreenShot 2025-09-08 at 4  34 21


ursbraem:
Btw, does it make sense to have a translation for "Swiss" german? (This is about Standard German for Switzerland, not swiss dialect). Apart from us not using `ß` I can't really think of anything that would be different in the "swiss" version.

brandonkelly:
Thanks! I made the same change via [CrowdIn](https://translate.craftcms.com) and merged it in for the next release.

> Btw, does it make sense to have a translation for "Swiss" german?

Not sure about that. Maybe open a [discussion](https://github.com/craftcms/cms/discussions) about it?

ursbraem:
Hi Brandon, thanks and sorry for posting it here! Chatgpt told me this was the right place 🤪

brandonkelly:
@ursbraem No worries! It’s the right place for everything besides translations :)

Issue #12796, opened 2023-03-06
[4.x]: `UrlHelper::buildQuery` decodes query parameter names resulting in invalid html attibutes


leevigraham:
### Description

`UrlHelper::buildQuery` decodes query parameter names resulting in invalid html attibutes. Square brackets are not allowed as part of the url and must be encoded.

### Steps to reproduce

In a twig template

```
{{ url('example', {catIds: [1,2,3]}) }}
```

### Expected behavior

should produce `/example?catIds%5B0%5D=1&amp;catIds%5B1%5D=2&amp;catIds%5B2%5D=3`

### Actual behavior

instead it produces: `/example?catIds[0]=1&amp;catIds[1]=2&amp;catIds[2]=3`

### Possible Fix

I'm pretty sure this line is the issue: https://github.com/craftcms/cms/blob/develop/src/helpers/UrlHelper.php#L87

Commenting it out fixes the issue… but I haven't tested side-effects.

### Craft CMS version

4.4

### PHP version

8.2

engram-design:
I'll mention that this now breaks a URL that contains shorthand Twig (like you mention `{draftId}` @brandonkelly). 

For example, people want to use `?submissionId={id}` as the redirectUrl for Formie forms. This is run through `UrlHelper::url($url, $request->getQueryStringWithoutPath())` so that it retains any existing query strings with what they supply. So the URL might end up being `https://my-site.test?form=contactUs&submission={id}`

But this now encodes the braces to be `submissionId=%7Bid%7D` which doesn't work for when we want to run it through `redirectToPostedUrl()`.

We'll work our way around it for now, but just wanted to flag this.

engram-design:
Also note that this breaks the multi-site dropdown for things like Globals, where changing the site results in:

```
/admin/globals/contact?site=%7Bhandle%7D
```
Due to the use of `{handle}` in the URL.

engram-design:
Cheers for the fix - https://github.com/craftcms/cms/issues/13603

brandonkelly:
@engram-design Yep! Thanks for pointing it out. 2.5.2 is out with this change reverted.

gregorterrill:
[...]

PR #17840, opened 2025-09-09
Don't decode query string names

timkelty:
### Description
`UrlHelper::buildQuery` can produce invalid urls with decoded brackets `?foo[]=bar`.

Because of some special cases with `?var={variables}` in URLs, we need to preserve some characters from encoding, but I'm not sure why we'd ever need to decode the _name_. Are there any known use cases of something like:
`?param[{id}]=value`?

There is a history of fixes & reverts, so even though this is one line, it shouldn't be merged lightly.

I tried to dig through the history a bit:

- initial use of `urldecode`: https://github.com/craftcms/cms/commit/89b54ed896d48123bc169d1cd731f685169b4eb3
- moved to `buildQuery`: https://github.com/craftcms/cms/commit/92fa26c2b379564beaa27070895078e53a015b04
- Deprecate `buildQuery` , _and_ drop urldecode  https://github.com/craftcms/cms/commit/2e97c56dea6231d2b30ad019ea8c10beca5dc40d
  - This PR brings us back to this state, but without deprecating `buildQuery`.
- Revert fix (adds `buildQuery` with `urldecode` back): https://github.com/craftcms/cms/commit/50f50ea84fafebecc97e2fdf92899055b2e2348a

I added some tests and notes where things changed.

### Related issues
- Fixes https://github.com/craftcms/cms/issues/12796
- https://github.com/craftcms/cloud-extension-yii2/issues/98


tests/unit/helpers/UrlHelperTest.php
```diff
@@ -291,15 +291,15 @@ public function buildQueryDataProvider(): array
             ['foo=0', ['foo' => false]],
             ['foo=1', ['foo' => true]],
             ['foo=1&bar=2', ['foo' => 1, 'bar' => 2]],
-            ['foo[0]=1&foo[1]=2', ['foo' => [1, 2]]],
-            ['foo[bar]=baz', ['foo[bar]' => 'baz']],
-            ['foo[bar]=baz', ['foo' => ['bar' => 'baz']]],
```
timkelty, line 296: These brackets should all be expected to be encoded now.

tests/unit/helpers/UrlHelperTest.php
```diff
@@ -291,15 +291,15 @@ public function buildQueryDataProvider(): array
...
+            ['foo%5B0%5D=1&foo%5B1%5D=2', ['foo' => [1, 2]]],
+            ['foo%5Bbar%5D=baz', ['foo[bar]' => 'baz']],
+            ['foo%5Bbar%5D=baz', ['foo' => ['bar' => 'baz']]],
             ['foo=bar%2Bbaz', ['foo' => 'bar+baz']],
-            ['foo+bar=baz', ['foo+bar' => 'baz']],
```
timkelty, line 298: This test had to change. Will it break anything?
I assume in this case, we mean a literal plus (not a space), in which case encoding is correct.

tests/unit/helpers/UrlHelperTest.php
```diff
@@ -441,14 +441,14 @@ public function urlWithParametersDataProvider(): array
                 '?param1=entry1&param2=entry2',
             ],
             '#' => [
-                self::ABSOLUTE_URL_HTTPS_WWW . '?param1=name&param2=name2#anchor',
+                self::ABSOLUTE_URL_HTTPS_WWW . '?param1=value&param2=value2#anchor',
                 self::ABSOLUTE_URL_HTTPS_WWW,
-                ['param1' => 'name', 'param2' => 'name2', '#' => 'anchor'],
```
timkelty, line 446: Tests had confusing transposition of `name` as the value…

tests/unit/helpers/UrlHelperTest.php
```diff
@@ -466,9 +466,34 @@ public function urlWithParametersDataProvider(): array
...
-                self::ABSOLUTE_URL_HTTPS_WWW . '?param1=name&param2=name2',
+                self::ABSOLUTE_URL_HTTPS_WWW . '?param1=value1&param2=value2',
                 self::ABSOLUTE_URL_HTTPS_WWW,
-                '?param1=name&param2=name2',
+                '?param1=value1&param2=value2',
+            ],
+            'query-string-with-token' => [
+                self::ABSOLUTE_URL_HTTPS_WWW . '?param1={value}',
+                self::ABSOLUTE_URL_HTTPS_WWW,
+                '?param1={value}',
```
timkelty, line 476: `{` `}` in value should stay unencoded.

tests/unit/helpers/UrlHelperTest.php
```diff
@@ -466,9 +466,34 @@ public function urlWithParametersDataProvider(): array
...
+            ],
+            'query-string-with-token-name' => [
+                self::ABSOLUTE_URL_HTTPS_WWW . '?param1%5B%7Bkey%7D%5D={value}',
+                self::ABSOLUTE_URL_HTTPS_WWW,
+                '?param1[{key}]={value}',
+            ],
```
timkelty, line 482: `{` `}` in _name_ should be encoded.

tests/unit/helpers/UrlHelperTest.php
```diff
@@ -466,9 +466,34 @@ public function urlWithParametersDataProvider(): array
...
+            'query-string-with-array' => [
+                self::ABSOLUTE_URL_HTTPS_WWW . '?param1%5Bkey%5D={value}&param1%5Bkey2%5D=value2&param2%5Bkey%5D={value3}',
+                self::ABSOLUTE_URL_HTTPS_WWW,
+                '?param1[key]={value}&param1[key2]=value2&param2[key]={value3}',
+            ],
+            'query-string-with-with-non-indexed-array' => [
+                self::ABSOLUTE_URL_HTTPS_WWW . '?param1%5B0%5D=value1&param1%5B1%5D=value2',
+                self::ABSOLUTE_URL_HTTPS_WWW,
+                '?param1[]=value1&param1[]=value2',
+            ],
```
timkelty, line 492: `buildQuery` adds indexes

tests/unit/helpers/UrlHelperTest.php
```diff
@@ -466,9 +466,34 @@ public function urlWithParametersDataProvider(): array
...
+            'query-string-with-forward-slash' => [
+                self::ABSOLUTE_URL_HTTPS_WWW . '?param1=some/path',
+                self::ABSOLUTE_URL_HTTPS_WWW,
+                '?param1=some/path',
```
timkelty, line 496: Forward slash in value is not encoded (though it wouldn't be anyway).

tests/unit/helpers/UrlHelperTest.php
brandonkelly: I don’t think so? Seems pretty likely in this case that `{key}` would end up getting replaced with something later on.

timkelty: We certainly can, but I was starting with the assumption that tokens only appear in the values.

We can decode the same `{` `}` `/` same as we do the value…though I'm not sure we need to do `/`

timkelty: Ok, I updated so those special chars get decoded from the param names as well.

PR #17869, opened 2025-09-17
Fixes an issue where clicking on an autosuggest option triggers the blur event and prevents selection via mouse

gcamacho079:
### Description
The fix addresses an issue where autosuggest options would disappear when clicking on them by:

1. Adding `tabindex="-1"` to the autosuggest container to allow click events to bubble properly
2. Updating the `onBlur` handler to only clear filtered options when focus moves outside the entire autosuggest container, rather than clearing on any focus change

This prevents the options list from disappearing when users interact with suggestion items while maintaining proper cleanup when focus moves to unrelated elements.

Thanks for identifying this and reviewing @i-just!


src/templates/_includes/forms/autosuggest.twig
```diff
@@ -163,8 +163,17 @@ new Vue({
...
-            if (e.relatedTarget) {
+            // Related target - element clicked or that focused moved to.
+            // Current target - the input element that is losing focus.
+            const { currentTarget, relatedTarget } = e;
+
+            // Click event bubbles from listbox to the .autosuggest-container (requires tabindex="-1")
+            // We only want to clear the filteredOptions if the focus moved outside the autosuggest container
+            // (i.e. not to the listbox or one of its children)
+
+            const targetIsOutsideAutosuggestContainer = relatedTarget && !relatedTarget.contains(e.currentTarget);
```
i-just:
```suggestion
            const targetIsOutsideAutosuggestContainer = relatedTarget && !relatedTarget.contains(currentTarget);
```

i-just (approved):

Issue #17892, opened 2025-09-26
[5.x]: Issues with multi-user editing with entries, including incorrect statuses


tomkiss:
### Description

We frequently encounter problems on websites where multiple-users have edited entries.

Normally, what happens is a user loads an article with previously unapplied edits, they apply them, and in the process overwrite edits from other users. This is an oddity we have come to mostly avoid through explanation to different editors.

However, today's example is a little different. It's of an article that was published two weeks ago, as follows:

 - User 1 created the entry. It is disabled.
 - User 2 at some point opened the article, made some sort of edit and exited the entry without saving.
 - User 1 made further edits, changed the title, enabled the entry and then published the entry.
 - Two weeks later, User 2 heads back in to the CMS. Because they have unsaved edits, the entry is appearing in the CMS listing with the little blue [Edited] flag. The title is the old title and the status says "Disabled".
 - User 2 tries to talk to User 1 about the entry, but the conversation is made harder because they're both seeing different things. User 2 can't see that the article has been published and is live, they also can't see that it's been renamed.
 - The problem is resolved by User 2 going in to the article and discarding their edits, but there's quite a bit of figuring out to be done to get to that part...

This seems like a complex area of the systems design but the user experience here can be quite frustrating. Not sure if this is a bug as such, but more of a UX issue. A user should be able to see if something is live or not, and it's current title, regardless of their unapplied edits.

### Craft CMS version

5.8.17

brandonkelly:
I guess part of the problem here is that we only update provisional drafts (which store the unsaved changes) with upstream changes when you actually open the entry edit page. So in theory, the problem would also be resolved if the user had simply opened the edit page; not necessarily saved it or discarded their changes.

In general people seem happy with the current behavior of showing provisional changes. Maybe there’s a better way to handle it though, so only actually edited values are getting pulled into the element index from the provisional draft…

brandonkelly:
Just made some changes that will improve this for Craft 5.9 (#17915). Going forward, only attributes/fields that were actually edited within the provisional draft will be shown on element indexes, etc., but the rest of the data will come from the canonical element.

brandonkelly:
Craft 5.9.0 is out with that change 🎉

tomkiss:
Amazing, thank you Brandon

PR #17915, opened 2025-10-02
Load provisional changes onto the canonical elements

brandonkelly:
### Description

When loading elements to be displayed in the control panel, previously elements would be replaced with their provisional drafts, if any existed for the current user.

With this change, any changed attributes/fields on the provisional drafts will get copied over to the canonical elements, without actually replacing the canonical elements in the array of elements to be shown.

The benefit is that any attributes/fields which haven’t changed in the provisional draft will still show their latest values within element indexes, relational fields, etc.

### Related issues

- #17892


mattbloomfield:
❤️ thank you!

PR #17930, opened 2025-10-08
Move asset bundles into separate package

brianjhanson:
Moves all of our existing asset bundle assets into a `packages/craftcms-asset-bundles` folder and `@craftcms/asset-bundles` subpackage. Much like #17924 this will allow us to split the build process between legacy and modern assets.

The number of changes makes GitHub's diff choke, so important things are:
- Register a new `@assetBundles` alias in the [yii2-adapter](https://github.com/craftcms/cms/blob/e3939d82cc2424ff3afe27fbfc38c2be6d219c59/yii2-adapter/bootstrap/bootstrap.php#L103-L108) that points to `packages/craftcms-asset-bundles`
- Replace all `$sourcePath`s with `@assetBundles` alias paths 
- Move `src` and `dist` directories into `packages/craftcms-asset-bundles`

The build command is now `build:bundles` so you can build all the asset bundles with `npm run build:bundles`. Passing arguments gets a little more awkward unfortunately. To build a specific bundle you have to double up your `--` so the command ends up being

```
npm run build:bundles -- -- --config-name=cp
```

This update unfortunately breaks the `npm run serve ...` commands because that currently relies on the front-end assets being colocated with the asset bundle class in order to resolve the dev server [via the /which-asset route](https://github.com/craftcms/cms/blob/3d2a427f53f1a14f404077e98cd0e20ba9c520ff/packages/craftcms-webpack/index.js#L123-L173). I haven't figured out the best way around that, but it might just be to ignore it. Ideally we won't be doing substantial dev on any of these asset bundles going forward.


brandonkelly:
@brianjhanson looks like this broke some asset bundle tests

brianjhanson:
Found some issues with this. Closing so I can fix those up (and maybe reevaluate this approach)

brandonkelly:
@brianjhanson I take it you’re happy with this now? Can you merge `6.x` into this branch?

Issue #17918, opened 2025-10-02
[3.x-5.x]: Save as New Entry Should Clear Slug


RitterKnightCreative:
### Description
Clients often don't want to start from scratch when creating new entries so one of the easiest ways to get started is to do a save as.  This mostly works. 

However when creating a new entry from an _existing_ entry, the donatee's  slug isn't automatically cleared out so we end up with slugs like `whatever-service-name-2 `instead of the actual title they do end up entering. This is rarely, if ever, optimal or what you want.

No matter how many times I've asked clients to clear the slug before saving, it's a detail they up forgetting.

I know there's ways to automatically set slugs based on the title every time an entry is saved or using URL formats  but those solutions have their own sets of drawbacks.

It would useful if the default slug generation was improved when doing a save as. Craft already knows the duplicated entry is a draft, it seems trivial to clear the slug before presenting the editing screen.   If the default behavior can't be edited, maybe it can check to see if the slug and the title don't match (or it's about to put -1 or -2  on it) and silently (or not so silently) fix the problem—it would probably really only need to do this on the first save/publish.

### Steps to reproduce

1. Pick an existing entry.
2. Use the red drop-down menu and choose "Save as a New Entry"
3. Experience the horror.

### Expected behavior

Entry slugs should match the title upon first save.

### Actual behavior

On duplicated entries the slug is already set even though the client may have entered a new title since.

### Craft CMS version

5.8.17

### PHP version

8.3

### Installed plugins and versions

-

janhenckens:
Seconded!

brandonkelly:
Craft 5.9.0 is out now with this change 🎉 Thanks for the suggestion.

PR #17931, opened 2025-10-09
clear our the slug when saving as new

i-just:
### Description
When saving an element via the “Save as new” alternative action, clear out the slug so that it gets generated from the title on save.

This PR is for v4 only. 

This change has a lesser impact in v4 than in v5, as it will only really affect a case where you have an entry, you edited the title and therefore created a provisional draft (assuming autosaving is on), and you choose to save it as new. In this case, the slug will be generated based on the new title instead of an incremented old slug. If you edited the slug, it will be retained.

### Related issues
#17918


brandonkelly:
Let’s not make this change for v4.

PR #17992, opened 2025-10-27
[6.x] Structures service

riasvdv:


src/Structure/Concerns/StructureNode.php
riasvdv, line 1: This adds the behavior previously provided by `NestedSetBehavior` and `NestedSetQueryBehavior`

PR #18005, opened 2025-10-29
[6.x] Inertia Dashboard

brianjhanson:
When I started this, I thought, "I'll do the dashboard, that's not too complicated.". Guess what world? I was wrong. Turns out there's a lot of functionality packed into our little dashboard. 

Even though this PR isn't finished, I wanted to open it to start getting feedback on the direction things are heading. 

The general plan is to use [Inertia](https://inertiajs.com/) with Vue to handle the rendering of the CP. The Vue pages will then render our web components from the `@craftcms/cp` package (#18010). General layout will be handled by Tailwind utility classes. 

All these together will hopefully help usher in a new era free from the quantum entanglement we find ourselves in with the current CSS. 

Even though this PR is large already, there's still a good list of things that need to be tackled before we'll reach feature parity with the current Dashboard. 

### To-Do
- [ ] New widget items must be keyboard selectable
- [ ] `settingsJs` may need to be evaluated when creating a widget
- [ ] Make sure the window scrolls to the newly created widget
- [ ] Create / Updates should happen in the queue
- [ ] Create an empty state for the widget manager
	- Text should say "You don’t have any widgets yet." to maintain translations
- [ ] Add the ability to undo deleting a widget
- [ ] The layout should be masonry (I had hoped we could get away with grid, but it's kind of yucky)
- [ ] Toast notifications for all operations
- [ ] Make sure to initialize UIElements in newly created widgets (may not be necessary)
- [ ] Append head and body HTML (may not be necessary)
- [ ] Colspan selectors should be `SlidePicker` components
- [ ] Accessibility audit / tests (make sure we're not going backwards)
- [ ] Improve the Widget Manager display and UX
- [ ] Update UI for existing widget types
	- [ ] My Drafts
	- [ ] New Users
	- [ ] Quick Post
- [ ] UI polish (won't be in this PR)


resources/js/composables/useAxios.ts
brianjhanson, line 1: I'm a bit tempted to replace this with [TanStack Query](https://tanstack.com/query/latest) but I'm also a bit wary of hitching ourselves to that bandwagon.

riasvdv: I've found that in general axios/fetch requests are barely necessary when using Inertia as intended and for the few cases where they're needed fetch is enough.

As soon as there's something that's clicked or interacted with by the user for the action you either create a [link](https://inertiajs.com/links) (even if it's a `POST` or `DELETE`) or a [form](https://inertiajs.com/forms).

Especially since it makes it possible to do [partial reloads](https://inertiajs.com/partial-reloads)

PR #18009, opened 2025-10-30
Deprecate categories, global sets, and tags

brandonkelly:
### Description

Deprecates support for categories, global sets, and tags, and moves them out of core, into the yii2-adapter package.

Support for them can be added to new Craft installs via the yii2-adapter-provided commands:

- `craft:add-categories-support`
- `craft:add-global-sets-support`
- `craft:add-tags-support`

And removed via the commands:

- `craft:drop-categories-support`
- `craft:drop-global-sets-support`
- `craft:drop-tags-support`


riasvdv (approved):
Looks good to me, found a few other references and deleted those.

There's also a bunch of translations still for Categories / Globals / Tags but I don't know if it's worth removing those

riasvdv (approved):
Tested the commands, one of them needed foreign keys to be disabled so added that.

I don't know why phpstan is failing on the codecept-build step though 🤔

PR #18011, opened 2025-10-31
various settings bugs (mostly integers and ignore trashed)

i-just:
### Description
1. unique validation (for section and entry type handle, site handle and name, site group name) should ignore soft-deleted sections/entry types
2. the `sections/create` command wasn’t allowing you to select “No” for the `enableVersioning` question
3. in section and entry type controllers, the `edit` and `store` methods should check if the `sectionId` and `entryTypeId` values are `null` and if not, cast those to an integer, as that’s what the `Sections` and `EntryTypes` services accept
4. in the `SitesController`, `store` and `destroy` methods need to cast `siteId` to an integer
5. in the `FieldsController`, the `store` method needs to cast `fieldId` to an integer
6. data objects were ignoring the site’s `baseUrl` and `enabled` private properties; also, the values of those should remain dependent on other values (`baseUrl` on `hasUrls` and `enabled` on `primary`)
7. it was not possible to set entry type’s `showSlugField` and `showStatusField` to `false`
8. it was not possible to create an entry type via slideout from the section’s edit page

### Related issues
n/a

### DIFF
--- a/src/Http/Controllers/Settings/EntryTypesController.php
+++ b/src/Http/Controllers/Settings/EntryTypesController.php
@@ -216,14 +216,13 @@ final class EntryTypesController
     {
         $entryTypeId = $request->input('entryTypeId');
 
-        abort_if(is_null($entryTypeId), 404, "Invalid entry type ID: $entryTypeId");
+        if ($entryTypeId) {
+            $entryTypeId = (int) $entryTypeId;
+            abort_if(is_null($found = $this->entryTypes->getEntryTypeById($entryTypeId)), 400, "Invalid entry type ID: $entryType");
 
-        $entryTypeId = (int) $entryTypeId;
-
-        abort_if(is_null($found = $this->entryTypes->getEntryTypeById($entryTypeId)), 400, "Invalid entry type ID: $entryType");
-
-        $entryType->uid = $found->uid;
-        $entryType->id = $entryTypeId;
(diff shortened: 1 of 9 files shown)


src/Http/Controllers/Settings/EntryTypesController.php
```diff
@@ -89,6 +89,10 @@ public function edit(Request $request, ?int $entryTypeId = null): CpScreenRespon
     {
         $entryTypeId ??= $request->input('entryTypeId');
```
riasvdv, line 90: We could use `$request->integer`, it returns `0` instead of `null` in case it wasn’t passed, that way we don’t have to manually cast it

i-just: Yes, I hesitated between using `->integer` and what’s in this PR. Ended up going with this solution as it makes the errors clearer (in case one is thrown). 

Since I know Brian was also asking about something similar, maybe we should discuss which approach we all prefer and then stick to it throughout.

i-just: That said, I made a few stupid errors here, so let me fix those first 🤦‍♀️ 

We can discuss the preferred approach on Monday and go from there.

i-just:
closing in favour of https://github.com/craftcms/cms/pull/18028

PR #18028, opened 2025-11-05
Bugfix/various settings pages

i-just:
### Description

1. unique validation (for section and entry type handle, site handle and name, site group name) should ignore soft-deleted records
2. the `sections/create` command wasn’t allowing you to select “No” for the `enableVersioning` question
3. use the `create` method for creating a new section, entry type and field
4. use the `edit` method with route model binding for editing the section, entry type, field and site
5. ensure section, entry type, field and site id are cast to int for store and destroy methods, where applicable
6. fixed an issue where the `baseUrl` and `enabled` private `Site` properties were ignored; also, those values should remain dependent on other values (`baseUrl` on `hasUrls` and `enabled` on `primary`)
7. fixed an issue where it was not possible to set entry type’s `showSlugField` and `showStatusField` to `false`
8. fixed an issue where it was not possible to create an entry type via slideout from the section’s edit page

### Related issues
n/a


src/Http/Controllers/Settings/EntryTypesController.php
riasvdv, line 263: Do we want to extend the url parameters for consistency to the destroy method as well? Or do we keep this a POST request with the entryTypeId as body?

I guess we'll see when we get to refactoring the frontend parts.

src/Http/Controllers/Settings/EntryTypesController.php
```diff
@@ -85,17 +86,17 @@ public function create(): CpScreenResponse
             ]);
     }
 
-    public function edit(Request $request, ?int $entryTypeId = null): CpScreenResponse
+    public function edit(EntryTypeModel $entryType): CpScreenResponse
     {
-        $entryTypeId ??= $request->input('entryTypeId');
+        abort_if(is_null($entryType->id), 404, 'Entry type not found');
```
riasvdv, line 91: Laravel already tries to fetch the model and throws a 404 if it doesn't exist, so this extra check isn't necessary. I've adjusted it for all controllers: https://github.com/craftcms/cms/pull/18028/commits/c09b46777e1429fe34475b7c638e953a82b5ccec

src/Section/Data/Section.php
```diff
@@ -93,7 +95,7 @@ public static function rules(?ValidationContext $context = null): array
                 'string',
                 'max:255',
                 new HandleRule(['id', 'dateCreated', 'dateUpdated', 'uid', 'title']),
-                Rule::unique(Table::SECTIONS)->ignore($context->payload['sectionId'] ?? null),
+                Rule::unique(Table::SECTIONS)->ignore($context->payload['sectionId'] ?? null)->withoutTrashed('dateDeleted'),
```
riasvdv, line 98: Added tests for this one and the others https://github.com/craftcms/cms/pull/18028/commits/1f4b462de594442fb0ab42469fe9340c7d390c74

riasvdv (approved):

src/Http/Controllers/Settings/EntryTypesController.php
i-just: I'd keep them as they are, but we can discuss further on Monday if you feel strongly about this?

PR #18063, opened 2025-11-13
Fix installing via composer

brianjhanson:
When I [moved assets into a package](https://github.com/craftcms/cms/pull/17930) I failed to realize that the `packages` directory wasn't included when downloading from composer which meant that the `@assetBundles` alias wouldn't resolve properly.

This updates the `craftcms-asset-bundles` build to throw the dist files into the `resources/build/legacy` folder so they can safely be referenced by Laravel.

This has me rethinking my plan of having asset bundles in a separate package, but that's a problem for another day. 

The majority of these changes are just dist files, the only relevant changes are
- Updating the `craftcms-asset-bundles/webpack.config.js` https://github.com/craftcms/cms/compare/6.x...asset-fix?expand=1#diff-03a17eac7513a47bfd95751597bf5b85f16d366897b3688056a33286be04d5f6
- Updating `bootstrap.php` to reference the new folder https://github.com/craftcms/cms/compare/6.x...asset-fix?expand=1#diff-8b8f19985b653f1ad4bfdebe2ff358d2ff70abded9d45e605ba0ffed0e4516c4


leevigraham:
Tried again to install via composer. No bueno…

Looks like "cms" is missing from the path.

Screenshot 2025-11-15 at 22 30 23

brianjhanson:
Bummer. I tried to test it against this branch but the `self.version` in the Yii2 Adapter foiled my plans. I'll keep looking into it next week and hopefully sort it out.

PR #18064, opened 2025-11-13
Move installer to inertia

brianjhanson:
Biting off a bit less than #18005 — moves the installer to Inertia. Still a lot of plumbing included in this one, but an easier first step.

Nothing super notable changed in the process the installer (for now) still behaves like you know and love. 

This isn't even super inertia powered because Inertia doesn't really love multi-step forms that all appear on the same page.

In the future we might want to refactor this to be separate pages while keeping track of where a user is in the session, but that's for another day. 

https://github.com/user-attachments/assets/3539c442-d896-4038-841f-a0b1e6e58eeb

This PR does not have an autocomplete component for the Site URL because we haven't built that component yet and it's a bit more complex than I wanted for this PR. 

This also doesn't include Playwright tests because those seem to be broken in 6.x at the moment, going to be looking into that soon.


resources/js/pages/Install.vue
```diff
@@ -0,0 +1,316 @@
...
+        </template>
+
+        <template v-else-if="isCurrent('installing')">
+          <InstallingScreen :data="formData" @success="goToNext()" />
+        </template>
+        <template v-else-if="isCurrent('complete')">
+          <div
+            class="tw:p-8 tw:grid tw:justify-center tw:content-center tw:gap-4"
+          >
+            <h2>{{ t('app', 'Craft is installed! 🎉') }}</h2>
```
riasvdv: The new `t()` php function puts the category behind the message and parameters and defaults to `'app'`, we might want to do this on the frontend as well for consistency. It's almost always 'app' in our context so it's less clutter so this would become `t('Craft is installed! 🎉')`

src/Http/Controllers/InstallController.php
```diff
@@ -60,38 +61,57 @@ public function index(): Response
...
         // Grab the license text
         $licensePath = Aliases::get('@craftcms/LICENSE.md');
         $license = file_get_contents($licensePath);
+        $licenseHtml = Str::markdown($license);
 
         // Guess the site name based on the server name
         $defaultSystemName = $this->defaultSiteName();
         $defaultSiteUrl = $this->defaultSiteUrl();
         $defaultSiteLanguage = $this->defaultSiteLanguage();
+        $locales = Craft::$app->getI18n()->getAllLocales();
```
riasvdv: There's a new `CraftCms\Cms\Support\Facades\I18N` that can be used to use the new service.

This is not something that needs to change in this PR though as I'd probably go through the controllers myself to refactor them to use the laravel-data Dto objects for validation and I can take other refactors in with that as well.

src/Http/Controllers/InstallController.php
```diff
@@ -60,38 +61,57 @@ public function index(): Response
...
+        $connection = config('database.default');
+        $dbConfig = config("database.connections.{$connection}");
```
riasvdv:
```suggestion
        $dbConfig = DB::getConfig();
```

I think this results in the same data

src/Providers/AppServiceProvider.php
```diff
@@ -84,6 +84,9 @@ public function boot(): void
         $this->setTimezone();
         $this->setNamespace();
         $this->bootAliases();
+        $this->loadViewsFrom("{$this->root}/resources/views", 'c');
+
+        Vite::useHotFile("{$this->root}/resources/hot");
```
riasvdv: We already have this one as well https://github.com/craftcms/cms/blob/6.x/src/Twig/TwigServiceProvider.php#L15

I'd probably add a `ViewServiceProvider` and merge some of the things into that.

src/Cms.php
```diff
@@ -41,4 +41,9 @@ public static function systemName(): string
 
         return $name ?: config('app.name', 'Craft');
     }
+
+    public static function cpLocale(): string
+    {
+        return config('app.locale', 'en-US');
+    }
```
riasvdv: There's `app()->getLocale()` you can use globally.

riasvdv (approved):

resources/js/pages/Install.vue
brianjhanson: Yeah, I noticed that, I'll tackle that refactor in another PR though (hopefully before we have _too_ many `t` calls in JS)

packages/craftcms-cp/src/components/select/select.ts
```diff
@@ -1,62 +1,36 @@
...
-import {LionSelectInvoker, LionSelectRich} from '@lion/ui/select-rich.js';
-import {css, html} from 'lit';
-import type CraftIcon from '../icon/icon.js';
-import {baseInputStyles} from '../../styles/form.styles.js';
+import {LionSelect} from '@lion/ui/select.js';
+import {html} from 'lit';
 import '../option/option.js';
 import '../icon/icon.js';
 
-export default class CraftSelect extends LionSelectRich {
```
brianjhanson, line 9: I had originally planned to use the [Rich Select](https://lion.js.org/components/select-rich/overview/) for everything in the CP, but ran into performance issues when rendering the number of options for the site locale.

routes/cp.php
brianjhanson, line 1: @riasvdv I think this is because of the rector rules. I forget, is this what we want?

riasvdv: Yes but we should just import it at the top 👍

Issue #18052, opened 2025-11-12
[5.8.19]: Mobile re-order assets doesn't work


white-franks:
### Description
If you go to Craft in mobile view, the re-ordering of Assets by moving the by the arrow button doesn't work. See the video attached. It works fine in desktop view.

https://github.com/user-attachments/assets/8c260896-aa30-4b33-be9d-ea990b2f3532

### Craft CMS version

5.8.19

### Installed plugins and versions

-

i-just:
Hi, thanks for reporting! On mobile devices, you can use the “move” actions from the 3-dots menu. 

<img width="949" height="802" alt="Image" src="https://github.com/user-attachments/assets/170bf19b-f4f5-4a93-9341-408594580887" />

We also made some improvements around this for 5.9.

brandonkelly:
Craft 5.9.0 is out now with the changes in #18067 🎉

PR #18067, opened 2025-11-14
Bugfix/18052 drag sort and devices without mouse pointer events

i-just:
### Description
1. adds move up/down actions to the default table columns field in the customise sources modal; the `Craft.ui.createCheckboxSelectField()` method now accepts new `includeSortActions`. When set to `true`, `Craft.SortableCheckboxSelect` is used for the checkboxes, instead of regular `Garnish.CheckboxSelect` with separate `Garnish.DragSort` initialisation.

2. adds move up/down actions to the `Craft.AdminTable`; this affects e.g. settings > sites > site group; dashboard widgets

3. The 3-dot menu that allows access to the “Customise sources” action is now visible on mobile devices too (element index page; `BaseElementIndex`). 

4. element index actions are now showing on mobile devices (`ElementIndexesController`)

5. prevents initialisation of `Garnish.DragSort` and hides the diamond icon (used for drag-sorting) when the device doesn’t have mouse pointer events.

    Here’s where the changes can be tested:
    - `AdminTable` => settings > sites > site group; dashboard widgets
    - `BaseElementSelectInput` => relation fields (e.g. entries field when viewed on entry edit page)
    - `CardsElementIndexView` => matrix in element index view mode when viewed on entry edit page
    - `ComponentSelectInput` => settings > section edit page entry types field
    - `CustomizeSourcesModal` => Entries index page > customise sources
    - `FieldLayoutDesigner` => settings > entry type's edit page > card view designer
    - `GroupedEntryTypeManager` => settings > matrix field > entry types section
    - `NestedElementManager` => matrix field in cards view mode when viewed on e.g. entry edit page
    - `SortableCheckboxSelect` => settings > entry type's edit page > card view designer
    - `MatrixInput` => matrix field in blocks mode when viewed on e.g. entry edit page
    - `routes` => settings > routes

### Related issues
Resolves PT-2007, PT-1970, PT-78, #18052


gcamacho079:
@i-just I'm reviewing on my iPad (no pointer devices) and it looks like ~~_Craft.hasMousePointerEvents_ is evaluating to true there as well. I'm playing around with some alternatives.~~

Strike that for the record. When I was testing in the Customize Sources modal, it seemed like the `.move` handles were removed from the DOM after the handles were hidden and replaced with new (visible) handles somewhere along the line. 🤔 Addressed in the PR I just made by adjusting where we were hiding them.

gcamacho079 (changes requested):
I’ve made a PR with this branch as the target, incorporating a couple of tweaks. I’m still experiencing some issues with drag handles being visible, and I haven’t yet been able to track the source.

For example, on the categories index page, I’m seeing drag handles on Safari/iOS but not in Chrome/macOS; but they _are_ visible in Firefox/macOS.

i-just:
Thanks @gcamacho079! I just merged your PR into this one and also merged the latest 5.9 in here.

I also made an adjustment for the editable table (e.g. generated fields table or table field settings).

As far as I can tell, the places that still have drag-sorting and diamond icons are:
- element index pages when viewing a structure
- places where `VueAdminTable` is used; e.g. Settings > Assets - reordering volumes; I know it’s still an accessibility issue, but fwiw that seems to work on the Android devices and on various iOS simulators I tried

When testing, it looks like `Craft.hasMousePointerEvents` will evaluate to: 
- in Firefox on macOS, => `true` => incorrect, but on a real mobile device, on Android, it’ll correctly evaluate to `false`
- in Chrome on macOS => `false` => correct
- in Safari on iOS (via a simulator) => `false` => correct

The categories index page - I can see the drag handles for that one everywhere, but that’s a structure, so it’s expected. Or am I misunderstanding what page you had in mind?

gcamacho079:
> The categories index page - I can see the drag handles for that one everywhere, but that’s a structure, so it’s expected. Or am I misunderstanding what page you had in mind?

Nope, that was a misunderstanding on my end! 🙏🏼

gcamacho079 (approved):

brandonkelly:
Great work!!

PR #18083, opened 2025-11-19
[6.x] Fix assets

brianjhanson:
We'll call #17930 a failed experiment. Something about having the asset bundles in their own package (and with their own build process) seems to cross a few wires. I'mg uessing it has to do with the mono repo setup, but I wasn't able to track down anything in particular. I just got things to work when I reverted what I did previously. 

This moves the asset bundles into the `yii2-adapter` package. I was originally trying to avoid this because I thought we might be able to use some of the asset bundles for the new CP, but I think that's becoming less likely. 

I'd still like to come up with some kind of system where we can maintain the old asset bundles and new front-end assets without a huge headache, but I think that's a problem we can solve another day.

Opening as a draft because I still need to work through some CSS variable issues, but I'm hoping this is far enough that @lukeholder and @riasvdv can make sure this is better and not worse.


riasvdv:
@brianjhanson I went ahead and merged this as it solves the most pressing issues

PR #18092, opened 2025-11-24
multi-instance relation fields; empty/notempty querying

i-just:
### Description
When querying a multi-instance relation field with an `:empty:`/`:notempty:` value, the query results don’t limit the results to only the instance used in the query. The exists condition only takes into consideration the field ID, not the field instance UID.

**Steps to reproduce:**
- clean 5.8.20 installation
- create a local filesystem, a volume that uses it and an Assets field (handle `assets`)
- create a section with an entry type that pulls in the above Assets field twice (`assets` and `assets2`)
- in that section, create five entries; entries 1, 2 and 4 should have a value in the `assets` field; entry 3 should have a value in the `assets2` field
- create a template with the following code in it:
```
{% set results = craft.entries().section('<sectionHandle>').assets2(':empty:').all() %}
    {% if results|length == 0 %}
        <p>no matches</p>
    {% else %}
        {% for result in results %}
            <p>{{ result.title }}</p>
        {% endfor %}
    {% endif %}
```
- the result will be only entry 5, but it should be entries 1, 2, 4 and 5
- for `:notempty:`, the results will be entries 1, 2, 3, 4, but it should be entry 3
- if you switch to querying the `assets` field, the results will also be wrong

### Related issues
n/a, reported by @tommysvr

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Fixed a bug where remove buttons within multi-select Selectize inputs weren’t working if the input wasn’t focusend and fully in view. ([#18079](https://github.com/craftcms/cms/issues/18079))
 - Fixed a bug where font icons weren’t hidden from screen readers. ([#18078](https://github.com/craftcms/cms/pull/18078))
+- Fixed a bug where relation fields weren’t handling `:empty:`/`:notempty:` element query params properly if the field had multiple instances within a field layout. ([#18092](https://github.com/craftcms/cms/pull/18092))
 
 ## 5.8.20 - 2025-11-18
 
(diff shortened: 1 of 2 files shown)


brandonkelly:
This is tricky, because the `content` column value for relational fields could be missing for relational fields that predate Craft 5.2, even for elements with saved relations.

Looks like we solved this for `RelationalFieldConditionRule::modifyQuery()` by only calling `BaseRelationField::existsQueryCondition()` for single-instance fields; otherwise going with a condition based on `getValueSql()`. So I just updated `BaseRelationField::queryCondition()` to be more like that when dealing with `:empty:`/`:notempty:`, which is working reliably in my testing for both pre-5.2 fields and post-5.2/multi-instance fields.

The only caveat is that once a pre-5.2 field goes multi-instance, you need to resave its elements (so the `content` table gets populated with the relation arrays) before those queries will work as expected. But that was already the case for “has a value” and “is empty” relation field condition rules, so that’s fine.

Guessing this will resolve #18095 as well.

brandonkelly (approved):

PR #18097, opened 2025-11-24
[6.x] Inertia General Settings

brianjhanson:
Updates the general settings page to be powered by Inertia. 

https://github.com/user-attachments/assets/bc9c42f9-a6d4-4415-9b63-8396f23f992d

This PR does take a first run at establishing what we mean by `appareance` and `variant` with the `<craft-callout/>` component. At the moment I'm thinking we'll try to hold the system to a handful of variants (`default`, `success`, `warning`, `danger`, and `info`) and a handful of appearances (`outline`, `fill`, `outline-fill`, `accent`, `plain`). This is pretty much the same system as Web Awesome employs. This doesn't establish a mixin or any way of applying those same properties to multiple components, but we'll want to do that before too long. 

This contains initial implementations of both the `craft-combobox` and `craft-input-file`. In the future, we may need to bring in some external libraries to build out more functionality for these, for that reason I didn't spend too much time making them perfect. 

For file uploads, I'm currently looking at [FilePond](https://pqina.nl/filepond/) which seems to be a pretty robust open source solution. They also have an [Image Editor](https://pqina.nl/pintura/?ref=filepond) we could potentially use to replace our custom  implementation. 

### To-Do
- [x] Read only mode for settings
- [x] If you have a file uploaded the "Choose file" doesn't work
- [ ] ~~When you have a combobox focused and save via `cmd+S` the list will close and re-open after save~~ — I'm punting on this one for now because I'll need to review the accessibility of this thing either way.
- [x] `CMD+S` hint should switch to `CTRL+S` for windows machines

### Related issues
Resolves PT-3002


src/Http/Controllers/Settings/GeneralSettingsController.php
```diff
@@ -22,17 +21,26 @@ public function __construct(
...
+                'readOnly' => ! $generalConfig->allowAdminChanges,
+            ]);
+        }
 
-        return view('craftcms::settings/general/_index', [
+        return Inertia::render('Settings/General/Index', [
             'system' => $this->projectConfig->get('system') ?? [],
+            'timezones' => DateTime::getTimeZoneOptions(),
             'readOnly' => ! $generalConfig->allowAdminChanges,
+            'save_url' => route('craft.cp.settings.general.store'),
```
riasvdv: Probably want to decide if we snake case or camel case props

src/Http/Controllers/Settings/GeneralSettingsController.php
```diff
@@ -46,6 +54,9 @@ public function store(Request $request): Response
 
         $this->projectConfig->set('system', $systemSettings, 'Update system settings.');
 
-        return $this->asSuccess(t('General settings saved.'));
+        $request->session()->flash('message', 'System settings saved.');
+
+        return to_route('craft.cp.settings.general.index')
```
riasvdv:
```suggestion
        return back()
```

src/Http/Controllers/Settings/GeneralSettingsController.php
riasvdv: With [Wayfinder](https://github.com/laravel/wayfinder) URLs like these wouldn't need to be passed

src/Http/Middleware/HandleInertiaRequests.php
```diff
@@ -39,8 +47,55 @@ public function version(Request $request): ?string
     #[\Override]
     public function share(Request $request): array
     {
+        $sitesService = Craft::$app->getSites();
+        $currentSite = $sitesService->getCurrentSite();
```
riasvdv:
```suggestion
        $currentSite = \CraftCms\Cms\Support\Facades\Sites::getCurrentSite();
```

src/Http/Middleware/HandleInertiaRequests.php
```diff
@@ -39,8 +47,55 @@ public function version(Request $request): ?string
...
+        $isInstalled = Craft::$app->getIsInstalled();
```
riasvdv:
```suggestion
        $isInstalled = \CraftCms\Cms\Shared\Models\Info::isInstalled();
```

src/Cp/Navigation.php
```diff
@@ -0,0 +1,203 @@
...
+use CraftCms\Cms\Utility\Utilities;
+use CraftCms\Cms\Utility\Utility;
+
+use function CraftCms\Cms\t;
+
+class Navigation
+{
+    public static function getItems(): array
+    {
+        $isAdmin = Craft::$app->getUser()->getIsAdmin();
```
riasvdv:
```suggestion
        $isAdmin = \Illuminate\Support\Facades\Auth::user()->isAdmin();
```

src/Cp/Navigation.php
```diff
@@ -0,0 +1,203 @@
...
+        $generalConfig = app(GeneralConfig::class);
```
riasvdv:
```suggestion
        $generalConfig = \CraftCms\Cms\Cms::config();
```

src/Cp/Navigation.php
```diff
@@ -0,0 +1,203 @@
...
+
+        $navItems = [
+            [
+                'label' => t('Dashboard'),
+                'url' => 'dashboard',
+                'icon' => 'gauge',
+            ],
+        ];
+
+        if (Craft::$app->getEntries()->getTotalEditableSections()) {
```
riasvdv:
```suggestion
        if (\CraftCms\Cms\Support\Facades\Sections::getTotalEditableSections()) {
```

src/Cp/Navigation.php
```diff
@@ -0,0 +1,203 @@
...
+            $navItems[] = [
+                'label' => t('Assets'),
+                'url' => 'assets',
+                'icon' => 'image',
+            ];
+        }
+
+        if (
+            Edition::get() !== Edition::Solo &&
+            Craft::$app->getUser()->checkPermission('viewUsers')
```
riasvdv:
```suggestion
            Auth::user()->can('viewUsers')
```

src/Cp/Navigation.php
```diff
@@ -0,0 +1,203 @@
...
+        ) {
+            $navItems[] = [
+                'label' => t('Users'),
+                'url' => 'users',
+                'icon' => 'user-group',
+            ];
+        }
+
+        // Add any Plugin nav items
+        $plugins = app(Plugins::class)->getAllPlugins();
```
riasvdv: I'd probably add all the necessary services to the constructor and have Laravel inject those when creating the navigation object

src/Edition.php
```diff
@@ -161,4 +162,13 @@ public static function require(Edition|int $edition, bool $orBetter = true): voi
             throw new WrongEditionException("Craft $edition->name is required for this.");
         }
     }
+
+    public function toArray()
```
riasvdv:
```suggestion
    public function toArray(): array
```

riasvdv (changes requested):
The Inertia part looks good to me, just some small suggestions / updated services

src/Http/Controllers/Settings/GeneralSettingsController.php
brianjhanson: I kind of like how explicit `to_route` is, but I don't feel super strongly about it either.

src/Http/Controllers/Settings/GeneralSettingsController.php
brianjhanson: I like camel. This one slipped through the cracks because I copied it from an example.

src/Cp/Navigation.php
brianjhanson: For no particular reason I wanted `getItems` to be `static` which is why I did this. I've updated all of these `app()` calls to be injected via the constructor now.

src/Http/Controllers/Settings/GeneralSettingsController.php
riasvdv: Not a strong opinion about it either, `to_route` is fine 👍

riasvdv (approved):

deleugpn:
probably worth it ignoring wayfinder compiled files so they don't pollute git diff (here in this PR or future PRs)

src/Cp/Navigation.php
```diff
@@ -0,0 +1,215 @@
...
+    public function __construct(
+        private readonly Plugins $plugins,
+        private readonly Utilities $utilities,
+        private readonly GeneralConfig $generalConfig
+    ) {}
+
+    public function getItems(): array
+    {
+        $isAdmin = Auth::user()?->isAdmin();
+        $generalConfig = Cms::config();
```
deleugpn: what's the difference between this `$generalConfig` and `$this->generalConfig`?

src/Cp/Navigation.php
```diff
@@ -0,0 +1,215 @@
...
+        // Fire a 'registerCpNavItems' event
+        // @TODO Bring this back
+        // if ($this->hasEventHandlers(self::EVENT_REGISTER_CP_NAV_ITEMS)) {
+        //     $event = new RegisterCpNavItemsEvent(['navItems' => $navItems]);
+        //     $this->trigger(self::EVENT_REGISTER_CP_NAV_ITEMS, $event);
+        //     $navItems = $event->navItems;
+        // }
+
+        // Figure out which item is selected, and normalize the items
+        $path = Craft::$app->getRequest()->getPathInfo();
```
deleugpn: would probably be nice if Request object could also be injected in the constructor. It makes it immediately clear that this class is bound to Http context

resources/js/pages/SettingsGeneralPage.vue
```diff
@@ -0,0 +1,260 @@
...
+  .preview--icon {
+    aspect-ratio: 1;
+    width: 32px;
+  }
+  .preview--logo {
+    aspect-ratio: 16/9;
+    width: 288px;
+    height: auto;
+  }
+</style>
```
deleugpn, line 409: minor suggestion: when working on large projects with hundreds of `Index.vue` files, it gets very annoying to "Open File" on IDE. Its nice to have file names matching the page that they represent.

On a similar note, I also like to add a suffix to every inertia page (i.e. `GeneralSettingsPage.vue`) so that its super easy to differentiate between Vue components that are rendered by Inertia vs Vue components that are used internally within Vue itself

src/Http/Controllers/Settings/GeneralSettingsController.php
```diff
@@ -41,11 +50,14 @@ public function store(Request $request): Response
...
+            $systemSettings['live'] = $request->boolean('live');
         }
 
         $this->projectConfig->set('system', $systemSettings, 'Update system settings.');
 
-        return $this->asSuccess(t('General settings saved.'));
+        $request->session()->flash('message', 'System settings saved.');
+
+        return back()
+            ->with('success', 'System settings saved.');
```
deleugpn, line 93: do you need both `$request->session()->flash()` and `back()->with()`? Wouldn't just one of them suffice?

src/Cp/Navigation.php
riasvdv: Those are the same, so `$this->generalConfig` should be used as it's injected in the constructor

src/Cp/Navigation.php
riasvdv: This is a legacy `Craft::` call which will definitely be refactored away, but always good to do in new code which means less work later

src/Http/Controllers/Settings/GeneralSettingsController.php
riasvdv: There's a `\CraftCms\Cms\Support\Flash` class to flash messages (currently calls legacy session flash messages for backward compatibility)

We could add another flash to the session in there to the "new" flash messages so we can keep using 1 helper and have flash messages show up correctly whether it's inertia or legacy.

brianjhanson: I don't think we need to worry about the legacy stuff here (correct me if I'm wrong) so I just removed the `$request->session()->flash()` call

routes/cp.php
```diff
@@ -81,13 +82,29 @@
...
+        // Rebrand
+        Route::get('rebrand/{type}/{filename}', function (string $type, string $filename) {
+            abort_unless(in_array($type, ['icon', 'logo']), 404);
+
+            $file = Storage::disk('rebrand')->path("$type/$filename");
+
+            abort_unless(file_exists($file), 404);
+
+            return response()->file($file);
+        })->where('filename', '.*');
```
brianjhanson, line 107: This route essentially just makes files in the `storage/rebrand/{icon|logo}` directory publically accessible. 

I know this will eventually be refactored, but if there's something we should do better now, let me know.

src/Cp/Rebrand.php
brianjhanson, line 1: This is much stripped down rebrand service. I would imagine all of this will be replaced when we move Rebrand assets more into the fold.

src/Http/Controllers/Settings/GeneralSettingsController.php
```diff
@@ -4,48 +4,113 @@
...
+                        Storage::disk('rebrand')->deleteDirectory($folder);
+
+                        if ($image) {
+                            $safeName = Assets::prepareAssetName($image->getClientOriginalName(), true, true);
+                            $image->storePubliclyAs($folder, $safeName, 'rebrand');
+                        }
+                    }
+                }
+            }
+        }
```
brianjhanson, line 78: This logic works, but it feels a bit cumbersome to me. I'm not sure how how best to make it better though. 

On initial file upload, we'll get a file we can happily access via `$request->hasFile()` but when we return a file to the front-end we just give it an array of the `name` and `url`. 

That means if you save the page and haven't edited  on an edit or something, the icon / logo will be returned as that array which Laravel doesn't need to do anything with and should ignore

The solution I came up with is to remove `siteIcon` and `siteLogo` from the POST if they haven't been messed with. It works but doesn't feel ideal.

src/Edition.php
```diff
@@ -140,6 +141,17 @@ public static function canUpgrade(): bool
...
+    public static function isAtLeast(Edition|int $edition): bool
+    {
+        try {
+            self::require($edition);
+
+            return true;
+        } catch (WrongEditionException) {
+            return false;
+        }
+    }
```
brianjhanson: @riasvdv I think we could replace a lot of conditionals with this if we like it. If we don't, I can just remove it from the PR and adjust my logic to match elsewhere's.

yii2-adapter/legacy/controllers/RebrandController.php
brianjhanson, line 1: We could probably just delete this controller entirely? I don't think anyone but us is using it

src/Edition.php
riasvdv: I'd change the signature to `is(Edition|int $edition, bool $orBetter = true)` to keep it consistent with the `require` one, except that this one returns a bool and doesn't throw

yii2-adapter/legacy/controllers/RebrandController.php
riasvdv: Yeah, I've been removing legacy controllers when there are replacements for the functionality

routes/cp.php
riasvdv: Seems fine for now 👍 maybe add a `/** @TODO: Remove when rebrand assets are refactored **/` just in case we'd forget

src/Shared/Rules/RequiresEditionRule.php
```diff
@@ -0,0 +1,25 @@
...
+namespace CraftCms\Cms\Shared\Rules;
+
+use Closure;
+use CraftCms\Cms\Edition;
+use Illuminate\Contracts\Validation\ValidationRule;
+
+class RequiresEditionRule implements ValidationRule
+{
+    public function __construct(
+        private readonly Edition $requiredEdition
```
riasvdv: I'd add the `bool $orBetter = true` to the constructor here as well to provide flexibility

src/Http/Controllers/Settings/GeneralSettingsController.php
riasvdv, line 78:
```suggestion
if (Edition::isAtLeast(Edition::Pro)) {
    foreach (['siteIcon', 'siteLogo'] as $key) {
        // Resolve the input name to the folder
        $folder = $key === 'siteIcon' ? 'icon' : 'logo';

        // Make sure the user actually wants to change this
        if (! $request->has($key)) {
            continue;
        }

        // If the user explicitly removed the file just delete the folder
        if ($request->input($key) === null) {
            Storage::disk('rebrand')->deleteDirectory($folder);
        }

        // If the user uploaded a new file
        if (! $request->hasFile($key)) {
            continue;
        }
    
        $image = $request->file($key);
        Storage::disk('rebrand')->deleteDirectory($folder);

        if ($image) {
            $safeName = Assets::prepareAssetName($image->getClientOriginalName(), true, true);
            $image->storePubliclyAs($folder, $safeName, 'rebrand');
        }
    }
}
```

Using some early returns would at least make it more readable with less nesting. I'd probably also refactor this to a method like `updateRebrand()` which does nothing when the edition isn't Pro

> The solution I came up with is to remove siteIcon and siteLogo from the POST if they haven't been messed with. It works but doesn't feel ideal.

I know we're using POST here, but that's a bit what a PATCH request by design would expect and doesn't seem like a wrong thing where you only send the data which you want updated. We could make this more explicit by having it use the PATCH verb as well.

riasvdv: Seems like the reply went to a comment instead https://github.com/craftcms/cms/pull/18097/files/8382d1e857ce02e1f40a48dbcff990840fb4109f#r2615602098

src/Edition.php
brianjhanson: I added the param, but left the name as `isAtLeast`. I don't love that it's hard to tell if it's inclusive or not on a light read, but I think it's better than `is` which would imply strict equality to me.

src/Http/Controllers/Settings/GeneralSettingsController.php
brianjhanson: @riasvdv I messed with making this a PATCH and only sending up changed values, but I don't think it's worth it at the moment. As far as I can tell, the inertia `form` helper doesn't have a way to just get the values that have changed, so you end up having to do that manually. I might revist down the road as we're doing more of this kind of thing though.

resources/js/components/MainNav.vue
```diff
@@ -6,7 +6,7 @@
 </script>
 
 <template>
-  <craft-navigation label="Primary">
+  <craft-nav-list label="Primary">
```
gcamacho079: Since we aren’t using the `<nav>` element anymore, we can remove the `label` property here.

riasvdv (approved):

PR #18106, opened 2025-11-26
[6.x] `craft-action-menu` component

brianjhanson:
Builds on #18097 and Adds a `craft-action-menu` component that can be used where `Garnish.DisclosureMenu` is used now. On the settings page that just means the "Save and continue edition" action item. 

CleanShot 2025-11-26 at 12 29 59

#### Exmaple usage
```html
<craft-action-menu ?opened="${open}">
  <craft-button type="button" icon type="button" slot="invoker">
    <craft-icon name="ellipsis"></craft-icon>
  </craft-button>
  
  <craft-action-item icon="up-right-from-square" href="#">View in a new tab</craft-action-item>
  <craft-action-item icon="eye" href="#">Preview File</craft-action-item>
  <hr />
  <craft-action-item icon="download">Download</craft-action-item>
  <craft-action-item icon="folder-open">Show in folder</craft-action-item>
  <craft-action-item icon="pen">Edit asset</craft-action-item>
  <hr />
  <craft-action-item icon="pen">Open in Image Editor</craft-action-item>
  <hr />
  <craft-action-item variant="danger" icon="rotate">Replace</craft-action-item>
  <craft-action-item variant="danger" icon="x">Remove</craft-action-item>
</craft-action-menu>
```

Which will result in:
CleanShot 2025-11-26 at 12 23 04


brianjhanson:
@gcamacho079 I added [CMS-1692](https://linear.app/craftcms/issue/CMS-1692/audit-craft-action-menu) to track the A11y audit of this. I still need to go through `Garnish.DisclosureMenu` and make sure I didn't miss anything super obvious but we're leaning on [Lion's Overlay functionality](https://lion.js.org/fundamentals/systems/overlays/overview/) so I'm hoping they've done the majority of the work for us

gcamacho079:
@brianjhanson, one issue I’m seeing with the icon buttons in general that we should address in the design system is to ensure they always have an accessible name.

Can we require that an icon button always have an accessible name? In the case of icons, this would usually be:
- A visually-hidden text element inside the button element, or
- Requiring that the icon itself have alternative text, if it’s not explicitly marked as decorative

Edited to add: I'm wondering if having an explicit label property and having a separate icon slot would make it easier to ensure there’s always an accessible name.

packages/craftcms-cp/src/components/action-menu/action-menu.ts
```diff
@@ -0,0 +1,74 @@
...
+    }
+  `;
+
+  @queryAssignedElements({selector: 'craft-action-item'})
+  actionItems!: CraftActionItem[];
+
+  // @ts-ignore
+  _defineOverlayConfig() {
+    return {
+      placementMode: 'global',
```
gcamacho079: Curious if there's a reason we're using global versus local placement. It looks like the `global` placement mode might be overriding the dropdown config as I'm seeing a `dialog` when I inspect this:

A dialog element inside the craft-action-menu shadow root

This may have to do with the [placement mode overriding some of the configs for dropdown-style overlays](https://lion.js.org/fundamentals/systems/overlays/configuration/#istooltip-placementmode-local)

gcamacho079: Nevermind, I read the [Considerations](https://lion.js.org/fundamentals/systems/overlays/rationale/#considerations) part of the Overlay docs that explains this. 🤦🏼 Still curious about the `global` placement though

brianjhanson: I used `global` mostly because I've found that to have fewer caveats in [Floating UI](https://floating-ui.com/) (which is the successor to the library used in Lion)

brianjhanson:
> Can we require that an icon button always have an accessible name? 

I think it would probably be best to have an explicit `<craft-icon-button>` element or something that has a "required" `label` field. I'm not sure how hard of a requirement we can make it, but we can try.

gcamacho079 (changes requested):
A couple things in addition to the accessible name:

- [ ] After the disclosure is opened and the user tabs through and _then out of_ the disclosure menu, the disclosure should close
- [ ] An `aria-controls` attribute that points to the ID of the action menu container

brianjhanson:
>  After the disclosure is opened and the user tabs through and then out of the disclosure menu, the disclosure should close

I refactored things a bit and this seems to be happening now. 

> An aria-controls attribute that points to the ID of the action menu container

I've added `aria-controls` to the `craft-button` that will point to the first element within the `content` slot. Is that correct? Or should the ids be placed on different elements? Is there a way for me to test this?

gcamacho079:
@brianjhanson there should be a top-level wrapper surrounding every element in the content slot, since the trigger button controls expansion of that entire container.

A disclosure button pointing to the first element inside the disclosure via aria-controls

So in the above screenshot, there should be another element wrapping both `div[slot="content"]`, and that should be where the ID is placed. (See the [Example Disclosure for FAQ](https://www.w3.org/WAI/ARIA/apg/patterns/disclosure/examples/disclosure-faq/) in the APG documentation for an example of the markup)

Something else - it seems the first element is also getting a `role="dialog"` attribute; this should be removed.

The disclosure still isn't closing for me after I tab through all the contents. 🤔

brianjhanson:
> So in the above screenshot, there should be another element wrapping both div[slot="content"], and that should be where the ID is placed. (See the [Example Disclosure for FAQ](https://www.w3.org/WAI/ARIA/apg/patterns/disclosure/examples/disclosure-faq/) in the APG documentation for an example of the markup)

Odd, in my markup I only have one `div[slot="content"]`.

CleanShot 2025-12-02 at 15 08 33@2x

If there's just one, the current behavior is correct, right?

> Something else - it seems the first element is also getting a role="dialog" attribute; this should be removed.

I went with `role="none"` to match what lion is doing elsewhere. They cover this in the [considerations](https://lion.js.org/fundamentals/systems/overlays/rationale/#considerations) of their overlay stuff. We can set that to something else if that would be better, but it looks like we need to se it to _something_ otherwise Lion will [automatically set it to `dialog`](https://github.com/ing-bank/lion/blob/abcc6fdd6949397ce4690551f453d238c2802edb/packages/ui/components/overlays/src/OverlayController.js#L729)

> The disclosure still isn't closing for me after I tab through all the contents. 🤔

I see what you're saying now. It works in the storybook but I think that's for unrelated reasons. I need to have a think on how best to handle this.

gcamacho079:
> Odd, in my markup I only have one `div[slot="content"]`.

Indeed! I was testing adding more items into that particular disclosure since you mentioned:

> I've added aria-controls to the craft-button that will point to **the first element within the content slot.**

If there's only ever one `div[slot="content"]`, the implementation is correct. 👍🏼

brianjhanson:
> I'm not sure how strictly we can enforce it, but there should only ever be one `content` slot. I wrote it oddly because I had just written the code logic and translated that to english directly. 

Makes sense! My immediate thought is to issue a loud, highly visible component failure when something breaks the accessibility that can't be ignored, like not showing the component at all. 😈

brianjhanson:
@gcamacho079 I went ahead and pulled this into #18097 so that PR can encompass the entire settings page.

PR #18108, opened 2025-11-28
get value of the json expression before casting

i-just:
### Description
`$orderBy` is a `JsonExtract` expression, which means we can’t use it as is when casting for MySQL.

This was causing an issue where the table on the Entries index page wasn’t loading any data when using MySQL. All worked as expected with PostgreSQL (and still does). 

### Related issues
n/a


riasvdv:
I made a Cast expression we can use for this: https://github.com/craftcms/cms/blob/6.x/src/Database/Expressions/Cast.php

that way we don’t have to rely on fetching the grammar ourselves (in theory the connection on the query could be different than the default DB:: one)

riasvdv (approved):

Issue #18125, opened 2025-12-03
[5.x]: Unable to override `$template` via `View::EVENT_BEFORE_RENDER_PAGE_TEMPLATE`


cbovers:
### Description
We are trying to implement a multi-theme setup where the rendered template path depends on a custom site setting (e.g. a “theme” field on a Site).
The idea was to prepend a theme folder to the template name before Craft resolves the actual template file.

However, it appears that in Craft CMS 5.x, the template property of the TemplateEvent triggered by View::EVENT_BEFORE_RENDER_PAGE_TEMPLATE cannot actually influence the template that gets rendered.

### Steps to reproduce

We registered the following event listener:

```
Event::on(
    View::class,
    View::EVENT_BEFORE_RENDER_PAGE_TEMPLATE,
    function (TemplateEvent $event) {
        if (Craft::$app->getRequest()->getIsCpRequest()) {
            return;
        }

        // Prepend the active theme folder
        $event->template = '{theme-name-here}/' . $event->template;
    }
);
```

This listener fires correctly.

### Expected behavior

Changing $event->template during EVENT_BEFORE_RENDER_PAGE_TEMPLATE should allow us to influence which template gets rendered.

### Actual behavior

$event->template is not used by the View service.

Looking at craft\web\View::beforeRenderPageTemplate():

```
public function beforeRenderPageTemplate(string $template, array &$variables, string &$templateMode): bool
{
    if ($this->hasEventHandlers(self::EVENT_BEFORE_RENDER_PAGE_TEMPLATE)) {
        $event = new TemplateEvent([
            'template' => $template,
            'variables' => &$variables,
            'templateMode' => $templateMode,
        ]);
        $this->trigger(self::EVENT_BEFORE_RENDER_PAGE_TEMPLATE, $event);
        $variables = $event->variables;
        $templateMode = $event->templateMode;
        return $event->isValid;
    }

    return true;
}
```

The original $template argument is not updated after the event.
Only $variables and $templateMode are overwritten.

This means any modifications to $event->template have no effect.

### Craft CMS version

5.x

brandonkelly:
Just made it possible to modify the template via `EVENT_BEFORE_RENDER_TEMPLATE` and `EVENT_BEFORE_RENDER_PAGE_TEMPLATE` for Craft 5.9.

brandonkelly:
Craft 5.9.0 is out with that change 🎉

PR #18162, opened 2025-12-10
[5.9] Add EVENT_DEFINE_BASEPATHS

cbovers:
### Description

This PR adds a new event to the `craft\web\View` class named `EVENT_DEFINE_BASEPATHS`.
The event is triggered right before Craft resolves a template via `_resolveTemplate()`, and exposes the array of base template paths that Craft will search when locating templates.

#### Why this event is needed
Craft currently provides no way for modules or plugins to modify the base template paths that are used during template resolution. All template lookup logic is fixed in `TemplateLoader`, which means:

- You cannot inject additional template roots dynamically
- You cannot reorder lookup priorities
- You cannot introduce theme-based overrides without modifying core behavior
- Template fallback patterns cannot be implemented cleanly

This PR introduces an officially supported extension point that allows developers to define custom template path strategies using a simple event listener.

#### What this enables

With `EVENT_DEFINE_BASEPATHS`, a developer can intercept and modify the list of base paths Craft uses to find templates. This makes it possible to create more flexible, DRY, and maintainable template architectures.

For example, a project may support a theme system where templates should be resolved in the following order:
1. /templates/{theme}/
2. /templates/base/
3. /templates/

Using the event, a listener can detect the current theme (via project config, environment settings, user preferences, etc.) and dynamically inject or reorder basepaths before template resolution takes place.

This enables a “theme override” mechanism similar to what many CMS platforms provide, while preserving full compatibility with Craft’s existing template loader.

#### How it works
The event listener receives an array of basepaths and is free to:
- Add new paths
- Remove unwanted paths
- Reorder paths to change lookup priority

Craft then proceeds to resolve the requested template using the modified basepath list.

This approach gives developers complete control over template resolution strategy, without requiring changes to the loader, Twig environment, or existing Craft configuration.

#### Example use case
```
Event::on(
    View::class,
    View::EVENT_DEFINE_BASEPATHS,
    function (Event $event) {
        $basePaths = [];
        $site = Craft::$app->getSites()->getCurrentSite();
        $globals = \craft\elements\GlobalSet::find()->site($site)->one();
        if ($globals->theme) {
            $basePaths[] = Craft::getAlias('@templates') . DIRECTORY_SEPARATOR . $globals->theme;
        }
        
        $basePaths[] = Craft::getAlias('@templates') . DIRECTORY_SEPARATOR . 'base';
        $event->basePaths = array_merge($basePaths, $event->basePaths);
    }
);
```

### DIFF
--- a/src/events/DefineBasePathsEvent.php
+++ b/src/events/DefineBasePathsEvent.php
@@ -1,0 +1,21 @@
+<?php
+/**
+ * @link https://craftcms.com/
+ * @copyright Copyright (c) Pixel & Tonic, Inc.
+ * @license https://craftcms.github.io/license/
+ */
+
+namespace craft\events;
+
+use craft\base\Event;
+
+/**
+ * Define basePaths event class.
+ */
+class DefineBasePathsEvent extends Event
+{
+    /**
+     * @var array The component definitions
+     */
+    public array $basePaths = [];
+}
--- a/src/web/View.php
+++ b/src/web/View.php
@@ -11,6 +11,7 @@ use Craft;
 use craft\base\ElementInterface;
 use craft\events\AssetBundleEvent;
 use craft\events\CreateTwigEvent;
+use craft\events\DefineBasePathsEvent;
 use craft\events\RegisterTemplateRootsEvent;
 use craft\events\TemplateEvent;
 use craft\helpers\App;
@@ -107,6 +108,12 @@ class View extends \yii\web\View
      * @since 4.5.0
      */
     public const EVENT_AFTER_REGISTER_ASSET_BUNDLE = 'afterRegisterAssetBundle';
+
+    /**
+     * @event DefineBasepaths The event that is triggered when base paths are defined before resolving templates
+     * @since 5.9.0
+     */
+    public const EVENT_DEFINE_BASEPATHS = 'defineBasePaths';
 
     /**
      * @const TEMPLATE_MODE_CP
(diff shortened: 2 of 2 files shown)


AugustMiller:
> You cannot inject additional template roots dynamically

Unless I've misunderstood something, this _is_ possible, using the `craft\web\View::EVENT_REGISTER_CP_TEMPLATE_ROOTS` and `EVENT_REGISTER_SITE_TEMPLATE_ROOTS` events.

> You cannot reorder lookup priorities

This is true, but out of necessity. We sort the template root "keys" so that the loading order is stable and the most specific match happens first. As a result, similarly-named _roots_ like (`theme/sherlock`, `theme/base`, and `theme/shared`) may end up alphabetized (as `theme/base`, `theme/shared`, then `theme/sherlock`).

A single root can contain multiple base paths, though…

```php
$event->roots['theme'] = [
    // Insert the first path based on some config or environment variable:
    __DIR__ . '/theme/sherlock',

    // Always register the "base" theme:
    __DIR__ . '/theme/base',
];
```
…which we do _not_ sort.

---
(Oop, I wrote this before realizing that you are also the author!)

5.9 will also ship with new `EVENT_BEFORE_RENDER_TEMPLATE` and `EVENT_BEFORE_RENDER_PAGE_TEMPLATE` events (see #18125) that can intercept resolved template paths and rewrite them.

cbovers:
Hi @AugustMiller 

I'm sorry I just totally overlooked the possibility to be able to add an array of strings to the roots.
In order for me to bypass the `strncasecmp` function implemented in `_resolveTemplateInternal`, i have to use an empty string as the array key:

```
$basePaths = [
    Craft::getAlias('@templates') . DIRECTORY_SEPARATOR . 'sherlock'
    Craft::getAlias('@templates') . DIRECTORY_SEPARATOR . 'base'
];

$event->roots[''] = $basePaths;
```

No need to add an additional event. Thank you!

PR #18183, opened 2025-12-17
[6.x] Settings Index page

brianjhanson:
Targeting the `refactor-t-function` branch just for a cleaner diff. 

Moves the main settings page into inertia. The main change here is setting up at least a first pass of our icon library to use with the `craft-icon` component (which is just a wrapper around `wa-icon`).

> [!WARNING]  
> I updated where assets are published which may break things for you.

Previously we published the `resources/build` folder to `public/vendor/craft` but now that we're also publishing icons, those have been moved to `public/vendor/craft/build`. Icons will end up in `public/vendor/craft/icons`.

I know at least @riasvdv and I have symlinks set up that links `public/vendor/craft` to `resources/build`. With this PR you'll need to update that. In my case I just symlinked the entire `resources` directory.

```sh
# From your Craft site (DDEV project)
cd public/vendor && ln -s ../../vendor/craftcms/cms/resources craft
```


src/Cp/Settings.php
```diff
@@ -0,0 +1,117 @@
...
+
+        // @TODO ask Rias for help with this
+        // // Fire a 'registerCpSettings' event
+        // $eventName = $readOnly ? self::EVENT_REGISTER_READ_ONLY_CP_SETTINGS : self::EVENT_REGISTER_CP_SETTINGS;
+        // if ($this->hasEventHandlers($eventName)) {
+        //     $event = new RegisterCpSettingsEvent(['settings' => $settings]);
+        //     $this->trigger($eventName, $event);
+        //
+        //     return $event->settings;
+        // }
```
brianjhanson: @riasvdv Can you remind me how I should be implementing events so both Yii and Laravel events work properly? Or maybe just point me towards an example?

src/Cp/Settings.php
```diff
@@ -0,0 +1,117 @@
...
+            'label' => t('Assets'),
+        ];
+        $settings[$label]['filesystems'] = [
+            'icon' => 'light/folder-open',
+            'label' => t('Filesystems'),
+        ];
+
+        $label = t('Plugins');
+
+        $pluginsService = app(Plugins::class);
```
riasvdv: Can move this to the constructor and inject the `Settings` class in the controller method where needed. That way everything gets dependency injected. The `GeneralConfig` can also be injected into the constructor here then

That way the `all()` method doesn't need to be static either.

src/Http/Controllers/Settings/SettingsIndexController.php
```diff
@@ -0,0 +1,20 @@
...
+
+class SettingsIndexController
+{
+    public function __invoke(GeneralConfig $generalConfig)
+    {
+        return Inertia::render('SettingsIndexPage', [
+            'readOnly' => ! $generalConfig->allowAdminChanges,
+            'settings' => Settings::all(),
+        ]);
+    }
```
riasvdv:
```suggestion
    public function __invoke(GeneralConfig $generalConfig, Settings $cpSettings)
    {
        return Inertia::render('SettingsIndexPage', [
            'readOnly' => ! $generalConfig->allowAdminChanges,
            'settings' => $cpSettings->all(),
        ]);
    }
```

src/Cp/Settings.php
riasvdv: 1. Create new events [RegisterReadonlyCpSettings](https://github.com/craftcms/cms/pull/18183/commits/e217f28ea3d3e1c2a06f6f1828a8db60194e3197#diff-fc3ee50514f54bedfd948ef3d2071126c97bb43e97fb1b7a0ebfc09558f04502) and [RegisterCpSettings](https://github.com/craftcms/cms/pull/18183/commits/e217f28ea3d3e1c2a06f6f1828a8db60194e3197#diff-c93fbb96e0c972bad44ab0802d4fab41fd09e57b7bdbaed8ee8fc252f42c318b)

2. [Dispatch the new events](https://github.com/craftcms/cms/pull/18183/commits/e217f28ea3d3e1c2a06f6f1828a8db60194e3197#diff-e54fcced8b5a7a6123b5852a6d7a8101024e47c8c0dd11df46a038e0a439b34dR110-R118)

3. Add a static `registerEvents()` [to the legacy class that used to dispatch them](https://github.com/craftcms/cms/pull/18183/commits/e217f28ea3d3e1c2a06f6f1828a8db60194e3197), where it listens for the new event and dispatches that in the Yii way.

4. Add a call to that static method [in the service provider of the adapter](https://github.com/craftcms/cms/pull/18183/commits/e217f28ea3d3e1c2a06f6f1828a8db60194e3197)

resources/js/pages/SettingsGeneralPage.vue
```diff
@@ -138,10 +139,37 @@
         </craft-button-group>
       </template>
 
-      <div
-        class="bg-white border border-border-subtle mx-4 rounded-sm shadow-sm"
-      >
+      <div class="bg-white border border-border-subtle rounded-sm shadow-sm">
         <template v-if="readOnly">
+          <<<<<<< HEAD
```
deleugpn: git conflict?

Issue #18175, opened 2025-12-16
[5.x]: `Unable to duplicate element` when copying/pasting Matrix entries


tommysvr:
### Description
When attempting to copy and paste Matrix entries on a new entry, a toast error displays and the paste fails.

https://github.com/user-attachments/assets/7628d030-9d77-4fb0-9689-0381d4359ff7

### Steps to reproduce

1. Create a new entry and add a nested Matrix entry
2. Copy the entry
3. Paste the entry

### Expected behavior
Entry is pasted

### Actual behavior
Entry is not pasted, toast warning says 'Element duplicated.'

Logs show a warning:

```
[web.WARNING] [craft\controllers\{closure}] Unable to duplicate element: {"type":"craft\\elements\\Entry","id":268,"draftId":59,"fieldId":4,"ownerId":267,"siteId":3} 
```

### Craft CMS version

5.8.21

brandonkelly:
Fixed in 5.8.22!

PR #18185, opened 2025-12-18
ensure return element

i-just:
### Description
When a new nested entry is created for the matrix field in inline-editable blocks mode, it’s first created as a draft and so has a `draftId`. That new entry, with `draftId`, is passed back to the view and rendered. Then the owner element is auto-saved, triggering `NestedElementManager->saveNestedElements()`, which triggers the `Drafts->removeDraftData()`, which causes the `draftId` for that matrix “block” to be nullified.

If you then use the “copy” element action without reloading the page, the element data will contain the `draftId` that’s no longer applicable. When pasting, the canonical element is still found, but instead of being returned, we redirect to its control panel URL, causing the described issue.

Fix: when triggering the “copy” action, the element info can contain an extra `ensureReturnElement` param, which, when set to true, will enforce the element being returned instead of the redirect.

### Related issues
#18175


brandonkelly:
Went with a check to `$this->request->getAcceptsJson()` instead (to see if it’s an Ajax request), since there’s never a case where an Ajax request should get a redirect response to the element’s edit page.

PR #18199, opened 2025-12-24
Return `null` duration instead of default from stopCollectingCacheInfo

timkelty:
### Description
If something calls `stopCollectingCacheInfo`, there is no way to know if the elements had a duration, or if the default duration from config was used.

This moves the application of the default to `\craft\services\TemplateCaches::endTemplateCache`.

## Use-case

Craft Cloud uses `\craft\services\Elements::stopCollectingCacheInfo` to collect tags and expiry for static cache. It has it's own default for static cache (1 year), but because a duration is always returned (even if it is just the default config fallback), there isn't a way to know if it actually came from an element.

## Notes

To avoid this kind of thing, future versions should separate cache duration settings by use-case, e.g. `templateCacheDuration`, instead of sharing one config value.

### DIFF
--- a/src/services/Elements.php
+++ b/src/services/Elements.php
@@ -691,16 +691,6 @@ class Elements extends Component
             return [null, null];
         }
 
-        // Only use the duration if it's less than the cacheDuration config setting
-        $generalConfig = Craft::$app->getConfig()->getGeneral();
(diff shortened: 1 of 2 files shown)


brandonkelly:
This is a breaking change, so retargetted to 6.x.

PR #18212, opened 2026-01-05
Fix money field focus issue

brianjhanson:
Fixes an issue where focusing the money field would place the cursor in the wrong place. 

**Before**

https://github.com/user-attachments/assets/a337721a-4fac-4244-baca-4edee465ecb9

**After**

https://github.com/user-attachments/assets/d22c810f-a23f-4c31-8865-01fcfed208ab


src/web/assets/money/src/Money.js
```diff
@@ -24,9 +24,7 @@ import './Money.scss';
           this.$clearBtn.on('click', $.proxy(this, 'onClearBtnClick'));
         }
 
-        if (this.$field.val() != '') {
-          this.updateInputMask();
-        }
+        this.updateInputMask();
```
brianjhanson: @nfourtythree looks like the value check was added by you in https://github.com/craftcms/cms/commit/db5e7405e47dfb0a757c1ae7139ee7f9c7211522. On the off chance you remember why, and it's something we need to keep, just let me know.

nfourtythree: Was a while ago so unfortunately can't remember specifically. 

But looking at the commit I think it was related to not displaying the clear button when the field is blank as it would be redundant. 

Might be worth testing the fix with Commerce variants price fields. As the promotional price field can be completely blank rather than `0.00`.

redburn:
@brandonkelly This bug seems to have snuck back in. I'm on 5.10.8.1 and seeing the same behavior as in the Before video.

brianjhanson:
@redburn I did some testing and definitely see an issue when tabbing into an empty money field but I checked out 5.8.22 and the issue is still present. I've created https://github.com/craftcms/cms/issues/19156 to describe what I'm seeing. Does that sound like what you're running in to as well?

redburn:
@brianjhanson Yep, that's it! Thanks. Pretty sure it was resolved with the earlier patch, but maybe I didn't test it well enough.

PR #18247, opened 2026-01-16
Bugfix/various security fixes

i-just:
### Description
- updated two docblocks in the `GeneralConfig`
- fixed two security issues that were recently addressed on v5 and look to have slipped through the cracks for v6
- updated changelog-wip to add missing “removed” and “deprecated” items

For the changelog updates, I found those while figuring out the situation of the v5 files that contained recent security fixes. Feel free to adjust/discard if there’s a reason those were not included in the changelog-wip.

### Related issues
n/a


riasvdv (approved):
I probably just forgot the changelog, thanks!

src/Http/Controllers/FieldsController.php
```diff
@@ -290,7 +290,7 @@ public function renderCardPreview(Request $request, Fields $fields): JsonRespons
             'thumbAlignment' => ['nullable', 'string'],
         ]);
 
-        $fieldLayoutConfig = $request->input('fieldLayoutConfig');
+        $fieldLayoutConfig = Component::cleanseConfig($request->input('fieldLayoutConfig'));
```
riasvdv, line 293: Is there any way we can add tests for these? That we're sure the security issue doesn't return

PR #18251, opened 2026-01-18
Bulk actions for Matrix fields in Cards view

brandonkelly:
### Description

Adds the ability to select nested entries in Matrix fields. When multiple entries are selected, they can be reordered together via drag-n-drop, like the Blocks view.

Both Cards and Blocks views also now have Duplicate and Delete field-level actions, when one or more nested entries are selected.

A Matrix field with two selected entries and one unselected entry. Its field-level action menu is expanded, showing the actions: “Copy selected entries”, “Duplicate selected entries”, “Field settings”, “Copy field handle”, and “Delete selected entries”.

### Related issues

- #17947
- #18001
- #18210


airidasbrikas23:
This is awesome!

PR #18253, opened 2026-01-18
Support including generated fields in table views

brandonkelly:
### Description

Makes it possible to select generated fields for inclusion within element table views, as well as makes generated fields available as sort options.

### Related issues

- #17507


thupsi:
Thank you Brandon! You are a true gentleman and a scholar! 🙂

PR #18293, opened 2026-01-23
Bugfix/legacy element namespace and bugs

i-just:
### Description
The first commit replaces the old element namespaces with new ones in the twig files. 
With the old namespaces, the wrong element type was set in the [`ElementIndexesController::beforeAction()`](https://github.com/craftcms/cms/blob/56f1e361c2693a1e1ff0b320b571faeb0d7e4a16/yii2-adapter/legacy/controllers/ElementIndexesController.php#L103). This caused errors on e.g. entry index page. Trying to view entries in a section caused the “Move to section is only available for Entries” error, the “New entry” button was not loading when the “All entries” source was selected, etc.

The second commit fixes an issue where it was not possible to create a new entry or edit an existing one due to an `Undefined property: stdClass::$siteId` error.

The third commit fixes an issue where it was not possible to save a new nested entry due to `CraftCms\Cms\Element\Element::getEnabledForSite(): Return value must be of type ?bool, string returned` error.

### Related issues
n/a

### DIFF
--- a/resources/templates/assets/_index.twig
+++ b/resources/templates/assets/_index.twig
@@ -1,6 +1,6 @@
 {% extends '_layouts/elementindex.twig' %}
 {% set title = "Assets"|t('app') %}
-{% set elementType = 'CraftCms\\Cms\\Entry\\Elements\\Asset' %}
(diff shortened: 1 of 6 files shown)


riasvdv (approved):
Ah damnit, thought I got all of those, thanks!

PR #18297, opened 2026-01-25
Use carets in Composer version constraints

brandonkelly:
When installing a plugin, or updating Craft/plugins, the updated `composer.json` constraints will now use the caret operator (e.g. `^1.2.3`), so the packages can be updated via `composer update`.

Previously only the exact version being installed/updated was specified, with no caret operator, so the packages would be locked at the expected version, and not updated automatically when `composer update` was run. That was meant as a protection, in case any associated licenses had expired, so running `composer update` wouldn’t cause them to run a version that they weren’t licensed to have. (We call these “astray licenses”.)

While that’s still a concern, we decided it’s not worth the added `composer update` friction, especially since most Craft/plugin licenses are set to auto-renew.

Instead, the `up` command now checks for any astray licenses up front, and warns about them, giving the developer a chance to revert an unintended package update before any migrations have been run for it.

An “up” command’s output, warning that a plugin isn’t licensed to run its installed version, and asking whether the user wishes to continue anyway.


MoritzLost:
@brandonkelly Just saw this change in the 5.9 announcement post. Any chance we can get a config setting to toggle between the old and the new behavior?

I actually preferred the old behavior, and our update process sort of expects it. Since Craft and plugins may also require migrations after updating, we have been using `craft update` to do that in one step. Since we know that Craft and plugins are fixed to a specific version, that allowed us to run `composer update` to update any other dependencies, without having to worry about migrations. So we had a clear separation where `craft update` is for Craft dependencies, `composer update` for everything else.

With the new behavior, we still need to run `craft up` after `composer update`, so there isn't really any reason not to use `craft update`. That leaves `composer update` in a weird spot where it doesn't have a clear boundary to `craft up`, and there's also no simple way to update any _non_-Craft dependencies.

brandonkelly:
@MoritzLost This was mostly a “paving the cow paths” sort of change. Most people seem more comfortable running `composer update` + `craft up` (often automatically via `post-update-cmd`), and had to add the caret operators manually after installing plugins, so it made sense for us to just start doing that automatically.

You could still separate the two actions if you run `craft update` first, then `composer update` once all plugins are updated. (`craft update` won’t touch anything besides the specific packages you’re telling it to update, plus their dependencies.)

MoritzLost:
@brandonkelly Yes absolutely, though we can never be sure that `composer update` won't update any Craft plugins … oh well, I guess it's just a matter of changing habits 👍

Issue #18324, opened 2026-01-28
[5.x]: craftid.svg for plugin store is missing


gglnx:
### Description

On opening the plugin store inside the control panel, the following error message is logged:

> Could not get the contents of /[…]/vendor/craftcms/cms/src/icons/solid/craftid.svg: The file doesn't exist

### Steps to reproduce

1. Open the plugin store
2. Review the error log

### Expected behavior

No error message

### Actual behavior

Error message is logged

### Craft CMS version

5.9.1

### Installed plugins and versions

-

i-just:
Hi, thanks for reporting! I raised a PR for this.

brandonkelly:
Craft 5.9.2 is out with a fix for that. Thanks again!

PR #18330, opened 2026-01-28
added missing craftid.svg icon

i-just:
### Description
Added missing `craftid.svg` icon and an alias for it. 

The icon is still present in the latest v4: https://github.com/craftcms/cms/blob/4.17.1/src/icons/craftid.svg, and it’s used [here](https://github.com/craftcms/cms/blob/5.9.1/src/templates/plugin-store/_index.twig#L12).

### Related issues
#18324


brandonkelly:
That button never shows up anymore, so I just removed it.

PR #18339, opened 2026-01-29
[5.x] Fixed link field data type in GQL

nfourtythree:
### Description
A GraphQL error can occur when querying if you have multiple link fields setup with different "allowed link types".

A way to replicate this is by have two link field setup:

- A Link Field - with allowed types of "URL" and "Entry"
- Z Line Field - with allowed type of just "URL"

On creating a GQL query of the following you will see the error.

```gql
{
  entries(section: "test") {
    id
    ... on testLinkField_Entry {
      title
      zLinkField {
        url
      }
      aLinkField {
        id
        entry {
          id
        }
      }
    }
  }
}
```

This PR fixes this issue.


stephanedemotte:
Breaking change @brandonkelly, @nfourtythree, my website goes down after the update to 5.9.4, all our client website will goes down with this update :/

screenshot-2026-01-30_05-59-32


```{
  "errors": [
    {
      "exception": "craft\\errors\\GqlException",
      "message": "Tried to load an unregistered type \"LinkData\". This can indicate both a typo in the query or an issue with the schema used.",
      "file": "/var/www/html/vendor/craftcms/cms/src/gql/TypeLoader.php",
      "line": 39,```

brandonkelly:
@stephanedemotte Thanks for pointing that out. We’ve fixed that for the next release via #18356.

PR #18380, opened 2026-02-05
[6.x] Workbench

riasvdv:
A barebones setup of Workbench for testing the package while developing without needing a separate "host" application.

The current seeder creates the following:
- fieldLayout with just an EntryTitleField
- `Page` entry type with this field layout
- `Home` single
- `Pages` structure
- `Posts` channel

I'll probably improve this further when I've had a chance to use it a bit more.

The test changes / fixes are related to the Laravel request not being available in tests.

# Workbench

Workbench allows you to run a live Craft CMS application during development without needing a separate Laravel project. This is useful for manually testing features, debugging, and exploring the CMS.

## Quick Start

```bash
composer run serve
```

This starts a development server at `http://127.0.0.1:8000`.

## How It Works

Workbench uses [Orchestra Testbench](https://packages.tools/workbench) to bootstrap a minimal Laravel application with Craft CMS loaded. The setup consists of:

| Directory                                 | Purpose                                                           |
|-------------------------------------------|-------------------------------------------------------------------|
| `vendor/orchestra/testbench-core/laravel` | The Laravel application skeleton                                  |
| `workbench/`                              | Your development workspace for routes, views, seeders, and config |
| `testbench.yaml`                          | Configuration that ties everything together                       |

When you run `composer run serve`, Workbench:
1. Loads the skeleton
2. Registers all service providers from `testbench.yaml`
3. Discovers routes, commands, and config from `workbench/`
4. Starts the PHP development server

## Directory Structure

```
workbench/
├── app/
│   └── Models/           # Eloquent models for testing
├── config/               # Custom config overrides
├── database/
│   ├── factories/        # Model factories
│   ├── migrations/       # Test migrations
│   └── seeders/          # Database seeders
├── resources/
│   └── views/            # Blade/Twig views
└── routes/
    ├── web.php           # Web routes (auto-discovered)
    └── console.php       # Artisan commands (auto-discovered)
```

## Configuration

The `testbench.yaml` file controls Workbench behavior:

```yaml
providers:                   # Service providers to load
  - CraftCms\Cms\Providers\CraftServiceProvider
  - CraftCms\Yii2Adapter\Yii2ServiceProvider
  # ...

workbench:
  start: /                   # Default route when opening the browser
  install: true              # Run Laravel's default migrations
  sync:                      # Symlink directories into the skeleton
    - from: resources
      to: public/vendor/craft
  discovers:
    web: true                # Auto-load workbench/routes/web.php
    commands: true           # Auto-load workbench/routes/console.php
    config: true             # Auto-load workbench/config/*.php
```

### Running artisan commands

```bash
vendor/bin/testbench migrate
vendor/bin/testbench tinker
vendor/bin/testbench route:list
```

### Rebuilding the workbench

```bash
composer run build
```

This runs the build recipes defined in `testbench.yaml`.

## Further Reading

- [Workbench Documentation](https://packages.tools/workbench)
- [Testbench Configuration](https://packages.tools/getting-started/configuration)


src/Http/Middleware/HandleInertiaRequests.php
```diff
@@ -46,15 +47,20 @@ public function version(Request $request): ?string
...
+        if (! $isInstalled) {
+            return parent::share($request);
+        }
+
+        $currentSite = Sites::getCurrentSite();
         $updates = app(Updates::class);
         $nav = app(Navigation::class);
 
-        if ($isInstalled && ! $updates->isCraftUpdatePending()) {
+        if (! $updates->isCraftUpdatePending()) {
```
riasvdv, line 63: @brianjhanson I had to return early here if not installed because the installer was throwing exceptions that there was no site.

Might want to double-check what is needed for the installer here or if just doing parent::share is enough

brianjhanson: I don't think the installer needs any of this stuff at the moment so this should be good. I've made a similar change in my branches but apparently forgot to commit it 🤦

brianjhanson:
Would we be able to point [Playwright](https://playwright.dev/docs/intro) tests at this?

brianjhanson (approved):

riasvdv:
> Would we be able to point [Playwright](https://playwright.dev/docs/intro) tests at this?

I'd maybe look into https://pestphp.com/docs/browser-testing as that runs on Playwright but is fully integrated with the test environment

brianjhanson:
> > Would we be able to point [Playwright](https://playwright.dev/docs/intro) tests at this?
> 
> I'd maybe look into https://pestphp.com/docs/browser-testing as that runs on Playwright but is fully integrated with the test environment

Yeah, I tried to get that working but didn't have any luck on my first attempt. I forget exactly what I ran int though. 

The other piece is that we already have some e2e tests written with Playwright that it would be good to point at 6.x and not have to rewrite in Pest

PR #18390, opened 2026-02-06
[6.x] Inertia utilities

brianjhanson:
Ports all of the utilities to inertia. 

For some reason, when I started this adventure, I didn't realize that all of the Utilities were rendered along one route. I foolishly thought I'd be able to port them one by one, but alas, that was not the case.

As this ended up being a fairly wide ranging PR, I've asked CoPilot to summarize the broad strokes.

### New Components and Stories

* Added a new `CraftProgressBar` component, including its styles and a comprehensive Storybook file with multiple usage examples and interactive demos. [[1]](diffhunk://#diff-0685eafd8f23d73687f50d0f1dec9720f000eaab29d93b3294d829df667739acR1-R327) [[2]](diffhunk://#diff-de0cbf132d8b32bea8993f5792d1917804eedc8be9a4e6fb4bc77438083dcac4R1-R100)
* Introduced the `CraftCheckboxIndeterminate` component, providing indeterminate checkbox functionality and custom styles. This component makes our handling of the "Select All" options much more predictable. The one downside is that previously we were sending up `'*'` to denote "all" but Lion's component doesn't seem to like this out of the box, instead preferring `''`. The good news is that by default it sends `''` plus all of the selected options. In the cases I've run into so far, that ends up simplifying the logic in PHP as we no longer have to worry about a special "all" case and instead just handle each array item as expected.

### Component Styling and Layout Enhancements

* Updated the `CraftCheckbox` component to use a CSS grid layout, improving alignment of input, label, and help text. [[1]](diffhunk://#diff-493aee39b3cd54c26a75679b4d50913dae187bb0f01f0af72b13b5861e7ef05cL11-R21) [[2]](diffhunk://#diff-493aee39b3cd54c26a75679b4d50913dae187bb0f01f0af72b13b5861e7ef05cR33)
* Improved the `CraftCheckboxGroup` component by adding a grid layout for `.input-group` for better spacing and structure.
* Enhanced the `CraftIcon` component with a default font size and removed an unused styles file. [[1]](diffhunk://#diff-86b1fc25b2e1d82798000c4bb3728b60e5e7dae4716df74c196d5d6e7d4943e9L2-R2) [[2]](diffhunk://#diff-86b1fc25b2e1d82798000c4bb3728b60e5e7dae4716df74c196d5d6e7d4943e9L11-R18) [[3]](diffhunk://#diff-05d13b41c7596ae5221c837f4c5c77729ae75a6fb8cc40c4f1ad76571c56d77fL1-L3)

### Navigation Item Improvements

* Refined `CraftNavItem` styles to support a `flush` property for better alignment, added new CSS variables for padding, and introduced a `nav-item--flush` class for negative margin adjustment. [[1]](diffhunk://#diff-8cad326b0e0c12beef90de1c3d5f0d3cdb83d693a7e23a1ee7a87b85d8a03d7aR4-R17) [[2]](diffhunk://#diff-8cad326b0e0c12beef90de1c3d5f0d3cdb83d693a7e23a1ee7a87b85d8a03d7aR27-R30) [[3]](diffhunk://#diff-007b9dbaa89edb2d078b0147550c888802af62dbd150bc055427c1b5ac7da16aR39-R42) [[4]](diffhunk://#diff-007b9dbaa89edb2d078b0147550c888802af62dbd150bc055427c1b5ac7da16aR134)
* Updated rendering logic in `CraftNavItem` to apply new classes for icon-only and flush variants.

### Miscellaneous

* Adjusted padding in the `Callout` component for improved spacing.


packages/craftcms-cp/src/services/AssetIndexer.ts
brianjhanson, line 1: This is pretty much a straight port of the previous AssetIndexer so it could probably use more love and more intertia-fying. It works for the moment though

packages/craftcms-cp/src/utilities/api/actionClient.ts
brianjhanson, line 1: This file + the `apiClient` probably aren't long for this world. Inertia and their helpers can probably handle all of the scenarios this is used in eventually

resources/js/bootstrap/cp.ts
brianjhanson, line 1: This file is doing all the setup work to create a Cp global object. It might make sense to replace the global object with some kind of context element either in Lit or in Vue, but for the moment a global is nice to make it easy to use these services in both legacy and Inertia pages.

resources/js/components/Date.vue
brianjhanson, line 1: This component will eventually respect your locale and potentially date formatting preferences. For now, we're just doing the bare minimum

src/Cp/Cp.php
brianjhanson, line 1: This will probably get folded into HandleInertia requests in some way. The goal is to provide any configuration the CP needs, ideally without exposing everything set in the general config.

src/Queue/Enums/JobStatus.php
```diff
@@ -4,15 +4,54 @@
...
+        return null;
+    }
+
+    public function jsonSerialize(): array
+    {
+        return [
+            'label' => $this->label(),
+            'value' => $this->value,
+        ];
+    }
```
brianjhanson, line 56: I imagine we'll want to do this with a lot of enums that make their way to the front-end, so it will probably be worth abstracting at some point.

src/Http/Middleware/HandleInertiaRequests.php
```diff
@@ -102,6 +113,7 @@ public function share(Request $request): array
                 ],
                 'cpUrl' => UrlHelper::cpUrl(),
                 'actionUrl' => UrlHelper::actionUrl(),
```
brianjhanson: @riasvdv I believe I saw that you switched these to `cp_url` and `action_url` instead of the helpers, but I've also noticed those global functions don't return a full URL. Is there a reason for that? 

I don't think I have strong opinions either way, but I do find full URLs are a little easier to deal with on the front end

src/Cp/Cp.php
```diff
@@ -0,0 +1,34 @@
...
+use craft\helpers\UrlHelper;
+use CraftCms\Cms\Cms;
+use Illuminate\Support\Collection;
+
+final readonly class Cp
+{
+    /**
+     * @TODO Could/should all this data just be handled in an inertia middleware?
+     * We'll need to render all legacy pages with inertia in that case.
+     */
```
riasvdv, line 16: I did a first stab at a more global sharing of data in this PR: https://github.com/craftcms/cms/pull/18448/changes/328ac7cbcd955ef3dd2d411e2db7b6bb5d90c4cd

We could do the same for the CP stuff, so both Twig, Blade and Inertia preferably have their global data available

src/Deprecator/Models/DeprecationError.php
```diff
@@ -4,10 +4,22 @@
...
+/**
+ * @property int $id
+ * @property string $key
+ * @property string $fingerprint
+ * @property Carbon $lastOccurrence
+ * @property string $file
+ * @property int $line
+ * @property string $message
+ * @property string $traces
+ */
```
riasvdv: These shouldn't be necessary when you have Laravel Idea installed and let it generate helper files, were you getting phpstan errors?

src/Http/Controllers/Settings/SitesController.php
```diff
@@ -168,7 +168,7 @@ public function store(Request $request): RedirectResponse
             $isNew = true;
         }
 
-        $site->groupId = $request->input('group');
+        $site->groupId = (int) $request->input('group');
```
riasvdv:
```suggestion
        $site->groupId = $request->has('group') ? $request->integer('group') : null;
```

`groupId` is nullable so casting `null` to `0` would have unexpected consequences

src/Http/Controllers/Updates/UpdaterController.php
```diff
@@ -54,6 +60,24 @@ public function __construct(
         }
     }
 
+    /**
+     * Renders the Updater page via Inertia.
+     */
+    #[Override]
+    public function index(#[Give('Craft')] Application $craft): Response
```
riasvdv, line 67:
```suggestion
    public function index(): Response
```

`$craft` isn't used anymore in this method

src/Http/Controllers/Utilities/ClearCachesController.php
```diff
@@ -52,10 +53,14 @@ public function clearCaches(Request $request, Application $app): JsonResponse
             }
         }
 
+        if ($request->inertia()) {
```
riasvdv: Are these endpoints used in non-inertia pages still? Otherwise we can just always return back here and in all the other `$request->inertia()` checks?

src/Http/Controllers/Utilities/SystemMessagesController.php
```diff
@@ -41,10 +39,13 @@ public function show(Request $request): Response
...
-                'message' => $message,
-                'language' => $message?->language,
-            ]),
+            'message' => [
+                'key' => $message->key,
+                'heading' => $message->heading,
+                'subject' => $message->subject,
+                'body' => $message->body,
+                'language' => $message->language,
+            ],
```
riasvdv:
```suggestion
            'message' => $message->toArray(),
```

I don't know what you prefer, just suggesting this as a possibility since `$message` is an eloquent model

src/Http/Controllers/ConfigSyncController.php
```diff
@@ -45,6 +52,24 @@ public function __construct(
         parent::__construct($request, $generalConfig, $composer, $plugins, $updates);
     }
 
+    /**
+     * Renders the Config Sync page via Inertia.
+     */
+    #[Override]
+    public function index(#[Give('Craft')] Application $craft): Response
```
riasvdv, line 59:
```suggestion
    public function index(): Response
```

src/Http/Middleware/HandleInertiaRequests.php
```diff
@@ -76,7 +80,13 @@ public function share(Request $request): array
                 'success' => $request->session()->get('success'),
                 'error' => $request->session()->get('error'),
             ],
-            'craft' => [
+            'queue' => fn () => [
+                'enabled' => Craft::$app->get('queue') instanceof QueueInterface,
```
riasvdv: I think we can have this always enabled now that we just track the queue through events

src/Http/Middleware/HandleInertiaRequests.php
riasvdv: What exactly makes them a little easier to deal with?

src/Utility/Utilities/DeprecationErrors.php
```diff
@@ -43,14 +41,21 @@ public static function badgeCount(): int
...
     #[Override]
-    public static function contentHtml(): string
+    public static function toolbarHtml(): string
     {
-        $view = Craft::$app->getView();
-
-        $view->registerAssetBundle(DeprecationErrorsAsset::class);
-
-        return template('_components/utilities/DeprecationErrors/index', [
+        return view('c::utilities.deprecation-errors.toolbar', [
```
riasvdv:
```suggestion
        return view('utilities.deprecation-errors.toolbar', [
```

We don't need to prefix these anymore.

src/Utility/Utilities/QueueManager.php
```diff
@@ -40,21 +40,45 @@ public static function icon(): string
     #[Override]
     public static function toolbarHtml(): string
     {
-        return template('_components/utilities/QueueManager/toolbar');
+        $progressService = app(JobProgress::class);
```
riasvdv, line 43: Note to self: check if we're resolving the utilities through the container, in that case we can inject the service into the constructor instead and use `$this->jobProgress`, etc...

riasvdv (changes requested):

src/Deprecator/Models/DeprecationError.php
brianjhanson: No PHPStan errors, my autocomplete just wasn't completing. That might have just been a glitch though, I'll remove them.

src/Http/Controllers/Utilities/SystemMessagesController.php
brianjhanson: I ended up going with just 

```'message` => $message``` 

since that will `toArray` anyways. 🤷‍♂️ 

I typically like using resources, but I'm not sure if we're going to need/want those in here or not.

src/Utility/Utilities/QueueManager.php
brianjhanson: The problem is that these methods are static so resolving them through the constructor doesn't help. I thought about making them non-static, but that felt like too large of a breaking change for plugins that define their own utilities.  I'm sure we could add the resolution into the function itself but that wasn't something I wanted to bite off.

src/Http/Middleware/HandleInertiaRequests.php
brianjhanson: Of course I don't have a good answer for you right now, just a general feeling. I went ahead and updated this to use the new global functions until I remember why I didn't like them in the first place.

src/Utility/Utilities/QueueManager.php
riasvdv: Ah missed that they were static 👍

src/Cp/Cp.php
brianjhanson: Nice! Once that one is merged in I'll shift some things around to utilize it

src/Http/Controllers/Utilities/SystemMessagesController.php
riasvdv: Resources are good for me as well 👍  I suspect we'll need them at some point so it might be good to be consistent

PR #18424, opened 2026-02-13
Update the db/convert-charset collation argument docs

johnnynotsolucky:
### Description

The docs for the db/convert-charset command are no longer accurate. This PR updates `$collation` arguments docs to reflect the defaults here: https://github.com/craftcms/cms/blob/2f5149d4d64a5dafae9db99357823655422ba77d/src/helpers/Db.php#L1916-L1936

### Related issues

None.


brandonkelly:
Thank you!

PR #18433, opened 2026-02-16
[6.x] Extract Input namespacing from View

riasvdv:


brandonkelly:
@riasvdv Let’s rename `namespaceInputId()` to just `namespaceId()` since the method itself doesn’t care where the ID came from.

riasvdv:
> @riasvdv Let’s rename `namespaceInputId()` to just `namespaceId()` since the method itself doesn’t care where the ID came from.

@brandonkelly Done!

PR #18458, opened 2026-02-20
I18n/greek translations

thupsi:
I started doing this manually a while ago, but would take forever to complete, so I took the plunge and used Claude for the remaining translations.

Did a quick check, both by scanning the translations file and by applying to an actual project, and seems pretty good.

There are places where a one-to-one mapping of english to greek does not make sense, or looks awkward, so I made some adjustments.

All-in-all I think it's a pretty decent first pass, certainly better than not having greek language as an option.


brandonkelly:
Thanks! Uploaded that to CrowdIn and enabled Greek for the next release.

brandonkelly:
Craft 5.9.13 is out with the Greek translation. Thanks again!

PR #18459, opened 2026-02-20
[6.x] inertia sections settings index

brianjhanson:
Implements the settings sections index page in Inertia. Mostly that means a new example for how we're handing tables.


src/Http/Controllers/Settings/SectionsController.php
```diff
@@ -200,9 +238,12 @@ public function destroy(Request $request, Sections $sections): Response
             'id' => ['required', Rule::exists(Table::SECTIONS, 'id')],
         ]);
 
-        $sections->deleteSectionById($request->input('id'));
+        $name = $sections->getSectionById($request->input('id'))->name;
+        $sections->deleteSectionById($request->integer('id'));
```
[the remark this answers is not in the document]
brianjhanson: This doesn't seem very likely to me, but @riasvdv or @brandonkelly if you think it's worth adding the condition here we can.

src/Http/Controllers/Settings/SectionsController.php
```diff
@@ -39,9 +40,46 @@ public function __construct(
         $this->readOnly = ! $generalConfig->allowAdminChanges;
     }
 
-    public function index(): View
+    public function index(Request $request, Sections $sections): \Inertia\Response
     {
-        return view('settings.sections._index', [
+        $page = (int) $request->input('page', 1);
+        $limit = (int) $request->input('per_page', 50);
```
riasvdv:
```suggestion
        $page = $request->integer('page', 1);
        $limit = $request->integer('per_page', 50);
```

src/Http/Controllers/Settings/SectionsController.php
```diff
@@ -39,9 +40,46 @@ public function __construct(
...
+        $searchTerm = $request->input('search');
+
+        $sort = $request->input('sort', [
```
riasvdv:
```suggestion
        $sort = $request->array('sort', [
```

riasvdv (approved):
Just nitpicks

PR #18464, opened 2026-02-23
[6.x] sqlite

riasvdv:
### Description

This is a first try at getting tests to run on sqlite, mostly for local development right now.


src/Database/DatabaseServiceProvider.php
```diff
@@ -32,6 +35,10 @@ public function register(): void
         Connection::macro('isMysql', fn (bool $strict = false) => $strict ? $this->getDriverName() === 'mysql' : in_array($this->getDriverName(), ['mysql', 'mariadb']));
         Connection::macro('isMaria', fn () => $this->getDriverName() === 'mariadb');
         Connection::macro('isPgsql', fn () => $this->getDriverName() === 'pgsql');
+        Connection::macro('isSqlite', fn () => $this->getDriverName() === 'sqlite');
+
+        Builder::macro('whereBool', fn ($column, bool $value) => $this->where($column, new QueryExpression(var_export($value, true))));
+        Builder::macro('orWhereBool', fn ($column, bool $value) => $this->orWhere($column, new QueryExpression(var_export($value, true))));
```
riasvdv, line 41: In QueriesUniqueElements, we use ->toRawSql to generate a subquery, Laravel doesn't cast 0 or 1 to bool for Postgres as it usually lets Pdo handle this.

This is why we need these specific macros, so the query gets the correct actual boolean values.

src/Database/DatabaseServiceProvider.php
```diff
@@ -45,6 +52,33 @@ public function boot(Repository $config, Connection $db, \Illuminate\Cache\Repos
             MigrateCommand::class,
         ]);
 
+        if ($db->getDriverName() === 'sqlite') {
+            $this->bootSqlite($config, $db, $cache);
+        } else {
+            $this->bootDefault($config, $db, $cache);
+        }
```
riasvdv, line 59: Preferably, an extra connection and the whole db2 handling wouldn't be done by us anymore. Laravel allows setting [a separate lock_connection](https://github.com/laravel/framework/issues/36280#issuecomment-780059924) in case you have cache set to database. (file or Redis are more common to use as cache)

PR #18470, opened 2026-02-24
fix: condition logic for (not) empty operator

bramnijssen:
### Description
Fixes conditional logic for the **(not) empty** operator on dropdown fields. Previously, conditions using **(not) empty** always evaluated to true because `$this->_values` was checked first. That check is only relevant for **(not) in** conditions.

### Steps to reproduce the fixed issue
1. Create an entry for an entry type that includes a Title field and a Dropdown field.
2. Add a Visibility Condition to the Title field using an Entry Condition where the Dropdown field is set to **(not) empty**.
3. Observe that the Title is always visible, regardless of whether the dropdown field had a value.


brandonkelly:
Thanks for the PR! Craft 4.17.7 and 5.9.13 are out with that fix.

PR #18488, opened 2026-02-27
Proposal: Craft core save optimization

boboldehampsink:
### Issue

I have multiple multi-site installs where saving entries can take more than 30 seconds.

### Proposal

Implements synchronous save-path performance optimizations for multi-site element saves, including propagation-loop reuse, nested-save dedupe, search queue aggregation, and duplicate change-tracking/caching reduction.

### Related issues

-

### DIFF
--- a/src/elements/NestedElementManager.php
+++ b/src/elements/NestedElementManager.php
@@ -824,6 +824,9 @@ JS, [
 
         $elementIds = [];
         $sortOrder = 0;
+        $managerKey = isset($this->field)
+            ? sprintf('field:%s', (string)($this->field->uid ?? $this->field->handle ?? $this->field->id))
+            : sprintf('attribute:%s', (string)$this->attribute);
 
         $transaction = Craft::$app->getDb()->beginTransaction();
         try {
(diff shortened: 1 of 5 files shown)


boboldehampsink:
@brandonkelly created this as a draft/proposal because I'm not confident this is something you'd want. But I thought it might be worth looking into at the least.

brandonkelly:
> nested-save dedupe

Can you provide an example where a nested element is getting saved multiple times in the same request?

boboldehampsink:
Yes. A concrete case is a multi-site owner with a Matrix field using non-`all` propagation (for example `language`):

1. Sites: `en-US`, `fr-FR`, `fr-CA`.
2. Owner entry is saved on `en-US` in a request where `fr-FR` is newly propagated (`newSiteIds` contains `fr-FR`), while `fr-CA` already existed.
3. In `NestedElementManager::saveNestedElements($owner)`, this enters the branch where:
   - `otherSiteIds` = new sites only (`fr-FR`)
   - `preexistingOtherSiteIds` = preexisting sites (`fr-CA`)
4. For localized owner `fr-FR`, its supported nested-site set (for `language`) includes both `fr-FR` and `fr-CA`, so `sharedPreexistingOtherSiteIds` is non-empty.
5. That hits the recursive call `saveNestedElements($preexistingLocalizedOwner)` (the `fr-CA` owner).

Result: nested elements for that owner/site can be saved again in the same request cycle, which is what the dedupe key (`owner + site + manager + nestedId + sortOrder`) now prevents.

PR #18491, opened 2026-02-27
[6.x] Inertia section edit

brianjhanson:
The main focus here is porting the sections edit page over to Inertia but maybe the real code was all the things we had to change along the way.

### Editable Tables

The biggest of those is figuring out how to manage editable tables. In the end, I ended up creating our own version of TanStack Table's `columnHelper` which will eventually have dedicated methods for each of the input types the current `editableTable.twig` supports. At the moment we only have the ones that were needed for this port (`text`, `lightswitch` and `checkbox`). 

### Vue Wrappers

Another fairly large shift in this PR is that we've added a script that will generate Vue components from the lit web components. I wanted to avoid this, but annoyingly Vue and Lion don't agree on some prop values. The Vue wrappers smooth this out so you can just `v-model` the inputs as you would expect. These components are _only_ to handle the `v-model` binding which is why we still have a `Select` component in the main `resources/js/components` folder. The `Select` component takes in an `options` prop and handles normalizing those before rendering the appropriate markup. 

Not ideal, but at least these layers keep things flexible.

It may end up that using Lion's form components isn't really worth it if we're living in a purely Vue world but for now, I still like the flexibility it allows us. That way if we have some views that we can't fully convert to Vue for some reason, we can ensure some visual consistency by using the `@craftcms/cp` components. 

### Copilot's summary

Including this because it did a pretty good job of summing things up.

**Component Enhancements and Features:**

* `craft-action-item` now supports checkable items with new `checked`, `active`, and `type` properties, updated rendering logic, and corresponding style adjustments to allow for checkbox-like menu items. [[1]](diffhunk://#diff-9245da7af305742edcb6384fd435ab04c6fcc7d7d289b66233967093903c252bR17-R31) [[2]](diffhunk://#diff-9245da7af305742edcb6384fd435ab04c6fcc7d7d289b66233967093903c252bL40-R69) [[3]](diffhunk://#diff-c6ba6fe5a1f3df58abe8973c1700ca3a5912e603525339b62680a52a4cb668c0R12-R15) [[4]](diffhunk://#diff-c6ba6fe5a1f3df58abe8973c1700ca3a5912e603525339b62680a52a4cb668c0R26-R29)
* `craft-chip` gains an `icon` property for easy prefix icon rendering, improved slot handling, and minor style adjustments for layout consistency. [[1]](diffhunk://#diff-7207a1fc404638d06c07d8e02539a6d1725fff890acb51063a882a4950476a6fR28-R48) [[2]](diffhunk://#diff-7207a1fc404638d06c07d8e02539a6d1725fff890acb51063a882a4950476a6fL43-R64) [[3]](diffhunk://#diff-c647546b8643ca2596def0871b52bb5116da99f40f1d2ecbd0ae57533e219b36L10-R10) [[4]](diffhunk://#diff-c647546b8643ca2596def0871b52bb5116da99f40f1d2ecbd0ae57533e219b36L22-R21) [[5]](diffhunk://#diff-c647546b8643ca2596def0871b52bb5116da99f40f1d2ecbd0ae57533e219b36L52-R51)
* `craft-button` introduces a new `dashed` appearance, with corresponding styles and property updates for more visual variety. [[1]](diffhunk://#diff-1de718031c8752a56d8a3bd66a33e827b71e96d48e74e3c2ce23c00581af5b4aR107-R118) [[2]](diffhunk://#diff-80b8cc8bc15c096d16c81622cd5988bfc3617bf6ee2b59019fe7db85c5f52169L30-R31)
* `craft-input` and `craft-select` components now support additional sizing and alignment options, including `small`, `medium`, and `large` sizes, a `center` property, and a corrected use of `maxlength` instead of `size` for input length. [[1]](diffhunk://#diff-db58986bb09a97650d3521cbe1781894417137e41509dc6befb1e9c5b273b802L11-R21) [[2]](diffhunk://#diff-f2db3dcfba47380b507298229ed3e3e4b299c9ccf223c505b6bd2b0a13e47141R11-R14) [[3]](diffhunk://#diff-124f316d91035e798acdb877f051b62ea6c7f270ff59656e8e4da5dd11bbd037R6-R14) [[4]](diffhunk://#diff-29426a1ae4e2d622ea6e9e8648673f525231e50bc1a9c03558d82953423e2dbfL5-R5)
* `craft-switch-button` and `craft-switch` have updated sizing logic for better consistency across controls, and a new Storybook story is added for `craft-switch` with various usage examples. [[1]](diffhunk://#diff-535ef073830842d5af3e529ed8d4689aadb0492b0a86260334ffa2a46d354893L10-R10) [[2]](diffhunk://#diff-535ef073830842d5af3e529ed8d4689aadb0492b0a86260334ffa2a46d354893L22-R22) [[3]](diffhunk://#diff-abbe0ac1d2de0a36c518d7c2034189c2c72d12bee1e5ef934ad0007f6d8221d6R1-R81)

**Build Process and Exports:**

* The build process now includes a step to generate Vue wrappers, with a new script and npm command (`generate:vue-wrappers`), and updates to `package.json` exports to better support consumption from different environments and frameworks. [[1]](diffhunk://#diff-c389df12f1e9b46cad8e51faca5e490c8f5e85bccabe916918ab282509e83ce5L30-R31) [[2]](diffhunk://#diff-c389df12f1e9b46cad8e51faca5e490c8f5e85bccabe916918ab282509e83ce5L39-R44) [[3]](diffhunk://#diff-e21684df3313d6b0718ab5dbff278934250524b43e59fdc58e4bf78f77b912dfR10) [[4]](diffhunk://#diff-e21684df3313d6b0718ab5dbff278934250524b43e59fdc58e4bf78f77b912dfR90-R105) [[5]](diffhunk://#diff-e21684df3313d6b0718ab5dbff278934250524b43e59fdc58e4bf78f77b912dfR159)

**Storybook and Documentation:**

* New stories and improved story rendering for `craft-action-item` and `craft-switch` components, demonstrating new features and usage patterns. [[1]](diffhunk://#diff-d6a8a7fc3910459ed9f589b858f40fa2efd8302277aa5f2a67c386e6f44210f4L14-R20) [[2]](diffhunk://#diff-d6a8a7fc3910459ed9f589b858f40fa2efd8302277aa5f2a67c386e6f44210f4R66-R95) [[3]](diffhunk://#diff-abbe0ac1d2de0a36c518d7c2034189c2c72d12bee1e5ef934ad0007f6d8221d6R1-R81)


src/Http/Controllers/Settings/SectionsController.php
```diff
@@ -35,84 +34,143 @@
...
+            default => SORT_ASC,
+        };
+
+        [$pagination, $tableData] = $sections->getSectionTableData(
+            page: $page,
+            limit: $limit,
+            searchTerm: $searchTerm,
+            orderBy: $orderBy,
+            sortDir: $sortDir,
+        );
```
brianjhanson: This doesn't feel great. Ideally we'd use Laravel's [`paginate` stuff](https://laravel.com/docs/12.x/pagination#introduction) but I'm not sure we're quite there yet.

src/Http/Controllers/Settings/SectionsController.php
```diff
@@ -35,84 +34,143 @@
...
+                'propagationMethod' => $section->propagationMethod->value,
+                'defaultPlacement' => $section->defaultPlacement->value,
+                'previewTargets' => $section->previewTargets ?? [],
+                /**
+                 * PHPStan thinks this can't be null because of the rules, but when a section is first created
+                 * it is `null`
+                 *
+                 * @phpstan-ignore nullCoalesce.property
+                 */
+                'entryTypes' => $section->entryTypes ?? [],
```
brianjhanson: @riasvdv I'm not sure if you have any better ideas here. PHPStan seems to be inferring types from the `$section->getRules()` which says that `entryTypes` is required, but [it's set to `null` by default](https://github.com/craftcms/cms/blob/883a3fea4561d25d584fb301ee994aad5b0f5da8/src/Section/Data/Section.php#L82)

riasvdv: The private internal value is null by default, but the actual property is not nullable: https://github.com/craftcms/cms/blob/883a3fea4561d25d584fb301ee994aad5b0f5da8/src/Section/Data/Section.php#L75

brianjhanson: I was getting some kind of error earlier but it's gone now 🤷. Removed

src/Entry/Data/EntryType.php
```diff
@@ -200,14 +201,7 @@ public function getRules(): array
...
             'id' => ['nullable', 'integer'],
-            'color' => [
-                'nullable',
-                'string',
-                Rule::in(array_merge(
-                    array_map(fn (Color $color) => $color->value, Color::cases()),
-                    ['__blank__']
-                )),
-            ],
+            'color' => ['nullable', new Enum(Color::class)],
```
brianjhanson: @riasvdv not sure if this change is a red flag or not. Without it, when trying to save an entry type with a color the validation would fail because the value was an enum but the validation rule was looking for a string.

src/Twig/NodeVisitors/SinglePreloader.php
```diff
@@ -89,7 +89,7 @@ public function leaveNode(Node $node, Environment $env): \Twig\Node\Node
         ]));
 
         foreach ($subNodes as $key => $subNode) {
-            $body->setNode($key, $subNode);
+            $body->setNode((string) $key, $subNode);
```
brianjhanson, line 92: @riasvdv this is another one I ran into when trying to save Entry Types with a color. Seems pretty harmless to me, but wanted to flag it just in case.

riasvdv: Yeah that’s probably just because we have declare strict types now

src/Entry/Data/EntryType.php
riasvdv: What about the `__blank__ ` case? Was this being set as a string in the old controller implementation?

src/Http/Controllers/Settings/SectionsController.php
```diff
@@ -84,73 +84,89 @@ public function index(Request $request, Sections $sections): \Inertia\Response
...
+                'value' => $type->value,
+                'label' => $type->label(),
+            ], SectionType::cases()),
+            'propagationOptions' => [
+                ['value' => 'none', 'label' => t('Only save entries to the site they were created in')],
+                ['value' => 'siteGroup', 'label' => t('Save entries to other sites in the same site group')],
+                ['value' => 'language', 'label' => t('Save entries to other sites with the same language')],
+                ['value' => 'all', 'label' => t('Save entries to all sites enabled for this section')],
+                ['value' => 'custom', 'label' => t('Let each entry choose which sites it should be saved to')],
+            ],
```
riasvdv: There's a `PropagationMethod` enum which we can probably use for this and give it a `label()` method

src/Http/Controllers/Settings/SectionsController.php
```diff
@@ -84,73 +84,89 @@ public function index(Request $request, Sections $sections): \Inertia\Response
...
+                'handle' => $section->handle,
+                'type' => $section->type->value ?? SectionType::Channel->value,
+                'enableVersioning' => $section->enableVersioning,
+                'maxAuthors' => $section->maxAuthors ?? 1,
+                'maxLevels' => $section->maxLevels,
+                'propagationMethod' => $section->propagationMethod->value,
+                'defaultPlacement' => $section->defaultPlacement->value,
+                'previewTargets' => $section->previewTargets ?? [],
+                'entryTypes' => $section->entryTypes,
+            ],
```
riasvdv: Should we introduce a `SectionResource`?

brianjhanson:
@riasvdv I made a `CraftCms\Cms\Cp\CanSelect` trait for enums that allows us to easily format them as `value`, `label` for selection on the frontend but if you have a better way to do that, or feel like that trait should live elsewhere I'm not attached to the name or location. 

I also added a `composer ci` script so I can run all the things GitHub does locally more easily, but if you think that script name is bad (it may be) we can change it

Issue #18500, opened 2026-03-02
[4.x]: Redactor content link tags getting target=_blank attributes after draft auto save


7ochem:
### Description
We're running Craft CMS 4.x with Redactor as a rich text editor. The CMS admin runs on a subdomain (e.g. https://cms.example.org) while the website front side runs on the main domain (e.g. https://www.example.org).

When I change some text in a rich text field and bring the focus to another field, then a draft auto save is being kicked off and all the links to other entries are getting `target="_blank"` and `rel="noopener"` attributes added wrongfully. 

I traced this back to [ElementEditor.js](https://github.com/craftcms/cms/blob/4.x/src/web/assets/cp/src/js/ElementEditor.js#L1837) which calls `Craft.initUiElements(this.$contentContainer)` with the idea to initiate potentially uninitiated fields.

Inside `initUiElements()` there's a piece of code that adds `target="_blank"` and `rel="noopener"` when the link hostname is not equal to the window location hostname.

https://github.com/craftcms/cms/blob/3262cfa628ade304f0e565f9e45b52fab0f32df3/src/web/assets/cp/src/js/Craft.js#L1920-L1930

Caused by this commit craftcms/cms@ab90c4bc5329164e2408b0c44caa4e26c81ce852 targeting to fix issue #17239. Released in 4.15.4 and 5.9.14.

### Steps to reproduce

1. Run a front side on a different (sub)domain than the CMS admin
2. Have an entry with a URL config setup on that front side domain
3. Have a Redactor field where you can add links
4. Link from one entry to another in the text, leave "open in new window" checkbox unchecked
5. Re-open the created entry and change something in the text (not the link)
6. Set the focus to a different field, so a "save draft" is triggered
7. Check the link again: attributes `target="_blank"` and `rel="noopener"` were added, as if I have checked the "open in new window" checkbox
8. Click the link and choose "Edit" and you'll see that the "open in new window" checkbox is now indeed checked

### Expected behavior

The CMS scripts should not add attributes unintended. Maybe, to initialise elements that are only shown under a condition, it should target those elements specifically instead of running initialisation on the whole container.

### Actual behavior

Attributes `target="_blank"` and `rel="noopener"` should only be added if I have intentionally checked the "open in new window" checkbox

### Craft CMS version

4.17.6

### PHP version

8.3.30

### Installed plugins and versions

-

i-just:
Hi, thanks for reporting! I raised a PR for this.

7ochem:
[...]

PR #18502, opened 2026-03-03
ignore wysiwyg content when adding targets to outbound links

i-just:
### Description
Exclude links located in the `contenteditable` part of a WYSIWYG field from automatically getting the `target` and `rel` attributes.

### Related issues
#18500


7ochem (approved):
I can confirm that this solution works. If I override `/cpresources/b23bdd32/cp.js` in my Developer Toolbar Sources tab and add `.not('[contenteditable="true"] a')`, then the issue mentioned in #18500 does not occur 👍🏻

PR #18510, opened 2026-03-04
Replace thamtech/yii2-ratelimiter-advanced with built-in IP rate limiter

timkelty:
`thamtech/yii2-ratelimiter-advanced` is outdated and breaks with cache components defined as functions or that don't extend `yii/caching/Cache`. So, this PR:

- Removes the `thamtech/yii2-ratelimiter-advanced` dependency, in favor of Yii's built-in `RateLimiter`.
- Adds `craft\filters\IpRateLimitIdentity` to extend IP-based rate limiting to Yii's `RateLimiter`.

This is intended as lateral change - the rate limiting behavior for `send-password-reset-email` remains equivalent.


src/controllers/UsersController.php
```diff
@@ -3131,14 +3116,14 @@ private function clearPassword(ModelInterface|Model $model): void
     }
 
     /**
-     * @throws BadRequestHttpException
```
timkelty: This wasn't really part of this change. Not sure why, but this made PHPstan angry.
[It was pre-existing](https://github.com/craftcms/cms/actions/runs/22601318674/job/65483597229).

PR #18511, opened 2026-03-04
Add option to hard delete elements via GraphQL mutations

eirabben:
### Description
There is currently no way to specify if elements should be hard deleted via GraphQL mutations.

This PR adds a `hardDelete` option to delete mutations for Entries, Assets, Categories and Tags. Here is an example of a query using the new parameter:

```graphql
mutation HardDeleteEntry {
  deleteEntry(id: 12345, hardDelete: true)
}
```

I could not find any tests related to these parameters, but let me know if there are any, and I will update the PR.

### DIFF
--- a/src/gql/mutations/Asset.php
+++ b/src/gql/mutations/Asset.php
@@ -59,7 +59,10 @@ class Asset extends Mutation
         if ($createDeleteMutation) {
             $mutationList['deleteAsset'] = [
                 'name' => 'deleteAsset',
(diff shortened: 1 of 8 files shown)


brandonkelly:
Thanks for the PR! I merged it into the `5.10` branch via c6854d944dfe8625af99b055d8bc727e3edbc73f.

brandonkelly:
Craft 5.10.0 is out with this change. Thanks again!

PR #18539, opened 2026-03-11
Broken plugin store when loading remote Cloud assets locally

angrybrad:
Fixed a bug where the Plugin Store would not load if you set up Craft to use remote Craft Cloud assets locally

Can be merged into 5.x

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Release Notes for Craft CMS 4
 
+## Unreleased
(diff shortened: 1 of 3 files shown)


timkelty (changes requested):
I think its better to just do this in craftnet.
See https://github.com/craftcms/craftnet/pull/1310

angrybrad:
@timkelty yup, i'm ok with this.

PR #18558, opened 2026-03-13
Updates on Sites page

gcamacho079:
### Description
**Marking as a draft because I temporarily created a `ModalFormNew.vue` component rather than refactoring/replacing the existing one everywhere it’s used. Only the modals on the Sites page have been replaced.**

- Updated `craft-modal` (previously `craft-dialog`) to use Lion’s overlay system. In addition, added a name property that applies an accessible name to the `dialog` element.
- Modified `ModalForm` (`ModalFormNew`) to fire a `close-modal` event that targets the web component it’s nested inside
- Rewrites `craft-action-menu`:
    - Uses Lion properties for accessing overlay nodes
    - Removes `close-overlay` events attached to action items. Instead, I handled this on the Vue side; alternatively, we could use the [`isBlocking` property](https://lion.js.org/fundamentals/systems/overlays/configuration/#isblocking) to achieve the same visual result.
- Adds a Storybook test for checking focus management on `craft-modal` open and close events
- Modifies colors slightly to meet minimum contrast requirements

**To Do**
- [ ] Refactor all usages of `<ModalForm>` to use new component
- [ ] Refactor `<Modal>` to use new Lion component
- [ ] Implement width property in Lion component
- [ ] Change name back to Dialog


packages/craftcms-cp/src/components/modal/modal.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  override firstUpdated(c: PropertyValues<this>) {
+    super.firstUpdated(c);
+
+    if (c.has('name')) {
+      this.__setAccessibleName();
+    }
+  }
+
+  override updated(c: PropertyValues<this>) {
```
brianjhanson: What does `c` stand for in this case? In general, I'm a fan of long, explicit varaible names over brevity.

resources/js/pages/SettingsSitesIndex.vue
```diff
@@ -32,33 +47,38 @@
...
   });
 
+  const MODAL_ACTIONS = {
+    rename: {
+      label: t('Rename Group'),
+    },
+    create: {
+      label: t('New Group'),
+    },
+  };
```
brianjhanson: I feel like it would be easier just to throw these into the components vs. abstracting them up here

resources/js/components/ModalFormNew.vue
```diff
@@ -0,0 +1,57 @@
...
+  function resetHandler(event: Event) {
+    const {target} = event;
+    target!.dispatchEvent(new Event('close-overlay', {bubbles: true}));
+    emit('close');
+  }
+</script>
+
+<template>
+  <form @submit.prevent="submitHandler">
+    <Pane :title="title">
```
brianjhanson: `Pane` takes an `as` parameter so you can determine what element will be used to render it.

```suggestion
    <Pane as="form" :title="title" @submit.prevent="submitHandler">
```

resources/js/pages/SettingsSitesIndex.vue
```diff
@@ -19,6 +19,21 @@
...
   import IndexLayout from '@/layout/IndexLayout.vue';
+  // Add refs for the two elements we need to control
+  const siteGroupActionsMenu = ref<HTMLElement & {opened: boolean}>();
+  const renameGroupModal = ref<HTMLElement & {opened: boolean}>();
+
+  async function openRenameModal() {
+    // Close the action menu first — modal is no longer nested inside it
+    if (siteGroupActionsMenu.value) {
+      siteGroupActionsMenu.value.opened = false;
+    }
```
brianjhanson: It seems a bit overbearing for the page level component to have to manage which action menus are open on any given page. Could we make action menus more independent? 

Is there some kind of blanket rule? Like when a modal opens, make sure all action menus are closed first? 

I'm just thinking on complex pages, we could have 20 – 30 lines of action menu state management making the component harder to parse.

resources/js/pages/SettingsSitesIndex.vue
```diff
@@ -19,6 +19,21 @@
...
+    // Wait one tick for the hide to begin before showing the modal
+    await nextTick();
+    if (renameGroupModal.value) {
+      renameGroupModal.value.opened = true;
+    }
```
brianjhanson: For this one, I think we can just use a regular old ref and throw it on the `:opened` prop

```
const renameModalOpen = ref(false);

// Then

<craft-modal :opened="renameModalOpen">
```

brianjhanson (commented):
I think the modal updates look great! I still lean towards using the name `dialog` instead of `modal`, mainly because of the `<dialog/>` element, but it’s not a huge deal to me.

Most of my comments are minor details, but my main question is about how we’re managing action menus. I think action menus should close when an item is selected (maybe that’s a bit of a stretch), but even if they don’t, I believe we need to find a way to handle their state that doesn’t require the parent to manage all the action menus inside it.

I’m not entirely sure what that entails. It seems like something that might be covered by Lion’s `OverlayController`, but I’d need to dive a bit deeper into how that works and what it does.

packages/craftcms-cp/src/components/modal/modal.ts
gcamacho079: Ah, agreed! Changed to `changed`

packages/craftcms-cp/src/components/action-menu/action-menu.ts
```diff
@@ -49,41 +38,30 @@ export default class CraftActionMenu extends OverlayMixin(LitElement) {
...
-    this._setupInvoker();
-    this._setupContent();
+    this.__setupInvoker();
+    this.__setupContent();
   }
 
   override firstUpdated() {
     this.uid = uuid();
-    this._addEventListeners();
   }
```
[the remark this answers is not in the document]
gcamacho079: Currently a work-in-progress. 👍🏼 Trying to think through the best way to handle this while maintaining the “trigger stack” so we don’t have to manually manage focus when a modal is closed and the trigger element can’t be found.

PR #18570, opened 2026-03-17
[6.x] Mail revisited

riasvdv:
### Description

@AugustMiller correctly identified I overlooked the site specific overrides in the email settings which I didn't account for when moving the mail system to Laravel's mailer, which resulted in removing the settings a bit too eagerly.

This PR re-introduces the email settings, including site overrides, with a `CraftMailable` that can be used by developers to opt-in to Craft's email settings, the `SystemMessageMailable` also extends this mailable.

The main difference between 5.x and 6.x is that there is no transport and transport settings anymore, all that's possible is to choose a Laravel configured `mailer`. If none is chosen then the default Laravel mailer is used.

Instead of reusing the old twig template and controller, I've re-implemented the settings screen in Inertia. [The Site overrides are very basic](https://github.com/user-attachments/assets/fb2fbe44-7ec6-4f27-9547-211d212e7cf8) and @brianjhanson should definitely take another look at it.


brianjhanson (approved):
Made two larger changes

1. Updated the table to be an `AdminTable` with the `useEditableTable` composable
2. Switched out `craft-combobox` for a `CraftCombobox` Vue component. 

I'd like to have our combobox be a web component but so far when I've tried using it there are weird bugs around models updating and form submissions. Keeping everything in Vue has been better for now. 

I'll likely delete the `craft-combobox` component to make things a little less confusing but this didn't feel like the right PR to do that in.

PR #18574, opened 2026-03-17
Removed validation that enforced the maxTime to be later than the minTime for the time field

bramnijssen:
### Description
The current validation conflicts with how the timepicker component behaves. The timepicker natively supports wrapping past midnight and allows selecting times up to 24 hours after the defined minTime.

For example: Setting only minTime to 09:00AM already produces a selectable range that continues past midnight and wraps until 08:55AM.

The proposed change enables valid use cases where the time range spans across midnight. For example:
- Start: 10:00AM
- End: 02:00AM

Previously, this configuration was blocked by validation despite being fully supported by the timepicker.

### DIFF
--- a/src/validators/TimeValidator.php
+++ b/src/validators/TimeValidator.php
@@ -9,7 +9,6 @@ namespace craft\validators;
 
 use Craft;
 use craft\helpers\DateTimeHelper;
-use craft\i18n\Locale;
 use DateTime;
 use yii\base\InvalidConfigException;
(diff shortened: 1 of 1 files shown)


brandonkelly:
Thanks for the PR! That change would have broken TimeValidator in general though.

I added support for overnight time ranges for Craft 5.10 via #18575.

brandonkelly:
Craft 5.10.0 is out with that change.

PR #18595, opened 2026-03-19
[6.x] Upgrade to Vite 8

riasvdv:
### Description

Seems like a nice speed boost


vite.config.js
```diff
@@ -2,7 +2,6 @@ import {defineConfig, loadEnv} from 'vite';
 import laravel from 'laravel-vite-plugin';
 import fs from 'fs';
 import vue from '@vitejs/plugin-vue';
-import tsconfigPaths from 'vite-tsconfig-paths';
```
brianjhanson, line 5: Does Vite8 include this or something like it by default? Without it we previously would have to define aliases both in the tsconfig and in the vite config

brianjhanson (approved):

vite.config.js
riasvdv: Yeah it had a warning about it and you now just add something to the config 👍

PR #18610, opened 2026-03-23
[6.x] A new CLI banner

riasvdv:
Thought it might be nice

image


brandonkelly:
We could probably ditch the “A new install” text  – that was meant to be a Star Wars “A New Hope” reference, to go with the Star Wars-styled “Craft CMS” text.

Issue #18605, opened 2026-03-23
[5.x]: infinite loop, when trying to fetch a Site in general config callback `setPasswordPath`


Anubarak:
### Description
When you try to fetch a site in `setPasswordPath` to set the password path depending on the current site you currently end up in an infinite loop.

### Steps to reproduce

1. have a Craft 5.9.x system
2. grab a site by handle or execute getAllSites in one way or another (happens when you fetch an entry as well)
3. see the error `Fatal error: Allowed memory size of 536870912 bytes exhausted (tried to allocate 53248 bytes) in /app/vendor/yiisoft/yii2/web/ErrorHandler.php on line 266`

### Expected behavior
No error

### Actual behavior
inifinite loop due to Code in `craft\services\Sites`

here is the stack trace

```
> Exception: Infinite loop detected in /app/vendor/craftcms/cms/src/web/Request.php:1858
Stack trace:
#0 /app/vendor/craftcms/cms/src/web/Request.php(1845): craft\web\Request->_checkIfActionRequestInternal(true)
#1 /app/vendor/craftcms/cms/src/web/Request.php(683): craft\web\Request->checkIfActionRequest()
#2 /app/vendor/craftcms/cms/src/services/Sites.php(1278): craft\web\Request->getIsActionRequest()
#3 /app/vendor/craftcms/cms/src/services/Sites.php(672): craft\services\Sites->_allSites(NULL)
#4 /app/modules/site/Module.php(121): craft\services\Sites->getSiteByHandle('default')
#5 /app/config/general.php(37): modules\site\Module::getSingle('setPassword', 'default')
#6 /app/vendor/craftcms/cms/src/helpers/ConfigHelper.php(118): craft\services\Config->{closure}('default')
#7 /app/vendor/craftcms/cms/src/config/GeneralConfig.php(7411): craft\helpers\ConfigHelper::localizedValue(Object(Closure), 'default')
#8 /app/vendor/craftcms/cms/src/web/Request.php(1906): craft\config\GeneralConfig->getSetPasswordPath()
#9 /app/vendor/craftcms/cms/src/web/Request.php(1845): craft\web\Request->_checkIfActionRequestInternal(true)
#10 /app/vendor/craftcms/cms/src/web/Request.php(683): craft\web\Request->checkIfActionRequest()
#11 /app/vendor/craftcms/cms/src/services/Sites.php(1278): craft\web\Request->getIsActionRequest()
[...]

PR #18613, opened 2026-03-24
don't check if action request for site requests

i-just:
### Description
When figuring out if getting all sites should return the disabled ones, we initially did this:
https://github.com/craftcms/cms/commit/e4165bd3276865e52a2d3f70ef294a5a6b5d7e5a
and then we changed it to this:
https://github.com/craftcms/cms/commit/a8faf8bb10d9b449ca8a2c365378f1f3c8828f82

That recent change triggers an infinite recursion for site requests if you’re trying to make one of the special paths site-dependent.

**Steps to reproduce:**
- clean 5.9.17 installation
- add this to the `config/general.php`:
```
	->setPasswordPath(function (string $handle) {
        $site = Craft::$app->sites->getSiteByHandle($handle);

        return 'my-set-password';
    })
```
- visit any site on the front end, e.g. the homepage

If you have Xdebug enabled, it will error quite quickly with “Xdebug has detected a possible infinite loop”.

The root cause is, as mentioned by the OP.

**Solution:** don’t include disabled sites for site requests, and don’t attempt to check for action requests in the process.

### Related issues
#18605

### DIFF
--- a/src/services/Sites.php
+++ b/src/services/Sites.php
@@ -1275,7 +1275,8 @@ class Sites extends Component
     {
         if ($withDisabled === null) {
             $request = Craft::$app->getRequest();
-            $withDisabled = !$request->getIsSiteRequest() || $request->getIsActionRequest();
+
+            $withDisabled = !$request->getIsSiteRequest() && $request->getIsActionRequest();
         }
 
         return $withDisabled ? $this->_allSitesById : $this->_enabledSitesById;


brandonkelly:
That change would have prevented disabled sites from being loaded on CP requests too, unless they were CP action requests.

I think it already has the right logic there – we do want to load disabled sites for all action requests (regardless of whether it’s front-end or CP).

Fixed by just setting `$this->_checkedRequestType = true` before calling `$this->_checkIfActionRequestInternal()`, so if one of the config settings ends up causing some recursion, the second call to `getIsActionRequest()` will immediately return `true` or `false`.

PR #18616, opened 2026-03-24
[6.x] Migration and Plugin migration improvements

riasvdv:
### Description

- Adds a way to configure the migrations path a plugin loads migrations from using `getMigrationsPath`, defaults to Laravel convention + Craft's previous location for those files.

- Refactors the installation so all Laravel's optional migrations are also published and ran during a Craft install. This ensures we can use sessions, cache, batched jobs, etc without needing users to publish the migrations themselves. Using the framework commands to publish them ensures we don't need to maintain parity.

### DIFF
--- a/src/Database/LaravelMigrations.php
+++ b/src/Database/LaravelMigrations.php
@@ -24,15 +24,21 @@ class LaravelMigrations
             return;
         }
 
-        $migrator->track('content');
+        $originalTrack = $migrator->getTrack();
 
-        $pendingMigrations = $migrator->getPendingMigrations($migrationPaths);
+        try {
+            $migrator->track('content');
 
-        if (empty($pendingMigrations)) {
-            return;
+            $pendingMigrations = $migrator->getPendingMigrations($migrationPaths);
+
+            if (empty($pendingMigrations)) {
+                return;
+            }
+
+            $migrator->run($pendingMigrations);
+        } finally {
+            $migrator->track($originalTrack ?? 'content');
         }
-
-        $migrator->run($pendingMigrations);
     }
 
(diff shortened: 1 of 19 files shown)


src/Database/LaravelMigrations.php
```diff
@@ -0,0 +1,103 @@
...
+    {
+        $this->publishMigrationFiles();
+
+        $migrationPaths = $this->migrationPaths();
+
+        if (empty($migrationPaths)) {
+            return;
+        }
+
+        $migrator->track('content');
```
AugustMiller: Should we restore the original track at the end of this method?

src/Plugin/Concerns/Installable.php
```diff
@@ -57,23 +60,47 @@ public function uninstall(): void
...
+        $migration = require $path;
+
+        if (is_object($migration)) {
+            return $migration;
+        }
+
+        return $this->findMigrationClassByFile($realPath);
+    }
+
+    private function findMigrationClassByFile(string $realPath): ?object
```
AugustMiller, line 86: This is required because Laravel’s migrations don’t actually have class names? (i.e. `return new class extends Migration {}`) 😬

AugustMiller (approved):

src/Plugin/Concerns/Installable.php
riasvdv: Yes it supports both but the recent convention is no class name

src/Database/LaravelMigrations.php
riasvdv: With current callsites it's not necessary, but a good call to do it regardless as it might impact future calls

PR #18633, opened 2026-03-26
[6.x] Inertia 3

brianjhanson:
Update to Inertia 3


src/Providers/AppServiceProvider.php
```diff
@@ -109,6 +109,13 @@ public function boot(): void
                 'report' => false,
             ],
         ]);
+
+        // Configure Inertia
+        Config::set('inertia.ssr.enabled', false);
+        Config::set('inertia.use_script_element_for_initial_page', true);
+        Config::set('inertia.page_paths', [
+            resource_path('js/pages'),
+        ]);
```
brianjhanson: @riasvdv I'm not sure if this is the best place for this kind of thing so let me know if you think it belongs elsewhere.

riasvdv: What is the reason we need to explicitly set those? I know we needed page_paths in tests in the previous version (should be in TestCase somewhere)

Preferably everything would work with the defaults though

riasvdv: `use_script_element_for_initial_page` doesn't seem to exist and it looks like the script element is the default now anyway

`inertia.page_paths` seems to be renamed to `inertia.pages.paths`, so setting this doesn't currently do anything and can be removed

`ssr` being enabled on the Laravel side doesn't mean it'll work right away, I'd prefer to just disable it in vite.config.js:

Image

brianjhanson: I agree it seems like `use_script_element_for_initial_page` doesn't exist. And yet...

https://github.com/user-attachments/assets/2e1962e4-64e6-4577-9ca1-eafb98ece644

brianjhanson:
@riasvdv not sure what happened but it just started working? I'm assuming I had a cache somewhere that I was failing to clear. Mildly concerning, but things are working on the defaults now. 

Edit: offline (well, still online, but away from here) we determined that my issues were because I didn't run `composer update` on my test project. After running that, things started working.

yii2-adapter/composer.lock
brianjhanson, line 1: @riasvdv should this file even be in this diff? Or should I reset its change?

riasvdv: Yeah that's fine

PR #18649, opened 2026-03-30
[6.x] Entry type handle validation

AugustMiller:
Placeholders were showing up in the `handle` validator on entry types; once that was fixed, the error was cryptic (like "“2” is not a valid handle.”), which led me to the second patch.

Checking `$saveAsNew` seemed appropriate before trying to increment the handle; if someone chooses an existing handle, we probably shouldn't quietly swallow the collision.

### DIFF
--- a/src/Http/Controllers/Settings/EntryTypesController.php
+++ b/src/Http/Controllers/Settings/EntryTypesController.php
@@ -268,7 +268,7 @@ class EntryTypesController
         $entryType->showStatusField = $request->boolean('showStatusField', $entryType->showStatusField);
 
         // If we're duplicating the entry type and the handle hasn't changed, find a unique one
-        if ($entryType->handle === ($originalEntryType->handle ?? null)) {
+        if ($saveAsNew && $entryType->handle === ($originalEntryType->handle ?? null)) {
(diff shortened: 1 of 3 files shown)


src/Http/Controllers/Settings/EntryTypesController.php
```diff
@@ -268,7 +268,7 @@ public function store(Request $request): Response
         $entryType->showStatusField = $request->boolean('showStatusField', $entryType->showStatusField);
 
         // If we're duplicating the entry type and the handle hasn't changed, find a unique one
-        if ($entryType->handle === ($originalEntryType->handle ?? null)) {
+        if ($saveAsNew && $entryType->handle === ($originalEntryType->handle ?? null)) {
```
riasvdv, line 271: This is logic straight from 5.x: https://github.com/craftcms/cms/blob/5.x/src/controllers/EntryTypesController.php#L240

Do we need to change it there as well then?

brandonkelly: `$originalEntryType` is only defined when `$saveAsNew` is `true`, so there’s no way the condition would pass when `$safeAsNew` is `false`. So the change isn’t necessary, but doesn’t hurt and probably makes it a little more clear.

Issue #18679, opened 2026-04-08
[5.x]: "Save and add another" results in confusing redirect after subsequent save


amphibian:
### Description

When using "Save and add another" to create an entry, then doing a regular "Create entry" on the subsequent entry screen, you will be redirected to the entry you initially came from.

### Expected behavior

Redirect to the entry listing for the channel in question.

### Actual behavior

Redirect to the referring entry.

I know there was recently a couple of updates which dealt with entry redirection behaviour and trying to make it consistent, so maybe this is actually intended behaviour, but in practice this particular workflow feels _off_, and I already have clients asking me about it as though it's a bug.

### Craft CMS version

5.9.18

### PHP version

8.4

### Installed plugins and versions

-

brandonkelly:
Thanks for pointing that out! I just PR’d a fix for this, for Craft 5.10: https://github.com/craftcms/cms/pull/18680

brandonkelly:
Craft 5.10.0 is out with that fix. Thanks again!

Issue #18404, opened 2026-02-10
[5.x]: Unexpected dashboard redirects after save when sites share Translation Keys


mandrasch:
### Description

When managing multiple sites within the same Site Group that share a similar Translation Key logic, the control panel unexpectedly redirects the user to a different site view after hitting the save button for an entry.

### Steps to reproduce

Create a Site Group (e.g., "Group A").

Create two sites within that group with similar languages (e.g., de-DE and de-CH).

On an Entry Type, set the Title Translation Method to "custom" and set the translation key format to a dynamic string that results in the same key for both sites, for example: {{ object.site.group.handle ~ '-' ~ object.site.language|slice(0,2) }}

Open an entry in "Site 1", make an edit, and click Save.

<details>
<summary>View Screenshots for project conf</summary>

<img width="1354" height="276" alt="Image" src="https://github.com/user-attachments/assets/75129707-d99c-4c25-8cd6-5382bdcd4791" />

<img width="937" height="954" alt="Image" src="https://github.com/user-attachments/assets/174a2c3f-e846-4d10-8bef-5d372262e013" />

<img width="493" height="222" alt="Image" src="https://github.com/user-attachments/assets/63ecc17d-dfd7-4471-be6c-6b5ba5901758" />

<img width="926" height="1133" alt="Image" src="https://github.com/user-attachments/assets/9b264cac-eda2-4307-8390-f6d8e57fe7c3" />

<img width="886" height="962" alt="Image" src="https://github.com/user-attachments/assets/77895a36-84b8-4461-9856-dab3fb694301" />

</details>

### Expected behavior

Regular "Entry saved" confirmation, redirect back to entries

### Actual behavior

The dashboard redirects to another language version of the entry:
https://github.com/user-attachments/assets/d56ece0f-1ca4-4599-a59c-187e5c6a7a62

Important: It's specifically when you click Save, rather than 'Save and continue editing'.

### Craft CMS version

5.9.5

### PHP version

8.3

### Operating system and version

OSX / DDEV version: v1.24.10

### Database type and version

8.0

### Installed plugins and versions

Fresh install

brandonkelly:
This is happening due to the changes in #16140. We ran into a similar strange behavior a couple days ago.

We decided to revert the change for now, until we can think it through a little more thoroughly. Craft 5.9.8 is out now with that change.

mandrasch:
Thanks very much for the quick fix, @brandonkelly!

brandonkelly:
[...]

PR #18680, opened 2026-04-08
Explicitly set `returnUrl` params on element edit pages

brandonkelly:
### Description

Currently after saving a full-page element edit form, the browser is redirected to the referrer URL, so long as it doesn’t look like the URL to the same exact element’s edit page.

That can lead to some awkward behaviors, depending on how you get to the element edit page to begin with. For example, if you press “Save and continue editing” or “Save and add another” from an element’s edit page, the subsequent page will redirect back to the initial edit page (see #18679).

This PR drops the automatic redirection based on the referral URL, in favor of explicitly setting a `returnUrl` param on element edit pages, when accessing the page from an element index.

It also updates the “Save and continue editing”, “Save and add another”, and “Revert content from this revision” actions to maintain the `returnUrl` param, so the original element index URL will be remembered even as authors interact with those actions.

### Related issues

- #16140
- #18404
- #18483
- #18679


i-just (changes requested):
that's the only issue I found when testing:
- clean installation running branch for this PR
- create a couple of sections (Blog, My Channel); the entry type can contain all the defaults and only the title
- create a custom page called “Pages”
- add a Blog section to it 
- go to CP > Pages > Blog; add an entry to it & fully save
- ensure you’re still in CP > Pages > Blog and edit the entry from the previous step
- choose “save as a new entry” alternative action
- press “Create new entry” (or “Delete draft”)
- you're redirected to the Entries > All entries page; you should be redirected to the Pages > Blog


not sure if this is expected behaviour:
- the setup listed above
- create an Entries field
- add that field to the entry type used by “My Channel”
- ensure you have an entry in that channel and that it has an entry from the Blog section selected in the Entries field
- go to CP > Entries > My Channel and edit the entry from the previous step
- double-click or use the actions menu to edit the entry listed in the Entries field, then open it in a new tab
- press “Save”
- you’re redirected to the Entries > My Channel index page

I get why this is happening, but I’m not sure if this is expected from the user’s POV.

PR #18719, opened 2026-04-15
Fix project config ordering for partial loaded elementSources entries

wvell:
### Description
When project config is applied to a clean install, pending changes are currently
sorted only by path depth. That can cause `sections` changes to be processed
before related `elementSources` entries are fully constructed.

In our case, `handleChangedSection()` ends up reading element sources while an
`elementSources` entry only contains nested values like `defaultSort` and
`tableAttributes`, but is still missing scalar keys such as `type`. That leads
to:

    Undefined array key "type"

in `ElementSources::defineSources()`.

| Before | After |
| --- | --- |
| `elementSources...defaultSort` | `elementSources...defaultSort` |
| `elementSources...tableAttributes` | `elementSources...tableAttributes` |
| `sections...siteSettings` | `elementSources...0` |
| `elementSources...0` | `sections...siteSettings` |
| `sections...entryTypes` | `sections...entryTypes` |

Before this change, `sections...siteSettings` could run before the parent
`elementSources...0` entry had been fully applied. After this change, all
`elementSources` paths are processed first, so section change handlers never
observe partially-built element source configs.

This change sorts pending changes by top-level config key first, and then by
depth within each group.

A minimal regression test has been added to reproduce that failure mode.

### Related issues
#18720

### DIFF
--- a/src/services/ProjectConfig.php
+++ b/src/services/ProjectConfig.php
@@ -1486,8 +1486,16 @@ class ProjectConfig extends Component
 
         unset($removedItem);
 
-        // Sort by number of dots to ensure deepest paths listed first
+        // Sort by top-level key first, then deepest paths first within each group.
         $sorter = function($a, $b) {
+            $aTop = explode('.', $a, 2)[0];
+            $bTop = explode('.', $b, 2)[0];
+            $topCmp = strcmp($aTop, $bTop);
+
+            if ($topCmp !== 0) {
+                return $topCmp;
+            }
+
             $aDepth = ProjectConfigHelper::pathDepth($a);
             $bDepth = ProjectConfigHelper::pathDepth($b);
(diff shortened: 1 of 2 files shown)


brandonkelly:
Can you send your project config files and Composer files into support@craftcms.com so we can look into this?

If I’m reading the changes right, the PR’d fix seems to be relying on the fact that `elementSources` comes alphabetically before `sections`, which doesn’t seem like a very solid solution.

brandonkelly:
Nevermind, I was able to reproduce the issue with the test you added, and wrapped my head around the issue.

Modified your fix a bit so that paths are sorted by their similarity across all path segments, not just the first one (cf50e0b0e8aeca53d925599a7439943676ae3dab), and merged it in for the next release.

Thanks for the PR!

PR #18721, opened 2026-04-15
[6.x] Internal legacy assets

riasvdv:
### Description

This ports a minimal implementation of asset bundles to make them work, so we have the ability to extract the yii2-adapter when the time comes while still being able to rely on the legacy frontend assets.

To get rid of the legacy AssetManager dependency, this also ports the Rebrand functionality to GeneralConfig instead of relying on an arbitrary folder


brianjhanson:
@riasvdv this looks good to me! I don't see anything in here related to the build process. I'm assuming you didn't mess with that for the moment?

Issue #18692, opened 2026-04-13
[5.x]: Generated fields on global sets are not accessible through reference tags


lexisother:
### Description

Basically as the title says, I'm not getting any text when trying to access generated fields through reference tags.

### Steps to reproduce

1. Set up a fresh Craft CMS instance with the following structure (the existence of a section for the `Content Item` entry type is implied):
```
Global Sets
|- My Global Set (myGlobalSet)
   |- Description (description)              | Plain text
   |- Linked item (linkedItem)               | Entry field for Content Item entries
   |- Linked description (linkedDescription) | Generated field: `{{ object.linkedItem.one().description }}`

Entry Types
|- Content Item
   |- Title (title)
   |- Description (description) | Plain text
```
2. Create a new Content item and fill in the description
3. Link the new entry to the global set's `linkedItem` field
4. Attempt to render the global set out to the frontend (either through a template or `View#renderString`)
5. Observe how the regular description works and the linked description remains empty
For clarity:

<img width="725" height="213" alt="Image" src="https://github.com/user-attachments/assets/2947cf70-6bb3-4dee-b0c3-720989f40e7a" />

### Expected behavior

The field value is correctly rendered.

### Actual behavior

The field renders as empty.

### Craft CMS version

5.9.19

### PHP version

8.3.30

### Operating system and version

macOS 26.3 (running DDEV with kernel `Linux version 6.17.8`)

### Database type and version

MySQL 8.0.40

### Image driver and version

[N/A]

### Installed plugins and versions

[N/A]

i-just:
Hi, thanks for reporting! I just raised a PR that adds the ability to parse generated fields in reference tags. The PR is targeting v5.10.

brandonkelly:
Craft 5.9.21 is out with a fix for this. Thanks again!

lexisother:
Thanks for the quick fix!

PR #18725, opened 2026-04-16
make reference tags parsing work with generated fields

i-just:
### Description
Adds the ability to parse generated fields in reference tags.

To test:
- create a section with an entry type that has a generated field with a handle of `myTest` and a value of `abcde`
- create an entry in that section
- create a twig template with the following code: `<p>{{ craft.app.elements.parseRefs('{entry:194:myTest}') }}</p>` (replace the number with the actual ID of the entry from the previous step)
- view that template in a browser

Before this change, the result would be the entry’s URL (if the section is configured with URIs) or an empty string (if configured without URIs).

With this change, you should see `abcde`.

### Related issues
#18692


brandonkelly:
This ended up being just a bug in `Element::__isset()`, which wasn’t checking for generated field values, so I fixed it there, and that felt like a reasonable 5.9.x change.

PR #18734, opened 2026-04-20
[5.10] Add "is one of" and "is not one of" operators to text/number condition rules

nfourtythree:
### Description
It has been requested that text condition rules have the ability to have the "is one of"/"is not one of" operators.

This is so you could have a filter that says "Title is one of 'Hello', 'Hey' or 'Yo'", which is not possible with the currently available operators.

Adding this to the `BaseTextConditionRule` meant that `BaseNumberConditionRule` got this feature for "free". Have kept this in as I do think it is useful to be able to say something like "ID is one of 100, 999 or 123456".

One thing this did mean was that the `MoneyFieldCondition` needed to opt out of the feature. This is a decision based on the fact that inputting the data is different (due to currency) and there is less likely for a need in something with a monetary value.


i-just (changes requested):
Nice 🙂  

I found the following issues (which are probably all connected):

1. type error when “is one of”/“is not one of” is used in a field layout > visibility conditions > entry condition
	- create a section with an entry type with a title and plain text fields; 
	- set visibility conditions > entry condition for the Title to something like “plain text is one of abc”
	- try to create a new entry in this section or edit an existing one
	- type error shows:
```
		TypeError: in_array(): Argument #2 ($haystack) must be of type array, string given in /var/www/html/packages/cms/src/base/conditions/BaseTextConditionRule.php:224
		Stack trace:
		#0 /var/www/html/packages/cms/src/base/conditions/BaseTextConditionRule.php(224): in_array()
		#1 /var/www/html/packages/cms/src/fields/conditions/TextFieldConditionRule.php(39): craft\base\conditions\BaseTextConditionRule->matchValue()
		#2 /var/www/html/packages/cms/src/fields/conditions/FieldConditionRuleTrait.php(256): craft\fields\conditions\TextFieldConditionRule->matchFieldValue()
		#3 /var/www/html/packages/cms/src/elements/conditions/ElementCondition.php(282): craft\fields\conditions\TextFieldConditionRule->matchElement()
		#4 /var/www/html/packages/cms/src/base/FieldLayoutComponent.php(363): craft\elements\conditions\ElementCondition->matchElement()
		#5 /var/www/html/packages/cms/src/models/FieldLayout.php(1519): craft\base\FieldLayoutComponent->showInForm()
		#6 /var/www/html/packages/cms/src/controllers/ElementsController.php(1089): craft\models\FieldLayout->createForm()
		#7 /var/www/html/packages/cms/src/controllers/ElementsController.php(446): craft\controllers\ElementsController->_prepareEditor()
		#8 [internal function]: craft\controllers\ElementsController->{closure:craft\controllers\ElementsController::actionEdit():446}()
		#9 /var/www/html/packages/cms/src/web/CpScreenResponseFormatter.php(127): call_user_func()
		#10 /var/www/html/packages/cms/src/web/CpScreenResponseFormatter.php(50): craft\web\CpScreenResponseFormatter->_formatTemplate()
		#11 /var/www/html/vendor/yiisoft/yii2/web/Response.php(1111): craft\web\CpScreenResponseFormatter->format()
		#12 /var/www/html/packages/cms/src/web/Response.php(358): yii\web\Response->prepare()
		#13 /var/www/html/vendor/yiisoft/yii2/web/Response.php(339): craft\web\Response->prepare()
		#14 /var/www/html/vendor/yiisoft/yii2/base/Application.php(397): yii\web\Response->send()
		#15 /var/www/html/web/index.php(12): yii\base\Application->run()
		#16 {main}
```

(same thing will happen with the editability conditions)

2. same scenario as above, but once you added the condition and saved the entry type, edit it again - the text you previously added to the entry condition doesn’t show; (same thing will happen with the editability conditions);

3. issue 2 also affects the customise sources modal
	- go to the entry index page, click the 3 dots under the list of sources and click customise sources
	- add a custom source and set entry criteria to e.g. plain text is one of xyz; save
	- view the customise sources modal again and note that the value of the condition is missing

	This then leads to the entries that actually match the criteria not showing when that custom source is selected.

4. filtering entries by number field (is one of 1,3,5) causes an error: `craft\fields\conditions\NumberFieldConditionRule::elementQueryParam(): Return value must be of type ?string, array returned`

fwiw, with issues 2 and 3, the values appear to be saved in the project config;

nfourtythree:
Thanks for the notes @i-just I have pushed up some changes to combat those issues.

i-just (approved):

PR #18745, opened 2026-04-22
[6.x] Implement a contrast-ratio-based adaptive color theme

gcamacho079:
### Description
- Uses [Leonardo](https://leonardocolor.io/index.html) to generate color palettes for light and dark themes with color swatches dependent on specified contrast ratios. In addition to adaptive colors (i.e., `--color-blue-500`), I added static colors that don’t adapt to the theme (i.e., `--color-static-blue-500`)
- For added consistency, I moved the semantic color definitions and the defined fill, border, and foreground color indexes to the `generate-colors.js` script. White and black 
- Fixed some Storybook a11y errors, and updated the a11y checks in Storybook to trigger a fail. When necessary, I added per-story a11y settings to prevent existing issues from triggering a fail

TODO: Some colors may still require adjustment, but I wanted to get this implemented. For example, we’ve had a lot of issues in the past with our light switches not looking “green” enough in the active state, so we may need to use the static palette for this case. I can tackle this in an upcoming PR.


brianjhanson:
I agree we'll still need to tweak some colors, but hopefully we can do that without too much trouble along the road here. 

At the moment I find the CP to feel just a little too dark. 

Here for example, it feels like all of the colors are just a few clicks too dark, like you're wearing sunglasses or something. 

CleanShot 2026-05-01 at 13 53 30@2x
CleanShot 2026-05-01 at 13 54 40@2x

After thinking about it for a few seconds, I think that particular case is probably fixed by making callouts `quiet` instead of `loud`

Anyway, this is great!

PR #18747, opened 2026-04-23
[6.x]: Plugin Path Adjustments

AugustMiller:
Proposing a slight re-restructuring of the recommended plugin package…

- `config/` is now a top-level directory;
- `resources/` is now a top-level directory, and this is standardized by a new `Plugin::getResourcesPath()` method;
- Template roots are registered based on the new `resources/` location;
- Icons and icon “masks” (`InteractsWithCp`) are now expected to be in the `resources/` directory (not floating, adjacent to the base plugin class);
- The new `HasFrontendAssets::getPublishablePath()` method consolidates a number of instances in which we were building paths in the `public/vendor/` dir;

```
vendor/org/package-name/
    src/
        Plugin.php
        ...
    config/
        plugin-name.php
    resources/
        views/
        templates/
        build/
        js/
        css/
        ...
```

The main issue I see here is that if a plugin’s base class is at the top level of its package (not in a `src/` directory), we'll be traversing above its root while looking for config and resources. (This is a preexisting assumption that we make elsewhere in the plugin architecture!) If a plugin has this structure, it should override `getResourcesPath()`.


riasvdv:
Seems good to me, we’ll have to make sure the structure of existing plugins still works to make it easier on devs porting their plugins

brandonkelly:
@AugustMiller can you review the failing checks?

AugustMiller:
We’re already requiring that templates be moved into a `resources/` folder, so that shouldn’t be any additional work.

Plugin config files have historically only provided as stubs for developers to copy into their local project, not for plugins to interact with… and while this is technically not "compatible" with that convention, discovery + publishing of config files is a new feature in 6.x.

The only additional "breaking" change ends up being that `icon.svg` and `icon-mask.svg` need to be relocated to the top-level `resources/` directory!

(Re: checks—sorry for all the noise. I will spend some time next week getting them to run locally.)

AugustMiller:
I don't think I can get the failing test to pass, because it's only doing a string comparison. Is it appropriate to use `realpath()` in test in situations like this?

riasvdv:
> Is it appropriate to use realpath() in test in situations like this?

Yes or https://github.com/craftcms/cms/blob/6.x/src/Support/File.php#L34

riasvdv (approved):

PR #18754, opened 2026-04-24
[6.x] don't use php.ini file for the backup test command

i-just:
### Description
I’m using `ddev` locally. When I run the whole suite of tests via `ddev exec --dir /var/www/html/vendor/craftcms/cms ./vendor/bin/pest`, it’ll hang after completing the `tests/Feature/Dashboard/Widgets/WidgetTest.php` test.

The next test file in line is `tests/Feature/Database/BackupsTest.php`, and running `ddev exec --dir /var/www/html/vendor/craftcms/cms ./vendor/bin/pest tests/Feature/Database/BackupsTest.php` just hangs.

Long story short, the solution seems to be to add the `-n` param to the backup command, which will cause no `php.ini` file to be used in the command.

Is anyone else having this issue? 
Does anyone have an opinion on whether the proposed change is safe in our case?

### Related issues
n/a

### DIFF
--- a/tests/Feature/Database/BackupsTest.php
+++ b/tests/Feature/Database/BackupsTest.php
@@ -30,7 +30,7 @@ afterEach(function () {
 
 function backupsTestPhpCommand(string $script, array $args = []): string
 {
-    $command = escapeshellarg(PHP_BINARY).' -r '.escapeshellarg($script);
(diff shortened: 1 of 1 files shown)


riasvdv:
This is just in tests so if that fixed the issue for you and it also works in CI then it's all good 👍

PR #18766, opened 2026-04-29
[6.x] Bugfix/tweaks

i-just:
### Description
I started fixing some small issues I found on `6.x`.

1. when editing an entry, the URL is something like: `https://6.ddev.site/admin/content/entries/blog/6-test1`, but if the entry is fresh (you just pressed the “new entry” button, it’ll be something like or `https://6.ddev.site/admin/content/entries/blog/8?draftId=4&fresh=1`, so the `-slug` is optional (also, at least historically, you were able to access an existing entry via `https://6.ddev.site/admin/content/entries/blog/6-test1` but also `https://6.ddev.site/admin/content/entries/blog/6`). It’s the same story with the “old” non-content route.

2. error was thrown when trying to create a Matrix-type field

3. partial fix for cms-1980; fixes the first UI issue so that now both UI and CLI installation throw the same error; there’s more info about this in Linear.

4. In the CP > Utilities > Deprecation Errors table, the Stack Trace button was throwing a 405 error. The action URL was pointing to a URI that was missing a slash after the action trigger. 
@brianjhanson, it looks like the `getActionUrl()` in `resources/js/composables/useCraftData.ts` already had the right code, but it was commented out; not sure about the history behind it.

5. fixed redirection loop [CMS-1982](https://linear.app/craftcms/issue/CMS-1982/accessing-admincontentpagesection-route-throws-a-redirection-error)

### Related issues
n/a


routes/cp.php
```diff
@@ -102,7 +102,7 @@
     ]);
     Route::get('assets/edit/{id}{slug}', EditElementController::class)->where($idSlugParams);
     Route::get('entries/{section}/{id}{slug}', EditElementController::class)->where($idSlugParams);
-    Route::get('content/{page}/{section}/{id}{slug}', EditElementController::class)->where([
+    Route::get('content/{page}/{section}/{id}{slug?}', EditElementController::class)->where([
```
riasvdv, line 113: Can we add a test case for this? Can be a small one that just asserts it doesn't 404

brianjhanson:
@i-just I pushed a few commits to clean up the deprecation warnings. I need to get rid of `useCraftData` as well as all of the action client stuff now that we have [Wayfinder](https://github.com/laravel/wayfinder) in here which is a better way of referencing routes

routes/cp.php
i-just: I updated the test and also made the same change with the slug being optional for the “old” non-content route.

PR #18770, opened 2026-04-30
[6.x] QueueManager & JobProgress improvements

riasvdv:
### Description

- Adds the missing endpoints for the release and retry buttons in the QueueManager.
- Adds `dateCompleted` and `dateFailed` dates to the records (not yet shown in the component @brianjhanson)
- Keeps completed jobs around for 7 days after which they get pruned using GarbageCollection


resources/build/assets/cp.css
riasvdv, line 1: @brianjhanson I don't know what triggered all these changes on the compiled files, I only adjusted the QueueManagerIndex 😅

brianjhanson: It looks like `npm run build` doesn't retain the same minified variable names between runs, which is mildly annoying but ultimately harmless I think.

brianjhanson:
We don't currently show `dateCompleted` or `dateFailed` within the queue table so I'm not sure we need to add them in this PR. Now that the records will stick around for 7 days we probably should add them, but I want to think about how exactly to handle it since successful and failed jobs could be in the same table.

brianjhanson (approved):

PR #18771, opened 2026-04-30
[6.x] Replace custom user photo picker with Asset field

riasvdv:
### Description

**One small behavioral change**: Removing the asset from the user no longer deletes it but keeps the asset as that is consistent with how the asset field operates everywhere else.

image

image


brandonkelly:
We intentionally didn’t make these normal asset select inputs, because you wouldn’t want to give users the ability to choose from all other users’ avatars.

PR #18774, opened 2026-04-30
[6.x] Fix conditions validation

brianjhanson:
This issue ended up being within the validatior on the constructor. The request input name (e.g. `sources[custom:...][condition]`) uses bracket notation, but Laravel's `request->input()` and Validator require dot notation to access nested array data. This moves the bracket-to-dot conversion before theinput retrieval so both the data access and validation work correctly.


src/Http/Controllers/ConditionsController.php
```diff
@@ -46,25 +46,26 @@ public function __construct(
             ]);
         }
 
-        $config = $this->request->input($baseConfig['name']);
+        $name = str_replace(['[', ']'], ['.', ''], $baseConfig['name']);
```
riasvdv: `CraftCms\Cms\Support\Arr::get()` does this in a slightly different way:

```
// Normalize the key into dot notation
        if (is_string($key) && preg_match('/^[\w\-]+(?:\[[^\[\]]+\])+$/', $key)) {
            $key = rtrim((string) preg_replace('/[\[\]]+/', '.', $key), '.');
        }
```

We can add a distinct method for that, and use it in the `get()` method and here?

brianjhanson: Done and done. I added `Str::toDotNotation` and replaced both with it.

riasvdv (approved):

PR #18776, opened 2026-05-01
[6.x] Installer cleanup

brianjhanson:
Cleans up some styling and code mostly around the installer, but there are some other changes in here as well. 

I think the biggest changes is that the vue wrappers now handle errors. I'm a bit torn on that change. Part of me wants those wrappers to be as thin as possible, but when that's the case I end up craving yet another layer that will handle errors, etc. 

We'll stand here for now, but the future may push us in another direction.

This branch also has some cleanup related to the system messages config. My goal is to remove `actionClient` and `useActionClient` and only use wayfinder + inertia. At the moment those two are only really used in the queue service, but that will need to stick around until we're able to ditch twig views.

Fixes CMS-2001

Includes [#18773](https://github.com/craftcms/cms/pull/18773)


resources/js/pages/Install.vue
```diff
@@ -47,10 +46,8 @@
     possibleSteps,
   } = useInstall();
 
-  const state = ref<'idle' | 'loading' | 'error'>('idle');
-
   watchEffect(() => {
-    possibleSteps.value.db.hidden = !props.showDbScreen;
+    possibleSteps.value.db.hidden = props.showDbScreen;
```
riasvdv:
```suggestion
    possibleSteps.value.db.hidden = !props.showDbScreen;
```

PR #18778, opened 2026-05-01
[6.x] Fixes user settings for solo edition

brianjhanson:
While clicking around I noticed that clicking the `users` item when using the solo edition would result in a "Not found" error. 

It turns out the `settings/users` route was only defined for team versions and up, but Solo still uses it in v5.

I played with a handful of approaches but this one seemed to work best despite not being my favorite. 

Feel free to let me know if there's a better way to do this.

### DIFF
--- a/routes/cp.php
+++ b/routes/cp.php
@@ -260,7 +260,7 @@ Route::middleware(['auth:craft', 'can:accessCp'])->group(function () {
                     ->name('settings.site-groups.destroy');
             });
 
-        // User settings
+        // User settings index
         if (Edition::isAtLeast(Edition::Team)) {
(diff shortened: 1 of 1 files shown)


riasvdv (approved):

routes/cp.php
```diff
@@ -258,9 +260,15 @@
                     ->name('settings.site-groups.destroy');
             });
 
+        // User settings index
+        if (Edition::isAtLeast(Edition::Team)) {
+            Route::get('settings/users', [UserGroupsController::class, 'index']);
+        } else {
+            Route::get('settings/users', fn () => redirect(cp_url('settings/users/fields')));
```
riasvdv, line 267: We could also use `Route::redirect(` but either is fine

PR #18791, opened 2026-05-05
[6.x] Introduce static color palette using Leonardo, update button variants and appearances, add Docs

gcamacho079:
### Description
- Adds a static color theme (example: `--color-static-blue-300`) for colors we don’t want to change based on light/dark theme. This theme can be viewed on the Color Palette Storybook page.
- Updates warning color value in `generate-colors.js` to be consistent with other warning colors in the control panel
- Made updates to variant and appearance names for clarity and consistency.
    - `accent` was previously a semantic variant as well as the name for the “most accented” component. In this PR, I retained the semantic name but updated the appearance name to be “solid.”
- Buttons were pared down to 3 appearances (solid, outline, and plain), which is consistent with the larger appearance subset (leaving out `outline-fill` and `fill`).
- Button hover and active colors are generated by dropping the `lightness` level of the colors, which should help to maintain contrast ratios.
- Rather than having `default` amongst the semantic color names, the default option has been renamed to `neutral`. I personally found the use of `default` really confusing, since it wasn’t clear whether it referred to an appearance or a variant. Sticking to a semantic naming style helped clarify this.
  - I think the only remaining thing we could clarify is that info/accent are used similarly and have a similar color scheme. 🤔 
- Updates the primary/accent color to blue to prevent using a similar color to the danger variant

Static color palette viewed in Storybook

Screenshot of the callout variants and appearances, and how they are consistent with the button variant/appearance subset

### Related issues
Resolves ACC-36


gcamacho079:
@brianjhanson, we discussed a few more changes in the last design meeting (such as opt-out/styleless buttons), but I wanted to keep this PR focused on the static color palette and working through some of the naming consistency we discussed.

brianjhanson:
@gcamacho079 I made a few adjustments to this

- Switched the accent color back to red (at least for now)
- Cleaned up some additional `primary` references
- Fixed the contrast on solid buttons

brianjhanson (approved):

packages/craftcms-cp/src/components/button/button.styles.ts
```diff
@@ -197,16 +199,16 @@ export default css`
   }
 
   /* Filled */
-  :host([appearance~='filled']) {
+  :host([appearance~='solid']) {
```
brianjhanson, line 208: @gcamacho079 I don't remember, what there a reason buttons get a `solid` appearance instead of `fill`?

packages/craftcms-cp/src/components/button/button.ts
```diff
@@ -7,6 +7,22 @@ import '../icon/icon.js';
...
+  Solid: 'solid',
+  Outline: 'outline',
+  Plain: 'plain',
+} as const;
+
+export const ButtonVariant = {
+  Accent: 'accent',
+  Neutral: 'neutral',
+  Danger: 'danger',
+} as const;
```
brianjhanson, line 20: I moved these into the button component because they're so specific

packages/craftcms-cp/src/components/callout/callout.styles.ts
```diff
@@ -72,7 +72,7 @@ export default css`
     border-end-end-radius: var(--_radius);
   }
 
-  :host([appearance~='accent']) {
+  :host([appearance~='solid']) {
```
brianjhanson, line 75: Feels like this should be `fill` too. Or were you wanting us to use `solid` instead of `fill`?

packages/craftcms-cp/src/stories/tokens/variants.stories.ts
```diff
@@ -105,40 +104,6 @@ export const ButtonMatrix: Story = {
   `,
 };
 
-/**
```
brianjhanson, line 108: I removed all of this, as of #19033 the indicators _do_ have appearances (but only `fill`, `outline-fill` and `outline`). We're going to need at least `outline-fill` and `outline` for element statuses. I included `fill` for good measure, but if that feels like too much we can ditch `fill`.

packages/craftcms-cp/src/styles/shared/tokens.css
```diff
@@ -109,6 +110,33 @@
   /**
   Components
    */
```
brianjhanson:
```suggestion
/**
 Static colors
  */
```

packages/craftcms-cp/src/components/callout/callout.styles.ts
gcamacho079: @brianjhanson I do like the idea of sticking with `solid`. I personally find it confusing when other components use `fill` ([like indicators](https://github.com/craftcms/cms/pull/18791/changes/3b67a59ce4e6583e3658f59191dbdfca53c43493#diff-e29924d28041679fcb6ef8ff294426c8a5c5040d9886ae3bb7d88ae7040b79dcR37)) to refer to the appearance as well as the actual fill color, and I like the appearances being consistent with the ones already defined in the callout matrix.

packages/craftcms-cp/src/stories/tokens/variants.stories.ts
gcamacho079: @brianjhanson seems we're on the same page about that! I do think fill seems unnecessary.

PR #18795, opened 2026-05-05
[6.x] Handle non-inertia responses as page visits

riasvdv:
### Description

This prevents Inertia from opening successful requests in a modal if it's visited through `<Link>` but doesn't return an Inertia request.

We can now always use `<Link>` and not worry about if it's a new controller/view we're accessing or a standard response

I've adjusted the settings index page as a test & example.

@brianjhanson I did not commit the built assets as to not pollute the diff

### DIFF
--- a/resources/js/bootstrap/cp.ts
+++ b/resources/js/bootstrap/cp.ts
@@ -100,24 +100,49 @@ const Cp = {
       },
     });
 
-    // Handle normal responses as page visits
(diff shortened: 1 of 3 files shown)


brianjhanson (approved):
I just updated `Link` -> `CpLink` which is our own wrapper around Link that allows us to control the styling more like our other components.

PR #18800, opened 2026-05-06
[6.x] Refactor Authentication front-end

brianjhanson:
Refactors the LoginForm.js file into a web component. Step one in a bit of a larger plan. For now, this replaces most of the code within `_special/login.twig` with a new `craft-login-form` web component. That web component handles the whole of the auth flow. It will submit the username/password, display a two factor challenge and pass you on to the return configured return URL. 

Eventually, I'd like to make this web component available to front-end requests so users can just throw a `<craft-login-form/>` into their templates and get the whole login flow. 

### How it works

The general flow hasn't changed from the previous LoginForm.js file. The auth flow is made up of three parts. There's the main login form, the password reset form, and the 2fa challenge form (when applicable). In this implementation these are three separate components that communicate with each other via some internal events. 

After the initial login, the LoginController will return a JSON object that includes the HTML for the next stage's auth form along with some other data. The full data shape is represented by the `TwoFactorData` typescript type. 

When data is return with `authMethod` the challenge form is also shown. A successful challenge will eventually emit a `craft:login:success` event you can listen for in order to act accordingly. I'm still working out / thinking about how to handle events globally in the system, so this may very well change.

By default, this will redirect you to the `returnUrl` but the event is cancelable so you can handle it as you see fit. For example, within our elevated session manager we prevent the default and just close the modal. 

```js
const $loginForm = $container.find('craft-login-form');
$loginForm.on('craft:login:success', (event) => {
  event.preventDefault();
  this.success = true;
  this.loginModal.hide();
});
```

This PR also removes the idea of an auth handler in favor of listening for events directly. Components that extend the `AuthChallengeForm` component will emit a `login-verified` event you can listen to in order to act on this part of the process. For example, within the auth setup of TOTP we use this to close the modal after a successful verification

```js
const totpForm = document.querySelector('craft-totp-form');
totpForm.addEventListener('login-verified', (event) => {
  event.preventDefault();
  Craft.Slideout.instances['{{ containerId }}'].showSuccess();

  // @TODO This will eventually be unnecessary when more of the CP is inertia
  Craft.authMethodSetup.refresh();
})
```

We recommend extending this `CraftAuthChallengeForm` base component in order to create your own challenge forms that will follow the convention of ours. Both the TOTP and RecoveryCode forms extend this base component as examples.

### Note

All code related to the `TotpAsset` and the `RecoveryCodesAsset` has been moved into the yii2-adapter and removed from the main codebase.


src/Auth/AuthMethods.php
```diff
@@ -198,8 +198,9 @@ public function setUser(?User $user): void
 
         if ($user) {
             Session::put('user.id', $user->id);
+            Session::put('user.pending_2fa_at', now()->timestamp);
```
brianjhanson, line 202: The 2fa session was previously lasting as long as a regular session. This creates a lil' 5 minute session for completing the 2fa flow

packages/craftcms-cp/src/components/pane/pane.ts
brianjhanson, line 1: This will eventually take over for `Pane.vue`

packages/craftcms-legacy/cp/src/js/LoginForm.js
brianjhanson: Maybe this can just be deleted?

src/Route/RouteServiceProvider.php
```diff
@@ -98,6 +100,18 @@ public function boot(Router $router, Routes $routes): void
...
+            $loginName = $request->input('loginName');
+
+            // Don't throttle if no loginName provided - validation will handle it
+            if (empty($loginName)) {
+                return Limit::none();
+            }
+
+            // Throttle by loginName to prevent enumeration attacks
+            return Limit::perMinute(1);
+        });
```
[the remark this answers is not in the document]
brianjhanson: We don't want to rate limit by loginName as that wouldn't prevent someone from testing a bunch of them.

src/Http/Controllers/Users/AuthMethodController.php
```diff
@@ -30,6 +31,8 @@ public function __construct(
 
     public function setupHtml(Request $request, HtmlStack $HtmlStack): JsonResponse
     {
+        Session::put('user.pending_2fa_at', now()->timestamp);
```
brianjhanson, line 34: With a short 2fa session, we need to start one to give you some time between when we show the setup HTML and when you confirm your 2fa method. Without starting a session here (and with the short 2fa session time) you're not able to set up a new auth method.

I think this is safe because this controller is behind the `auth:craft` middleware, but if this seems like a bad idea I'm open to suggestions.

We can also ditch the idea of opening a short session for finishing the 2fa process but I do feel like it's probably a good idea not to allow you to finish the 2fa process anytime during your (potentially long) session.

resources/templates/_special/login.twig
```diff
@@ -1,151 +1,17 @@
...
-  </div>
-</div>
+<craft-login-form {{ attr({
+  'show-reset-password': showResetPassword ?? false,
+  'show-remember-me': showRememberMeCheckbox ?? false,
+  'static-email': staticEmail,
+  'username': username,
+  'initial-error': initialError ?? null,
+  'use-email-as-username': generalConfig.useEmailAsUsername
+}) }}>
```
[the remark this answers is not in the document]
brianjhanson: Fixed in https://github.com/craftcms/cms/pull/18800/changes/e2b0b8c2c4e58377f1e69cc650e3788d200f6a9d#diff-a9c3db516028fcb3f06d59a49cad9eac5db5d77e8a9c44a117132b022f229d84R7

resources/js/components/Auth/login/login-form.ts
```diff
@@ -0,0 +1,428 @@
...
+                  class="login-username"
+                  name="username"
+                  .value="${this.staticEmail}"
+                />`
+              : html`
+                  <div class="field">
+                    <craft-input
+                      label="${this.#usernameLabel()}"
+                      id="login-username"
+                      type="${this.useEmailAsUsername ? 'email' : 'text'}"
```
riasvdv, line 269: Image

I don't have `useEmailAsUsername` enabled in the config, but this still shows as an email input

resources/js/components/Auth/login/login-form.ts
```diff
@@ -0,0 +1,428 @@
...
+        {attribute: t('Password'), min: this.minPasswordLength},
+        'yii'
+      );
+    }
+
+    if (passwordLength > this.maxPasswordLength) {
+      return t(
+        '{attribute} should contain at most {max, number} {max, plural, one{character} other{characters}}.',
+        {attribute: t('Password'), max: this.maxPasswordLength},
+        'yii'
```
riasvdv: The `yii` translation category has been removed and any messages there should have been added to the `app` category

resources/js/components/Auth/login/login-form.ts
```diff
@@ -0,0 +1,428 @@
...
+                class="login-password"
+                name="password"
+                autocomplete="current-password"
+                required
+                @user-input-changed="${this.#onInput}"
+              ></craft-input-password>
+
+              ${this.showResetPassword
+                ? html`
+                    <craft-button
```
riasvdv, line 292: Image

The forgot password button has a larger interactive area than is visible, which cause me to click it when trying to enable "Stay signed in"

resources/js/components/Auth/login/login-reset-password.ts
```diff
@@ -0,0 +1,168 @@
...
+
+  override firstUpdated() {
+    this._input?.focus();
+  }
+
+  #label() {
+    return this.useEmailAsUsername ? t('Email') : t('Username or Email');
+  }
+
+  #validate(): string | true {
```
riasvdv: Should we leave validation up to the controllers with Inertia?

src/Http/Controllers/Auth/LoginController.php
```diff
@@ -35,7 +37,10 @@ public function showLogin(Request $request): Response|View
             return redirect()->action([TwoFactorAuthenticationController::class, 'showForm']);
         }
 
-        return $this->renderViewWithFallback('login');
+        return Inertia::render('LoginPage', [
```
riasvdv: Currently it's possible to override the login view (`renderViewWithFallback()`), we should probably preserve that functionality somehow

src/Http/Middleware/EnsureTwoFactorChallengeIsRecent.php
```diff
@@ -0,0 +1,48 @@
...
+        $pendingAt = Session::get('user.pending_2fa_at');
+
+        if ($pendingAt && now()->timestamp - $pendingAt <= self::TTL) {
+            return $next($request);
+        }
+
+        Session::forget(['user.id', 'user.pending_2fa_at']);
+
+        if ($request->wantsJson()) {
+            return new JsonResponse(['message' => t('Your verification session has expired. Please sign in again.')], 401);
```
riasvdv: Generally status code `419` is used for expired pages

src/Http/Middleware/EnsureTwoFactorChallengeIsRecent.php
```diff
@@ -0,0 +1,48 @@
...
+        }
+
+        if ($request->isSiteRequest() && $loginPath = $this->generalConfig->getLoginPath()) {
+            return redirect($loginPath);
+        }
+
+        return redirect(CpAuthPath::Login->value);
```
riasvdv: This might need a `Url::cpUrl(` wrapped around it

src/Route/RouteServiceProvider.php
```diff
@@ -98,6 +100,18 @@ public function boot(Router $router, Routes $routes): void
...
+
+        RateLimiter::for('password-reset', function ($request) {
+            $loginName = $request->input('loginName');
+
+            // Don't throttle if no loginName provided - validation will handle it
+            if (empty($loginName)) {
+                return Limit::none();
+            }
+
+            // Throttle by loginName to prevent enumeration attacks
```
riasvdv:
```suggestion
            // Throttle to prevent enumeration attacks
```

src/Route/RouteServiceProvider.php
```diff
@@ -98,6 +100,18 @@ public function boot(Router $router, Routes $routes): void
...
+            return Limit::perMinute(1);
```
riasvdv: A rate limiter without a key is global, it's probably best to add some kind of key

```suggestion
            return Limit::perMinute(1)->by($request->ip());
```

riasvdv (changes requested):

src/Route/RouteServiceProvider.php
brianjhanson: Resolved in [3750e5d](https://github.com/craftcms/cms/pull/18800/commits/3750e5d72dd40622c397d9e2bc9392df63e1c07b)

src/Http/Middleware/EnsureTwoFactorChallengeIsRecent.php
brianjhanson: Resolved in [f0aa91f](https://github.com/craftcms/cms/pull/18800/commits/f0aa91f6d7109859f3335981ed65f7b98d5a4e59)

src/Http/Middleware/EnsureTwoFactorChallengeIsRecent.php
brianjhanson: Resolved in [4275561](https://github.com/craftcms/cms/pull/18800/commits/4275561287a098f24e5814c98d0bfd675249f9ad)

resources/js/components/Auth/login/login-form.ts
brianjhanson: Resolved in [62c7b3c](https://github.com/craftcms/cms/pull/18800/commits/62c7b3cab15a6481b2e81ee6d9fdc748a04fb8a6)

resources/js/components/Auth/login/login-reset-password.ts
brianjhanson: I've moved it all to be server side buts stopped short of using the [precognition stuff](https://inertiajs.com/docs/v3/the-basics/http-requests#precognition). Eventually I'd like to be able to distribute these components so people can create their own login templates and just include our component in it in addition to the custom login page thing. In that case I don't really want to assume inertia is around.

src/Http/Controllers/Auth/LoginController.php
brianjhanson: Added in [799c2f6](https://github.com/craftcms/cms/pull/18800/commits/799c2f61bff9721100fc15547e0b858002abbc61)

riasvdv (approved):

src/Http/Controllers/Auth/LoginController.php
```diff
@@ -87,8 +88,12 @@ public function showLoginModal(Request $request, Impersonation $impersonation, H
...
     {
         $request->validate([
-            'loginName' => ['required', 'string'],
-            'password' => ['required', 'string'],
+            'loginName' => [
+                'required',
+                'string',
+                Rule::when($this->generalConfig->useEmailAsUsername, 'email'),
+            ],
+            'password' => ['required', 'string', 'min:6', 'max:160'],
```
riasvdv: Ah missed this one, we should use `\Illuminate\Validation\Rules\Password::required()` here which makes sure the rules are the same everywhere (and configurable)

https://laravel.com/docs/13.x/validation#validating-passwords

PR #18814, opened 2026-05-07
Replace custom fields within existing field layout elements

brandonkelly:
### Description

Adds the ability to replace the custom field for an existing field layout element.

The selected custom field chip within a field layout element’s settings, with an expanded action menu that includes “Settings”, “Replace”, and “Remove” action items.<br>

When a new field is selected, the Label, Handle, and Instructions field values will be automatically populated with the previous field’s Name, Handle, and Default Instructions, if the values were previously blank, to maintain authoring/development consistency. (The developer is free to remove/edit those values if that’s not desired, of course.)

### Related issues

- #14566
- #17893
- #18024


alexanderloewe:
❤️

PR #18861, opened 2026-05-13
[6.x] Add configurable compiled templates path

riasvdv:
### Description

Adds a `compiledTemplatesPath` general config setting so Craft’s compiled Twig templates directory can be configured while preserving the default `storage/runtime/compiled_templates` path. The path supports aliases and environment-variable parsing via Craft’s existing path service.

### Related issues

- PT-3115

### DIFF
--- a/src/Config/GeneralConfig.php
+++ b/src/Config/GeneralConfig.php
@@ -162,6 +162,8 @@ class GeneralConfig extends BaseConfig
      * :::
      *
      * @group Environment
+     *
+     * @since 6.0.0
      */
     public ?string $compiledTemplatesPath = null;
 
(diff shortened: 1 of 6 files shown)


src/Config/GeneralConfig.php
```diff
@@ -147,6 +147,24 @@ class GeneralConfig extends BaseConfig
...
+     * ->compiledTemplatesPath('@storage/runtime/templates')
+     * ```
+     * ```shell Environment Override
+     * CRAFT_COMPILED_TEMPLATES_PATH=@storage/runtime/templates
+     * ```
+     * :::
+     *
+     * @group Environment
+     */
+    public ?string $compiledTemplatesPath = null;
```
timkelty, line 168: Why `?string` with `null` default, and not just the static default value (`storage/runtime/compiled_templates`), like `ideHelperPath`?

Issue #18810, opened 2026-05-07
[6.x]: Vite::useHotFile() called globally in ViewServiceProvider breaks host application's Vite integration


ivowk:
### Description
ViewServiceProvider::register() calls Vite::useHotFile() globally, redirecting Laravel's shared Vite singleton to the package's own hot file path (vendor/craftcms/cms/resources/hot). This overwrites the host application's hot file path (public/hot), breaking frontend Vite integration - the @vite directive in Blade templates no longer detects the dev server, so the HMR client is never injected and Blade refresh stops working.

### Steps to reproduce

1. Configure Vite with refresh: true in vite.config.js as per the Laravel and DDEV docs
2. Start the Vite dev server
3. Edit a Blade template and observe the browser

### Expected behavior
The host application's @vite directive detects the dev server via public/hot and injects the HMR client, enabling full-page refresh on Blade template changes as documented by the laravel-vite-plugin.

### Actual behavior
@vite looks for the hot file at vendor/craftcms/cms/resources/hot instead of public/hot, does not detect the dev server, and falls back to the asset manifest. No HMR client is injected, Blade refresh does not work, and frontend assets may fail to load entirely if no manifest exists.

The only workaround is to manually re-override the hot file in the host app's AppServiceProvider, scoped to non-CP requests:

```
public function boot(): void
{
    if (!app()->runningInConsole() && !request()->isCpRequest()) {
        Vite::useHotFile(public_path('hot'));
    }
}
```

### Craft CMS version

6.0.0-alpha.1

### PHP version

8.5

### Operating system and version

ddev on macOS

### Installed plugins and versions

-

brandonkelly:
This was fixed in Craft 6.0.0-alpha.2. Thanks for reporting!

PR #18871, opened 2026-05-14
[6.x] Moves vite config into AppServiceProvider

brianjhanson:
Previously our hotFile and build directory config weren't getting picked up by the @vite directive in app.blade.php. Most of the time things worked fine, but if you built and then started the dev server it wouldn't find the hot file and would continue serving built assets

@riasvdv maybe I missed something about why these had to be cloned and configured individually. If that's the case we can just add the lines in `AppServiceProvider` and leave the rest alone. It seems like things are working with this, but admittedly I didn't do extensive testing.

### DIFF
--- a/src/Providers/AppServiceProvider.php
+++ b/src/Providers/AppServiceProvider.php
@@ -23,6 +23,7 @@ use Illuminate\Contracts\Config\Repository;
 use Illuminate\Foundation\Application;
 use Illuminate\Foundation\Console\AboutCommand;
 use Illuminate\Foundation\Events\LocaleUpdated;
+use Illuminate\Foundation\Vite;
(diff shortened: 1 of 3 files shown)


riasvdv:
Cloning them is a fix for this issue: https://github.com/craftcms/cms/issues/18810

brianjhanson:
> Cloning them is a fix for this issue: #18810

The clones have been restored

brianjhanson:
Moved this into a draft, I need to test it against the issue in #18810 to make sure that still works. The current state is a bit broken for me, but I'm not confident this won't just reintroduce #18810

brianjhanson:
Closing in favor of #18880

PR #18922, opened 2026-05-20
[6.x] Add back the auth page powered by logo

riasvdv:
### Description

image


resources/js/layout/AuthBase.vue
```diff
@@ -1,5 +1,6 @@
 <script setup lang="ts">
   import useCraftData from '@/composables/useCraftData';
+  import craftCmsLogoUrl from '../../legacy/cp/dist/images/craftcms.svg';
```
brianjhanson: We should probably move this out of legacy and put it with the images for the installer.

brianjhanson (changes requested):

brianjhanson (approved):

PR #18927, opened 2026-05-20
[6.x] Inertia route settings

riasvdv:
### Description

Ports the route settings page to Inertia

I've decided to not port it in the same way with the edit happening in a modal, this isn't really a pattern used anywhere else in the control panel.

This ports the edit screen to a normal Inertia page with a form, which we can render in a slideout once that functionality is implemented.


resources/js/common/form/MixedInput.vue
riasvdv, line 1: This is mostly a strict port of the legacy MixedInput component, it works but I'm not 100% confident on how idiomatic Vue this is

resources/js/common/components/ActionMenu.vue
riasvdv, line 1: I extended this to make it possible to pass in `actionMenuItems` (in this case a "Delete" action) which show up next to the save button

resources/js/common/composables/useReorderableItems.ts
riasvdv, line 1: This was crashing the browser when used on the Routes index, I don't exactly know why it was crashing but Codex made these changes to memoize or delay some things and now it's not crashing.

resources/js/common/layouts/AppLayout.vue
```diff
@@ -206,40 +207,51 @@
...
-                              >
-                                <craft-icon
-                                  name="chevron-down"
-                                  :label="label"
-                                ></craft-icon>
-                              </craft-button>
-                            </template>
-                          </ActionMenu>
-                        </craft-button-group>
+                            v-if="formAdditionalActions?.length"
```
riasvdv, line 245: Added an extra ActionMenu here

resources/js/pages/settings/routes/Index.vue
```diff
@@ -0,0 +1,299 @@
...
+  }
+
+  function handleReorder(startIndex: number, finishIndex: number) {
+    const routes = reorderedRoutes(props.routes, startIndex, finishIndex);
+
+    router
+      .optimistic<{routes: Array<RouteIndexData>}>(() => {
+        return {
+          routes,
+        };
```
riasvdv, line 58: First time I used Inertia optimistic, but a pretty good use-case for it 😄

resources/js/common/form/MixedInput.vue
brianjhanson: I'm fairly sure at the moment this is the only place using this mixed input, so it's fine if it's a little wonky.

PR #18932, opened 2026-05-20
[6.x] Inertia filesystems

brianjhanson:
Brings the filesystem config over to inertia. To handle the whole "dynamic form data" situation, we have two solutions in here. Modern and legacy. 

For legacy settings, we do what we're currently doing with the fields page and serialize all the form data found within a certain container via the `serializeFormInputsAsObject` function. 

For modern settings, the `Edit.vue` template provides a `fsTypeSettings` object that you can inject into a custom Vue component registerred for your settings. 

I think we can make the modern process more elegant, but this works for the moment


resources/js/pages/SettingsFilesystemsEditPage.vue
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    return props.fsInstances[fsType] ?? null;
+  }
+</script>
+
+<template>
+  <AppLayout>
+    <Pane appearance="raised">
+      <div class="grid gap-3">
```
riasvdv: It's probably good to have a `<FormFields>` wrapper that defines the spacing, just so it's consistent everywhere and if it has to change it's only in 1 file, or have `<Pane>` be a grid by default

resources/js/pages/SettingsFilesystemsEditPage.vue
```diff
@@ -0,0 +1,130 @@
+<script setup lang="ts">
```
riasvdv: It might make sense to start putting pages in folders, for example `pages/settings/filesystems/EditFilesystemPage.vue` 

[Seb wrote a blog on how Spatie structures Inertia apps](https://spatie.be/blog/how-to-structure-the-frontend-of-a-laravel-inertia-react-application) and there might be some ideas in there we could use as well.

For example if there are components needed just for filesystems we could put them in `pages/settings/filesystems/components`

resources/js/components/Filesystems/FilesystemSettings.vue
```diff
@@ -0,0 +1,73 @@
...
+      <template v-if="hasUrls && filesystem.showUrlSetting">
+        <CraftCombobox
+          :label="t('Base URL')"
+          :help-text="t('The base URL to the files in this filesystem.')"
+          v-model="url"
+          :options="page.props.baseUrlSuggestions"
+          name="url"
+          :required="true"
+          placeholder="//example.com/path/to/folder"
+          data-error-key="url"
```
riasvdv: Do we always have to provide this or could we fall back to what's provided in `name` for example and only need to provide this when it's different than the input name?

resources/js/pages/SettingsFilesystemsEditPage.vue
brianjhanson: Yeah, I'm planning to add a `<craft-field-group/>` component that will handle the spacing as well as the layout system for fields.

resources/js/components/Filesystems/FilesystemSettings.vue
brianjhanson: Yeah, we'll want something like that. I just haven't quite worked out exactly where we'll want the logic. `data-error-key` will likely be obsolete relatively soon (as far as I know it's only to link the error summary to the field itself). 

We could create a `useCraftField` that would return all of these attributes. Ideally I'd like a place to centralize this logic so we don't have to worry (as much) about keeping all our fields up to date with things like this, and so we can make sure they all follow the same logic for setting those. 

My other thought is to eventually have some kind of `<CraftForm/>` component that would look over the fields within it and makes sure they have everything we want. 

Since I haven't quite figured that out, I've just been adding things manually for the moment.

resources/js/pages/SettingsFilesystemsEditPage.vue
brianjhanson: Opened https://github.com/craftcms/cms/pull/18964 to restructure things based on their opinions. I'll hopefully pull that in soon so we don't have too many PRs to rebase

PR #18943, opened 2026-05-21
[6.x] Inertia asset transforms settings

riasvdv:
### Description

The index was already ported, this ports the create/edit screen

Created a new `ColorInput` which has the same functionality as the legacy one.

image


resources/js/components/form/ColorInput.vue
brianjhanson, line 1: Ideally this would live in the `@craftcms/cp` package as a web component instead of in the Vue side of things so we can bring support for the new version to pages rendered via twig a bit easier.

brianjhanson (approved):

PR #18944, opened 2026-05-21
[6.x] Validation improvements & fixes

riasvdv:
### Description

Fixes two validation lifecycle gaps:

- `Validatable::prepareForValidation()` is now called consistently from Craft ruleset validation, including element rulesets and custom rulesets that override their own preparation hook.
- Plain `Validatable` classes using `Validates` without a configured ruleset now fall back to `ComponentRules` instead of failing when `ruleset` resolves to `false`, so their `getRules()` contract works as expected.

- Also fixes an issue with the legacy yii model serializing rulesets.

Also improves the following
- `setAttributes()` is now moved from `Component` to the `Validates` trait, which makes plain class validation easier
- `ComponentRules` has been renamed to `ValidatableRules` to more clearly describe its intent.

### Related issues

- #18946

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Fixed an error that occurred when Updates were cached and deserialized.
 - Fixed an error that prevented link fields from saving.
 - Fixed a bug where Money fields could throw an error during element validation when the field value was falsy.
+- Fixed a bug where `CraftCms\Cms\Validation\Contracts\Validatable::prepareForValidation()` wasn’t called consistently, and plain `Validatable` classes without a configured ruleset couldn’t be validated. ([#18944](https://github.com/craftcms/cms/pull/18944))
 
 ## 6.0.0-alpha.4 - 2026-05-19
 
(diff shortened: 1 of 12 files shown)


janhenckens:
@riasvdv this merged might be causing the error I'm seeing on composer update right now:

````
Class craft\config\BaseConfig contains 1 abstract method and must therefore be declared abstract or implement the remaining method (CraftCms\Cms\Validation\Contracts\Validatable::passedValidation)
````

````
PHP Fatal Error 'yii\base\ErrorException' with message 'Class craft\config\BaseConfig contains 1 abstract method and must therefore be declared abstract or implement the remaining method (CraftCms\Cms\Validation\Contracts\Validatable::passedValidation)'
````

riasvdv:
@janhenckens Yeah I caught this and fixed it in https://github.com/craftcms/cms/commit/9e962b196c9cb599e4ca481d3157c88f6071ac1e

janhenckens:
> @janhenckens Yeah I caught this and fixed it in [9e962b1](https://github.com/craftcms/cms/commit/9e962b196c9cb599e4ca481d3157c88f6071ac1e)

Weird, composer isn't picking up that change on my end - with this "version" `"craftcms/cms": "6.x-dev as 6.0.0-alpha.4"` (wiped vendor, reinstalled, same error)

PR #18960, opened 2026-05-26
[6.x] Markdown field

riasvdv:
### Description

A native markdown field built on [Overtype](https://overtype.dev/) which renders through Craft's Markdown service with custom flavors support.

Built as a web component.

image

With a popover containing a link field:

image


brianjhanson (approved):
There are probably a few things in here that I'll want to tweak, but I think that will be easier and cleaner to do after it's merged.

gcamacho079 (commented):
I encountered a blocking issue for keyboard users while reviewing this PR. I [opened an issue in the Overtype repository](https://github.com/panphora/overtype/issues/113) and am awaiting a response before continuing with the complete field audit.

brianjhanson:
Do you think this is a simple enough issue that we could submit a PR to fix it?

riasvdv:
I've managed to override the behavior on our side, so we don't necessarily have to wait on Overtype to change the behavior: [https://github.com/craftcms/cms/pull/18960/changes/5b6d1d699156a04a902cb1f0c5982ce0b4afd7b5](<https://github.com/craftcms/cms/pull/18960/changes/5b6d1d699156a04a902cb1f0c5982ce0b4afd7b5>)

* Tab and shift-tab always escape the editor
* Added Ctrl/Cmd+\] and Ctrl/Cmd+\[ shortcuts for indent/outdent (most common ones across other editors)

gcamacho079 (approved):

PR #18969, opened 2026-05-26
[6.x] Add TypeScript transformer generation

riasvdv:
### Description

Adds a dev-only Spatie TypeScript transformer setup for the Workbench/Testbench app, configured to generate PHP-mirrored TypeScript declarations under `resources/js/generated`. 

The transformer starts from an explicit seed list and recursively includes referenced Craft CMS classes

Vite runs the Testbench transformer command before JS builds and on relevant PHP hot updates.


workbench/app/Providers/TypeScriptTransformerServiceProvider.php
```diff
@@ -0,0 +1,34 @@
...
+{
+    protected function configure(TypeScriptTransformerConfigFactory $config): void
+    {
+        $config
+            ->outputDirectory(dirname(__DIR__, 3).'/resources/js/generated')
+            ->writer(new GlobalNamespaceWriter('types.d.ts'))
+            ->replaceType(DateTimeInterface::class, 'string')
+            ->provider(new ClassListTransformedProvider(
+                [
+                    Updates::class,
```
riasvdv, line 26: Here we can add additional classes we want to transform, I've kept it as a specific opt-in for now, if it becomes too unwieldy we can improve on this later.

Any classes referenced by a class that is transformed will also be transformed alongside it.

PR #18973, opened 2026-05-27
[6.x] Rename PluginsLoaded and add PluginsBooted event

riasvdv:
### Description

Renames `PluginsLoaded` to `PluginsRegistered`, and wires the Yii `EVENT_AFTER_LOAD_PLUGINS` compatibility bridge to the booted lifecycle.

### Related discussion

#18962


brandonkelly:
@riasvdv Couldn’t we just do this instead?

Replace

```php
Event::listen(
    PluginsLoaded::class,
    fn() => $pluginService->trigger(self::EVENT_AFTER_LOAD_PLUGINS),
);
```

with

```php
app()->booted(fn() => $pluginService->trigger(self::EVENT_AFTER_LOAD_PLUGINS));
```

brandonkelly:
(Seems unnecessary to introduce a new event that is basically just an alternative for `booted()`.)

riasvdv:
We could just do that, it does add an event people less familiar with Laravel can more easily discover

AugustMiller:
I think there is a benefit to providing a 1:1 event mapping from 5.x &rarr; 6.x. @Anubarak’s [original concern](https://github.com/craftcms/cms/discussions/18962) was basically that `app()->booted(...)` isn't a strong ergonomic signal to plugin developers, even when properly documented. 😅

This also leaves open an opportunity to recloate it closer to the _real_ “last plugin booted” moment, at a later date, which might actually make it distinct…

I'm open to immediately deprecating `PluginsBooted` with a note saying that it's equivalent to `app()->booted(...)`.

AugustMiller:
As far as the chained triggers, I suspect we need to be able to test the adapter’s response to that event, wherever it is triggered from?

brandonkelly:
> We could just do that, it does add an event people less familiar with Laravel can more easily discover

Meh, I think it’s better that they learn the proper way to handle this in Laravel.

Anubarak:
> > We could just do that, it does add an event people less familiar with Laravel can more easily discover
> 
> Meh, I think it’s better that they learn the proper way to handle this in Laravel.

Kinda agree, but in this case: it might still be better to rename it at least. We basically have an event that provides entirely different purposes with the same name.

brandonkelly:
@Anubarak True, `PluginsRegistered` is more semantic.

PR #18994, opened 2026-05-28
[6.x] Namespace plugin-defined action routes by plugin handle

riasvdv:
### Description
Namespace legacy plugin-defined `actions.php` routes by plugin handle so plugins can define route tails without colliding under the shared action trigger.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Fixed a bug where legacy plugin-defined `actions.php` routes could collide between plugins. ([#18994](https://github.com/craftcms/cms/pull/18994))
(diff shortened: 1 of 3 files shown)


AugustMiller (approved):
A bit of context, for observers…

This is considered a legacy feature, intended to mimic Yii’s "action" pattern of `module-id/controller-name/action-id`. Automatically-registered action paths in Craft 3.0 through 5.x were globally unique, based on the “controller namespace” they existed in, and plugins/modules were not able to register out-of-band _action_ routes.

Nothing prevented (or will prevent) a developer or application from manually defining a conflicting route like `admin/actions/...` or `actions/...`, but this PR at least makes the "no-conflict" mode a default.

This is functionally equivalent to having used a top-level `Route::prefix('my-plugin', ...)` in your `routes/actions.php` file.

Issue #18997, opened 2026-05-29
[6.x]: forms.dateTimeField() throws $(...).datepicker is not a function on utility page for a plugin


vnali:
### Description

There is an issue with forms.dateTimeField() on utility page for a plugin.

When rendering:
```
{{ forms.dateTimeField({
    ...
}) }}
```
the page throws the following console error: `Uncaught TypeError: $(...).datepicker is not a function`.

I already updated to the [this commit](https://github.com/craftcms/cms/commit/1ec593b91dc1e0711b45a0995cc65b398ece3708) .

### Craft CMS version

6.0.0-alpha.5

### Installed plugins and versions

-

brianjhanson:
Thanks for reporting! I spent a chunk of time on this this morning hoping that I'd be able to open a PR for a quick patch but alas, a good solution has escaped me for the moment. 

I haven't quite figured out the best way to provide compatibility for all this legacy code in the new inertia pages of the CP.

PR #19003, opened 2026-05-29
[6.x] Refactors the legacy JS a bit

brianjhanson:
A constant work in progress. 

#18997 surfaced a problem with our previous approach. On inertia pages we don't currently load the `CpAsset`, mostly because that comes with a lot of stuff, a good chunk of which we don't need or want in the inertia context. 

This updates our approach to include a shim for inertia pages that will bridge the gap between what legacy pages expect and what inertia pages provide. 

There's a fair amount of gymnastics in here that the robots helped me out with. I imagine over time we'll see if those things are necessary or not. 

The styles are still rough for legacy fields, but they should at least be functional with these changes.

Fixes #18997


brianjhanson:
I mucked up this PR somehow, going to close and create a clean one.

PR #19008, opened 2026-05-31
[Security] Add allowed_classes => false to LogTarget unserialize() to prevent PHP Object Injection

XananasX7:
## Summary

The Craft CMS debug LogTarget serializes and deserializes panel data to/from the filesystem. All four unserialize() calls in LogTarget.php lack an allowed_classes restriction.

All panel save() methods return plain PHP arrays of scalars — no PHP objects are ever legitimately stored. Adding allowed_classes => false to all four call sites prevents PHP Object Injection if an attacker can write to the debug data path.

Note: ProjectConfig already uses allowed_classes => false correctly (line 1857). This PR applies the same pattern to LogTarget.

### DIFF
--- a/src/debug/LogTarget.php
+++ b/src/debug/LogTarget.php
@@ -73,7 +73,7 @@ class LogTarget extends \yii\debug\LogTarget
         $content = $this->module->fs->read($indexFile);
 
         if ($content !== '') {
-            return array_reverse(unserialize($content), true);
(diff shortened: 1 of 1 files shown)


brandonkelly:
Thanks for the PR!

brandonkelly:
FFR, please submit security-related reports/fixes privately, per our [security policy](https://github.com/craftcms/cms/security/policy).

PR #19009, opened 2026-05-31
[6.x] Move Laravel auth off of User element into CraftUser interface

riasvdv:
### Description

Hooking up the User element into Laravel's auth service wasn't the best idea. It causes issues when trying to use Craft in combination with other ecosystem packages.

Using the default eloquent provider, and tying authentication methods and services around a `CraftUser` solves this problem and allows developers to use their own User models and authentication systems, as long as they implement the interface.

[Filament](https://filamentphp.com/docs/5.x/users/overview#authorizing-access-to-the-panel) also does this through one or more interfaces depending on the functionality you need. 

Our integration surface is still mostly through the User Element which makes the `asElement()` method fairly load-bearing, but at least provides _some_ extension point and separates Authentication from Elements.

---

**Main Changes**
- Removed `CraftCms\Cms\Auth\UserProvider`; the `craft` guard now defaults to Laravel’s `eloquent` provider using `CraftCms\Cms\User\Models\User`.
- Added `CraftCms\Cms\User\Contracts\CraftUser` and `CraftUserTrait`.
- Both the user element and Eloquent user model now implement `CraftUser`.
- Added `Auth::craftUser()` / `auth('craft')->craftUser()` and `request()->craftUser()` as the Craft-safe way to access the authenticated user.
- Internal policies, auth methods, passkeys, user preferences, notifications, and CP code now type against `CraftUser` where they only need auth/user identity behavior.
- Twig/global `currentUser` is still normalized back to the user element via `currentUserElement()`, so most templates should not change.

**Integration Impact**
Most plugins/integrations probably do not need changes if they use `currentUser`, Craft user elements, or normal `Gate`/`can()` checks.

Code should change if it uses Laravel auth directly and assumes `Auth::user()` or `$request->user()` is a `CraftCms\Cms\User\Elements\User`. Use:

```php
$user = Auth::craftUser();
// or
$user = $request->craftUser();
```

Then use contract methods:

```php
$userId = $user->getCraftUserId();
$isAdmin = $user->isAdmin();
$element = $user->asElement();
```


brandonkelly (approved):

brandonkelly:
Not sure what to do about this line:

https://github.com/craftcms/cms/blob/0cd7f7a1ab3e51a4b73fac7a5339ecde0f23520c/tests/Feature/Element/DeletionBlockersTest.php#L56

PHPStorm and Rector both think that `find()` resolves to `CraftCms\Cms\User\Elements\User::find()`, so I [modified it](https://github.com/craftcms/cms/pull/19009/changes/9f6ea7c78fd80642cbd9060dcec399cff558a3f3) based on that to work around a [rector issue](https://github.com/craftcms/cms/actions/runs/26829463484/job/79110856226?pr=19009). But judging from [this error](https://github.com/craftcms/cms/actions/runs/26839996126/job/79144818094?pr=19009), it’s actually resolving to `Illuminate\Database\Eloquent\Builder::find()`.

riasvdv:
I think the rector cache in GitHub actions is messing things up, might have to delete the cache entries

brandonkelly:
@riasvdv I got the same Rector issue locally, so I don’t think it’s a cache problem.

PR #19014, opened 2026-06-01
[6.x] Add field references + deletion blockers

riasvdv:
### Description

This brings the [field references & deletion blocker](https://github.com/craftcms/ckeditor/pull/576/changes) that was added to CKEditor into the core.

This introduces a new `fieldreferences` table, a `TracksReferencesFieldInterface` marker contract, field lifecycle listeners that keep tracked references in sync, and a generic deletion blocker/replacement flow for referenced elements. 

The Markdown field now opts-in to this system, so element reference tags in Markdown content can block deletion and be replaced through the delete-elements flow.

Plugin developers can add reference tracking to custom fields by:

1. Implementing the `CraftCms\Cms\Field\Contracts\TracksReferencesFieldInterface` marker interface.
2. Ensuring `serializeValue()` returns the raw serialized field content that should be scanned for element reference tags. (This should already be the case for almost all custom fields)
3. Ensuring the serialized value can be passed back through `setFieldValue()` / normalization after numeric reference IDs are replaced. (Should also already be the case for all custom fields)

When tracked references are replaced, Craft updates the serialized field value and saves the source element without validation; the normal field save lifecycle then refreshes `fieldreferences`.

Existing content is not backfilled automatically. Sites should resave elements containing tracking-enabled fields to populate reference rows.

### TODO

- [x] Add method & trait so fields can provide references they contain and control how replacement is handled


src/Field/Concerns/TracksReferences.php
```diff
@@ -0,0 +1,42 @@
...
+        }
+
+        $element->setFieldValue($this->handle, $newValue);
+
+        return true;
+    }
+
+    private function elementRefs(): ElementRefs
+    {
+        return app(ElementRefs::class);
```
riasvdv, line 40: @brandonkelly Do plugins use element refs often? If so we could probably make a Facade for this as well

brandonkelly: Yeah probably worth it.

PR #19018, opened 2026-06-01
[6.x] Element index inertia pages

brianjhanson:


brianjhanson:
Renamed the branch and opened https://github.com/craftcms/cms/pull/19095

PR #19019, opened 2026-06-02
Bugfix/options fields empty

i-just:
### Description
- fixes an issue that could occur when querying multi-option fields that were empty but used an empty array notation
- fixes an issue that could occur when querying multi-option fields with `['not', ':empty:']` notation


brandonkelly:
Fixed this via f97507c9cf5ed539c27e2cd474facb2b800190fa by having multi-option fields stop storing an empty array if no options are selected.

`resave --with-fields …` commands can be used to update existing elements.

Issue #19004, opened 2026-05-29
[5.x]: Twig 3.27 sandbox ensureToStringAllowed() materializes ElementQuery, causing unbounded recursion / memory exhaustion


ryanleichty:
### What happened?

After updating to Craft 5.10.4.1 (which pins `twig/twig` to `~3.27.0`), specific entries began throwing `Allowed memory size exhausted` on both the front-end **and** in the control panel (entry edit / preview). It does not affect all entries — only ones whose relation graph reachable from a sandbox-rendered element is large and/or circular.

The fatal error surfaces at misleading locations (e.g. `vendor/yiisoft/yii2/web/ErrorHandler.php:310`, `craft/elements/db/ElementQuery.php:2760`) because those are just wherever the final allocation lands; the real cause is unbounded recursion.

**Root cause**

Twig 3.27.0 shipped the fix for CVE-2026-48807, which makes `Twig\Extension\SandboxExtension::ensureToStringAllowed()` recurse into **any `Traversable`** while sandboxing is active — it materializes it via `iterator_to_array()` and recursively policy-checks every element:

```php
// vendor/twig/twig/src/Extension/SandboxExtension.php (3.27.0)
if ($obj instanceof \Traversable) {
    if (isset($seen[$obj])) {
        return $obj;
    }
    $seen[$obj] = true;
    $array = iterator_to_array($obj);
    $this->ensureToStringAllowedForArray($array, $lineno, $source, $seen);
    ...
}
```

Craft's `craft\base\Element` and `craft\elements\db\ElementQuery` are both `Traversable`. So when **any sandboxed template** outputs/coerces an element (object templates, title formats, `{% sandbox %}`, plugins that render in the sandbox), Twig materializes the element, iterates its field values — each relation field is an `ElementQuery` — runs the query, loads the related elements, then iterates **those**, and so on, walking the entire relation graph and **executing a DB query at every hop**.

The cycle guard `isset($seen[$obj])` keys on the **object instance**, but Craft returns a **brand-new `ElementQuery` object on each relation access**, so the guard never matches. With any circular relation reachable from the rendered element (e.g. Entry → Asset → Entry), the recursion is effectively unbounded → hundreds of element queries → memory exhausted. This still OOMs at a 512MB limit.

This is the same class of regression that occurred when this recursion was first introduced in Twig 3.14.1 (see twigphp/Twig#4439, which caused identical OOMs in Drupal).

### How to reproduce

[...]

PR #19023, opened 2026-06-02
prevent ElementQuery in ElementQuery recursion when iterating over element's properties

i-just:
### Description
A potential fix that prevents infinite recursion when iterating over an element's properties.

### Related issues
#19004


brandonkelly:
Resolved via e23f27ccb7b7876fe545d61b1e30ecb4c270cb88

PR #19032, opened 2026-06-03
[6.x] POC PHP Web Components

brianjhanson:
Something I've been mulling over for a bit and finally put together a proof of concept today to get an idea of what it feels like. 

Along with our web components we generate a [Custom Elements Manifest](https://github.com/webcomponents/custom-elements-manifest) which contains a JSON representation of the components in our system. Using that, I wanted to see how reasonable it was to generate PHP classes that would mirror the API of our web components. The goal would be to use those components instead of the various `*Html` methods we have around the codebase. 

It turns out it's pretty feasible. 

This PR sets up a proof of concept of that, mostly using Claude for the dirty work. 

I'm still on the fence on if this is a good idea or not. I'm worried that the burden of maintenance will be larger than the benefit of doing something like this instead of `Html::tag('craft-web-component')`


brianjhanson:
Closing in favor of #19248

PR #19040, opened 2026-06-05
Relocate ValidationRulesResolving

AugustMiller:
100% breaking changes! 😈

- The `ValidationRulesResolving` is no longer emitted for only Elements, so I've relocated it into the `Validation` namespace.
- It also had some compatibility issues when [Rulesets were used in place of FormRequests](https://github.com/craftcms/laravel-ruleset-validation#2-you-can-inject-a-ruleset-directly-for-a-request)… specifically that `ValidationRulesResolving::$component` (now `$subject`) wasn't permissive enough for Request objects.
- The Ruleset responsible for `$rules` on the event is now included, as well.

CHANGELOG.md contains specific descriptions, and the event class docblock has been updated with an example of expected usage.


AugustMiller:
Testing locally, I got a bizarre failure with `Tests\Unit\Support\DiffTest` 🙃

riasvdv (approved):

PR #19043, opened 2026-06-05
[6.x] Remove ConfigService, consolidate onto window.Craft global

brianjhanson:
This PR addresses a bit of a larger change that's been bothering me for a bit. When we named the front end package `@craftcms/cp` we were thinking that all of the code related to the CP would be held in there. This was pre-inertia when the idea of the new CP was a bit more more fluid. 

Now that we've settled on inertia, it became a little difficult to draw the line of what belonged in `@craftcms/cp` and what belonged in the main `resources` folder.

This switches things up a to make that more clear.

The `@craftcms/cp` package has been renamed to `@craftcms/ui` and is more narrowly scoped to the web components and a handful of utilities they rely on (most notably the `t` function). I've moved the services over to the main `resources/js` file where they're more appropriate. I don't think anyone is going to want or need to use our `QueueService` anywhere but in our CP. it makes sense to keep it closer to the chest. 

This PR also clarifies the configuration story. I had originally thought we'd have some kind of `Cp` export from the `@craftcms/cp` package that you would configure with data from the server, and then import and use throughout the codebase. However, for the moment, that isn't terribly feasible. We still have a lot of code that expects a global `Craft` variable to be around, and to have the configuration data tied to it. Rather than fight that uphill battle now, this moves everything back to a single `window.Craft` variable that can be used anywhere. This variable will likely get a few new tricks as we go along, but for the moment it's largely the way it used to be. 

---

#### Changes

1. **Remove ConfigService, consolidate onto window.Craft (5386a30)**

 - Eliminates the ConfigService class and window.Cp/window.CpConfig globals
 - All runtime config is now accessed via the single window.Craft object
 - Simplifies how CP JavaScript accesses URLs, CSRF tokens, and feature flags

2. **Decouple @craftcms/cp from window.Craft global (a0ade64)**

 - Introduces a configure() / getCraft() pattern so the package doesn't directly reference window.Craft
 - App bootstrap calls configure(window.Craft) once at startup

3. **Rename @craftcms/cp → @craftcms/ui, extract services (603dedc)**

 - Package is now a pure UI library: web components, design tokens, translations, utilities
 - Services (QueueService, AssetIndexer, Csrf) and API clients (actionClient, apiClient) move to resources/js/common/
 - auth-challenge-form uses native fetch; disclosure uses the package's own Cookies class directly
 - Removes axios as a package dependency

4. **Decouple legacy CP.js from QueueService (1c8e0f7)**

 - Legacy Garnish CP controller no longer imports from the resources layer
 - Queue methods delegate to Craft.$queue (set by bootstrap)

5. **Move CP-specific styles to resources (cdb2801)**

 - cp.css (skip-link, error-list, sr-only, etc.) moves from the package to resources/css/ui.css
 - Design system primitives (tokens, variables, colorable, base) remain in @craftcms/ui
 - New base.css entry in the package for Storybook / consumers needing just the design system


brianjhanson:
I opened #19260 for renaming `@craftcms/cp` to `@craftcms/ui`. I think the configuration refactor will come a bit later

PR #19050, opened 2026-06-06
[4.x] Encode returnUrl slashes

### LINKED ISSUE #19037: [5.x]: `returnUrl` parameter on edit page is not fully urlencoded
### Description

The `returnUrl` URL param added in PR #18680 does not urlencode the forward slashes so you end up with a URL like `returnUrl=https%3A//test.com/admin/content/entries/singles%3Fsource%3Dsingles`. Which is fine in many environments, but can cause issues in some. For example we have a Fastly VCL that collapses that double-slash down to a single one, meaning you get redirected back to an invalid URL after saving the entry. 

### Steps to reproduce

1. Edit any entry in Craft 5.10.x. 
2. Check the return URL in the address bar

### Expected behavior

`returnUrl=https%3A%2F%2Ftest.com`

### Actual behavior

`https%3A//test.com`

### Craft CMS version

5.10.5

### PHP version

8.4

_No response_

### Database type and version

_No response_

### Image driver and version

_No response_

-

timkelty:
- Removes the `/` character exception for encoding query strings.
- `pathParam` values are kept intentionally unencoded to keep this fix as minimal as possible.

### Related issues

Fixes https://github.com/craftcms/cms/issues/19037

### DIFF
--- a/src/helpers/UrlHelper.php
+++ b/src/helpers/UrlHelper.php
@@ -80,7 +80,8 @@ class UrlHelper
     /**
      * Returns a query string based on the given params.
      *
-     * Param values will be encoded, except for `/`, `{`, and `}` characters.
+     * Param names and values will be encoded, except for `{` and `}` characters.
+     * Path param values will also preserve `/` characters.
      *
      * @param array $params
      * @return string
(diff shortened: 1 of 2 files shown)


brandonkelly:
No need to fix this in v4.

PR #19065, opened 2026-06-10
[6.x] fix `ValidateMixin::addErrors()` to accept Yii2-format error array

lukeholder:
The addErrors() compatibility shim was incorrectly typed as `(string $attribute, string $error)` — matching addError(), not addErrors().

Yii2's `addErrors()` takes `['attribute' => ['msg', ...]]`, which maps directly to `MessageBag::merge()`. Add a test covering the merge behaviour.


lukeholder:
@riasvdv not sure we are wanting to add array shape phpdocs?

riasvdv (approved):

PR #19071, opened 2026-06-10
Stream the Set Status bulk action with Db::each() to avoid OOM

### LINKED ISSUE #19070: [5.x]: SetStatus bulk action loads the entire selection into memory, goes OOM on large selections
### Description

I'm running into a memory exhaustion issue when using the "Set status" bulk element action on a large selection of elements.

Currently, `craft\elements\actions\SetStatus::performAction()` loads the entire selection into memory at once using `$query->all()`. Because the array of elements is retained during the entire foreach loop where `saveElement()` is called, peak memory usage grows linearly with the number of selected elements.

With a large enough selection, this hits the PHP `memory_limit` and causes the request to die with a fatal error: `Allowed memory size of … bytes exhausted.`

### Steps to reproduce

1. Create a Structure section with a realistic content model (e.g. ~2 Matrix fields containing CKEditor blocks, dropdowns, and a few relation fields).
2. Seed ~9,500 entries into this section.
3. Set memory_limit = 512M.
4.  In the Entries index, select all entries and run Set status → Disabled. (On a Structure section the index lazy-loads as you scroll rather than paginating, so select-all quickly covers the whole section, which is how our customers hit this in production.)

### Expected behavior

All selected entries are updated successfully without hitting the memory limit. Memory usage should remain stable regardless of the selection size, consistent with how `resaveElements()` handles bulk operations.

### Actual behavior

The request runs out of memory before it can process all selected elements. With memory_limit = 512M, the request runs out of memory after processing about 2,800 entries.

Memory usage increases steadily as more elements are loaded and processed, eventually exceeding PHP's memory_limit and resulting in a fatal "Allowed memory size exhausted" error.

Note: When testing a local fix that streams elements using Db::each(), the same 9,500-entry selection completes successfully with a stable memory footprint.

I'll follow up with a PR that streams the loop with `Db::each()`, mirroring how `resaveElements()` handles this.

### Craft CMS version

5.10.5

### PHP version

8.4.20

_No response_

### Database type and version

MySQL 8.0 and MariaDB 10.11.16

### Image driver and version

_No response_

- None required (reproduced on a clean Craft 5 install)

nvdkemp:
Fixes #19070

**Problem**
Currently, `SetStatus::performAction()` loads the entire selection into memory using `$query->all()` and holds it during the `saveElement()` loop. This causes peak memory to grow linearly with the selection size, meaning large bulk status changes easily hit the `memory_limit` (HTTP 500). See #19070 for a repro.

**Solution**
This PR switches the query to stream the elements using `Db::each()`, exactly like `Elements::resaveElements()` does. This ensures each element is freed from memory right after it gets saved. 

This is strictly a performance fix: events, validation, revisions, and results remain completely unchanged. With this fix, the ~9,500-entry selection from the issue completes with a stable ~175 MB peak instead of crashing partway through.

*Note:* Since we no longer have an array to count, I added a `$total` query to replace `count($elements)`. This ensures the 'all failed' check and singular/plural flash messages continue to work exactly as before.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Fixed a bug where the “Delete” element edit page action wasn’t working properly when editing a provisional draft.
 - Fixed a bug where `craft\helpers\App::parseEnv()` wasn’t returning boolean values for environment variable names that resolved to `true`/`false` values. ([#19029](https://github.com/craftcms/cms/issues/19029))
 - Fixed a styling issue.
+- Fixed a bug where the “Set Status” bulk element action could exhaust the memory limit on large selections. ([#19070](https://github.com/craftcms/cms/issues/19070))
 
 ## 5.10.5 - 2026-06-02
 
--- a/src/elements/actions/SetStatus.php
+++ b/src/elements/actions/SetStatus.php
@@ -12,6 +12,7 @@ use craft\base\Element;
 use craft\base\ElementAction;
 use craft\base\ElementInterface;
 use craft\elements\db\ElementQueryInterface;
+use craft\helpers\Db;
 
 /**
  * SetStatus represents a Set Status element action.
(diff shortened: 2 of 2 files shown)


brandonkelly:
Thanks for the PR!

I went ahead and applied similar fixes to other bulk actions that could potentially be affected by the same bug. (79c49f4eb8a2c0816791286af9db9a4b20553a9a)

PR #19072, opened 2026-06-10
[6.x] translatable asset alt text amends

i-just:
### Description
v6 version of https://github.com/craftcms/cms/pull/19067

In addition to the changes mentioned in PR #19067, this PR also drops the `assets.alt` column.

I bumped the `MIN_VERSION_REQUIRED` (src/Cms.php) to `5.11.0` as that will ensure that values from the `assets.alt` are copied over to the `assets_sites.alt` (details around that are on the other PR), and then when you upgrade to v6, that column can be dropped.

### DIFF
--- a/src/Asset/Elements/Asset.php
+++ b/src/Asset/Elements/Asset.php
@@ -316,12 +316,6 @@ class Asset extends Element
 
     public function __construct($config = [])
     {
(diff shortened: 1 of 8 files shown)


i-just:
I messed up the merge, and it was easier to close this one and open a new one: https://github.com/craftcms/cms/pull/19371

Issue #19069, opened 2026-06-10
[5.10.x]: Uploading images with a © symbol in their EXIF/IPTC metadata fails with SQL error 1366 since 5.10


johnvanhulsen:
### Description

Since updating to Craft 5.10, we can no longer upload certain JPEGs. The upload fails with this error:

SQLSTATE[HY000]: General error: 1366 Incorrect string value: '\xA9 Cor ...' for column 'alt' at row 1
The SQL being executed was: INSERT INTO `craft_assets` (..., `alt`, ...) VALUES (..., '... - Photo Vincent Kalut /PN /� Cor Vos 2026', ...)

These are press agency photos (cycling photos from Cor Vos / Photo News) that have a caption embedded in their metadata, ending with something like "© Cor Vos 2026". We've been uploading these photos for years without any problems. It only started failing after updating to 5.10.

After some digging I think I understand what's happening:
- 
- Craft 5.10 introduced a new feature that automatically fills the asset's alt text from the image's embedded metadata. That explains why this never happened before.
- The caption in these photos is apparently not saved as UTF-8 by the photo agency's software, but in an older encoding (Latin-1, I believe). In that encoding the © symbol is the byte `\xA9`, which matches the byte MySQL is complaining about in the error.
- So Craft reads the caption from the file and tries to save it into the database as-is, MySQL refuses it because it's not valid UTF-8, and the whole upload fails.

At first I thought it was our database's fault, because our tables were still on `utf8mb3`. So I ran `php craft db/convert-charset` and converted everything to `utf8mb4`, but the error stays exactly the same. From what I understand that makes sense: `\xA9` on its own simply isn't valid UTF-8, no matter which of the two charsets the table uses (in UTF-8 the © symbol would be `\xC2\xA9`).

So it looks like Craft needs to either convert the metadata text to proper UTF-8 before saving it, or just skip filling in the alt text when the text isn't valid UTF-8 instead of failing the entire upload.

This is a real problem for us, because basically every photo we get from this agency has "©" in the caption, so right now we can't upload any of them.

### Steps to reproduce

1. Craft 5.10.0 or later, with MySQL.
2. Get a JPEG with a caption in its IPTC metadata that contains a © symbol and is saved in Latin-1 encoding instead of UTF-8. Press agency photos often have this. I've upload one original file here: https://oym-general.b-cdn.net/production/CORVOS_00039036-072.jpg

### Expected behavior

[...]

PR #19073, opened 2026-06-10
getting alt text from IPTC and encoding

i-just:
### Description
This PR adds functionality that tries to detect if IPTC data is UTF-8 encoded and, if not, converts it to such. When a conversion occurs, it is assumed that the most likely encoding for the IPTC data was Windows-1252.

### Related issues
#19069


brandonkelly:
Fixed this by just passing the alt value through `StringHelper::convertToUtf8()`, like we do for all GET/POST params in `craft\web\Request`. (31f04bf7bc36e4d30d8e50c42436937e73d9ebb3)

PR #19079, opened 2026-06-11
[6.x] Fix return type error

brianjhanson:
Fixes a return type error (below) that happened when creating a user via the CLI

```
   TypeError

  CraftCms\Cms\Support\DateTimeHelper::toDateTime(): Return value must be of type DateTime|false, Carbon\CarbonImmutable returned

  at /tmp/packages/cms/src/Support/DateTimeHelper.php:86
     82▕             return $value;
     83▕         }
     84▕
     85▕         if ($value instanceof DateTimeInterface) {
  ➜  86▕             return Date::instance($value);
     87▕         }
     88▕
     89▕         if (! $value) {
     90▕             return false;

  1   /tmp/packages/cms/src/Element/Operations/ElementWrites.php:973
      CraftCms\Cms\Support\DateTimeHelper::toDateTime()

  2   /tmp/packages/cms/src/Element/Operations/ElementWrites.php:431
      CraftCms\Cms\Element\Operations\ElementWrites::updateModel()

Failed to run artisan craft:users/create: exit status 1
```

I'm pretty sure this is related to converting all of the date times to Carbon, but I didn't spend much time tracking the reason down.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - Fixed a bug where the `craft:install` command would hang if run within a production environment.
+- Fixed an error that could occur when `CraftCms\Cms\Support\DateTimeHelper::toDateTime()` returned a `DateTimeInterface` implementation other than `DateTime`. ([#19079](https://github.com/craftcms/cms/pull/19079))
 - Fixed a bug where “Replace relation” action buttons weren’t working.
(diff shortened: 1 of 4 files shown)


riasvdv (approved):

brianjhanson:
The errors on this seemed to get deeper as I went (the current ones are in the adapter). I was trying to just do a quick fix between other things which is why I haven't circled back yet.

Issue #19082, opened 2026-06-12
[5.x]: cannot merge multiple configuration files for per-environments.


peterspicer-unfold:
### Description

The documentation makes it clear that it should be possible to have general.php for standard config and a general.console.php file for console specific use, e.g. to allow overrides that wouldn't normally be possible.

My use case for example is to have CRAFT_ALLOW_ADMIN_CHANGES to be false generally but permit it for command line to deploy changes from central repo.

### Steps to reproduce

1. Install CraftCMS 5.
2. Set CRAFT_ALLOW_ADMIN_CHANGES to false in .env
3. Verify that no admin changes are possible in either admin or command line (e.g. from `craft up`)
4. Create a general.console.php file - it only needs minimal content, e.g. return ['allow_admin_changes' => true];
5. Try to `craft up`

### Expected behavior

`craft up` should run normally.

### Actual behavior

Exception is thrown:

```
Fatal error: Uncaught Error: Class "Craft" not found in /path/to/site/vendor/craftcms/cms/src/services/Config.php:171
Stack trace:
#0 /path/to/site/vendor/craftcms/cms/src/services/Config.php(106): craft\services\Config->_createConfigObj('general', 'general.console', Object(craft\config\GeneralConfig))
#1 /path/to/site/vendor/craftcms/cms/src/services/Config.php(240): craft\services\Config->getConfigSettings('general')
#2 /path/to/site/vendor/craftcms/cms/bootstrap/bootstrap.php(69): craft\services\Config->getGeneral()
#3 /path/to/site/vendor/craftcms/cms/bootstrap/console.php(35): require('/path/to..')
#4 /path/to/site/craft(12): require('/path/to...')
#5 {main}
  thrown in /path/to/site/vendor/craftcms/cms/src/services/Config.php on line 171
```

The reference is to Craft::configure, but this looks like it should App::configure. Changing this over lets craft up run, while disallowing changes in admin in the environments I'd like it to not be web accessible.

### Craft CMS version

5.10.5

### PHP version

8.3.27

### Operating system and version

MacOS something

### Database type and version

MySQL 8.4

### Image driver and version

n/a

### Installed plugins and versions

- CKEditor 5.5

brandonkelly:
`CRAFT_*` environment variables always take precedence over what’s in `config/general.php`.

[...]

PR #19083, opened 2026-06-12
[5.x] Use App for merging configurations.

peterspicer-unfold:
### Description

Switches from using Craft global object (which doesn't exist) to App global object (which does) to use the configure method for merging configuration files.

### Related issues

#19082


brandonkelly:
Thanks!

PR #19085, opened 2026-06-12
[5.x] Keep stale previews non-indexable

timkelty:
### Description

Keeps preview-intent URLs non-indexable even after their preview token becomes invalid or expires, without changing valid preview detection.

Related: https://github.com/craftcms/cloud/issues/166


brandonkelly:
Resolved via 4741fad1dfef3da44d4512d960d972e570c5abae

PR #19122, opened 2026-06-18
[6.x] Fix Twig string template exception reporting

riasvdv:
### Description

Fixes Twig exception mapping for inline string templates so Laravel exception reporting keeps valid string filenames instead of receiving null trace file values.

### DIFF
--- a/tests/Unit/Twig/TwigExceptionMapperTest.php
+++ b/tests/Unit/Twig/TwigExceptionMapperTest.php
@@ -2,7 +2,10 @@
 
 declare(strict_types=1);
 
+use CraftCms\Cms\Twig\TemplateRenderer;
+use CraftCms\Cms\Twig\Twig;
 use CraftCms\Cms\Twig\TwigExceptionMapper;
+use CraftCms\Cms\View\TemplateMode;
 
 it('ignores missing compiled template files', function () {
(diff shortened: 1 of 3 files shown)


tests/Unit/Twig/TwigExceptionMapperTest.php
```diff
@@ -11,3 +14,27 @@
...
+
+        $files = collect($trace)
+            ->filter(fn (array $frame) => array_key_exists('file', $frame))
+            ->pluck('file')
+            ->all();
+
+        expect($files)
+            ->not->toBeEmpty()
+            ->each->toBeString()
+            ->and(collect($files)->contains(fn (string $file) => str_starts_with($file, '__string_template__')))->toBeTrue();
```
AugustMiller, line 34: Twig also exposes `template_from_string()`, directly, which can change how template strings are named…

```twig
{{ include(template_from_string('{{ name }}', 'buggedTemplateString'), {}) }}
```

…which produces:

```
Variable "name" does not exist in "buggedTemplateString (string template 36556b15baf74d7ef8c1ddb07b3384f5)" at line 1.
```

Given this detection is only in a test (and the correct template error is indeed reported, either way), it would only start failing if Twig changed its automatic naming scheme. :+1:

AugustMiller (approved):

PR #19095, opened 2026-06-15
[6.x] Inertia element indexes

brianjhanson:

PR #19128, opened 2026-06-18
[6.x] Inertia entry type

brianjhanson:
First off, apologies for the large PR. I ended up making a lot of larger decisions in the process of porting this page. 

On the surface, this is the port of the Entry Types edit page over to inertia, but dig a bit deeper and there's a lot of decisions about how we'll handle Javascript, building on the work started in #19129. 

## The javascript side 
We continue brining Garnish into the modern age where we can. Porting to typescript, removing jQuery references and updating to the ESM lifestyle. At the moment all of that is in a separate `@craftcms/garnish` package but we may end up just folding it into the `@craftcms/cp` package for simplicity. The packages are getting a little sprawling. As time goes on, we'll pull more and more of the `@craftcms/garnish` files into `@craftcms/legacy` so the legacy code will use the new code as the internals. That will ensure that all of the legacy `Garnish.*` and `Craft.*` code paths will use the new code.

The most notable module migration here is the field layout designer (which comes with the card view designer and generated fields table). These three components outline my general approach to modernization, which is made up of a few parts:

- `modules/field-layout-designer/field-layout-designer.ts` — the main class, represents a straight port of the old `FieldLayoutDesigner.js` file into typescript and removing JQuery whenever possible. You should be able to use this class just as you could call `Craft.FieldLayoutDesigner`. If you have code that still calls `Craft.FieldLayoutDesigner` you can import the `field-layout-designer/index.ts` file which will shim the new class into the old structure.
- `modules/field-layout-designer/field-layout-designer.ce.ts` — A custom element that can be used in server generated code which will initialize the `FieldLayoutDesigner`. The goal of this it to save ourselves from namespaces when initializing FLD. Each web component is a totally separate instance, so you don't have to worry about multiples on the same page. 

There are a few other modules following this pattern in here (`generated-fields`, `listbox`, `sortable-checkbox-select`) which hopefully give a sense of the process. 

## The twig side

This PR also replaces the markup within the `buttonGroup.twig` file with the new `craft-button-group` web component. This change should be transparent and represents how I'm thinking about replacing the "leaf node" twig templates as we move forward. 

For more complex twig components, we may want to introduce a setting or service provider that lets you opt into the new markup instead of the old.

### New `@craftcms/cp` components  
- **`craft-select-rich` + `craft-select-invoker`** — rich `<select>` replacement (lifted from #19095).  
- **`craft-action-menu`** (+ `action-menu.types.ts`, `action-item` updates) — replaces the  
  legacy disclosure menu and now supports a data-driven rendering approach. 
- **`craft-icon`** — enhancements to support the icon picker.  
- **`craft-input`** — gained hidden-input support via the `hidden-input` attribute. It also gained a `monospace` attribute to display the text in the monospace font
- **`craft-reorder-button`** —  A special button variant that will automatically render an action menu with "Move up" and "Move Down" options within it. We'll need to expand this to support left / right eventually but this covers most use cases we have currently


src/Http/Controllers/Settings/EntryTypesController.php
```diff
@@ -127,94 +105,80 @@ public function edit(Request $request, ?EntryTypeModel $entryType = null): CpScr
...
+            'html' => $this->fieldLayoutDesigner->html($fieldLayout, [
+                'disabled' => $this->readOnly,
+                'withGeneratedFields' => true,
+                'withCardViewDesigner' => true,
+                'autoBoot' => false,
+            ]),
+        ];
+
+        $translationMethodOptions = [
+            ['value' => TranslationMethod::None->value, 'label' => t('Not translatable')],
```
riasvdv: `TranslationMethod` already has a `description` method, probably good to add a `label` method as well

src/Cp/FieldLayoutDesigner/FieldLayoutDesigner.php
```diff
@@ -139,81 +148,21 @@ public function html(FieldLayout $fieldLayout, array $config = []): string
...
-                    'class' => array_filter([
-                        ! $this->showFldUiElementSelector($fieldLayout, $element) ? 'hidden' : null,
-                    ]),
-                ]), $availableUiElements)).
-                Html::endTag('div') // .fld-ui-library
-                : '').
-            Html::endTag('div'). // .fld-library
-            Html::endTag('div'). // .fld-container
-            Html::endTag('div'); // .layoutdesigner
+        return view('c::forms.fld.designer', [
```
riasvdv, line 135: So good to get rid of that PHP HTML soup

packages/craftcms-cp/src/components/icon/icon.ts
brianjhanson, line 1: Updated to handle scenarios where we get the SVG markup from the server (of which there are at least a handful). Eventually the `craft-icon` component can/should handle this automatically but for the moment this felt like a better middle ground.

packages/craftcms-cp/src/components/input/input.ts
brianjhanson, line 1: Adds the ability to hide the input completely with `<craft-input hidden-input>`. Useful for situation where we want some of the input functionality (label, errors, etc.) but the way the user interacts with the input is a bit more complex (like the icon picker). I imagine we'll need this or something like it for things like relations and assets as well

packages/craftcms-cp/src/components/reorder-button/reorder-button.ts
brianjhanson, line 1: Creates a new `craft-reorder-button` which was previously implemented as a Vue component. The goal is for this to be used whenever items can be reordered as it will show a drag handler but will display a dropdown of "Move ..." options when clicked.

This will need to be expanded to handle left / right eventually, but up / down handles the vast majority of our use cases. 

Setting the position to determine when each item should be disabled or not is fairly annoying but I can't thing of a way around it. The good news is, if you're using this component in something like the `<craft-sortable-checkbox-list />` component we'll handle all of that for you.

packages/craftcms-cp/src/components/select-color/select-color.ts
brianjhanson, line 1: A specialty rich select component that will display all of the enum colors we support.

packages/craftcms-cp/src/components/select-rich/select-rich.ts
brianjhanson, line 1: A specialty select that allows markup within the options (and will carry it  over into the display of the invoker).

packages/craftcms-cp/tailwind.css
```diff
@@ -110,4 +110,20 @@
...
+  --color-fill-normal: var(
+    --c-color-fill-normal,
+    var(--c-color-neutral-fill-normal)
+  );
+  --color-fill-loud: var(--c-color-fill-loud, var(--c-color-neutral-fill-loud));
+
+  /* ——— Text utilities ——— */
+  --color-on-quiet: var(--c-color-on-quiet, var(--c-color-neutral-on-quiet));
+  --color-on-normal: var(--c-color-on-normal, var(--c-color-neutral-on-normal));
+  --color-on-loud: var(--c-color-on-loud, var(--c-color-neutral-on-loud));
```
brianjhanson, line 128: Allows you to do things like `class="bg-fill-loud text-on-loud"` within the CP

PR #19129, opened 2026-06-19
[6.x] Modernize garnish

brianjhanson:
Rumors of my death are greatly exaggerated

I've taken a few runs at avoiding this, but nothing's felt quite right. So this is how I learned to stop worrying and love Garnish. 

At the start of this endeavor it felt like a gargantuan task to bring Garnish into the modern age. That's all kind of changed now that we have agents that can do large swaths of the grunt work. 

This PR represents a partial PoC but more likely the start of bringing Garnish into the modern age. Converting it to typescript, removing jQuery and adding better documentation via a Storybook. 

While I don't think Garnish is perfect, it is pretty good, and has been consistently improved over time. I don't want to discard all those learnings from the start, but by splitting and modernizing Garnish we gain some flexibility in replacing the inner workings. 

Still very much a work in progress, just opening to get the idea out there. I'm not sure I want this to live in a separate package but doing it this way made it easier to work through without messing up too much other stuff.


brianjhanson:
Closing in favor of #19128  which has this branch pulled in

PR #19133, opened 2026-06-19
Capitalize first word in deletion blocker messages

mmikkel:
### Description

A couple of the strings added for the deletion blocker feature in 5.10 begin with an element type name produced by `lowerDisplayName()` and `pluralLowerDisplayName()`. When the count is 1, the sentence starts with a lowercase word. In English this goes unnoticed because the strings are prefixed with `'The'`, but languages without an article render a sentence starting in lowercase. 

For example, in Norwegian the message renders as:

`artikkel er relatert til 1 annen artikkel.`

Wrapping each formatted message in `StringHelper::upperCaseFirst()` capitalizes the first letter regardless of locale, fixing the casing without affecting the element type names elsewhere in the sentence.

Again, examples in Norwegian:

Before:
CleanShot 2026-06-19 at 23 08 12

After:
CleanShot 2026-06-19 at 23 08 40

_As a side note:_ even with this fix, the Norwegian translation specifically, isn't quite grammatically correct 🫠  The element type names are inserted in their indefinite base form ("artikkel", "relasjon"), but in these sentences Norwegian would use the definite form ("artikkel**en**", "relasjon**en**") — equivalent to the English "The article…" / "The relation…".

I don't think his can be solved in the static translation itself, because the ICU message string only receives the base form of the name and has no way to derive the definite form (the suffix varies by gender and inflection: artikkel → artikkelen, side → siden, produkt → produktet). I'm guessing this problem would have to be addressed in the code, e.g. by passing the definite form as a separate value. I'm unsure how to best approach that, though – but, just thought I'd mention it.

(FWIW, I'm also submitting some proposed changes to the Norwegian translations in Crowdin, that should help improve the above situation a little bit 😃)

### DIFF
--- a/src/elements/deletionblockers/RelationDeletionBlocker.php
+++ b/src/elements/deletionblockers/RelationDeletionBlocker.php
@@ -13,6 +13,7 @@ use craft\elements\ElementCollection;
 use craft\helpers\ArrayHelper;
 use craft\helpers\Cp;
 use craft\helpers\Html;
+use craft\helpers\StringHelper;
 
 /**
  * Class RelationDeletionBlocker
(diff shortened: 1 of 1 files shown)


brandonkelly:
Thanks!

brandonkelly:
Craft 5.10.8 is out with that fix.

PR #19134, opened 2026-06-19
Adds missing "Delete {type}" static translation string to CpAsset

mmikkel:
### Description

Adds missing `'Delete {type}'` and `'Delete {types}'` translation strings to `CpAsset`, which affected the new deletion blocker in 5.10.

Examples in Norwegian:

Before
CleanShot 2026-06-19 at 23 41 22

After
CleanShot 2026-06-19 at 23 41 37


brandonkelly:
Doh, thanks!

brandonkelly:
Craft 5.10.8 is out with that fix.

PR #19145, opened 2026-06-22
Locale name normalization

brandonkelly:
Laravel locales use `_` by convention (`en_US`) whereas Craft uses `-`.

Replaces all calls to `app()->getLocale()` / `app()->setLocale()` with calls to new `CraftCms\Cms\getLocale()` / `CraftCms\Cms\setLocale()` methods, which normalize locale names from/to the Laravel format.


brandonkelly:
Resolved via bc44cf0514446710bda15b1b987bcabdc3777a84

PR #19148, opened 2026-06-22
[6.x] Blade parity

riasvdv:
### Description

Adds initial Blade directives and an engine-agnostic renderer for Craft templates which brings near-parity for Blade templates

See [docs/blade.md](https://github.com/craftcms/cms/blob/feature/blade-parity/docs/blade.md)

Of course we can add more in the future or upon request


src/View/TemplateRenderer.php
```diff
@@ -0,0 +1,84 @@
...
+use Illuminate\Container\Attributes\Scoped;
+
+use function CraftCms\Cms\t;
+
+#[Scoped]
+readonly class TemplateRenderer
+{
+    public function __construct(
+        private BladeRenderer $blade,
+        private TemplateResolver $templateResolver,
```
AugustMiller: I find it a _little_ awkward that all template resolution still flows through a class ( `CraftCms\Cms\Twig\TemplateResolver`) in the `Twig` namespace. Is it possible/responsible to move into the same generic `View` space?

src/View/TemplateRenderer.php
```diff
@@ -0,0 +1,84 @@
...
+            return $this->blade->renderPageFile($resolvedTemplate, $variables, $templateMode, $template);
+        }
+
+        return $this->twig->renderPageTemplate($template, $variables, $templateMode);
+    }
+
+    private function isBlade(string $path): bool
+    {
+        return Str::endsWith($path, '.blade.php');
+    }
```
AugustMiller: Could we make this the responsibility of the renderer or engine, instead? Something like…

- Resolve template path
- Define list of renderers
- Emit event so plugins can add new ones
- Call (e.g.) `RendererInterface::matchPath()` on each one, returning the first one that says it can render the provided path

src/Twig/Extensions/LaravelExtension.php
```diff
@@ -24,6 +25,7 @@ public function getFunctions(): array
     {
         return [
             new TwigFunction('app', app(...)),
+            new TwigFunction('blade', fn (string $view, array $variables = []): string => app(BladeRenderer::class)->renderView($view, $variables), ['is_safe' => ['html']]),
```
AugustMiller: Dope. 😎

src/Edition/Exceptions/WrongEditionException.php
AugustMiller, line 1: Sort of off-topic, but this is a cool refactor from `RequireEdition`!

src/View/ViewServiceProvider.php
```diff
@@ -71,4 +83,38 @@ private function registerTemplateRoots(): void
...
+    private function bladeCompiler(): BladeCompiler
+    {
+        $blade = $this->app->make(CompilerInterface::class);
+
+        if (! $blade instanceof BladeCompiler) {
+            throw new RuntimeException('The blade.compiler container binding must resolve to a BladeCompiler instance.');
+        }
+
+        return $blade;
+    }
```
AugustMiller: Why can’t we just `make()` the expected concrete class?

docs/blade.md
```diff
@@ -0,0 +1,427 @@
...
+- `html`
+- `blade.php`
+
+That means `templates/news.twig` will be chosen before `templates/news.blade.php` when both exist. Change `defaultTemplateExtensions` if a project needs a different order:
+
+```php
+<?php
+
+return [
+    'defaultTemplateExtensions' => ['blade.php', 'twig', 'html'],
```
AugustMiller, line 23: Do we also need to update `CraftCms\Cms\View\TemplateMode::defaultTemplateExtensions()` so that the `Cp` context supports/loads/resolves `blade.php`, as well?

If an upgraded project defines this setting (say, to `['html', 'twig']`), they've effectively opted out of _site_ Blade template resolution, correct?

docs/blade.md
```diff
@@ -0,0 +1,427 @@
...
+];
+```
+
+Craft's template resolver still owns template lookup before either engine renders anything. Public/private template filtering, path containment, localized site template lookup, index template lookup, and registered template roots all apply before Craft decides whether the resolved file should be rendered by Twig or Blade.
+
+Control panel template lookup still defaults to `twig` and `html`. Blade CP and plugin views are rendered through Laravel's view system, either by calling `view()` directly or by using `BladeRenderer`.
```
AugustMiller, line 29: Ah, ignore the above! I missed this on my first pass and thought we were looking at a mode-agnostic feature. :v:

src/View/ViewServiceProvider.php
riasvdv: We need to do it when blade.compiler is resolved by the container, that is typed as the interface but if we use that directly phpstan complains 

We could add a phpstan ignore, but this guards against a misconfigured blade compiler (I know it’s rare, but I don’t see any harm in the extra few lines of code)

src/View/TemplateRenderer.php
riasvdv: Thanks for this suggestion! I've refactored into an interface now with a BaseTemplateRenderer which does a lot of the ceremony, the BladeRenderer is just a small implementation of that now. In theory other renderers could also be added.

src/Edition/Exceptions/WrongEditionException.php
riasvdv: Yes! I noticed there were a lot of try-catches around it with the same 404 if debug mode is disabled, so this nicely consolidates it

There's probably some other places we could use this

src/View/ViewServiceProvider.php
riasvdv: Solved by only using `afterResolving` which really is enough

docs/blade.md
riasvdv: Added blade to the CP version of `defaultTemplateExtensions` as well now

AugustMiller (approved):
No further suggestions! Very cool interoperability "olive branch" for Laravel folx who want to experiment with Twig, and vice-versa.

PR #19155, opened 2026-06-25
[6.x] User settings Inertia

riasvdv:
### Description

image


resources/js/pages/settings/users/Settings.vue
```diff
@@ -0,0 +1,335 @@
...
+                :options="defaultGroupOptions"
+                :disabled="readOnly"
+              />
+              <p class="mt-1 text-sm text-neutral-text-quiet">
+                {{
+                  t(
+                    'Choose a user group that publicly-registered members will be added to by default.'
+                  )
+                }}
+              </p>
```
brianjhanson: This should probably be in the help text instead of below the input.

src/Http/Controllers/Settings/Users/UserSettingsController.php
```diff
@@ -0,0 +1,101 @@
...
+            ] : []),
+            ...(Edition::get()->supportsPublicRegistration() ? [
+                'requireEmailVerification' => (bool) ($data['requireEmailVerification'] ?? false),
+                'validateOnPublicRegistration' => (bool) ($data['validateOnPublicRegistration'] ?? false),
+                'allowPublicRegistration' => (bool) ($data['allowPublicRegistration'] ?? false),
+                'deactivateByDefault' => (bool) ($data['deactivateByDefault'] ?? false),
+                'defaultGroup' => $data['defaultGroup'] ?? null,
+            ] : []),
+        ];
+    }
```
brianjhanson: Should this get moved into the `UserSettingsRequest`?

brianjhanson (approved):

src/Http/Controllers/Settings/Users/UserSettingsController.php
riasvdv: Good call 👍

PR #19166, opened 2026-06-26
[5.x] Fix generated fields preview

brianjhanson:
While porting the generated fields and card previews over to Inertia I noticed a bug that was preventing your generated field label from being added to the card view designer's checkboxes properly. Once I got it added, I noticed that if you checked it the preview wouldn't update properly. This fixes both of those issues.

The fix for updating the preview isn't the most elegant, but it works. 

**Before**

https://github.com/user-attachments/assets/7949eb12-354f-4828-bbd8-bed3b49d51eb

**After**

https://github.com/user-attachments/assets/e858a89a-d2bc-47f7-ae14-ba8eab0c4cc4


brianjhanson:

PR #19167, opened 2026-06-27
[6.x] Port GraphQL screens to Inertia

riasvdv:
### Description

Ports the GraphQL token and schema control panel screens to Inertia, moves their routes to normal CP routes, updates shared permission data/types, and fixes the main navigation active state for the new GraphQL pages.


resources/js/common/layouts/AppLayout.vue
```diff
@@ -279,7 +281,17 @@
               <template v-if="readOnly">
                 <CalloutReadOnly />
               </template>
-              <slot></slot>
+              <template v-if="hasDetails">
```
riasvdv, line 285: There was no way to have a sidebar on the right yet, hope this kind of fits what you were planning

resources/js/modules/permissions/components/PermissionList.vue
```diff
@@ -49,7 +103,7 @@
     <li>
       <CraftCheckbox
         :label="item.label"
-        :model-value="modelValue.includes(key.toLowerCase())"
+        :model-value="modelValue.includes(key)"
```
riasvdv: I don't know specifically why these were being lowercased, but the GQL permissions are case sensitive so otherwise the settings didn't work

resources/js/modules/settings/composables/useSettingsSave.ts
```diff
@@ -25,13 +25,14 @@ export function useSettingsSave<T extends Record<string, any>>(
...
-        preserveState: true,
-      };
-    }
+    const submitOptions = redirect
+      ? {
+          preserveScroll: true,
+          preserveState: true,
+        }
+      : {
+          replace: true,
```
riasvdv, line 34: This makes sure there's no browser history added for save and continue

brianjhanson (approved):

resources/js/common/layouts/AppLayout.vue
brianjhanson: Something like this. I still need to do a pass on `AppLayout` vs the existing twig implementation to make sure all (or at least most) of the same extension points are supported.

resources/js/modules/permissions/components/PermissionList.vue
brianjhanson: There was some kind of issue with UserPermissions. I think those are stored lower cased specifically which was different from the values on the page. 

Just make sure the permissions on the user page aren't broken and we should be good.

riasvdv: Ok yeah it broke them, I've added an option to preserve case now instead

brianjhanson: Do you think it's worth adding a ticket to make both of these permission situations behave the same server-side?

riasvdv: I've gone ahead and just done that, it's more straightforward to do this server-side anyway

PR #19170, opened 2026-06-27
[6.x] Port GraphiQL to Inertia

riasvdv:
### Description
Port the GraphiQL control panel page to the Inertia app shell, move it to `/graphql/explore`, and remove the core legacy GraphiQL asset bundle while keeping adapter compatibility self-contained.

image


riasvdv:
Something weird happened with the branches here, so I just went ahead and merged

PR #19172, opened 2026-06-27
[6.x] Port CP set-password and verify-email into Inertia

riasvdv:
### Description

Pretty straightforward views to port, also straightened up the `id` vs `uid` usage. Fallbacks for frontend routes still work as well with the Twig views.


resources/js/pages/auth/SetPassword.vue
brianjhanson, line 1: Ideally this would be a web component (like `login`, `totp` and `recovery-codes`). My thinking there is that we could expose them to end users who could include those web components in whatever template they want as a more robust way to customize the login flow and pages

brianjhanson (approved):
One comment, but something that we could also do down the line as it's mostly a "wouldn't it be nice if..." thought and not fully realized.

resources/js/pages/auth/SetPassword.vue
riasvdv: I've moved them to web components, also updated the Twig fallback views (which are used on the site routes) to use those components.

The components themselves now have styles though, which people might not want when using them in their own implementations? The login components do as well so I've followed that for now

resources/js/modules/auth/components/login/login-form.styles.ts
brianjhanson, line 1: We should probably rename this file to `auth-form.styles.ts` and move it out of the `login` folder

resources/js/pages/auth/SetPassword.vue
brianjhanson: Yeah, I imagine these components would be for people who are okay with that or are good with the amount of customization that will be available from the component library. People who don't want any of that will still be able to implement an entirely custom flow.

brianjhanson (approved):

Issue #19177, opened 2026-06-29
[6.x]: Craft DynamicRoute uses View::file(), causing path-based Blade view names and breaking named view composers


JasonStainton:
### Description

On a **fresh install** of **Laravel v13** + **CraftCMS v6** (no added plugins), Blade templates rendered via Craft dynamic routing use `view()->file(...)`, which sets the view **name** to the full file path rather than the logical view name.

This affects custom Laravel view composers/hooks that target named views (e.g. `View::composer('welcome', ...)`), because the composer event name no longer matches Laravel’s standard naming behavior.

For custom view composer compatibility, rendering via `view()->make($template, $variables)` appears to preserve expected Laravel behavior.

---

### Steps to reproduce

1. Create a fresh install of Laravel v13 and CraftCMS v6 (no additional plugins).
2. Create a CraftCMS site with an entry that routes to the default `welcome.blade.php` blade template, eg home page.
3. Ensure the default Laravel `web.php` route for home is removed so routing is going through Craft.
4. Add a custom composer:   - `View::composer('welcome', function (View $view) { dd($view->name(), $view->getPath(), $view->getData()); });` to the boot method of the `AppServiceProvider.php`
5. Hit `https://example.com` so Craft resolves and renders the page template through `DynamicRoute`.
4. Observe that with `view()->file($resolvedTemplate, ...)`, the composer either doesn’t match `welcome` as expected or receives path-based naming behavior (if set to wildcard `*`).
5. Compare by switching to `view()->make($template, ...)` in Craft’s Blade render in `DynamicRoute`.

---

### Expected behavior

Laravel-style behavior for named views should be preserved in this flow:

- `$view->name()` should be logical (e.g. `welcome`)
- `$view->getPath()` should be the resolved file path
- Named composers like `View::composer('welcome', ...)` should work consistently

---

### Actual behavior

With `view()->file($resolvedTemplate, ...)`:

- The view name is set to the full file path (e.g. `/var/www/app/resources/views/welcome.blade.php`)
- Composer matching becomes path-based rather than name-based
- Custom named composers/hooks are harder to target and behave differently from standard Laravel expectations

---

### Craft CMS version

6.0.0-alpha.9

### PHP version

8.5.5

### Installed plugins and versions

None

riasvdv:
Should already be fixed with https://github.com/craftcms/cms/pull/19148

JasonStainton:
[...]

PR #19178, opened 2026-06-29
Use View::make() for Blade template rendering in DynamicRoute

JasonStainton:
### Description
This updates Blade rendering in DynamicRoute to use `view()->make($template, $variables)` instead of `view()->file($resolvedTemplate, $variables)`.

#### Why:

`View::file()` sets both the view name and path to the resolved file path.
That results in path-based Blade view names (e.g. `/var/www/app/resources/views/welcome.blade.php`).
Path-based names make named view composers (e.g. `View::composer('welcome', ...)`) not behave as expected.
Using `View::make()` preserves Laravel’s logical view naming for composer matching while still resolving the same Blade template file for rendering.

### Related issues
#19177

### DIFF
--- a/src/Route/DynamicRoute.php
+++ b/src/Route/DynamicRoute.php
@@ -64,7 +64,7 @@ class DynamicRoute
         }
 
         if (Str::endsWith($resolvedTemplate, '.blade.php')) {
-            return view()->file($resolvedTemplate, $variables)->render();
+            return view()->make($template, $variables)->render();
         }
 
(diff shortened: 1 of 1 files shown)


riasvdv:
Thanks for the PR! However in https://github.com/craftcms/cms/pull/19148 I've already made some changes to this which should go out in the next alpha!

JasonStainton:
Perfect, even better 🙌

PR #19179, opened 2026-06-30
adds isRestrictedDir method

i-just:
### Description
This PR adds `Security::isRestrictedDir()` method, which is a wrapper for the `Security::isSystemDir()`, and additionally checks if the given path is located within some OS-specific directories.

### Related issues
n/a


src/services/Security.php
```diff
@@ -219,4 +219,59 @@ public function isSystemDir(string $path): bool
...
+
+        // is it located within a Craft-specific system directory
+        if ($this->isSystemDir($path)) {
+            return true;
+        }
+
+        // is it located within a sensitive os directory
+        // windows-based sensitive directories
+        if (PHP_OS_FAMILY === 'Windows') {
+            $winRoot = rtrim($_SERVER['SystemRoot'] ?? 'C:\\Windows', '\\');
```
angrybrad: You'll probably want to normalize the path separators here `/` and `\` are both valid on windows, but there are also file:// protocol wrappers to consider.

Actually, this should cover most all of them: https://github.com/craftcms/cms/blob/5.x/src/helpers/FileHelper.php#L54

angrybrad (approved):

PR #19192, opened 2026-07-01
[6.x] Inertia user permissions

riasvdv:


resources/js/modules/permissions/components/PermissionList.vue
riasvdv, line 1: Added support for "locked" items here, which is needed on the permissions screen when you select a user group and the user inherits its permissions

resources/js/modules/user/components/UserGroupSelect.vue
riasvdv, line 1: This is mostly copied from the `EntryTypeSelect` without its reordering capabilities. We should probably have a more generic `ComponentSelect`?

resources/js/pages/users/Permissions.vue
```diff
@@ -0,0 +1,153 @@
...
+  import CraftSwitch from '@craftcms/cp/vue/CraftSwitch.vue';
+  import {useSettingsSave} from '@/modules/settings/composables/useSettingsSave';
+  import {update} from '@actions/Users/PermissionsController';
+
+  defineOptions({
+    inheritAttrs: false,
+  });
+
+  const props =
+    usePage<CraftCms.Cms.Http.ViewModels.UserPermissionsViewModel>().props;
```
riasvdv: A ViewModel makes typing pretty easy for a page

src/Http/ViewModels/UserPermissionsViewModel.php
riasvdv, line 1: I feel like more pages could benefit, and will benefit, from using this pattern, it makes it easy to generate a TypeScript type for a full page and consolidating the logic for what it takes to prepare all the data for the frontend.

resources/js/modules/user/components/UserGroupSelect.vue
brianjhanson: Yeah, we should, that's what the 5 codebase does I was just waiting until we needed it to abstract it. We also probably to revert the dragging to reorder, etc. to our modernized garnish. That way when we replace the internals of that `DragSort` (or whichever it is) the rising tide will lift all boats.

src/Http/Controllers/Users/PermissionsController.php
```diff
@@ -41,37 +42,31 @@ public function index(Request $request, ?int $userId = null): CpScreenResponse
...
+            ->prepareScreen(function (CpScreenResponse $response, string $containerId) {
+                HtmlStack::jsWithVars(
+                    fn ($containerId) => <<<JS
+                        new Craft.ElevatedSessionForm('#' + $containerId, [
+                            '[name="admin"]',
+                            '[data-user-groups-input]',
+                            '[data-user-permissions-input]'
+                        ]);
+                    JS,
+                    [$containerId],
```
brianjhanson, line 56: We can leave this for now, but this is the kind of shennanigans that we can avoid by creating a web component wrapper for the `Craft.*` modules.

src/Http/Controllers/Users/PermissionsController.php
```diff
@@ -41,37 +42,31 @@ public function index(Request $request, ?int $userId = null): CpScreenResponse
...
-                        'param' => 'sendActivationEmail',
-                        'value' => '1',
-                    ],
-                ])
-            );
-        }
-
-        return $response;
+        return $this->asEditUserScreen($user, self::SCREEN_PERMISSIONS)
+            ->inertiaPage('users/Permissions', new UserPermissionsViewModel($user, $currentUser))
```
brianjhanson, line 46: I like the ideas of view models, but I'm wondering they could/should be combined with the `CpScreenResponse` idea. They feel a bit similar to me

riasvdv: I don’t think they’re that similar, CpScreenResponse is for structuring the response but gathering and formatting the data isn’t really a concern of it, we could create a class that extends CpScreenResponse but that would be hard to generate types for (all properties would end up in the type)

src/Http/Controllers/Users/PermissionsController.php
riasvdv: Yeah I thought I’d leave it like this and we could create the component and replace the usages in a different PR

packages/craftcms-cp/src/utilities/dom.ts
```diff
@@ -8,7 +8,14 @@ let existingJs: string[] | null = null;
...
 
-async function appendHtml(
+function waitForScript(script: HTMLScriptElement): Promise<void> {
+  return new Promise((resolve) => {
+    script.addEventListener('load', () => resolve(), {once: true});
+    script.addEventListener('error', () => resolve(), {once: true});
+  });
+}
+
+export async function appendElementHtml(
```
brianjhanson, line 18: What's the reasoning for the function name change here? I'm not against it, but I had just kept the names from Craft 5 for familiarity

brianjhanson (approved):

packages/craftcms-cp/src/utilities/dom.ts
riasvdv: Since it's being exported now I thought just `appendHtml` would conflict a bit with `useAppendHtml`

Issue #18209, opened 2026-01-05
[5.x]: View field selection list for singles, duplicates fields to select which are used multiple times


knufri:
### Description
I have created several entries as singles. I then assigned a text field multiple times to these entries (this has been possible since Craft 5.5+). Everything works fine, but when I display the table columns in the overview (view), they appear multiple times even though they are fields of the same type and have the same handle in the entries. There has been a similar issue before: [link](https://github.com/craftcms/cms/issues/16177)

<img width="2333" height="1287" alt="Image" src="https://github.com/user-attachments/assets/c2af7798-73b6-4c78-8933-ba2bc5f27c1d" />

### Steps to reproduce

1. Create entries in single section
2. Add same field multiple times
3. Go to the view field selection
4. See the duplicates

### Expected behavior

Show only once in the view field selection when the handle and field type are equal 

### Actual behavior

Duplicates are shown

### Craft CMS version

5.8.21

### PHP version

8.4.16

### Database type and version

MySQL 8.0.33

### Installed plugins and versions

```
   Name            Handle          Package Name                          Version  Installed  Enabled
   --------------  --------------  ------------------------------------  -------  ---------  -------
   Child Me!       child-me        mmikkel/child-me                      2.2.3    Yes        Yes    
   CKEditor        ckeditor        craftcms/ckeditor                     4.11.0   Yes        Yes    
   Elements Panel  elements-panel  putyourlightson/craft-elements-panel  3.0.1    Yes        Yes    
   Feed Me         feed-me         craftcms/feed-me                      6.11.0   Yes        Yes    
   Formie          formie          verbb/formie                          3.1.6    Yes        Yes    
   Image Resizer   image-resizer   verbb/image-resizer                   4.0.4    Yes        Yes    
   Maps            simplemap       ether/simplemap                       5.0.4    Yes        Yes    
   SEOMate         seomate         vaersaagod/seomate                    3.2.0    Yes        Yes    
   Video Utils     video-utils     aodihis/video-utils                   3.0.2    Yes        Yes    
   Vite            vite            nystudio107/craft-vite                5.0.1    Yes        Yes
```

i-just:
Hi, thanks for reporting! I raised a PR for this.

knufri:
Hi @i-just thank you for the fast reply and the PR 😁

brandonkelly:
Craft 5.8.22 is out with that fix. Thanks again!

PR #19197, opened 2026-07-01
Fix duplicated Content Block nested fields in element index columns

fabianhaef:
### Description

When the same Content Block field is used in the field layouts of multiple entry types within one section, the entry index’s **Table Columns** picker (under the “View” button) lists each of the Content Block’s nested fields **once per entry type**. With, say, 10 entry types sharing one Content Block field, every nested field shows up 10 times, making it impossible to tell which one to pick.

Ordinary fields shared across multiple layouts are already deduplicated for table columns — that was added in #18220 (for #18209), mirroring the condition-builder work in #15781. Content Block fields were never included in that dedup, so their nested fields still multiply.

#### Steps to reproduce

1. Create a Content Block field (e.g. “CTA”) with a couple of nested fields.
2. Create several entry types that each include that Content Block field in their layout.
3. Put them all in one channel/structure section.
4. On that section’s entry index, click **View** → look at **Table Columns**.
5. Each nested field is listed once per entry type.

### This PR

Applies the same treatment #18220 gave multi-instance fields, to Content Block fields:

- **`ElementSources::getTableAttributesForFieldLayouts()`** — Content Block instances are now grouped across the source’s field layouts by field UID + effective handle (`$groupedContentBlocks`), so each nested field is contributed once, keyed off a single representative layout element.
- **`Element::contentBlockAttributeHtml()`** — because the deduplicated column key now references one representative layout element UID (which may belong to a different entry type’s layout than the element being rendered), a new private helper `_contentBlockFieldFromLayoutElementUid()` resolves the Content Block field by field UID + effective handle within the element’s own layout. This parallels the existing `_getFieldFromAlternativeLayouts()` used for multi-instance fields, but matches on the *effective* handle (that method matches the raw layout-element handle override, which is null for non-overridden Content Blocks).

Card and thumbnail rendering are intentionally untouched: those paths resolve against each element’s own layout using per-entry-type card config keys, so they never hit the cross-layout case.

### Related

- #18220 / #18209 (the same fix for multi-instance fields)
- #15781 (the same fix for the condition builder)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Fixed a bug where Content Block fields’ nested fields could be listed multiple times in element index column settings, when the same field was used by multiple entry types. ([#19197](https://github.com/craftcms/cms/pull/19197))
 - Fixed a bug where the `site`/`siteId` params weren’t being respected on eager-loaded `localized` queries. ([#18588](https://github.com/craftcms/cms/issues/18588))
 - Fixed a bug where it wasn’t possible to enter decimal values after tabbing into a Money field. ([#19156](https://github.com/craftcms/cms/issues/19156))
 - Fixed a bug where the search inputs on the Entry Types and Fields settings index pages were case-sensitive on PostgreSQL. ([#19158](https://github.com/craftcms/cms/issues/19158))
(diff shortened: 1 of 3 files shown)


brandonkelly:
Thank you!

brandonkelly:
Craft 5.9.10 is out now with that fix. Thanks again.

PR #19212, opened 2026-07-06
[6.x] Entry index page

brianjhanson:
Ports the entry index page over to inertia and Vue. Functionality is 100% on par with Craft 5 yet (further work is under the "Element Index" tag in Linear), but it's a big enough chunk of work and works well enough that I think it's worth a merge. 

In Craft 5, when an element index was rendered, an XHR request with data plucked from local storage was immediately sent back to the server with information about how we wanted to render that view. With Vue and inertia, the initial Vue is now rendered with the data up front. 

Element indexes are made up of a hand full of composables that will cover off on the functionality. All of these composables are route agnostic, so hopefully we'll be able to re-use them to provide their functionality to any table or route in the future and not have everything coupled as tightly as it is today.

### Misc. updates

- When you pass a `name` attribute to a `<craft-button-group/>` it will essentially function as a radio button. 
- `craft-action-menu` now extends `craft-popover` which is a generic popover web component. Much like the corresponding element in Web Awesome
- `craft-button` will now only apply icon styles when `icon` is empty which will now allow you to pass in an `icon` name to create a button with an icon prefix or suffix
- `<craft-button/>` now supports an `icon-position` attribute which you can use to render the icon in either the `prefix` or `suffix` slot
- `<craft-checkbox/>` now supports the `label-sr-only` property
- `<craft-nav-item/>`s can now disable the subnav toggle or render it into a prefix slot
- URL query params are the source of truth throughout so CP index urls can be shared amongst users. 
- All six element-index XHR controllers now method-inject `ElementIndexRequest`; Laravel caches controller instances on the Route object, so constructor-injected FormRequests served stale data across in-process requests.
- Legacy notifications (`Craft.cp.displayNotification`, the "Entry copied" notice) were silently no-oping on Inertia pages — the `#notifications` container now gets created outside the Vue root at bootstrap.

The legacy Twig index and `element-indexes/*` endpoints remain untouched for all other element types and modal/field contexts.


src/Element/ElementIndexes.php
brianjhanson, line 1: @riasvdv Feels like you should probably put some eyes on this one. I'm not sure this is the best place for it, but I wanted somewhere to get some logic out of the controller.

riasvdv: This is all very HTTP bound so I think this could work as a ViewModel, or a base `ElementIndexViewModel` which can be extended for entries as `EntryIndexViewModel`, accepting the necessary variables that change the result (like `elementType`) in its constructor.

Any variables from the request can be retrieved directly inside the ViewModel so wouldn't need to be passed into the methods, or an `ElementRequest` could be passed into the constructor to make it explicit.

Any [public properties and public methods](https://github.com/craftcms/cms/blob/6.x/src/Http/ViewModels/ViewModel.php) get turned into props through `toArray` if you extend the base `ViewModel`

brianjhanson: Refactored in https://github.com/craftcms/cms/pull/19212/changes/fee058ffbf8d94bf2348e8ae4f7ec3dc33a8f201

src/Section/Sections.php
```diff
@@ -269,6 +271,62 @@ public function getEditableSections(): Collection
...
+     *     name: string,
+     *     sites: array<int>,
+     *     type: SectionType|null,
+     *     uid: string|null,
+     *     canSave: bool,
+     * }>
+     *
+     * @throws AuthenticationException
+     */
+    public function getPublishableSections(): array
```
riasvdv: Would it make more sense to have this return Section objects and do the actual mapping to an array (or a resource?) in the controller/viewmodel?

src/Cp/Cp.php
```diff
@@ -285,41 +283,9 @@ private static function previewIframeResizerOptions(GeneralConfig $generalConfig
         return $generalConfig->previewIframeResizerOptions;
     }
 
-    private static function publishableSections(User $currentUser): array
+    private static function publishableSections(): array
```
riasvdv: This method can probably be removed and just use `Sections::getPublishableSections()`

src/Section/Sections.php
brianjhanson: Refactored in https://github.com/craftcms/cms/pull/19212/changes/c2faa2712831b86cbbd1771037f1c2f3c3d6fc6b but I'm not sure how I feel about using a resource here, mostly because we're not really tied to a request (and resources seem to want to be). 

Do you have any feelings about it?

riasvdv: It seems to be used in all request related things (Cp, ViewModel) so a resource seems fine there for me, the goal of building the array is also so it can be consumed by the frontend

PR #19221, opened 2026-07-07
[6.x] Add signular dev command

brianjhanson:
Updates `npm run dev`. Now with more of what you need!

When running `npm run dev` we'll first check if you have everything all built up. If not we'll build them, and then send you on your way into Vite server land.


brianjhanson:
Closing in favor of #19222

PR #19229, opened 2026-07-08
Fixed rendering fallback CP views on FE requests

brandonkelly:
Fixes a bug where front end routes that went through `AuthenticationController::renderViewWithFallback()` were getting redirected to the CP equivalent URL, rather than rendering the CP view on the requested FE URL.

Also updates the views to submit to a FE-compatible route, when rendered on the front end.

### DIFF
--- a/resources/js/pages/auth/Login.vue
+++ b/resources/js/pages/auth/Login.vue
@@ -3,13 +3,13 @@
   import {usePage} from '@inertiajs/vue3';
   import useCraftData from '@/common/composables/useCraftData';
   import AuthBase from '@/common/layouts/AuthBase.vue';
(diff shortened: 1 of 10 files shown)


resources/js/pages/auth/Login.vue
```diff
@@ -29,7 +29,7 @@
 <template>
   <AuthBase>
     <craft-login-form
-      :action="attemptLogin().url"
+      :action="props.action"
```
brianjhanson: This can just be `action`, Vue will automatically resolve that to `props.action` within the `<template>` tags

brianjhanson (approved):

PR #19237, opened 2026-07-10
[6.x] Modernize elevated session confirmation

riasvdv:
## Description

- Replace the legacy elevated-session prompt with a Vue host shared by Inertia and legacy control-panel pages.
- Centralize password-confirmation preflight and 423 retry handling across settings, GraphQL, permissions, and sign-in-provider actions.
- Add Laravel-native confirmed-password middleware and login rate limiting while preserving the legacy JavaScript APIs.
- Preserve alternative login-method HTML fragments and initialize their legacy UI assets after rendering.
- Fix the Testbench workbench SQLite, session, and asset setup used for local browser testing.


riasvdv:
See #19238

PR #19243, opened 2026-07-13
Cleanup `Volume::attributeLabels()`

mspirkov:
### Description

I was checking my PHPStan extension [mspirkov/yii2-phpstan-rules](https://github.com/mspirkov/yii2-phpstan-rules) on your project and found this error. It seems to be dead code that can be removed, as the `url` property in the `Volume` model does not exist, and I did not find any implicit use of this code anywhere.

image


brandonkelly:
Thanks!

PR #19248, opened 2026-07-14
[6.x] PHP UI components

brianjhanson:
A follow on to #19215 but the how is much more interesting than the what. 

This establishes a set of PHP UI Components that have a fluent API which somewhat closely resembles the way [Filament](https://filamentphp.com/) approaches this problem. These components are all backed by our web components in the `@craftcms/cp` package and have the same set of functionality. I've explored automatically generating these classes in https://github.com/craftcms/cms/pull/19032 but for the moment these are written by hand. I think someday we may be able to generate them, but it also may not prove to be worth the overhead. 

With new components joining the party, how we handle legacy code becomes a bit of a challenge. In an ideal world, all of the CP would use the new rendering path, but we don't to break all of the `forms.*` calls out there in the world. 

The solution in this PR is to add a bit of a translation layer in the `FormFields` class. Let's look at the button for example. 

If you have some twig code that loads the macro and creates a button, it probably looks like this.

```twig
{% import '_includes/forms.twig' as forms %}

{{ forms.button({
  name: 'myButton',
  id: 'my-button',
  label: "I'm a button!"|t('app'),
}) }}
```

Previously, `forms.button` would go through a few layers and end up at the `_includes/forms/button.twig` file to eventually output our `button` element. 

The change here is that the `_includes/forms/button.twig` internals have been replaced with `{{- craft.cp.button(_context)|raw -}}`. That CP function calls the `Cp::button()` method which calls `FormFields::buttonHtmlFromConfig()`, which takes the legacy configuration and maps it into the new `Button` PHP UI component. 

That general pattern is followed for ~10 initial components here. 

`FormFields::*FromConfig()` methods are established as the translation layer between the legacy configurations and the new UI components. 

The bulk of the work here is creating PHP components with a fluent API that can be used to render our web components from the PHP and twig. To facilitate the migration, our twig includes will now call a corresponding helper in `FormFields` which will map old config values to the new API and render the component. 

The other goal of this is to have all our render contexts covered off on ways to render these components. 

### Twig

```twig
{{ ui('button', {
  appearance: 'plain', 
  label: 'Advance'|t('app'), 
  icon: 'chevron-down' 
}) }}
```

### PHP
```php
Button::make()
    ->label(t('Advanced'))
    ->appearance(Appearance::Plain)
    ->icon('chevron-down')
    ->toHtml();
```

### Blade / HTML / Vue
```html
<craft-button appearance="plain" icon="chevron-down">{{ t('Advanced') }}</craft-button>
```

### Vanilla JS
```javascript
const button = document.createElement('craft-button');
button.appearance = 'plain';
button.icon = 'chevron-down';
button.innerText = 'Advanced';

document.body.appendChild(button);
```

## Component Updates

- The `<craft-action-menu/>` now has a `searchable` attribute that will show a search field and allow you to filter the available options. [Storybook](https://604ccc89.craftcms-ui-storybook.pages.dev/?path=/story/components-action-menu--searchable-data-driven)
- `<craft-button/>` will now apply better styling between icon options. In practical terms, that means you can display a button with only an icon either with `<craft-button icon="x" aria-label="Delete"></craft-button>` or `<craft-button icon><craft-icon name="x" label="Delete"></craft-icon></craft-button>`
- `<craft-checkbox-group/>`, `<craft-radio-group/>` added
- `<craft-disclosure/>` API and component has changed. It now inherits from [LionCollapsible](https://lion.js.org/components/collapsible/overview/)
- `<craft-field/>` component added for situations where you need a `label`, `help-text`, etc. but the input focused web components are too simplistic. 
- `<craft-reorder-button/>` now supports an `orientation` which will let you render the options as "Move forward" or "Move backward" instead of "Move up" and "Move down"

## Module updates
- GroupedEntryTypeManager ported
- ComponentSelect ported

### Related
https://github.com/craftcms/cms/pull/19247


brandonkelly:
> `FormFields::*FromConfig()` methods are established as the translation layer between the legacy configurations and the new UI components.

Should those methods be defined somewhere in `yii2-adapter` then? (And probably everything in `_includes` too?)

brianjhanson:
> > `FormFields::*FromConfig()` methods are established as the translation layer between the legacy configurations and the new UI components.
> 
> Should those methods be defined somewhere in `yii2-adapter` then? (And probably everything in `_includes` too?)

That's the goal, we're just not quite ready yet. We still have too much twig in core that has old configuration objects. Moving this into the adapter now would mean any `forms.*` twig calls would error when the `FormFields::*FromConfig` method is trying to be called. 

I figure that will be one of the many cleanup passes once we have the CP fully migrated to Inertia / Vue.

PR #19306, opened 2026-07-24
[6.x] Button single variant

brianjhanson:
Previously buttons has two customization axis. You could specify either a `variant` (`accent`, `neutral` or `danger`) and an `appearance` (`solid`, `outline`, `fill`, `plain`) to determine what your button looked like. 

After revisiting (and looking at [plugin kit](https://docs.verbb.io/plugin-kit/web/components/button#variants), I feel like that's a bit overkill. This flattens `appearance` and `variant` into a single `variant` property which can be one of: `primary`, `danger`, `danger-plain`, `solid`, `fill`, `outline`, `dashed`, `plain`, `link` or `none`.

CleanShot 2026-07-24 at 10 23 47@2x

`danger-plain` is the only one that bugs me a bit, but I think that wart is better than having the ability to create buttons that may only appear once in the UI. 

This also refactors the `appearance=inherit` property on buttons that will allow a button to take on the color of its colorable parent. That behavior is now created by specifying a single `inherit` value on the button.

CleanShot 2026-07-24 at 10 26 09@2x

In this mode `primary` and `danger` buttons retain their normal fill, but other buttons will try to blend in.


brianjhanson:

PR #19337, opened 2026-07-29
[6.x] Replace Imagine with Intervention Image

riasvdv:
### Description
Replaces `pixelandtonic/imagine` with Intervention Image 4 for raster image manipulation. 

Adds GD, Imagick, and optional Vips drivers; expands supported transform formats; and updates image capability detection, metadata handling, reporting, CI, and tests.


riasvdv:
@AugustMiller Should probably note somewhere that we now support libvips in addition to imagick and gd

PR #19341, opened 2026-07-29
[6.x] Inertia Passwords & Verification

brianjhanson:
Ports the per-user Password & Verification account screen (myaccount/password) to Inertia/Vue, along with the supporting CP UI infrastructure it needed.

- Password & Two-Step Verification page — renders through Inertia::render via UserPasswordViewModel. The Two-Step Verification section is a Vue AdminTable listing each method's status and actions; setup for every method (TOTP, Recovery Codes, and plugin-provided) is server-rendered through getSetupHtml() and hosted in the slideout, preserving the plugin extension point.
- InputCopy component — a new <craft-input-copy> web component (read-only input with an integrated copy-to-clipboard button) and its PHP InputCopy counterpart, wired into the copyText form field so the migrated templates delegate to it.
- Actionable mixin — extracts the action-running behaviour (the action property, click handling, loading/success/error state, and the action:change-state event) out of craft-action-item into a shared Lit mixin, and applies it to craft-button so <craft-button action=…> runs actions directly. runAction gains a POST→blob download primitive, used by the recovery-codes download button.
- Base slideout styles — moved from the legacy craftcms-legacy package into resources/css/slideout.css so the Inertia CP styles slideouts without depending on the (no-longer-loaded) legacy CP stylesheet.
- actionClient — fixes bare action-path resolution on multi-site installs, where the ?site= query on the action base URL was corrupted by naive base-URL concatenation.
- Markdown filter — dedents md/markdown filter input so indented {% apply md %} blocks parse as Markdown instead of being captured as indented code blocks.


brianjhanson:

PR #19346, opened 2026-07-30
[6.x] Move more action routes

riasvdv:
### Description

This is the last batch (for now). Inertia-ported controller methods aren't really "action" routes so porting them to the cp.php routes instead.


resources/js/modules/slideout/cp-screen-slideout.ts
riasvdv, line 1: @brianjhanson Slideouts weren't loading when an inertia response was returned, this fixes it but I don't know if that's what you'd expect or do

PR #19370, opened 2026-08-05
Group the Matrix block action menu’s “Add … above” items by entry type group

### LINKED ISSUE #17594: [5.x]: Entry type groups missing from actions menu
### Description

Currently, all entry types are listed without the group headings in a block's actions menu (to add a new block above).

### Steps to reproduce

1. Create a Matrix field with two entry type groups.
2. Edit an entry, create a Matrix block, and then click on its actions menu.

### Expected behavior

All group headings are also visible in the actions menu, like in the create discloure menu:

Image

### Actual behavior

All entry types are listed ungrouped:

Image

### Craft CMS version

5.8.5

### PHP version

_No response_

_No response_

### Database type and version

_No response_

### Image driver and version

_No response_

-

fabianhaef:
### Description

Fixes #17594.

In the `blocks` view mode, each Matrix block’s action menu appends one “Add {type} above” item per entry type, in one flat list — also reported in #17594. We have Matrix fields with 35+ entry types, and the menu becomes a long scrolling list where the only way to find a type is to read every row.

The field’s “+” button already solves this: `create-button.twig` groups the same entry types by their entry type group and shows a search input above five types. This PR brings the block action menu in line with it:

1. **Group the “Add {type} above” items by entry type group** — same `collect().groupBy()` and `{hr, padded: false}` / `{heading, padded: false, items}` structure as `create-button.twig`. Fields with a single group (or no groups set) render exactly as before.
2. **Show a search input above five entry types** (`withSearchInput: not staticEntries and entryTypes|length > 5`), same threshold as the create button. The search filters *all* menu items (Collapse, Delete, …), like other searchable disclosure menus — it’s a separate commit so it can be dropped if that’s not wanted.

Drive-by: the trailing `{hr: true}` is now only added when add items follow, removing a dangling separator at the end of the menu for static blocks (e.g. revision view).

### Testing

- 20 entry types across 6 groups: headings and separators render in field-settings order; activating a grouped item still inserts the block above it; arrow-key navigation skips headings.
- Searching narrows the menu and hides emptied group headings; activating a filtered item works.
- Single group / no groups set: unchanged flat menu.
- Revision view (`staticEntries`): no add items, no search input, no trailing separator.
- `maxEntries` reached: `DisclosureMenu.updateVisibility()` hides the emptied groups including their headings.

### DIFF
--- a/src/templates/_components/fieldtypes/Matrix/block.twig
+++ b/src/templates/_components/fieldtypes/Matrix/block.twig
@@ -143,19 +143,34 @@
   }) %}
 {% endif %}
 
-{% set actionMenuItems = actionMenuItems|push({hr: true}) %}
-
 {% if not staticEntries %}
-  {% for entryType in entryTypes %}
-    {% set actionMenuItems = actionMenuItems|push({
+  {% set groupedEntryTypes = collect(entryTypes).groupBy(et => (et.group ?? 'General'|t('app'))) %}
+  {% if groupedEntryTypes.count() > 1 %}
+    {% set actionMenuItems = actionMenuItems|merge(groupedEntryTypes.map((entryTypes, group) => [
+      {hr: true, padded: false},
+      {
+        heading: group|t('site'),
+        padded: false,
+        items: entryTypes|map(entryType => {
+          icon: entryType.icon ?? 'plus',
+          color: entryType.color,
+          label: 'Add {type} above'|t('app', {type: entryType.name|t('site')}),
+          attributes: {
+            data: {action: 'add', type: entryType.handle},
+          },
+        }),
+      },
+    ]).flatten(1).all()) %}
+  {% else %}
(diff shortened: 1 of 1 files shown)


fabianhaef:
There’s a follow-up discussion for #17594 in #17600 with a few community suggestions — how this PR relates to them:

- **Nested submenus per group (Notion-style):** Craft’s disclosure menus don’t support nested submenus today, so this PR groups inline with headings — the same pattern `create-button.twig` uses. If submenu support lands later (e.g. via the #17731 direction), the grouped structure here maps onto it 1:1.
- **“Delete” being hard to reach:** the add items sit below the standard actions, so grouping doesn’t change Delete’s position — and the search input (second commit) gets you to any item by typing a few letters.
- **Keeping the flat view:** fields without entry type groups render exactly as before, so the flat list stays available by simply not assigning groups. A dedicated per-field toggle would be a small addition if you’d prefer the choice to be explicit.

PR #19374, opened 2026-08-05
Fix legacy widget interface methods

lukeholder:
### Description

WIP


lukeholder:
Replaced by https://github.com/craftcms/cms/pull/19375

PR #19389, opened 2026-08-10
[6.x] Initial port of the entry edit page to Inertia

brianjhanson:
### Description

Ports the full-page entry edit screen to Inertia, rendering the field layout through the Form system.

`ElementEditViewModel` is the shared payload for element edit screens, mirroring `ContentIndexViewModel`: public methods are payload keys, and element types extend it with their own keys plus a `saveUrl()`. `EntryEditViewModel` and `content/Edit.vue` are the entry-specific pieces, while `useElementEditPage` and `ElementEditPage.vue` hold the shared client pipeline so the remaining element types only supply what their save action needs.

The field layout is compiled to a `FormPayload` through the same `FieldLayoutCompiler` the legacy editor and slideouts already use, and rendered by `FormRenderer`. Saving posts through Inertia to the existing `entries/save-entry` action — the Form payload submits ordinary nested input names, so the save controller reads it without a translation layer.

Drafts, revisions, and provisional drafts still fall through to `EditElementController`, keeping their notices and apply/revert controls intact until those screens are ported later in this stack. That controller also continues to serve slideouts and the element types that haven't been ported.

The sidebar meta fields remain a server-rendered island here and move onto the Form system in #19390.

### DIFF
--- a/resources/js/modules/elements/components/ElementEditPage.vue
+++ b/resources/js/modules/elements/components/ElementEditPage.vue
@@ -1,0 +1,75 @@
+<script setup lang="ts">
+  import {t} from '@craftcms/ui';
+  import DynamicHtmlRenderer from '@/common/components/DynamicHtmlRenderer.vue';
+  import LayoutSlot from '@/common/components/LayoutSlot.vue';
+  import Pane from '@/common/components/Pane.vue';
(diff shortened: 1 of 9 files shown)


brianjhanson:

PR #19390, opened 2026-08-10
[6.x] Render element editor meta fields as a Form

brianjhanson:
### Description

Moves the editor sidebar's meta fields onto the Form system, so they submit as Inertia form state instead of being read out of the DOM at save time.

Adds `ElementInterface::sidebarForm()`, implemented on `HasControlPanelUI` with `metaFieldsNodes()` as the per-element-type extension point — the Form-system counterpart to `getSidebarHtml()` and `metaFieldsHtml()`, which stay in place for the legacy editor and slideouts. `Entry::metaFieldsNodes()` covers entry type, slug, parent, authors, post date, and expiry date under the same visibility rules as `metaFieldsHtml()`; the base implementation contributes the status and notes fields.

The sidebar renders as a second `FormRenderer` bridged into the same Inertia form as the field layout. Two root-scoped `useInertiaFormRenderer` bridges can share one form because each instance only clears the root keys it wrote itself, and both are constructed before either receives a mutation — `useInertiaFormRenderer.test.ts` pins that behavior. With every value now in Inertia state, the unsaved-changes guard reduces to `form.isDirty`.

Per-site statuses are not included. The expand control behind them is wired up by `Craft.ElementEditor`, which never ran on the Inertia screen, so it lands with the multi-site work later in the stack.

### DIFF
--- a/resources/js/modules/elements/components/ElementEditPage.vue
+++ b/resources/js/modules/elements/components/ElementEditPage.vue
@@ -20,10 +20,13 @@
     form,
     formPayload,
     onMutation,
+    onSidebarMutation,
     props: payload,
     renderer,
     save,
-    sidebarEl,
+    sidebarErrors,
+    sidebarPayload,
+    sidebarRenderer,
   } = useElementEditPage({saveData: props.saveData});
 
   useAppLayout(() => ({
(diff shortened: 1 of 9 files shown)


brianjhanson:

PR #19395, opened 2026-08-11
[6.x] Add multi-site controls to the Inertia element editor

brianjhanson:
### Description

Restores the element editor's multi-site controls on the Inertia screen.

Breadcrumbs now lead with a site crumb whose menu links to the same element on every other editable site it propagates to, carrying the current query params across so switching sites doesn't drop the draft, revision, or return URL. The crumb is omitted for non-localized elements and anywhere fewer than two editable sites apply.

The sidebar's status control follows the legacy editor: with two or more editable sites it renders an "Enabled for all sites" switch plus a per-site switch for each of them, collapsed into a group. Sites the element hasn't propagated to yet are absent from `siteStatusesForElement()`, so they default to enabled the same way the legacy editor defaults them. A single editable site keeps the plain `enabled` switch.

`useSiteStatuses` reconciles the two directions: the global switch applies its value to every site, and per-site changes roll back up — `true` when all are on, `false` when none are, indeterminate when they disagree. Because both are ordinary Form Controls writing into the shared Inertia form, this is value reconciliation with no DOM involved, unlike the lightswitch bookkeeping the legacy editor needs.

Copying a field's value to other sites is not included; it hangs off per-field controls rather than the sidebar and is better handled with the field-level work.

### DIFF
--- a/resources/js/modules/elements/composables/useElementEditPage.ts
+++ b/resources/js/modules/elements/composables/useElementEditPage.ts
@@ -4,6 +4,7 @@ import {t} from '@craftcms/ui';
 import {computed, onBeforeUnmount} from 'vue';
 import type {FormPayload} from '@/modules/forms/types';
 import {useInertiaFormRenderer} from '@/modules/forms/useInertiaFormRenderer';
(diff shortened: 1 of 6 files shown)


brianjhanson:

PR #19396, opened 2026-08-11
[6.x] Render provisional drafts and autosave in the Inertia editor

brianjhanson:
### Description

Renders provisional drafts in the Inertia editor and autosaves into them, which is what makes the port usable: the legacy editor creates a provisional draft the moment anyone types, so before this every previously-edited entry fell through to the old screen.

`EditEntryController` now only defers to `EditElementController` for named drafts and revisions. Provisional drafts render here, with a notice and a Discard control, and `saveUrl`/`applyDraftUrl` let the client pick its target at submit time — autosave can create a draft partway through editing a canonical element, and from that point saving means applying the draft rather than saving the element underneath it.

`useElementAutosave` posts to the shared `elements/save-draft` action, debounced and serialized so a burst of typing collapses into one trailing request. It is driven by the Form renderers' change callbacks rather than a watch on the Inertia form, because the form starts empty and gains its keys dynamically. Re-baselining after a save is suspended so it can't recreate the draft the save just consumed, and the draft id resyncs from the server after each visit. With edits safe in a draft, the unsaved-changes prompt now only fires when autosave hasn't caught up.

Also fixes a save-blocking bug from the meta-fields PR: the `DateTime` control defaults to a 30-minute increment, so an entry whose post time fell between steps failed native form validation and the Save button silently did nothing. The legacy editor never hit this because it submits through JS. Post and expiry dates now accept the stored precision.

### DIFF
--- a/resources/js/modules/elements/components/ElementEditPage.vue
+++ b/resources/js/modules/elements/components/ElementEditPage.vue
@@ -1,5 +1,6 @@
 <script setup lang="ts">
   import {t} from '@craftcms/ui';
+  import {computed} from 'vue';
   import DynamicHtmlRenderer from '@/common/components/DynamicHtmlRenderer.vue';
   import LayoutSlot from '@/common/components/LayoutSlot.vue';
   import Pane from '@/common/components/Pane.vue';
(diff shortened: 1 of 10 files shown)


brianjhanson:

PR #19397, opened 2026-08-11
[6.x] Add header and alternate save actions to the Inertia editor

brianjhanson:
### Description

Adds the header buttons and the alternate save actions beside the Save button.

`headerActions()` covers "Create a draft" (a canonical element the user may branch, posting to `elements/save-draft` with `dropProvisional` so a named draft is created rather than the provisional one being promoted) and "Save as a new …" for users who can't save the element but may duplicate it. `formActions()` serializes `getAltActions()` into the Save button's menu — Save and continue editing, Save and add another, Save as a new … — with redirects encrypted the way the save controllers expect to decrypt them.

Both render through `FormActions`, which already supported additional buttons and action items. The layout's own default action is switched off for this screen, since the element supplies its own "Save and continue editing" and the two would otherwise both appear.

`submitAction()` routes every one of these through the existing save pipeline rather than reimplementing submission per action: a pending-action reference overrides the target URL and merges the action's params for exactly one submission, so they inherit the elevated-session handling, error plumbing, and processing state already in place. Action params take precedence over the provisional-draft targeting, which is what lets "Create a draft" opt out of it.

Actions posting to a shared `elements/*` endpoint now carry the generic identity params those endpoints resolve by. The type-specific save controllers key off their own params, so a request built for one is not understood by the other — this previously surfaced as a 400 when applying a draft, and again here.

Applying a named draft and reverting a revision belong beside these buttons, but those screens still render through the legacy editor, so they arrive with the draft and revision work rather than as buttons nothing can reach.

### DIFF
--- a/resources/js/modules/elements/components/ElementEditPage.vue
+++ b/resources/js/modules/elements/components/ElementEditPage.vue
@@ -30,7 +30,24 @@
     sidebarErrors,
     sidebarPayload,
     sidebarRenderer,
+    submitAction,
   } = useElementEditPage({saveData: props.saveData});
+
+  // Alternate saves in the Save button's menu, and the buttons beside it.
+  const formActionItems = computed(() =>
+    payload.formActions.map((action) => ({
+      label: action.label,
+      onClick: () => submitAction(action),
+    }))
+  );
+
+  const headerButtons = computed(() =>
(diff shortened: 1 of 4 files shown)


brianjhanson:

PR #19399, opened 2026-08-11
[6.x] Render named drafts and revisions in the Inertia editor

brianjhanson:
### Description

Renders named drafts and revisions in the Inertia editor, removing the last fallback to the legacy `EditElementController` for entries.

Revisions render read-only with the notice explaining why, and offer "Revert content from this revision". Named drafts get "Apply draft" and a Save button labelled for what it saves — "Save draft" for a draft, "Create entry" for an unpublished one — which required a `submitButtonLabel` pass-through on the layout and form actions. An outdated draft still merges the canonical element's newer changes on the way in, and now says so above the editor; that logic moved back out of the legacy controller along with everything else.

The drafts-and-revisions switcher returns beside the breadcrumbs. Its groups arrive flattened, with headings carried as `display` items, because the action menu's item contract has no nested-group shape.

Also fixes a bug that predates this PR: `provisional` was sent on every draft-targeted request once a draft existed. The server narrows its draft lookup when that param is present, so a named draft stopped resolving — applying one appeared to succeed and silently did nothing. It's now sent only when the draft really is provisional, or when the request is the one creating it, matching what the legacy editor does.

### DIFF
--- a/resources/js/common/components/FormActions.vue
+++ b/resources/js/common/components/FormActions.vue
@@ -11,6 +11,8 @@
     actionItems?: Array<ActionItem>;
     additionalActions?: Array<ActionItem>;
     additionalButtons?: Array<ActionItemButton>;
+    /** Overrides the submit button's text (e.g. "Save draft"). */
+    submitLabel?: string;
     readOnly?: boolean;
   }>();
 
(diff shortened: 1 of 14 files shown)


brianjhanson:

PR #19403, opened 2026-08-11
[6.x] Render the element action menu in the Inertia editor

brianjhanson:
### Description

Brings the element action menu to the Inertia editor: Validate, Copy, the entry type and section settings slideouts, Delete for this site, and Delete.

`actionMenuDescriptors()` is the counterpart to `getActionMenuItems()` — the same actions, but each item names what it does rather than pairing markup with an inline script. The client dispatches those behaviors itself, so nothing depends on registered jQuery handlers. Element types extend it through `extraActionMenuDescriptors()` the way they extend the HTML items; entries add their settings slideouts there, and the entry type slideout still follows the sidebar's unsaved selection rather than the stored value.

Deletion continues to run through `ElementDeletionManager`, so blocking relations and references can be reassigned before the element goes. The manager was already ported to TypeScript, so the Vue handler constructs it directly.

"Edit" is never offered — this is the edit screen — and "View in a new tab" drops out once preview targets exist, matching the legacy rules.

Custom icons are qualified with their family. The action menu's icon renderer resolves a bare name against the default set, so `clone-dashed` was 404ing for the Copy item until its `custom-icons/` prefix was carried through.

Plugin-provided items still come through the HTML pairing and aren't rendered here yet; that contract is unchanged and needs its own decision.

### DIFF
--- a/resources/js/modules/elements/components/ElementEditPage.vue
+++ b/resources/js/modules/elements/components/ElementEditPage.vue
@@ -8,6 +8,7 @@
   import {useAppLayout} from '@/common/composables/useAppLayout';
   import FormRenderer from '@/modules/forms/FormRenderer.vue';
   import {useElementEditPage} from '@/modules/elements/composables/useElementEditPage';
+  import {useElementActionMenu} from '@/modules/elements/composables/useElementActionMenu';
 
   const props = defineProps<{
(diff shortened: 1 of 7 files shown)


brianjhanson:

PR #19404, opened 2026-08-11
[6.x] Add preview target links to the Inertia editor

brianjhanson:
### Description

Adds the View control to the Inertia editor, linking each of the element's preview targets at the front end.

The legacy editor mints preview tokens from JavaScript, juggling a token queue and rewriting hrefs once a token is live. Every input to that is known server-side, so the links arrive ready to follow instead: a live element points at its own URL, and anything not publicly visible — a draft, a disabled entry, a future post date — points at `preview/create-token`, which mints the token and redirects to the tokenized URL. The controller grants the matching session authorization, as the legacy screen did.

`PreviewController::createToken()` read its params from the request body only. Craft 5 read them with `getParam()`, i.e. query or body, which is what the View links depend on — they're plain GET hrefs. It now reads either, taking only the token's own keys, since a control panel URL carries others (`site`) that `RouteToken` rejects outright.

Live Preview itself is not included. `Craft.Preview` is built around a `Craft.ElementEditor` instance — it reads its settings, serializes its form, and asks it for tokenized URLs — and none of that exists on this screen. Wiring it up means writing an adapter that impersonates the legacy editor, which is worth doing deliberately rather than as a side effect of this change.

### DIFF
--- a/resources/js/modules/elements/components/ElementEditPage.vue
+++ b/resources/js/modules/elements/components/ElementEditPage.vue
@@ -59,6 +59,22 @@
     currentEntryTypeId: () => form.typeId,
   });
 
+  // "View" opens the element on the front end. The hrefs arrive ready to
+  // follow — a live element points at its own URL, anything else at a
+  // token-minting redirect that lands on the tokenized preview.
+  //
+  // These ride in `formAdditionalButtons` because the layout's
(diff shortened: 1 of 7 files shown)


brianjhanson:

PR #19405, opened 2026-08-11
[6.x] Poll for element activity and upstream changes in the Inertia editor

brianjhanson:
### Description

Adds collaborative awareness to the Inertia editor: who else is working on the element, and whether it has changed underneath the person editing it.

`useElementActivity` polls `elements/recent-activity` on the same fifteen-second cadence the legacy editor uses, rendering each active user's avatar beside the save controls with their activity message. When the element's or its canonical's last-modified stamp comes back different from what the screen rendered with, a notice offers to reload — naming the draft when the draft itself moved, and the element type otherwise.

Polling pauses while the tab is hidden and re-reads on return. The legacy editor keeps polling regardless, but a backgrounded tab has nobody to show avatars to, and the timestamps are compared fresh on the next visible poll either way. A failed poll is swallowed: this is ambient information nobody asked for, and the next tick may well succeed.

Revisions don't poll. They're read-only, so there is nothing to collaborate on and nothing to go stale.

### DIFF
--- a/resources/js/modules/elements/components/ElementEditPage.vue
+++ b/resources/js/modules/elements/components/ElementEditPage.vue
@@ -1,6 +1,7 @@
 <script setup lang="ts">
   import {t} from '@craftcms/ui';
   import {computed} from 'vue';
+  import {router} from '@inertiajs/vue3';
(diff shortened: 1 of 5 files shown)


brianjhanson:

PR #19406, opened 2026-08-11
[6.x] Add dismissible tips and additional-site statuses to the Inertia editor

brianjhanson:
### Description

Two remaining pieces of the element editor.

Dismissible tips now dismiss. `CalloutNode` renders a dismiss control for tips marked dismissible and remembers them by layout element UID, under the same local storage key the legacy editor uses, so a tip dismissed on either editor stays dismissed on both. Storage failures are swallowed — dismissal is a convenience, not something worth interrupting anyone over. The payload already carried `dismissible`; nothing acted on it.

The sidebar's site statuses now also cover supported sites the element does not propagate to, switched off, so the element can be added to them. The legacy editor hides these behind an "Add a site…" select that builds fields on demand; listing them reaches the same outcome without the dynamic DOM work.

Note that entries don't currently produce non-propagating supported sites — a section site either propagates or isn't supported at all — so that path is dormant for them and matters only for element types whose `getSupportedSites()` marks a site as non-propagating. The accompanying test covers the ordinary multi-site case rather than claiming to exercise it.

Copying a field's value from another site is not included, and isn't a gap in this editor: Craft 5 renders a "Copy value from site…" item into each translatable field's action menu, and the Craft 6 port never carried it over, so the feature is unreachable on the legacy editor too. `elements/copy-values-from-site` still exists and still works. Restoring it means giving Form-rendered fields an action menu, which is a Form system decision rather than an element editor one.

### DIFF
--- a/resources/js/modules/forms/CalloutNode.vue
+++ b/resources/js/modules/forms/CalloutNode.vue
@@ -1,7 +1,10 @@
 <script setup lang="ts">
+  import {t} from '@craftcms/ui';
+  import {ref} from 'vue';
+  import {dismissTip, isTipDismissed} from './dismissedTips';
   import type {FormNodePayload} from './types';
 
-  defineProps<{
+  const props = defineProps<{
     node: FormNodePayload<{
       html: string;
       variant: string;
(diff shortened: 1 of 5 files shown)


brianjhanson:

PR #19412, opened 2026-08-11
[6.x] Port User Group settings to Form

riasvdv:
### Description

Ports the User Group settings screens to the Form system.


brianjhanson (approved):
We could take this opportunity to create a view model for `UserGroupsController` but that can also come later.

PR #19420, opened 2026-08-11
[6.x] Port the asset edit page to Inertia

brianjhanson:
### Description

Ports the asset edit screen onto the shared Inertia element editor, replacing the `CpScreenResponse` the generic `EditElementController` produced for it. `assets/edit/{id}{slug}` now routes to `EditAssetController`, which renders `assets/Edit` from an `AssetEditViewModel`.

Assets have no drafts, revisions, or editable status, so most of the editor's machinery stays dormant here. What's left is the compiled field layout (title, alt text, custom fields), the Filename meta field, the file preview, and the asset's own actions. Saving posts to the generic `elements/save` — assets have no store action of their own — resolved from the identity attributes every element edit screen already submits.

`Asset::metaFieldsNodes()` renders Filename as a Form Control alongside the existing `metaFieldsHtml()`. Renaming validates the folder path and the filename together as `newLocation`, but the field posts `newFilename`, so the messages have to move to the name the Control answers to or they never reach the field that produced them. That remapping is a new `ElementInterface::formErrors()` hook, defaulting to the raw error bag; `ElementResponse::failure()` uses it for the Inertia branch only, since the legacy editor and the slideouts still apply errors by the validated attribute.

The asset's own actions — preview, download, show in folder, replace file, image editor, volume and filesystem settings — arrive as behavior descriptors rather than markup plus an inline script. The three that hand off to a legacy modal or uploader reload the screen afterwards rather than patching the file's details into the DOM, which is both shorter and more correct than the legacy JS that hand-edits the filename input, file size, dimensions, and thumbnail.

Two fixes to shared editor code, both surfaced by this screen because an in-place save with no redirect is the normal path for an element with no drafts:

- `useElementEditPage` captured `page.props` once. It's a computed, so the payload froze at whatever the screen first rendered with — the title, notices, and timestamps never updated after saving in place. Now read through `toReactive`.
- The activity poller compared against the timestamps it was constructed with, so a successful save reported our own write as someone else's change. It re-baselines on save.

`canAutosave` now also requires the element type to have drafts. Autosaving means writing a provisional draft; without one there's nothing to write to.

The file preview is still server-rendered HTML. Previewing and image editing both open legacy modals and the markup carries the JS that wires them up, so it arrives as a fragment rather than something the Vue side rebuilds.

### DIFF
--- a/resources/js/common/types/globals.d.ts
+++ b/resources/js/common/types/globals.d.ts
@@ -148,7 +148,12 @@ interface CraftStatic {
       message?: string | CpNotificationSettings,
       settings?: CpNotificationSettings
     ) => object;
+    displayNotice?: (
+      message?: string,
+      settings?: CpNotificationSettings
+    ) => object;
   };
+  broadcaster?: {postMessage(message: Record<string, unknown>): void};
   defaultIndexCriteria: Record<string, any>;
   systemUid?: string;
   canAccessQueueManager?: boolean;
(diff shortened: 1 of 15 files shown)


brianjhanson:

PR #19421, opened 2026-08-11
[6.x] Port the user edit page to Inertia

brianjhanson:
### Description

Ports the account Profile screen onto the shared Inertia element editor. Both `myaccount` and `users/{userId}` now render `users/Edit` from a `UserEditViewModel`, replacing two different things: the DOM-scraping `users/Profile` page for the current user, and the `EditElementController` `CpScreenResponse` for everyone else. `UserProfileViewModel` and `Profile.vue` are removed with them.

The account's native fields — username, full name, email, photo, affiliated site — are already native field-layout elements with Form Controls, so the compiled field layout covers the whole form. Saving posts to `users/save-user`, which the generic element save can't stand in for: it refuses the sensitive attributes a user account is mostly made of.

The screen list and the account navigation move out of `EditUserTrait` into an `EditUserScreens` service, since the ported screen and the sibling `CpScreenResponse` screens (Permissions, Preferences, Addresses, Password, Passkeys, Sign-in Providers) both need them. The trait keeps its behavior and delegates; the screen-name constants now live on the service.

`User::showStatusField()` returns false. `statusFieldHtml()` already rendered nothing — a user's status follows from the account actions, not an editable switch — but the Form path had no equivalent signal and would have rendered a Status lightswitch the legacy screen never showed.

The account actions (activate, suspend, unsuspend, unlock, deactivate, the password-reset emails and requirements, impersonation) arrive as behavior descriptors. Two need dispatch the client didn't have:

- `submit` takes `requireElevatedSession`, for signing in as another user.
- `copyUrl` fetches a single-use URL behind an elevated session and offers it for copying — the activation, password-reset, and impersonation URLs, which grant access to the account.

`redirectUrl` is encrypted, since the save controllers decrypt whatever `redirect` a request posts rather than trusting it. A brand-new account still lands on its permissions after being created.

Two of the tests this replaces were passing vacuously: `action([UsersController::class, 'edit'])` resolves to `myaccount`, which is registered first, so "for other users" was never actually rendering another user's screen, and the `assertDontSee('users/Profile')` it asserted would have passed regardless against the escaped page JSON. They now address the screen by URL.


brianjhanson:

PR #19427, opened 2026-08-12
[6.x] Add text expander

riasvdv:
### Description

Adds a generic `craft-text-expander` component for trigger-based autocomplete on text, textarea, and Markdown controls, with static options and same-origin asynchronous action sources.

image


brianjhanson (commented):
Pretty excited about this. It feels like this could be a better option than our combobox for a lot of scenarios. Usually we accept either arbitrary text or a selected option which combobox has struggled with a bit. 

@gcamacho079 I'm curious what the accessibility implications of something like this would be.

packages/craftcms-ui/src/components/text-expander/text-expander.ts
```diff
@@ -0,0 +1,770 @@
...
+  }
+
+  #bindTarget(): void {
+    const target = this.#resolveTarget();
+    this.#unbindTarget();
+
+    if (!isTextTarget(target)) {
+      throw new TypeError(
+        'craft-text-expander requires an input[type="text"] or textarea target.'
+      );
```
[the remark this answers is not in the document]
riasvdv: Confirmed independently with a regression that reproduces the Markdown host ID handoff after the expander’s first update. The expander now observes its DOM root until the referenced native target appears, then binds and stops observing.

packages/craftcms-ui/src/components/text-expander/text-expander.ts
brianjhanson, line 1: Is there a reason the `TextExpander` doesn't use `<craft-popup/>` or `<craft-listbox/>` directly? I'm guessing a lot of functionality in here is duplicated from those components.

riasvdv: `craft-popover` doesn't work as it currently requires an element and the text expander one is positioned based on the caret position (virtually calculated using `dom-input-range`). The text expander also keeps focus in the input, which popover also takes away right now.

I didn't consider listbox as it's in the CP module and not in craftcms/ui, but this is what the 🤖 says:

- Uses buttons and `aria-pressed`, not `role="option"` and `aria-selected`.
- Moves focus among buttons instead of retaining focus in the text input with `aria-activedescendant`.
- Only handles click selection; it lacks combobox keyboard navigation.
- Discovers a static set of options during construction, while suggestions are replaced on every query.
- Lives in the CP application, so `@craftcms/ui` cannot depend on it without reversing the package dependency.

So using either would require a lot of changes to the existing components

brianjhanson: Ah, I forgot that our currently listbox is a port from garnish and not an extension of the Lion listbox. I feel like it will make sense to have a `craft-listbox` that handles all of these kinds of things (responsible for the keyboard navigation, etc.) but that can come in a bit later as a separate effort. <br><br>I also feel like this *should* be a `craft-popover` (making whatever changes we need there to make it so) but that can be a separate effort anyways. I'm just thinking that from a functionality standpoint, they accomplish the same goals (showing some overlay content relative to a specific position).

riasvdv: I'll look into `craft-popover` 👍

gcamacho079:
@riasvdv @brianjhanson I did some poking around to figure out how we would make something like this accessible to screen reader users. To do this accessibly, here are the primary things that will need to be taken into account:

- The input MUST remain `<input type="text">` or `<textarea>`, since the listbox selection applies only to a _portion_ of the input value. This shouldn’t be a _replacement_ for a combobox component, but rather a feature added to text-type components.
- The options MUST be shown via a listbox, with keyboard support and attributes matching those [documented in the APG docs](https://www.w3.org/WAI/ARIA/apg/patterns/listbox/examples/listbox-scrollable/)
- Focus MUST remain on the text input when the listbox is opened
- Since we don’t have the built-in screen reader support you get when combining a `combobox` input with a `listbox`, we MUST be managing status updates via a live region. For example, typing the “@” and having the listbox display should be matched with a screen reader announcement: “@(username), 1 of (number) options”, and changing the `aria-activedescendant` should also trigger similar announcements.

riasvdv:
@gcamacho079  looks like the first three were already ok, added extra announcements for that last bullet point

a lot of this implementation was referenced from GitHub’s text expander

gcamacho079:
My go-to machine for SR testing is in the middle of long-running updates ⏳, but I can test it after that. If I’m the blocker, I can add it to my testing queue post-merge.

packages/craftcms-ui/src/components/text-expander/text-expander.ts
```diff
@@ -0,0 +1,826 @@
...
+
+// Queries may contain Unicode letters or numbers, underscores, and hyphens;
+// any other character terminates the active query.
+const queryPattern = /^[\p{L}\p{N}_-]*$/u;
+const defaultListboxLabel = t('Suggestions');
+// Label announced for the suggestion listbox, keyed by trigger character.
+// Add an entry here whenever a new trigger character is introduced.
+const triggerLabels: Readonly<Record<string, string>> = {
+  '#': t('Environment'),
+  '@': t('Users'),
```
riasvdv, line 72: @gcamacho079 i don’t think we should hard code these like that, `@` could be users in one context but we also use it for Aliases in some comboboxes currently.

brandonkelly: Doh, missed this comment before merging! Agree, this is something that should be configured.

riasvdv: https://github.com/craftcms/cms/pull/19440

PR #19443, opened 2026-08-16
Improve queue job reservation performance

### LINKED ISSUE #19097: [5.x]: Queue with lots of large jobs degrades performance
### Description
When the queue contains a large number of jobs and the job column is large, the queue performance slows.

I believe this is due to the reserve method in the queue service sorting by priority, but the queue table not having an index for the priority column.

To test this, I created a dummy job class with an empty execute method and a large (200kb+) string parameter.

Obviously this is an extreme example, but we initially noticed this with Commerce jobs for order emails & catalog pricing.

When I add a composite index for channel, priority and id, performance is restored.

### Steps to reproduce

1. Push a large number of jobs into the queue, where the job column blob is also relatively large
2. Observe query to reserve the job taking a longer amount of time

### Expected behavior
Queue runs fast

### Actual behavior
Queue runs slower than it could do

### Craft CMS version

5.10.5

### PHP version

8.5

_No response_

### Database type and version

MySQL 8.4.7

### Image driver and version

_No response_

_No response_

atirna:
## Why

The queue reserve query filters by channel, fail, and `timeUpdated`, then orders by priority and ID. This adds an index for that access pattern so it can find the next runnable job without sorting through large job payloads.

## Verification

- Ran the reserve query against a MySQL 8.4 queue with 20,000 8 KiB jobs: `EXPLAIN ANALYZE` went from a 541 ms filesort to a 0.008 ms ordered index lookup.
- Ran `php -l` on the changed PHP files and `git diff --check`.

Fixes #19097

### DIFF
--- a/CHANGELOG-5.11.md
+++ b/CHANGELOG-5.11.md
@@ -17,6 +17,7 @@
 ### System
 - Added support for `.well-known/passkey-endpoints` requests. ([#19364](https://github.com/craftcms/cms/pull/19364))
 - The front-end login page no longer returns a redirect response for logged-in users, if it’s a preview request. ([#19360](https://github.com/craftcms/cms/discussions/19360))
+- Improved queue job reservation performance when queues contain many large jobs. ([#19097](https://github.com/craftcms/cms/issues/19097))
 - Updated Twig to 3.28.
 - Updated yii2-debug to 2.1.28.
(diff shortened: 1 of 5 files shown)


brandonkelly:
Thank you! Merged for Craft 5.11.

brandonkelly:
Craft 5.11.0 is out with this change. Thanks again!

PR #19454, opened 2026-08-18
[6.x] Move alias support to Yii adapter

riasvdv:
### Description

Removes path alias support and the alias dependency from the core codebase, replacing core usages with Laravel helpers and the Path service. Moves legacy configuration, helper, Twig, plugin, Composer, and icon alias compatibility into the Yii 2 adapter, and updates affected settings copy and tests.

### Breaking change

The core UI fields that explicitly advertised or suggested aliases were:

| Page | Fields |
|---|---|
| Installer | Base URL |
| Settings → Sites | Base URL |
| Settings → Filesystems | Base URL |
| Settings → Filesystems → Local Folder | Base Path |

Several additional fields implicitly accepted aliases because their values passed through `Env::parse()`:

- Settings → General: System Name, System Status, Time Zone
- Settings → Sites: Name, Status, Base URL, Language
- Settings → Site Groups: Name
- Settings → Email: System Email Address, Sender Name, Reply-To Address, HTML Email Template, Mailer, and corresponding site overrides
- Settings → Assets → Volumes: Asset Filesystem, Subpath, Transform Filesystem, Transform Subpath
- Section/entry-type preview targets: URL Format
- Plugin license keys
- General config-backed fields such as `compiledTemplatesPath`, `cpIconUrl`, `brokenImagePath`, and `tempAssetUploadFs`

Third-party settings fields also lose alias suggestions if they use `suggestAliases`, `cpEnvSuggestions(true, …)`, or `getEnvSuggestions(true, …)`.


riasvdv:
@brandonkelly Adding just two methods to the `Path` service and replacing everything else with Laravel helpers seems to cover all of our own Alias usage

This does break sites that use Aliases in a way we can't easily migrate automatically or intercept using the yii2-adapter so will require extra work from developers.

riasvdv:
Not removing aliases, will remove `@web`

PR #19464, opened 2026-08-20
[6.x] Asset transformers

riasvdv:
See https://github.com/craftcms/cms/blob/feature/asset-transformers/docs/asset-transformers.md


docs/asset-processors.md
```diff
@@ -0,0 +1,356 @@
+# Asset Processors
+
+Asset Processors turn an Asset and an Asset Transform definition into a rendition. A rendition can be an image generated
+by Craft, a URL produced by an external image service, or another representation supplied by a plugin.
+
+The system separates two concepts:
+
+- An **Asset Processor** is a project-configured profile. It has a name, handle, driver, and driver-specific settings.
+- An **Asset Processor driver** is the PHP implementation that validates its supported operations and produces rendition
```
timkelty: "rendition" feels awkward to me…not inaccurate, just odd. Does the term end up in code?

docs/asset-processors.md
```diff
@@ -0,0 +1,356 @@
...
+Craft selects one processor for every transform request, in this order:
+
+1. The `processor` handle in the transform definition.
+2. The processor assigned to the Asset’s volume.
+3. The [`defaultAssetProcessor`](../src/Config/GeneralConfig.php) general config setting, which defaults to `craft`.
+
+An explicit processor can be combined with an inline transform:
+
+```php
+$result = $asset->transform([
```
timkelty: If we're renaming this to "processor"s…shouldn't be consistent?

docs/asset-processors.md
```diff
@@ -0,0 +1,356 @@
...
+    'processor' => 'remote-images',
+    'width' => 1200,
+    'format' => 'webp',
+]);
```
timkelty: what about considering splitting this up, so `processor` isn't flattened with transform options?

```
$result = $asset->transform([
    'width' => 1200,
    'format' => 'webp',
])->withProcessor('remote-images')
```

docs/asset-processors.md
```diff
@@ -0,0 +1,356 @@
...
+```
+
+The same override is available through `getUrl()` and Twig:
+
+```twig
+<img src="{{ asset.getUrl({
+    processor: 'remote-images',
+    width: 1200,
+    format: 'webp',
+}) }}" alt="">
```
timkelty: I would argue this should be deprecated, legacy-compatibility only.

Overloading the args of `getUrl` gets messy, esp. when we're defining the output of these as not necessarily url.

docs/asset-processors.md
```diff
@@ -0,0 +1,356 @@
...
+| `height` | Integer greater than zero |
+| `interlace` | An `ImageTransformInterlace` value |
+| `mode` | An `ImageTransformMode` value |
+| `position` | An `ImageTransformPosition` value |
+| `quality` | Integer from 1 through 100 |
+| `upscale` | Boolean |
+| `width` | Integer greater than zero |
+
+A driver can declare further operations. Unknown operations are removed, invalid values cause an
+`InvalidAssetTransformException`, and the driver receives operations sorted by handle.
```
timkelty: What if a driver wanted to accept _anything_?, i.e. defer validation to the cloud processor

That can make sense for a remote processor that may add or change option frequently. Otherwise, you're stuck waiting for the model to add the prop.

docs/asset-processors.md
timkelty: "operations" also feels weird - options or settings?

docs/asset-processors.md
```diff
@@ -0,0 +1,356 @@
...
+
+Named Image Transforms store custom operation values against the processor UUID. This keeps settings for processors that
+use different operation sets separate, even if the processors share a driver.
+
+## Immediate and deferred generation
+
+`Asset::transform()` and `Asset::getUrl()` accept an optional `immediately` argument:
+
+```php
+$result = $asset->transform(['width' => 800], immediately: true);
```
timkelty: I’m not convinced `immediately` belongs in the shared transform request. It isn’t portable across processors, and the spec already permits on-demand processors to treat both values identically.

Why not make it a processor-level setting? The Craft processor could implement the existing `generateTransformsBeforePageLoad` behavior, while other processors could define the generation strategy appropriate for their service.

docs/asset-processors.md
```diff
@@ -0,0 +1,356 @@
...
+production driver should use the service’s authenticated source-upload or signing flow when source Assets can be private.
+
+### The driver definition
+
+`AssetProcessorDriverDefinition` describes the driver’s Control Panel and validation surface:
+
+- `name` is the human-readable driver name.
+- `operations` maps custom operation handles to Laravel validation rules.
+- `settings` contains `Field` nodes for configuring each processor profile.
+- `operationFields` maps operation handles to `Field` nodes shown in named Image Transform forms.
```
timkelty: operations/operationFields feels awkward

docs/asset-processors.md
```diff
@@ -0,0 +1,356 @@
...
+
+Core operations are already available to every driver and should not be redeclared. A custom operation that uses a core
+handle must use exactly the core validation rules; conflicting rules cause an `InvalidAssetTransformException`.
```
timkelty: What about a scenario where validation inherently conflicts?

E.g. `quality` is an `int` for core, but for Cloudflare it is `int | 'high' | 'medium-high' | 'medium-low' | 'low'`.

docs/asset-processors.md
```diff
@@ -0,0 +1,356 @@
...
+
+1. The `processor` handle in the transform definition.
+2. The processor assigned to the Asset’s volume.
+3. The [`defaultAssetProcessor`](../src/Config/GeneralConfig.php) general config setting, which defaults to `craft`.
+
+An explicit processor can be combined with an inline transform:
+
+```php
+$result = $asset->transform([
+    'processor' => 'remote-images',
```
timkelty: what about considering splitting this up, so `processor` isn't flattened with transform options?

```
$result = $asset->transform([
    'width' => 1200,
    'format' => 'webp',
], 'remote-images')
```

or 

```
$result = $asset->transform([
    'width' => 1200,
    'format' => 'webp',
])->withProcessor('remote-images')
```

docs/asset-processors.md
```diff
@@ -0,0 +1,356 @@
...
+    'format' => 'webp',
+]);
+
+$url = $result->url;
+```
+
+The same override is available through `getUrl()` and Twig:
+
+```twig
+<img src="{{ asset.getUrl({
```
timkelty: I’d prefer to deprecate transform arguments on `getUrl()`. Backward compatibility could remain, but new code and documentation should use `asset.transform(...).url`.

Accepting transform arguments conflates Asset’s URL retrieval with requesting a transform and creates a second, less explicit transform API.

docs/asset-processors.md
```diff
@@ -0,0 +1,356 @@
+# Asset Processors
+
+Asset Processors turn an Asset and an Asset Transform definition into a rendition. A rendition can be an image generated
+by Craft, a URL produced by an external image service, or another representation supplied by a plugin.
```
timkelty: This makes it sound like the result could be anything, but `AssetTransformResult` like like it always should be URL-addressable.

timkelty (commented):
FYI: I only reviewed the spec doc, not the implementation.

I’m not sold on some of the terminology:

- `processor` reads like a new synonym for `transformer`. `transform` and `transformer` may carry baggage, but they’re established terms and clearly describe the relationship.

I understand why the model needs three levels, but names rooted in the existing transform terminology feel clearer:

  - `AssetTransform` / `AssetTransformRequest`
  - `AssetTransformer`
  - `AssetTransformDriver`

`rendition` feels awkward. It doesn’t appear in the public API and only occurs in production code in an error message. What about `transform result`, `output`, or even `variant` (Commerce Variants)?

`operations` and `operationFields` feel awkward. `options` / `optionFields`, `settings`, `config` feel more natural. “Operations” also seems to suggest discrete steps applied in succession, but these appear to be parameters that are normalized and passed to the driver.

docs/asset-processors.md
riasvdv: It does belong in the request, it's a hint to the processor that we expect it to be available right away, if a cloud based processor always has it available, then that's great, otherwise the processor has the choice between deferred or queued generation and immediate generation.

docs/asset-processors.md
riasvdv: A named transform gets a section for each processor and validates each processor with its own rules

docs/asset-processors.md
riasvdv: Good point, will make it so all parameters are passed to the transformers

docs/asset-processors.md
riasvdv: Adjusted the implementation so transformer rules now also take priority over the core rules

docs/asset-processors.md
riasvdv: Good call, named argument is better

```php
$result = $asset->transform([
    'width' => 1200,
    'format' => 'webp',
], transformer: 'remote-images')
```

It has to be named as `$immediately` is also an argument still.

docs/asset-processors.md
riasvdv: Done 👍

docs/asset-processors.md
riasvdv: Reworded

PR #19466, opened 2026-08-20
[6.x] Improve types

riasvdv:
This is the result of running the https://github.com/dmmulroy/anti-slop rules and having the 🤖 fix the issues.

Lots of `any` and `unknown` removed and replaced with better typings

@brianjhanson I didn't install it as actual oxlint rules in the repository, but maybe that's worth it?


brianjhanson:
@riasvdv I'd probably just add the rules. I can't think of a reason we wouldn't want them keeping things in check throughout

brianjhanson (approved):

riasvdv:
@brianjhanson Seems like it currently requires to copy all the rules into your own repo (it's not published to npm as an oxlint plugin), I'd wait a bit and see if he publishes it before copying that all in

PR #19468, opened 2026-08-20
Always call parent::setEagerLoadedElements()

brandonkelly:
Simpler and more comprehensive solution for #19459.

Always call `parent::setEagerLoadedElements()` when we’re overriding it, so that `getEagerLoadedElements()` continues to work as expected in those cases.


brandonkelly:
(This was reverted in 5.11.1 via #19539.)

PR #19511, opened 2026-08-29
fix: ensure slideout cleanup completes without transition events

### LINKED ISSUE #9912: Opening and closing slideout too fast causes elements on browser not accessible
**Description**
 I noticed clicking on an asset field several times quickly makes HTML output below without 'so-visible' class. As a result, elements on that page are not accessible anymore.
`<div class="slideout-shade" style="display: block;"></div>`

Maybe this is not an issue that happens all the time, and it took me some time to reproduce that, so please close this issue if this is not a common scenario.

**Additional info**
- Craft version: 3.7.134
- Chrome and Firefox desktop

mvanhorn:
### Description

Make `Slideout` transition completion resilient when reduced motion is disabled but an expected transition event never arrives: retain the event-driven path, add a duration-aware fallback that invokes each completion callback exactly once, and treat targets with no effective transition as immediately complete. Track fallback work with the existing transition listeners so `_cancelTransitionListeners()` cancels both parts when a slideout is reopened or rapidly toggled; this prevents an obsolete close callback from hiding the active shade or container. Keep this logic inside the existing production-used `_afterTransition()` lifecycle rather than introducing a test-only seam, and preserve the current shade fade and container motion when events fire normally.

`Craft.Slideout.close()` hides its shade and outer container only from callbacks registered by `_afterTransition()`. When a browser does not emit the expected `transitionend`, those callbacks remain pending, leaving an invisible, full-viewport shade with `display: block` and active pointer events, and preventing the slideout's close lifecycle from finishing. The thread now includes a deterministic `Craft.Slideout` reproduction on Craft 5.10.13.2 and identifies the missing transition event, although the maintainer could not reproduce it in the same browser version and requested a safe-mode check. There are no assignees, competing PRs, or closed prior attempts in the supplied issue evidence.

Fixes #9912

### Related issues

Not applicable to this change.


brandonkelly:
Thanks!

brandonkelly:
Craft 5.11.0 is out with this change. Thanks again!

PR #19520, opened 2026-08-31
[6.x] Updates to element index table

gcamacho079:
### Description
- Manages keyboard focus when reordering table rows by focusing the loading spinner and adds relevant tests
- Adds label CSS tokens in field styles
- Adds an accessible label to the “Items per page” select


resources/js/modules/elements/components/DataTable.test.ts
```diff
@@ -0,0 +1,87 @@
...
+  let container: HTMLElement | undefined;
+
+  afterEach(() => {
+    app?.unmount();
+    container?.remove();
+  });
+
+  it('moves focus to the spinner while re-sorting reloads, then back to the same sort button', async () => {
+    const table = createSampleTable();
+    const loading = ref(false);
```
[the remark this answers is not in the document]
gcamacho079: Rather than enabling and handling manual sorting in our Storybook, I think it makes sense to just update the `loading` prop manually to replicate the server-side sorting in the control panel. But at some point, we should handle this in our E2E tests.

resources/js/modules/elements/components/DataTable.vue
```diff
@@ -103,6 +113,54 @@
...
+        pendingSortFocusHeaderId.value = null;
+        await nextTick();
+        document
+          .getElementById(`header-${headerId}`)
+          ?.querySelector<HTMLButtonElement>('button')
+          ?.focus();
+      }
+    }
+  );
+
```
brianjhanson, line 163: This feels like a lot of workaround code. Is there a way we could/should structure the table and loading state to provide a better experience? Or is this just how these things are typically handled?

gcamacho079: This is fairly typical for handling focus management when focus gets dropped during a re-render. We have quite a few examples of this in 5.x, including the same focus drop fix for reordering element index table rows.

Unless we ensure that only the rows are re-rendered (not the `<thead>` or any buttons inside), but my understanding was that it wouldn’t be possible to do that with TanStack Tables. Because of the re-render, we need to manually move focus to the loading spinner, then back to the last-focused sort button. But the good thing is, workaround code like this only applies when the loading state is triggered from inside the `table` itself.

brianjhanson: 👍 

Yeah, we have to re-render the `<thead/>` because the columns can be defined per source.

brianjhanson (approved):

PR #19547, opened 2026-09-03
Fix currency label height and border radius in Money and Number fields

redburn:
Fixes a minor height discrepancy between the currency labels of Money and Number (with currency preview format) fields and their adjoining inputs, as well as a left border radius being applied to the Number input when it has a label in front:

2026-09-03 18_07_58-2026-09-03 18_06_30-Greenshot png @ 400% (Layer 2 copy 2, RGB_8#) _

Using `min-height: var(--input-height)`, like `.text` does, instead of `height:100%`, makes both the same height. Adding the `& + .text` rule that the Money label already has to the Number label fixes the border radius.


brandonkelly:
Thank you!

brandonkelly:
Craft 5.11.2 is out with that fix. Thanks again!

PR #19548, opened 2026-09-03
[6.x] Fix bugs in the section and entry type creation flow

brianjhanson:
### Description

A pass over bugs found while creating a section and its entry types. Each one
sat behind the previous, so they're grouped by where they surfaced rather than
by size.

**Creating a field from an entry type**

Clicking **New field** in the field layout designer prompted “Are you sure you
want to close this screen? Any changes will be lost.” before the editor opened.
`HUD.show()` re-parents the field library to `<body>`, so the button is no
longer inside the slideout panel; the store found no `[data-slideout-id]`
ancestor, read the click as base-page chrome, and marked every open panel for
closing — including the dirty entry type panel. The editor now opens from the
tab's add button, which stays inside the panel, and the library popup closes as
soon as the editor appears.

**Switching a field's type**

The type combobox silently kept the previous type, so `render-form` posted the
old one and the settings never changed. Its option list was rendered with an
unkeyed `render()`, so replacing the options patched the existing
`<craft-option>` elements in place; because they never disconnect, Lion never
re-registers them, and a `modelValue` naming one of the new options could never
be adopted. Keying them by value fixes the handoff, and the combobox no longer
announces a value it hasn't adopted yet.

Selecting a type was also reported as typing, so it sat on the 1000ms typing
debounce before anything happened. A value naming an option is now treated as a
committed choice; free text on its way to a match still waits, which is what
keeps a half-typed value from being posted as a field type.

**Relation field settings**

With the type switch working, the relation settings form could be reached for
the first time — and threw `options.map is not a function`. The sites
collection is keyed by ID, so `map()->all()` produced a keyed array that
`json_encode` emits as an object. It's a list now, matching how `RecentEntries`,
`QuickPost` and `ElementSourceForm` already build their site options.

“Which site should X be related from?” also rendered unconditionally, and “Show
the site menu” never hid — both halves of a toggle pair the legacy markup had
but the Form-node path didn't. Turning the switch on now lands on the primary
site, so it sticks: nothing is stored for the switch itself, it's derived from
whether a target site is set.

**Supporting changes**

`Choice` can now label its leading blank option (`placeholder()`, defaulting to
“Select…”) or drop it entirely (`withoutPlaceholder()`) for a setting with no
valid empty state — Assets' **Preview Mode** is the first user, since its
property is a non-nullable string validated by `Rule::in()`.

A form can also say where it stops being fixed. Controls declare
`rebuildsForm()`, and a new `Loader` node marks the boundary; while such a
refresh is in flight the renderer stands a loading indicator there and withholds
the nodes after it, so the settings about to be replaced don't sit there looking
current. The field editor places one straight after `field-settings-separator`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01NjeigCjRgjTJpnhbvyrxv8


src/Form/Controls/Control.php
```diff
@@ -123,6 +125,24 @@ final public function mode(ControlMode|string $mode): static
...
+
+        return $this;
+    }
+
+    final public function getRebuildsForm(): bool
+    {
+        return $this->rebuildsForm;
+    }
+
     final public function getMode(): ControlMode
```
riasvdv: Looks like the mode comment got pushed too high

src/Form/Controls/Control.php
```diff
@@ -17,6 +17,8 @@ abstract class Control implements ControlContract
     /** @var string|list<string>|null */
     protected string|array|null $deltaGroup = null;
 
+    protected bool $rebuildsForm = false;
```
riasvdv: I'm not sure I'm completely on board with this one. Forms need an endpoint to be able to refresh/rebuild. Not all forms have this, so what happens when you have a control that rebuilds a form, but no endpoint that goes along with it?

Or is rebuilding something different than refresh?

brianjhanson:
@riasvdv opened up https://github.com/craftcms/cms/pull/19554 to address his concerns.

PR #19551, opened 2026-09-03
[6.x] Don’t use the logout path as the post-login redirect

brianjhanson:
### Description

Visiting the logout path while signed out (or with an expired session) stashed it as the post-login redirect, so the next successful sign-in immediately signed the user back out — leaving them stuck in a loop they couldn’t log out of.

The logout routes required the `auth` middleware, so a guest request tripped Laravel’s `Authenticate` middleware, whose guest redirect stores the current URL in `url.intended`. Craft 5 allowed anonymous access to `users/logout` for exactly this reason.

- Moved the control panel logout route out of the `['auth', 'can:accessCp']` group and in alongside the other auth paths, and dropped `->middleware('auth')` from the front-end logout routes. Hitting logout while signed out now just redirects to the login page / post-logout redirect.
- Added `CraftCms\Cms\Support\Url::isLogoutUrl()`, which matches a URL’s path (minus any subfolder) against the control panel logout path and the localized `logoutPath` values.
- `URL::returnUrl()` now discards and forgets an intended URL that points at a logout path, so already-affected sessions heal themselves and no other flow can stash one.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01NjeigCjRgjTJpnhbvyrxv8


riasvdv:
Yeah so this is why you shouldn't allow GET for logout routes 😅 I do feel the solution is a bit over-engineered though, will have a look

PR #19561, opened 2026-09-04
[6.x] Updates to navigation, adds accessibility rules and instructions for agents

gcamacho079:
### Description
- Removes `craft-nav-item` anchor wrapping to prevent nested interactive elements
- Moves skip links so they are first focusable items in the DOM and are layered over the global navigation

### Related issues
Resolves ACC-257, ACC-259


packages/craftcms-ui/src/components/nav-item/nav-item.styles.ts
```diff
@@ -25,6 +44,12 @@ export default css`
     inset-block-end: 0;
   }
 
+  .nav-item__prefix craft-button,
+  .nav-item__suffix craft-button {
+    position: relative;
+    z-index: 1;
+  }
```
brianjhanson, line 51: Can we just add these styles to * within the prefix / suffix? Ideally the nav item doesn't have specials rules for specific items within those slots  (but I also recognize that isn't always feasible).

packages/craftcms-ui/src/utilities/navigate-event.ts
brianjhanson, line 1: Inertia already emits a handful of [events](https://inertiajs.com/docs/v3/advanced/events), whats the advantage of having our own?

PR #19563, opened 2026-09-05
Element condition refactoring

brandonkelly:
Laying the groundwork for nested condition groups, which could be combined with `OR` operators.

- [x] Move element query param application logic into static methods, so it’s accessible for arbitrary `Builder`s.
- [x] Split `modifyQuery()` out of `ElementConditionRuleInterface`, into `ElementQueryConditionRuleInterface()`, and modify signature to accept a `Builder` arg instead of the main element query, which it should modify directly (e.g. via new query param static methods).
- [x] Update `ElementCondition` to call condition rules’ `modifyQuery()` methods from within `$query->beforeQuery()`, passing the `Builder` object to them.

> [!CAUTION]
> This is a breaking change for plugins that provide element condition rules. Element condition rules must now explicitly implement `ElementConditionRuleInterface` and/or `ElementQueryConditionRuleInterface`, if they want to be applicable to element instance conditions and element query conditions, respectively.


src/Element/Queries/Concerns/QueriesRelatedElements.php
```diff
@@ -37,60 +38,60 @@ trait QueriesRelatedElements
...
+
+        $applied = new ElementRelationParamFilter(
+            fields: $elementQuery->customFields
+                ? Arr::keyBy(
+                    $elementQuery->customFields,
+                    fn (FieldInterface $field) => $field->layoutElement?->getOriginalHandle() ?? $field->handle,
+                )
+                : []
+        )->apply(
+            query: $elementQuery->getQuery(),
```
riasvdv: I think this should be `$query`

riasvdv (approved):

src/Element/Queries/Concerns/QueriesRelatedElements.php
brandonkelly: Yep you’re right, good catch.

PR #19564, opened 2026-09-05
[6.x] Port Dashboard to Vue and Inertia

riasvdv:
### Description

Felt like we finally have all the technology to port the dashboard.

Kept support for HTML widgets (`getBodyHtml()`), I don’t know if we want to move that into the adapter and deprecate that? (@brandonkelly @brianjhanson)

Some of our default widgets should probably get another styling pass.

Dashboard with Vue widgets and HTML widget support

Widget manager popover


brandonkelly:
> I don’t know if we want to move that into the adapter and deprecate that?

Feel like it makes sense to follow suit with fields, so probably worth making it a legacy support thing only.

PR #19572, opened 2026-09-07
Avoid locale-aware project config sorting on PostgreSQL

boboldehampsink:
### Description

Reloading project config on an internal config cache miss orders all stored paths using the database’s default collation. On PostgreSQL with a locale-aware collation, this sort can dominate the reload time and delay both control panel and frontend requests.

On an affected production database containing 38,633 project-config rows, the normal query took approximately 790 ms, compared with 55 ms for the same query ordered by `path COLLATE "C"`. Most of the original query time was spent sorting.

Use PostgreSQL’s bytewise `C` collation specifically when ordering paths in `_loadInternalConfig()`. This preserves the parent-before-child ordering needed to reconstruct nested config without paying for language-aware sorting of machine identifiers. The existing config cleanup still applies its natural key sorting afterward.

The change requires no schema migration and leaves MySQL ordering unchanged. It targets config-cache misses; it does not imply that every slow request reloads project config.

### Related issues

No existing GitHub issue.

### DIFF
--- a/src/services/ProjectConfig.php
+++ b/src/services/ProjectConfig.php
@@ -36,6 +36,7 @@ use yii\base\InvalidArgumentException;
 use yii\base\InvalidConfigException;
 use yii\base\NotSupportedException;
 use yii\caching\ExpressionDependency;
+use yii\db\Expression;
 use yii\web\ServerErrorHttpException;
 
 /**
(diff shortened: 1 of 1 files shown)


brandonkelly:
Thanks!

brandonkelly:
Craft 5.11.2 is out with that fix. Thanks again!

PR #19583, opened 2026-09-08
[6.x] Add chunked asset uploads and direct S3 multipart uploads

riasvdv:
### Description

Adds support for chunked file uploads and direct s3 uploads through `Uploaders`. Includes an `upload-client` JS component which handles the chunked uploading to the URL, with the headers, supplied by the uploader.

Also adds a persisted asset upload queue, so navigating away keeps uploads going

https://github.com/user-attachments/assets/16dae0ce-262b-48d1-aafe-b5fa8e6a3f98

```
Browser
  │
  │ create upload session + destination parameters
  ▼
Uploads::start()
  ├─ authorize destination
  ├─ resolve tempAssetUploadFs
  ├─ choose uploader from that temporary filesystem
  └─ store session:
       disk     = tempAssetUploadFs
       uploader = selected uploader name
       parameters = final destination
  │
  │ transfer bytes
  ▼
Temporary upload filesystem
  ├─ Tus: browser → PHP → temporary disk
  └─ S3: browser → temporary S3 disk directly
  │
  │ complete session
  ▼
UploadHandler
  ├─ Assets: copy temporary file → volume filesystem
  └─ User photo: download local copy → save photo
  │
  ▼
Uploader cleans up temporary files
```

See https://github.com/craftcms/cms/blob/feature/chunked-s3-uploads/docs/uploads.md


riasvdv:
@timkelty can you check if this satisfies the needs of Cloud?

resources/js/upload-client.ts
```diff
@@ -0,0 +1,307 @@
+import axios, {type AxiosRequestConfig, type AxiosResponse} from 'axios';
+
+const client = axios.create({adapter: 'xhr'});
```
brianjhanson, line 3: Could we do this with native fetch? I have a dream to ditch axios if we can just because it's popular and frequently targeted for exploits. Not sure if that will make it in to 6, but it would be good to avoid hanging more functionality on it if we can.

resources/js/uploads.ts
brianjhanson, line 1: Rather than creating a new global object, I think I'd rather either stick this on the `Craft.` global

brianjhanson:
It looks like this creates its own queue for uploads but I’m wondering if it should just use the JS queue already established in the CP. Admittedly I haven’t checked on the state of that in a bit, but on the surface having two feels redundant

resources/js/modules/uploader/upload-notification.ts
brianjhanson, line 1: This should probably be a web component with the amount of UI it's handling.

brianjhanson (approved):
Nothing we can't fix or adjust later, just dropped a few thoughts throughout

riasvdv:
Craft.Queue? I think that's still fully Garnish so a bit hard at the moment to integrate with for something new

brianjhanson:
Yeah, my thinking was to modernize & use that vs. creating a separate queue. Probably a separate PR at this point but just a thought

riasvdv:
Going to try and split things up a bit

Issue #3122, opened 2018-07-20
Backend is using `eval` which is incompatible with CSP


XhmikosR:
### Description

It seems the backend is loading frabric.js, which is using `eval`. In order for one to apply a CSP, they need to use `'unsafe-eval` which sort of beats the point IMO.

### Steps to reproduce

1. Visit the dashboard with a CSP applied without `'unsafe-eval'`
2. Enjoy the console errors

If you guys want to reproduce this, you can try using @april's [laboratory](https://github.com/april/laboratory) extension and you will see this:

![2018-07-20_18-37-41](https://user-images.githubusercontent.com/349621/43011610-aed48cdc-8c4c-11e8-82e8-3ef64808dd2f.jpg)

### Additional info

- Craft version: 3.0.16.1
- PHP version: 7.2
- Database driver & version: N/A
- Plugins & versions: N/A

There is an open issue on https://github.com/fabricjs/fabric.js/issues/1621 since 2014.

brandonkelly:
Are you able to identify where the `eval()` call is?

brandonkelly:
@andris-sevcenko if I’m reading the GH issue right, it sounds like this issue is mostly solved in Fabric 2.0, as the `Function` code was moved into separate modules that aren't included by default and may not be necessary in our case.

Here’s the search results for "new function" in Fabric: https://github.com/fabricjs/fabric.js/search?q=%22new+function%22&unscoped_q=%22new+function%22

So if we can use Fabric 2 without those files we should be able to close this issue.

XhmikosR:
Alternatively, I was looking into a way to enforce a different CSP for the backend, but I don't think it's possible with .htaccess (yes, I don't want to use htaccess either but I don't have control of it :)).

So, I'd need to do this via PHP in order to separate the frontend and the backend CSPs. So, if we could get rid of this, I might just get away with just a few extra bytes in a common CSP :)

brandonkelly:
Yeah you should be able to set a custom CSP from a module. If you haven’t made any changes to the module scaffolding in your Craft project yet, you can do it like this:

1. Uncomment this line from `config/app.php`:

    ```php
    //'bootstrap' => ['my-module'],
    ```

2. Add this to your `init()` method in `modules/Module.php`:

    ```php
    if (Craft::$app->request->isCpRequest) {
        Craft::$app->response->headers->add('<HeaderName>', '<HeaderValue>');
    }
    ```

XhmikosR:
[...]

PR #19600, opened 2026-09-10
[6.x] Modernize image editor

brianjhanson:
Brings the image editor into the new world. I shifted a few things around in the process which hopefully doesn't ruffle too many feathers. 

https://github.com/user-attachments/assets/1690846e-90b7-4f6f-84c9-bc319b9d7d04

A few changes that are worth noting:

- Opening the image editor will append an `editing=true` query param to the URL (via replace state, so the back button won't close it) which is what opens the modal. If a user tries to visit the image editor but doesn't have permissions to edit it, the server will prevent the modal from rendering open. 
- Fabric has been updated to the latest (Fixes #3122)
- `craft-button` now has a toggle mode. Set `toggle` on the button and it will fire a `craft-toggle` event and gets an `aria-pressed` attribute. That attribute should be used to style the active state. 
- Individual sections can now opt to be light or dark mode. Adding `data-theme="dark"` to an HTML element will flip it into dark mode. The functionality is still basic at the moment, but this lays the groundwork. 
- Undo / Redo is now available as you edit the image
- There's a "Revert to original" button which allows you to go back to the original without closing the modal


brianjhanson:
Fix for the flaky failing test is [over here](https://github.com/craftcms/cms/pull/19630)

PR #19604, opened 2026-09-11
[6.x] Add upload sessions and Uppy transport API

riasvdv:
### Description

Adds a shared upload-session API with ownership checks, expiry cleanup, resumable tus transfers, direct S3 multipart transfers, and asset destination handling. A standalone Uppy client exposes upload, progress, pause/resume, retry, and cancellation through Craft.Uploads.

```
Browser
  │
  │ create upload session + destination parameters
  ▼
Uploads::start()
  ├─ authorize destination through UploadHandler
  ├─ resolve tempAssetUploadFs
  ├─ choose uploader:
  │    explicit uploader setting, or inferred from temporary filesystem
  ├─ create session with temporary disk, uploader, and destination parameters
  └─ Uploader::start() returns UploadSetup:
       chunkSize + state → saved on session
       transportType + transportOptions → returned to browser
  │
  │ transfer bytes using the selected Uppy transport
  ▼
Temporary upload filesystem
  ├─ Tus: browser → PHP → temporary disk
  └─ S3: browser → temporary S3 disk directly
  │
  │ complete session
  ▼
Uploads::complete()
  ├─ reauthorize destination
  ├─ Uploader::complete() → staged UploadedFile
  ├─ verify file size
  └─ UploadHandler::complete()
       ├─ Assets: copy temporary file → volume filesystem
       └─ User photo: obtain local copy → save photo
  │
  ▼
Uploads calls Uploader::abort() to clean up temporary files
```

See https://github.com/craftcms/cms/blob/feature/upload-session-transports/docs/uploads.md

Upload-session event discussion: https://ampcode.com/threads/T-01a0ab6e-73bb-72ea-b504-9954678a7049


timkelty (approved):
No blockers for Cloud here, but here's what we would need from the follow-up implementation:

## craftcms/cloud requirements

- Support different upload mechanisms for different destination filesystems within the same installation. For example, Craft Cloud may define its own transport, while a project may use a combination of Craft Cloud, S3, R2, or other filesystem types.
  - One approach would be to select the upload transport per session after resolving the destination filesystem, while retaining a default transport for filesystems without a custom integration.
- Ensure uploaded file bodies do not need to pass through PHP during finalization. Provider integrations should be able to finalize uploads using storage-native operations where available.

## Other considerations

- In Craft 5.x, `tempAssetUploadFs` is broader than its name suggests. It backs Craft’s temporary Asset workspace, including temporary Asset elements, per-user folders, unresolved field uploads, and assets awaiting validation or relocation. If upload sessions also use it for transport staging, we should reassess whether one setting should represent both responsibilities or whether the naming and configuration should distinguish them.

brianjhanson (approved):

PR #19611, opened 2026-09-11
Drop redundant Assets/Entries field settings

brandonkelly:
Removes the “Show unpermitted files” and “Restrict allowed file types” settings from Assets fields, in favor of “File Type” and “Viewable” condition rules in the Selectable Assets Condition setting.

And ditto for “Show unpermitted entries” on Entries fields – fields with that disabled get a “Viewable” condition rule added to their Selectable Entries Condition setting.

a “Selectable Assets Condition” setting with File Type and Viewable condition rules


thupsi:
> And ditto for “Show unpermitted entries” on Entries fields – fields with that enabled get a “Viewable” condition rule added to their Selectable Entries Condition setting.

Shouldn't that be the opposite though? 🤔

brandonkelly:
@thupsi Sorry, yes, the PR description was incorrect.

PR #19626, opened 2026-09-14
Make element fixture revision creation opt-in

boboldehampsink:
### Description

Element fixtures currently use normal element saves, which create revision history and recursively copy nested Matrix and CKEditor content. This adds substantial work when seeding content that does not need revision history.

This change makes revision creation opt-in for both `BaseElementFixture` (including `EntryFixture`) and `BaseContentFixture`. Set the fixture property to retain the previous behavior:

```php
public bool $createRevisions = true;
```

A shared trait temporarily suppresses versioning for entries encountered during each fixture save, including nested Matrix entries with independent versioning enabled. It restores the in-memory settings and removes its event handler in `finally`, including when saving fails. Opting in respects existing section and field versioning settings. Validation, propagation, field hooks, and indexing retain their existing behavior; persistent versioning configuration is unchanged.

**Measured performance impact**

On a Craft CMS 5.11.1 multisite project, loading all 51 fixtures with automatic section revisions suppressed produced:

| Full fixture load | Duration |
| --- | ---: |
| Normal revision creation | 374.66s (6m 15s) |
| Revision creation suppressed | 167.71s (2m 48s) |

That is **206.95 seconds saved per seed run: 55.2% less time, or 2.23× faster**. The baseline added 737 revision records. Both runs finished with matching counts for canonical elements (798), canonical site rows (9,085), canonical relations (2,169), canonical ownerships (529), and search rows (2,960).

Both runs used the same source and container image, with the same database snapshot restored and the isolated Redis cache cleared before each run. This was one sequential local A/B comparison of revision suppression, rather than a benchmark of this exact event-based implementation. The gain depends on the project's revision and nested-content workload; optimized CI timing has not been measured. Matching structural counts do not establish full content equivalence.

### Related issues

None.


boboldehampsink:
Thanks @brandonkelly! Once this lands all of our builds will be faster :D

brandonkelly:
Thanks for the PR!

PR #19628, opened 2026-09-14
Update fabric to 7.4

brianjhanson:
Ports the fabric 1.7 -> 7 API changes from the 6.x image editor to the legacy AssetImageEditor:

- Promise-based SVG/image loading and setSrc
- An _animate() helper, since fabric now fires onComplete per property
- Size the image with scaleX/scaleY, since a FabricImage's width/height are its natural size
- Settle flip flags after the flip animation

I did all this work against 6.x but in such a way that it was easy to split off for 5.x which made me think we might as well update there so LTS version doesn't have the eval issue. 

This is _technically_ a breaking change if a plugin is using the existing `FabricAsset` as that will now load 7.4.x instead of 1.7.x.

### Related issues
Fixes https://github.com/craftcms/cms/issues/3122


riasvdv:
@copilot are there any public Craft v5 plugins that use `FabricAsset`?

PR #19632, opened 2026-09-14
[6.x] Freeze animated thumbnails in craft-thumbnail

gcamacho079:
### Description

Animated GIF/WEBP thumbnails on the Assets index and asset edit page were playing indefinitely, which fails WCAG 2.2.2 (Pause, Stop, Hide). Craft already freezes animated images elsewhere in the control panel, but that mechanism can't reach `craft-thumbnail`'s image since it's rendered inside a shadow root.

- Adds an `animated` property to `craft-thumbnail`, driven by the `animated` flag Craft already computes server-side, with a `.gif`/`.webp` extension fallback for markup that doesn't set it.
- When animated, freezes the source to its first frame by drawing it onto a decorative cover canvas sized to match the visible image, instead of letting it autoplay. The underlying image itself is left untouched, so its alt text still works normally.
- No PHP changes needed — the server already provided the needed signal.
- Adds a repo convention for documenting accessibility requirements alongside new or accessibility-relevant component changes going forward.

Tests: `thumbnail.test.ts` and a new `Animated` story in `thumbnail.stories.ts`; requirements tracked in the new `thumbnail.a11y.md`.


gcamacho079:
@brianjhanson I noticed `<craft-thumbnail>` isn't currently being used on element detail pages (in the details sidebar). Is this in progress on another branch, or should this PR be updated to include that?

brianjhanson:
> @brianjhanson I noticed `<craft-thumbnail>` isn't currently being used on element detail pages (in the details sidebar). Is this in progress on another branch, or should this PR be updated to include that?

Updating this PR to include that would be great.

brianjhanson (approved):

PR #19637, opened 2026-09-15
[6.x] Automatically lazy eager-load element relations

riasvdv:
### Description

Automatically lazy eager-load relational field queries across the originating element query result set during site requests rendered in site template mode. Automatic loading is skipped for Control Panel requests and CP template mode, while explicit `eagerly()` calls remain available.

Query criteria are isolated so different operations and filters do not share incompatible results, while customized query-builder state and incompatible result modes fall back to normal execution.

Adds the `autoEagerLoadElements` general config setting and supports opting individual queries out with `eagerly(false)`.

### Related issues

- https://github.com/craftcms/cms/discussions/16910


src/Config/GeneralConfig.php
```diff
@@ -412,6 +412,22 @@ class GeneralConfig extends BaseConfig
...
+     * ->autoEagerLoadElements(false)
+     * ```
+     * ```shell Environment Override
+     * CRAFT_AUTO_EAGER_LOAD_ELEMENTS=false
+     * ```
+     * :::
+     *
+     * @group System
+     */
+    public bool $autoEagerLoadElements = true;
```
riasvdv, line 429: Should this default to `true`?

riasvdv: Only frontend templates by default

PR #19648, opened 2026-09-16
[6.x] Design updates

brianjhanson:
- The page shell layout has been rearranged and cleaned up. 
- Pages can narrow their content column with `contentMaxWidth` and center it with `centerContent`. Settings and form pages now use both.
- You can debug the slots on the page via a ?debug-slots query param. That will stick until you make another request to turn them off (`?debug-slots=0`)
- The legacy CP stylesheet is no longer loaded on Inertia pages.

**Navigation**
- The whole nav tree is built on the server and sent once. Element sources hang off Entries, Assets and Users, grouped under their headings.
- The branch you're on expands in place, and every other branch opens in a flyout. Flyouts share hover timing and fit the space below them.
- Selection marks the source you're on rather than the index beside it, and matches query strings as well as paths.
- A collapse item sits at the bottom of the sidebar. When the sidebar floats on small screens, it layers above sticky page content.
- Breadcrumbs repeat the secondary nav's trail. Each source crumb has a switcher that matches the nav, including on an entry's edit page.
- The bare assets index redirects to its first source.

**Element editor and indexes**
- `ElementEditor` handles full pages as well as slideouts.
- Details tabs can be opened and closed from code, and each tab has a close button.
- New Entry, New User and Upload files have been morved into the element index toolbar.

**Components and tokens**
- `craft-button` has a `flush` option that offsets its padding, and spaces an icon from its label when the variant contains `plain`. Otherwise it does nothing.
- Popovers open in the browser's top layer, so sticky page content can't cover them, and take the theme's text color.
- Menu group headings are styled like nav headings.
- New semantic z-index tokens (`--c-layer-*`) and Tailwind utilities for surfaces, border colors, spacing and layers. A Storybook page documents the utilities.
- Quiet borders are translucent, with one variable to adjust the amount.
- Admin tables run full width on a shared `--c-container-padding` gutter.
- The CP language's writing direction is shared through Inertia and set as `dir` on the root element.

**Housekeeping**
- Removed husky and lint-staged from the legacy package. Its `prepare` script ran husky outside the repository root, where it could have redirected git hooks away from `vp`'s.


brianjhanson:
### Code review

Found 2 issues:

1. The default page heading renders `{{ title }}` instead of `{{ pageTitle }}`, dropping the shared-title fallback. `title` is an optional prop with no default, and `pageTitle` is `props.title?.trim() ?? page.props.title` — so any page that doesn't pass `:title` now renders an empty `<h1>` where the shared Inertia prop used to fill it. For example [`settings/users/Fields.vue`](https://github.com/craftcms/cms/blob/bc617183c0bc00372dab409151661f8463ea6a63/resources/js/pages/settings/users/Fields.vue#L41-L43) renders `<AppLayout :form="form" …>` with no `:title`, while [`UserFieldsController`](https://github.com/craftcms/cms/blob/bc617183c0bc00372dab409151661f8463ea6a63/src/Http/Controllers/Settings/Users/UserFieldsController.php#L37-L40) sets `->title(t('User Settings'))`, which reaches the page as `page.props.title`. The browser tab title is unaffected, since `<Head :title="pageTitle">` still uses the computed.

https://github.com/craftcms/cms/blob/bc617183c0bc00372dab409151661f8463ea6a63/resources/js/common/layouts/screens/PageScreen.vue#L284-L288

2. Focus is no longer returned to the sidebar toggle when the floating sidebar closes. `toggleButton` is never assigned anywhere at this head — on `6.x`, `PageScreen.vue` registered it with `:ref="registerToggle"`, but the rewrite drops that and the toggle moved into `CpTopBar.vue` (`id="sidebar-toggle"`, no ref binding) alongside a new close button in `CpSidebar.vue`. The optional-chained `toggleButton.value?.focus()` is therefore a silent no-op and focus falls to `<body>` on close, which is a WCAG 2.4.3 regression ([`.ai/rules/accessibility.md`](https://github.com/craftcms/cms/blob/bc617183c0bc00372dab409151661f8463ea6a63/.ai/rules/accessibility.md) requires WCAG 2.2 AA). The docblock at L44 still describes the contract as "Registered by whichever component renders it".

https://github.com/craftcms/cms/blob/bc617183c0bc00372dab409151661f8463ea6a63/resources/js/common/composables/useGlobalSidebar.ts#L111-L115

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #19665, opened 2026-09-19
[6.x] Form component compatibility

lukeholder:
### Description

Various fixes and restoration of functionality for Tables and other form components that existed in 5.x


src/Form/Controls/ComboboxCreate.php
```diff
@@ -0,0 +1,101 @@
...
+ * client needs to act on it once chosen; it doesn't inject the option itself, the same way
+ * every other Combobox/Choice leaves option-building entirely to the caller.
+ *
+ * On save, the created record is read from the response's {@see self::resultKey()} — matching
+ * the `$modelName` the create screen's own save action passes to
+ * `RespondsWithFlash::asModelSuccess()` — and appended to the option list as the new selection,
+ * using {@see self::labelField()}/{@see self::valueField()} (default `name`/`id`) to build its
+ * label/value.
+ */
+class ComboboxCreate extends Combobox
```
riasvdv: Should this really be a separate component from the Combobox? @brianjhanson what do you think?

brianjhanson: I'd say the same component. Ideally, I don't think the Combobbox needs to know it has a "Create" option at all. I'd say the option itself should be able to carry all this functionality and then the developer can just add that specific option into a regular combobox

PR #19678, opened 2026-09-21
Fix translation for deleting other users' types

fabianhaef:


brandonkelly:
Looks like the formatting changed in your PR, so it’s not really possible to tell what changed.

In any case, we manage translations via CrowdIn, at [translate.craftcms.com](https://translate.craftcms.com). Please create an account there and submit the change – and also leave a comment about it so we are notified about the suggestion.

